PR #19, opened 2016-03-02
insert script for organisation start date

sachinkulkarni12:


fineract-provider/src/main/resources/sql/migrations/core_db/V292__update_organisation_start_date.sql
```diff
@@ -0,0 +1 @@
+INSERT INTO `c_configuration` ( `name`, `value`, `date_value`, `enabled`, `is_trap_door`, `description`) VALUES ('organisation-start-date', 0, NULL, 1, 0, NULL);
```
rajuan, line 1: Default value should be disabled

PR #23, opened 2016-03-03
Display Full Name instead of First Name, Middle Name and Last Name if…

jinjurajan:
… configuration is enabled


rajuan:
In the code, there is only addition of new global configuration. This global configuration is not used for any validations or data management. There will not be any change to existing APIs whatsoever. Such global configurations are not going to help in any way.

PR #28, opened 2016-03-07
payment type applicable for disbursement charge configuration added

ghost:


fineract-provider/src/main/resources/sql/migrations/core_db/V293__configuration_for_paymnettype_application_forDisbursement_charge.sql
```diff
@@ -0,0 +1 @@
+INSERT INTO `c_configuration` (`name`, `value`, `enabled`, `is_trap_door`, `description`) VALUES ('paymenttype-applicable-for-disbursement-charges', NULL, 0, 0, 'Is the Disbursement Entry need to be considering the fund source of the paymnet type');
```
nazeer1100126, line 1: Migration file with this number already exists.

PR #37, opened 2016-03-15
Note functionality while Editing a Transaction and undoing a Transaction and on click of a particular Transaction adding a Note row

KoustavMuhuri:
Code changes for enabling Note functionality while editing a Transaction and undoing a Transaction and adding notes Row while on click of a particular transaction.


KoustavMuhuri:
Instead of making two commits I will be making 1 single commit and create another pull request.

PR #41, opened 2016-03-16
FINERACT-76 adding fund source to the payment type and adding payment…

ghost:
… type to client charge if selected will be using fund source of payment type for debit

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/accounting/journalentry/data/ClientTransactionDTO.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/accounting/journalentry/data/ClientTransactionDTO.java
@@ -22,6 +22,7 @@ import java.math.BigDecimal;
 import java.util.Date;
 import java.util.List;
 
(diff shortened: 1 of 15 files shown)


fineract-provider/src/main/java/org/apache/fineract/accounting/journalentry/data/ClientTransactionDTO.java
```diff
@@ -22,6 +22,7 @@
 import java.util.Date;
 import java.util.List;
 
+import org.apache.fineract.accounting.glaccount.domain.GLAccount;
```
rajuan: unused import

rajuan:
As of now, Income account client charges are defined while defining the charges. We would suggest the same place to add advanced GLaccount mappings for different payment types as well. Solution as provided in PR, in my opinion, splits the way GL accounts mappings are defined across portfolio as of now. Also allowing to define fund source account id at payment type definition could mislead the user into believing that the definition is valid across the portfolios. In the future also new use cases could come up which might make this solution non-extendable.

PR #48, opened 2016-03-23
Skipping Meeting which Falls On First Day Of Month

jinjurajan:


rajuan:
This PR has compilation issues

PR #60, opened 2016-04-03
[MIFOSX-2310] User allowed to leave Max age for LOSS criteria blank.

ddc22:
Under category we have STANDARD, SUB-STANDARD, DOUBTFUL and LOSS options. In which if the Loss is considered as the last category it would be better to set the Max Age as Infinity (Blank) so that for any loans which are due from very long period falls under that category.

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/organisation/provisioning/serialization/ProvisioningCriteriaDefinitionJsonDeserializer.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/organisation/provisioning/serialization/ProvisioningCriteriaDefinitionJsonDeserializer.java
@@ -104,10 +104,20 @@ public class ProvisioningCriteriaDefinitionJsonDeserializer implements Provision
                         .parameterAtIndexArray(ProvisioningCriteriaConstants.JSON_MINIMUM_AGE_PARAM, i + 1).value(minimumAge).notNull()
                         .longZeroOrGreater() ;
 
-                Long maximumAge = this.fromApiJsonHelper.extractLongNamed(ProvisioningCriteriaConstants.JSON_MAXIMUM_AGE_PARAM, jsonObject);
-                baseDataValidator.reset().parameter(ProvisioningCriteriaConstants.JSON_MAXIMUM_AGE_PARAM)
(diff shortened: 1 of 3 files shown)


fineract-provider/src/main/java/org/apache/fineract/organisation/provisioning/serialization/ProvisioningCriteriaDefinitionJsonDeserializer.java
```diff
@@ -104,10 +104,20 @@ public void validateForCreate(final String json) {
                         .parameterAtIndexArray(ProvisioningCriteriaConstants.JSON_MINIMUM_AGE_PARAM, i + 1).value(minimumAge).notNull()
                         .longZeroOrGreater() ;
 
-                Long maximumAge = this.fromApiJsonHelper.extractLongNamed(ProvisioningCriteriaConstants.JSON_MAXIMUM_AGE_PARAM, jsonObject);
-                baseDataValidator.reset().parameter(ProvisioningCriteriaConstants.JSON_MAXIMUM_AGE_PARAM)
-                        .parameterAtIndexArray(ProvisioningCriteriaConstants.JSON_MAXIMUM_AGE_PARAM, i + 1).value(maximumAge).notNull()
-                        .longGreaterThanNumber(ProvisioningCriteriaConstants.JSON_MINIMUM_AGE_PARAM, minimumAge, (i+1));
+
+                //Allow Max age of "Loss Category" to be blank => Symbolises infinity// NotNull check removed when  categoryId=4
+                if(categoryId == 4) {
```
rajuan, line 109: Categories are user configurable. Some organisations can have 3 categories and some other 10. So such hardcoded check will not help. This solution cannot be taken as is. This solution should be generic to allow only one nullable max age param and it doesn't violate overlap conditions.

ddc22, line 109: I was also confused at this! But the requirement in the issue tracker specifically assumes that there are only 4 categories 
https://mifosforge.jira.com/browse/MIFOSX-2310

> Currently under category we have "STANDARD", "SUB-STANDARD", "DOUBTFUL" and "LOSS" options.  In which if the Loss is considered as the last category it would be better to set the Max Age as Infinity (Blank)

This is why I uneasily did this fix. Will think of a way to abstract his out!

PR #67, opened 2016-04-08
Client data Added fields

anuragmath:
For those who need to add more personal details about a client and also a additional contact address the following fields have been added
1. Father Name
2. Email Address
3. Marital Status (Collection)
   (create codes through manage codes in admin->system and code title should be same as given    clientdataconstat.java file)
4. Religion (Collection) 
5. Dependent (Collection)
6. Education (Collection)
   same code creation from rest below fields

This kind of data will useful for NBFC/MFI to collect more information about there clients.


sheyden:
Hi anurgamath, Thanks for your PR. However the platform already supports extending the datamodel through datatables. This allows different implementers to setup exactly those fields they need for their specific setup. In this way we can keep the core client object (and code) clean and tidy, and only put properties in it that are required by other objects to function (mobile_no etc)

In addition we prefer for people to squash their commits before making a PR, this allows us to keep the commit history clean.

PR #70, opened 2016-04-12
Secure Update password

nikpawar89:
User now has to authenticate the update password request by password.

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/useradministration/api/UsersApiResource.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/useradministration/api/UsersApiResource.java
@@ -46,9 +46,14 @@ import org.apache.fineract.infrastructure.security.service.PlatformSecurityConte
 import org.apache.fineract.organisation.office.data.OfficeData;
 import org.apache.fineract.organisation.office.service.OfficeReadPlatformService;
 import org.apache.fineract.useradministration.data.AppUserData;
+import org.apache.fineract.useradministration.exception.IncorrectPasswordException;
 import org.apache.fineract.useradministration.service.AppUserReadPlatformService;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.context.annotation.Scope;
+import org.springframework.security.authentication.UsernamePasswordAuthenticationToken;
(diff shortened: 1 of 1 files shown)


fineract-provider/src/main/java/org/apache/fineract/useradministration/api/UsersApiResource.java
```diff
@@ -140,30 +148,45 @@ public String create(final String apiRequestBodyAsJson) {
...
-        final CommandWrapper commandRequest = new CommandWrapperBuilder() //
-                .updateUser(userId) //
-                .withJson(apiRequestBodyAsJson) //
-                .build();
-
-        final CommandProcessingResult result = this.commandsSourceWritePlatformService.logCommandSource(commandRequest);
-
-        return this.toApiJsonSerializer.serialize(result);
+  @PUT
+  @Path("/{userId}/usrnme/{usrnme}/currentPass/{currentPass}")
```
rajuan, line 152: 1. update API is used for all types of updates including firstname..... we cannot expect that a firstname update by back office user would need him to know the password of the user. This change cannot be accepted.
2. If the expectation is that an existing user is trying to modify the password of their own userId, then basic/oauth authentication itself would have validated the current password.... so no need to check again

nikpawar89:
After implementing this code, I did few tests to change the user details. Nothing accept 'updatepassword case' was asked password. For now I have just proposed the change, there are still vulnerabilities in this approach which I discovered  later. I am going to Fix them and would come back later. Thanks for the feedback

fineract-provider/src/main/java/org/apache/fineract/useradministration/api/UsersApiResource.java
nikpawar89, line 152: After implementing this code, I did few tests to change the user details. Nothing accept 'updatepassword case' was asked password. For now I have just proposed the change, there are still vulnerabilities in this approach which I discovered later. I am going to Fix them and would come back later. Thanks for the feedback

MexinaD: Hell @nikpawar  did you work on this?

vorburger:
https://github.com/apache/fineract/pull/639/

PR #76, opened 2016-04-22
254:In tranche loans, if repayment date is same as tranche disbursement date then allow to change the emi amount.

sachinkulkarni12:


fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/data/LoanTermVariationsDataWrapper.java
```diff
@@ -55,26 +55,34 @@ public LoanTermVariationsDataWrapper(final List<LoanTermVariationsData> exceptio
...
 
-    private boolean hasNext(final LocalDate date, ListIterator<LoanTermVariationsData> iterator) {
+    private boolean hasNext(final LocalDate date, ListIterator<LoanTermVariationsData> iterator, 
+            final Boolean isChangeEmiIfRepaymentDateSameAsDisbursementDateEnabled) {
         boolean hasVariation = false;
         if (iterator.hasNext()) {
             LoanTermVariationsData loanTermVariationsData = iterator.next();
-            if (!loanTermVariationsData.getTermApplicableFrom().isAfter(date)) {
-                hasVariation = true;
+            if (!isChangeEmiIfRepaymentDateSameAsDisbursementDateEnabled) {
```
pramodn02, line 68: you can modify the effective date of emi amount  with the configuration.. no need to check all the time

PR #103, opened 2016-05-09
293_post_interest_job

AvinashKumarG:


fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountReadPlatformServiceImpl.java
```diff
@@ -200,6 +200,30 @@ public SavingsAccountReadPlatformServiceImpl(final PlatformSecurityContext conte
         return this.paginationHelper.fetchPage(this.jdbcTemplate, sqlCountRows, sqlBuilder.toString(), finalObjectArray,
                 this.savingAccountMapper);
     }
+    
+    //Pagination Read service for Active Savings Account Post Interest Job
+    @Override
+    public Page<SavingsAccountData> retrieveActiveSavingsAccount(final Integer maxPazeSize,final Integer pageNumber, final Integer status) {
```
nazeer1100126, line 206: The method itself is to retrieve active savings accounts. Do we need to pass status as param?

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularServiceImpl.java
```diff
@@ -37,42 +41,73 @@
...
+	@Override
+	public void postInterestForAccounts() throws JobExecutionException {
+		final int maxPazeSize = 10;
+		boolean hasErrorOccurred = false;
+		boolean tempHasErrorOccured = false;
+		int offSet = 0;
+		final Integer status = SavingsAccountStatusType.ACTIVE.getValue();
+		List<SavingsAccount> savings = new ArrayList<>();
+
+		Page<SavingsAccountData> savingsAccounts = this.savingAccountReadPlatformService
```
nazeer1100126, line 66: You never using this savingsAccounts .

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularServiceImpl.java
```diff
@@ -37,42 +41,73 @@
...
+
+		while (savingsAccounts.getPageItems().size() < savingsAccounts.getTotalFilteredRecords()) {
+			offSet += savingsAccounts.getPageItems().size();
+			savingsAccounts = this.savingAccountReadPlatformService.retrieveActiveSavingsAccount(maxPazeSize, offSet,
+					status);
+			for (SavingsAccountData savingsAccountData : savingsAccounts.getPageItems()) {
+				SavingsAccount savingsAccount = this.savingAccountAssembler.assembleFrom(savingsAccountData.id());
+				savings.add(savingsAccount);
+			}
+			postInterestInBatch(savings);
```
nazeer1100126, line 77: Why do we need to call postInterestInBatch() method two times?

PR #110, opened 2016-05-10
FINERACT-37 Reschedule payments to the next payment date on holidays

jroweboy:
A simple patch that adds a field to the database to enable rescheduling repayments that land on holidays to the next repayment date instead of a set date. 

Note: This patch does not contain any way to enable this feature (in testing I just set the field to true in the database by hand). This is just a request for comments on my method of implementation, and while that getting settled, I'll add in the other things required to make this a full change.

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/organisation/workingdays/domain/WorkingDays.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/organisation/workingdays/domain/WorkingDays.java
@@ -41,15 +41,19 @@ public class WorkingDays extends AbstractPersistable<Long> {
 
     @Column(name = "extend_term_daily_repayments", nullable = false)
     private Boolean extendTermForDailyRepayments;
+
+    @Column(name = "extend_term_holiday_repayment", nullable = false)
+    private Boolean extendTermForRepaymentsOnHolidays;
     
     protected WorkingDays() {
 
     }
 
-    protected WorkingDays(final String recurrence, final Integer repaymentReschedulingType, final Boolean extendTermForDailyRepayments ) {
+    protected WorkingDays(final String recurrence, final Integer repaymentReschedulingType, final Boolean extendTermForDailyRepayments, final Boolean extendTermForRepaymentsOnHolidays) {
         this.recurrence = recurrence;
         this.repaymentReschedulingType = repaymentReschedulingType;
         this.extendTermForDailyRepayments = extendTermForDailyRepayments;
+        this.extendTermForRepaymentsOnHolidays = extendTermForRepaymentsOnHolidays;
     }
 
     /**
@@ -73,6 +77,8 @@ public class WorkingDays extends AbstractPersistable<Long> {
     public Boolean getExtendTermForDailyRepayments(){
         return this.extendTermForDailyRepayments;
     }
+
+    public Boolean getExtendTermForRepaymentsOnHolidays() { return this.extendTermForRepaymentsOnHolidays; }
 
     public Map<String, Object> update(final JsonCommand command) {
         final Map<String, Object> actualChanges = new LinkedHashMap<>(7);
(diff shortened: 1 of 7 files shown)


fineract-provider/src/main/java/org/apache/fineract/organisation/workingdays/domain/WorkingDays.java
```diff
@@ -74,6 +78,8 @@ public Boolean getExtendTermForDailyRepayments(){
         return this.extendTermForDailyRepayments;
     }
 
+    public Boolean getExtendTermForRepaymentsOnHolidays() { return this.extendTermForRepaymentsOnHolidays; }
+
     public Map<String, Object> update(final JsonCommand command) {
```
rajuan, line 83: No code changes that would set this variable is written..... no one can change this attribute

fineract-provider/src/main/java/org/apache/fineract/organisation/workingdays/domain/WorkingDays.java
```diff
@@ -41,15 +41,19 @@
 
     @Column(name = "extend_term_daily_repayments", nullable = false)
     private Boolean extendTermForDailyRepayments;
+
+    @Column(name = "extend_term_holiday_repayment", nullable = false)
+    private Boolean extendTermForRepaymentsOnHolidays;
```
rajuan, line 46: Read platform changes to be able to read this variable is required

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/DefaultScheduledDateGenerator.java
```diff
@@ -96,10 +96,37 @@ public LocalDate generateNextRepaymentDate(final LocalDate lastRepaymentDate, fi
...
+                        holiday.getFromDateLocalDate().toDateTimeAtStartOfDay(),
+                        holiday.getToDateLocalDate().toDateTime(LocalTime.parse("23:59:59"))
+                ));
+            }
+            while (intervalListContainsDate(holidayInterval, dueRepaymentPeriodDate)) {
+                LocalDate previousDate = dueRepaymentPeriodDate;
+                dueRepaymentPeriodDate = getRepaymentPeriodDate(loanApplicationTerms.getRepaymentPeriodFrequencyType(),
+                        loanApplicationTerms.getRepaymentEvery(), dueRepaymentPeriodDate, loanApplicationTerms.getNthDay(),
+                        loanApplicationTerms.getWeekDayType());
+                if (previousDate == dueRepaymentPeriodDate) {
```
rajuan: As long as repayment period frequency type is valid, this condition will never satisfy.... I do not understand why if condition is added

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/DefaultScheduledDateGenerator.java
```diff
@@ -96,10 +96,37 @@ public LocalDate generateNextRepaymentDate(final LocalDate lastRepaymentDate, fi
                         loanApplicationTerms.getNumberOfdays());
             }
         }
-        
+        if (currentCalendar == null && holidayDetailDTO.getWorkingDays().getExtendTermForRepaymentsOnHolidays()) {
```
rajuan, line 99: What if calendar is already present and a holiday is newly added?

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/DefaultScheduledDateGenerator.java
```diff
@@ -18,9 +18,10 @@
  */
```
rajuan, line 18: I have generally given the comments.... I will let other code experts to comment on sufficiency of the solution and also if there are any more problems with usecases

fineract-provider/src/main/java/org/apache/fineract/organisation/workingdays/domain/WorkingDays.java
jroweboy, line 83: I apologize that I didn't mention the fact that I was asking for feedback on how I added the changes. That said, I will add in the rest of the backend changes in order to make this an acceptable PR when I get a little extra time.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/DefaultScheduledDateGenerator.java
jroweboy, line 99: This sounds like a design decision that I will need to ask about. But I agree that something needs to change. Imagine the case where a loan has been dispersed, and then a holiday is added that affects a future repayment. This patch does not account for this case, and I'm guessing that it should. Good catch.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/DefaultScheduledDateGenerator.java
jroweboy: Hmmm... I wrote this patch 6 months ago and looking at it again, I can't remember why I added this either. I will remove this then.

PR #116, opened 2016-05-13
Adding multiple selection option for reportparameters

jinjurajan:


fineract-provider/src/main/resources/sql/migrations/core_db/V310__multiselect_for_reportparameters.sql
```diff
@@ -0,0 +1,10 @@
+UPDATE `report`.`stretchy_parameter` SET `parameter_sql`='(select lo.id, lo.display_name as `Name`
```
nazeer1100126, line 1: Why this script required?

avikganguly01: @jinjurajan : Can you explain this query and share a link to the JIRA ticket so that we can pull this in?

myrle-krantz:
ping

Is this pull request still "alive"?  Will Nazeer and Avik's question receive an answer?  If there's no response in a week this pull request will be closed.

PR #117, opened 2016-05-14
Easier Setup

avikganguly01:
Change in default exclusion so that ServerApplication and EmbeddedTomcatWithSSLConfiguration are default available in build path.
Inclusion of commons-io dependency as FileUtils in the excluded EmbeddedTomcatWithSSLConfiguration requires this dependency.

Changing sort function used in InterestRateChart to be compatible with Java 7 as well since I believe not all users would have moved to Java 8 by now.


nazeer1100126:
Can you explain bit more why this change is required? 
Prerequisite for Fineract is Java 8. So no need to change the InterestRateChart  sort function.

avikganguly01:
Well, it is always great to be on top of things and use all the latest technologies but if you feel breaking backward compatibility because of one line is fine, then it is your call.

The other part of the PR is about easier setup where ServerApplication is by default included in the sourceSet, so for beginners, there won't be any need to change classpath exclusions before firing the server up. Dependency inclusion is because of excluding `EmbeddedTomcatWithSSLConfiguration` and commons-lang forking out commons-io as a separate dependency.

PR #124, opened 2016-05-26
Agging notes,createdBydate and createdbyuser m loan transaction table

jinjurajan:


nazeer1100126:
Please make use of existing notes feature instead of adding new columns in loan transaction table.

PR #125, opened 2016-05-26
fineract collection sheet

KoustavMuhuri:
Adding Total Due,Total Collected,Total OverDue and Installment Due in Json on click of Productive Collection Sheet.

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/group/api/CentersApiResource.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/group/api/CentersApiResource.java
@@ -66,6 +66,7 @@ import org.apache.fineract.portfolio.group.data.CenterData;
 import org.apache.fineract.portfolio.group.data.GroupGeneralData;
 import org.apache.fineract.portfolio.group.data.StaffCenterData;
 import org.apache.fineract.portfolio.group.service.CenterReadPlatformService;
+import org.apache.fineract.portfolio.group.service.CenterReadPlatformServiceImpl;
 import org.apache.fineract.portfolio.meeting.data.MeetingData;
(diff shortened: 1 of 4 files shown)


fineract-provider/src/main/java/org/apache/fineract/portfolio/group/api/CentersApiResource.java
```diff
@@ -155,6 +156,8 @@ public String retrieveAll(@Context final UriInfo uriInfo, @QueryParam("sqlSearch
         final ApiRequestJsonSerializationSettings settings = this.apiRequestParameterHelper.process(uriInfo.getQueryParameters());
         if (meetingDateParam != null && officeId != null) {
             Date meetingDate = meetingDateParam.getDate("meetingDate", dateFormat, locale);
+			LocalDate date = new LocalDate(meetingDate);
+			CenterReadPlatformServiceImpl.datePassed = date;
```
nazeer1100126, line 160: Why are you assigning meeting date to a static variable in CenterReadPlatformServiceImpl

fineract-provider/src/main/java/org/apache/fineract/portfolio/group/service/CenterReadPlatformServiceImpl.java
```diff
@@ -85,6 +86,7 @@
     private final CodeValueReadPlatformService codeValueReadPlatformService;
     private final ConfigurationDomainService configurationDomainService;
     private final CalendarReadPlatformService calendarReadPlatformService;
+    public static LocalDate datePassed;
```
nazeer1100126, line 89: This variable will be null if this datePassed is not set from API resource.

KoustavMuhuri:
Re sending the PR after taking care of the suggested comments

PR #129, opened 2016-06-03
504 loan write off reason added

pramodconflux:


fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -1039,12 +1045,20 @@ public CommandProcessingResult writeOff(final Long loanId, final JsonCommand com
 
         this.loanEventApiJsonValidator.validateTransactionWithNoAmount(command.json());
 
+        Long writeoffReasonId = command.longValueOfParameterNamed("writeoffReasonId");
         final Map<String, Object> changes = new LinkedHashMap<>();
         changes.put("transactionDate", command.stringValueOfParameterNamed("transactionDate"));
         changes.put("locale", command.locale());
         changes.put("dateFormat", command.dateFormat());
+        changes.put("writeoffReasonId", writeoffReasonId);
```
nazeer1100126, line 1053: If writeoffReasonId is not passed from client, why do we need to add writeoffReasonId to changes map?

PR #135, opened 2016-06-09
validating disbursement date against expected disbursement date

jinjurajan:


nazeer1100126:
This code is merged. But not reflecting.

avikganguly01:
@nazeer1100126 : I think that this validation is already there but the code seems to be attempting some sort of sync. Closing this as I could not find any open JIRA ticket related to this.

PR #155, opened 2016-06-22
FINERACT-177 individual client share in group loan

pramodconflux:


fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/api/GroupLoanIndividualMonitoringApiResource.java
```diff
@@ -0,0 +1,119 @@
...
+		this.apiRequestParameterHelper = apiRequestParameterHelper;
+		this.toApiJsonSerializer = toApiJsonSerializer;
+		this.commandsSourceWritePlatformService = commandsSourceWritePlatformService;
+		this.groupLoanIndividualMonitoringReadPlatformService = groupLoanIndividualMonitoringReadPlatformService;
+	}
+    
+    @GET
+    @Consumes({ MediaType.APPLICATION_JSON })
+    @Produces({ MediaType.APPLICATION_JSON })
+    public String retrieveAll(@Context final UriInfo uriInfo) {
```
nazeer1100126, line 72: Better to give pagination support. Instead of giving complete data.

ottoman91:
hi, is this feature still being used? i have a query regarding it, let me know if it is. thanks.

PR #170, opened 2016-07-14
commit for FINERACT-46 (Throw an exception if staff to be deactivated is attached to an entity with status enum id not equal to 600)

emmanuelnnaa:


fineract-provider/src/main/java/org/apache/fineract/organisation/staff/domain/StaffRepository.java
```diff
@@ -27,10 +27,36 @@
 
     public final static String FIND_BY_OFFICE_QUERY = "select s from Staff s where s.id = :id AND s.office.id = :officeId";
 
+    public final static String IS_ANY_ACTIVE_ENTITY_ASSOCIATED_WITH_STAFF_QUERY = "select " +
```
nazeer1100126, line 30: Instead of adding a native query in StaffRepository, add new method in StaffReadPlatformService and use JDBC template to run this query

fineract-provider/src/main/java/org/apache/fineract/organisation/staff/service/StaffWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -91,6 +92,17 @@ public CommandProcessingResult updateStaff(final Long staffId, final JsonCommand
 
             final Staff staffForUpdate = this.staffRepository.findOne(staffId);
             if (staffForUpdate == null) { throw new StaffNotFoundException(staffId); }
+            
+            final boolean isActive = command.booleanPrimitiveValueOfParameterNamed("isActive");
+            
+            if(staffForUpdate.isActive() && !isActive) {
+                final boolean isAnyActiveEntityAssociatedWithStaff = this.staffRepository.
```
nazeer1100126, line 99: Use StaffReadPlatformService instead of repository

fineract-provider/src/main/java/org/apache/fineract/organisation/staff/domain/StaffRepository.java
```diff
@@ -27,10 +27,36 @@
...
+            "join m_staff ms on ml.loan_officer_id = ms.id " +
+            "where ms.id = :staffId " +
+            "and ml.loan_status_id in (100,200,300,700)) " +
+            "union all " +
+            "(select count(msa.id) as total " +
+            "from m_savings_account msa " +
+            "join m_staff ms on msa.field_officer_id = ms.id " +
+            "where ms.id = :staffId and " +
+            "msa.status_enum in (100,300))) as t1 ";
+
```
avikganguly01, line 50: Client and share accounts not checked.

PR #172, opened 2016-07-18
commit for FINERACT-49 (Waived charges and zero installment charges amounts not excluded from total charges amount on reschedule loan preview)

emmanuelnnaa:


nazeer1100126:
Hi Emmanuel, Can you check whether this issue is still valid after multi reschedule changes are merged.

PR #175, opened 2016-07-18
commit for FINERACT-51 (Add the 'is_reversed' property of the loan transaction entity to the 'LoanTransactionData' object)

emmanuelnnaa:


fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanReadPlatformServiceImpl.java
```diff
@@ -247,7 +247,7 @@ public LoanScheduleData retrieveRepaymentSchedule(final Long loanId,
              ***/
             final String sql = "select "
                     + rm.LoanPaymentsSchema()
-                    + " where tr.loan_id = ? and tr.transaction_type_enum not in (0, 3) and  (tr.is_reversed=0 or tr.manually_adjusted_or_reversed = 1) order by tr.transaction_date ASC,id ";
+                    + " where tr.loan_id = ? and tr.transaction_type_enum not in (0, 3) and  (tr.is_reversed=0 or tr.is_reversed = 1) order by tr.transaction_date ASC,id ";
```
nazeer1100126, line 250: what happens if manually_adjusted_or_reversed=0 or 1?

emmanuelnnaa, line 250: @nazeer1100126 
The main reason for the change was to have the query return all results (both system reversed and manually reversed). That can also be achieved without the following line "and  (tr.is_reversed=0 or tr.is_reversed = 1)". Consequently, I will remove the above line and create a new pull request.

PR #176, opened 2016-07-18
commit for FINERACT-52 (Null value not handled in 'ClientWritePlatformServiceJpaRepositoryImpl.CloseClient' date comparison)

emmanuelnnaa:


fineract-provider/src/main/java/org/apache/fineract/portfolio/client/service/ClientWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -631,7 +631,7 @@ public CommandProcessingResult closeClient(final Long clientId, final JsonComman
                 throw new InvalidClientStateTransitionException("close", "is.under.transfer", errorMessage);
             }
 
-            if (client.isNotPending() && client.getActivationLocalDate().isAfter(closureDate)) {
+            if (client.isNotPending() && client.getActivationLocalDate() != null && client.getActivationLocalDate().isAfter(closureDate)) {
```
nazeer1100126, line 634: Instead of checking activation date to null why can't we add a check if client status is active and client activation date is after closure date as error message is about closure date can not be before activation date?

emmanuelnnaa, line 634: @nazeer1100126 
There's no guarantee that the "activationDate" property of all clients will have a value other than NULL or an empty string. During migration of a new tenant, some data might be missed. Moreover, it is good practice to carry out such checks to avoid null pointer exceptions.

vorburger:
@emmanuelnnaa thank you very much for your interest in Mifos!  I'm an (old) volunteer who got reminded of Mifos this week-end at the FOSDEM conference, and I'm taking a moment to help the project clean up some of it's old issues and pull requests etc.  

We were wondering whether you are still alive, and would be willing to rebase this PR and take any previous review feedback into account?  If we don't hear from you within 2 weeks, we will take the liberty to close this PR, to avoid confusion for future new contributors looking at this repository - hope you understand and that's fair.

vorburger:
Friendly service reminder: I will close this PR in 3 days to help this project have a clean list of active PRs.

PR #200, opened 2016-08-09
address_module

nikpawar89:


fineract-provider/src/main/java/org/apache/fineract/infrastructure/configuration/api/GlobalConfigurationApiResource.java
```diff
@@ -47,73 +47,74 @@
 @Scope("singleton")
 public class GlobalConfigurationApiResource {
```
nazeer1100126, line 48: Since there is no change in this file, this file should not be part of this PR

fineract-provider/src/main/resources/sql/migrations/core_db/V312__address_module_tables_metadat.sql
```diff
@@ -0,0 +1,227 @@
...
+
+-- code inserts
+INSERT INTO `m_code` (`code_name`, `is_system_defined`) VALUES ('STATE',1);
+INSERT INTO `m_code` (`code_name`, `is_system_defined`) VALUES ('COUNTRY',1);
+INSERT INTO `m_code` (`code_name`, `is_system_defined`) VALUES ('ADDRESS_TYPE',1);
+
+-- inserts for STATE
+INSERT INTO `m_code_value` (`code_id`, `code_value`,`order_position`, `is_active`) 
+VALUES(
+ (select id from m_code where code_name = 'STATE'), 'Andhra Pradesh', 0,1);
```
nazeer1100126, line 29: I don't think we need to add states of India as part of SQL migration. This is admin job and he will be adding countries and states from the UI.

fineract-provider/src/main/java/org/apache/fineract/portfolio/address/api/AddressConfigurationApiResources.java
```diff
@@ -0,0 +1,96 @@
...
+import org.apache.fineract.portfolio.address.service.AddressReadPlatformServiceImpl;
+import org.apache.fineract.portfolio.address.service.FieldConfigurationReadPlatformService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Scope;
+import org.springframework.stereotype.Component;
+
+@Path("/clients/addresses")
+@Component
+@Scope("singleton")
+public class AddressConfigurationApiResources {
```
nazeer1100126, line 51: I believe this template api should be part of ClientAddressApiResources. And this class can be removed

fineract-provider/src/main/java/org/apache/fineract/portfolio/address/domain/Address.java
```diff
@@ -0,0 +1,348 @@
...
+
+	/*@OneToMany(mappedBy = "address", cascade = CascadeType.ALL)
+	private List<ClientAddress> clientaddress = new ArrayList<>();*/
+	
+	@OneToMany(mappedBy = "address", cascade = CascadeType.ALL)
+	private Set<ClientAddress> clientaddress;
+
+	private String street;
+
+	private String address_line_1;
```
nazeer1100126, line 50: Give proper names to all variables in entity classes. You can refer other entity classes

Nayan:
@nikpawar89 please add the API docs :)

PR #202, opened 2016-08-11
address_module

nikpawar89:


fineract-provider/src/main/java/org/apache/fineract/portfolio/address/service/AddressWritePlatformServiceImpl.java
```diff
@@ -0,0 +1,279 @@
...
+			final Long addressid = add.getId();
+			final Address addobj = this.addressRepository.getOne(addressid);
+
+			final boolean isActive = jsonObject.get("isActive").getAsBoolean();
+
+			final ClientAddress clientAddressobj = ClientAddress.fromJson(isActive, client, addobj, addressTypeIdObj);
+			this.clientAddressRepository.save(clientAddressobj);
+
+		}
+		final long typ = 1;
```
nazeer1100126, line 151: why typ is hard coded to 1 and why you are always returning entity id with 1 ?

fineract-provider/src/main/java/org/apache/fineract/portfolio/address/service/AddressWritePlatformServiceImpl.java
```diff
@@ -0,0 +1,279 @@
...
+		this.fromApiJsonDeserializer.validateForUpdate(command.json());
+
+		final long addressId = command.longValueOfParameterNamed("addressId");
+
+		final ClientAddress clientAddressObj = this.clientAddressRepositoryWrapper
+				.findOneByClientIdAndAddressId(clientId, addressId);
+
+		final Address addobj = this.addressRepository.getOne(addressId);
+
+		if (!(command.stringValueOfParameterNamed("street").isEmpty())) {
```
nazeer1100126: If street is not passed in json, NullPointerException will thrown. This is same for all parameters in update json.

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/service/ClientWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -90,734 +92,803 @@
...
+	private final SavingsProductRepository savingsProductRepository;
+	private final SavingsApplicationProcessWritePlatformService savingsApplicationProcessWritePlatformService;
+	private final CommandProcessingService commandProcessingService;
+	private final ConfigurationDomainService configurationDomainService;
+	private final AccountNumberFormatRepositoryWrapper accountNumberFormatRepository;
 	private final FromJsonHelper fromApiJsonHelper;
+	private final ConfigurationReadPlatformService configurationReadPlatformService;
+	private final AddressWritePlatformService addressWritePlatformService;
+
+	@Autowired
```
nazeer1100126, line 118: This file is having formatting issue. Import mifosx preferences and format it

PR #208, opened 2016-08-15
commit for FINERACT-72 (Speed up the retrieval of journal entries when 'transactionDetails' is set to true)

emmanuelnnaa:


nazeer1100126:
Hi Emanuel,

Can you explain bit more, how it is going to increase the performance.

avikganguly01:
@nazeer1100126 : I don't think it makes sense to add payment detail to JE table.

PR #209, opened 2016-08-16
feature implementation of undoreject and undowithdrawal for client

jyothsnag:


fineract-provider/src/main/java/org/apache/fineract/portfolio/client/service/ClientWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -844,4 +844,59 @@ public CommandProcessingResult reActivateClient(Long entityId, JsonCommand comma
...
+	@Override
+	public CommandProcessingResult undoRejection(Long entityId, JsonCommand command) {
+		final AppUser currentUser = this.context.authenticatedUser();
+		this.fromApiJsonDeserializer.validateUndoRejection(command);
+
+		final Client client = this.clientRepository.findOneWithNotFoundDetection(entityId);
+		final LocalDate undoRejectDate = command
+				.localDateValueOfParameterNamed(ClientApiConstants.reopenedDateParamName);
+
+		if (!client.isRejected()) {
```
nazeer1100126, line 857: This validation logic can be moved to validator

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/service/ClientWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -844,4 +844,59 @@ public CommandProcessingResult reActivateClient(Long entityId, JsonCommand comma
...
+	@Override
+	public CommandProcessingResult undoWithdrawal(Long entityId, JsonCommand command) {
+		final AppUser currentUser = this.context.authenticatedUser();
+		this.fromApiJsonDeserializer.validateUndoWithDrawn(command);
+
+		final Client client = this.clientRepository.findOneWithNotFoundDetection(entityId);
+		final LocalDate undoWithdrawalDate = command
+				.localDateValueOfParameterNamed(ClientApiConstants.reopenedDateParamName);
+
+		if (!client.isWithdrawn()) {
```
nazeer1100126, line 885: This validation logic can be moved to validator

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/service/ClientWritePlatformServiceJpaRepositoryImpl.java
rajuan, line 857: I do not think moving the validation to validator is absolutely required here. We are still avoiding entity creation, which is a good benefit.

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/service/ClientWritePlatformServiceJpaRepositoryImpl.java
rajuan, line 885: I do not think moving the validation to validator is absolutely required here. We are still avoiding entity creation, which is a good benefit.

PR #219, opened 2016-08-25
FINERACT-37 Reschedule repayment on holidays+extend term

terencemo:
Based on PR #110.
Adds a WorkingDays flag extendTermForRepaymentOnHoliday when this is set
If a repayment occurs on a holiday, shift it to the next repayment date and
extend the loan term by the repayment interval, e.g
if repayment is weekly and one repayment occurs on holiday, move that to the next original schedule repayment date and extend loan term by 1 week

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/DefaultScheduledDateGenerator.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/DefaultScheduledDateGenerator.java
@@ -18,9 +18,10 @@
  */
 package org.apache.fineract.portfolio.loanaccount.loanschedule.domain;
 
-
+import org.apache.fineract.organisation.holiday.domain.Holiday;
 import org.apache.fineract.organisation.holiday.service.HolidayUtil;
 import org.apache.fineract.organisation.workingdays.domain.RepaymentRescheduleType;
+import org.apache.fineract.organisation.workingdays.domain.WorkingDays;
 import org.apache.fineract.organisation.workingdays.service.WorkingDaysUtil;
 import org.apache.fineract.portfolio.calendar.data.CalendarHistoryDataWrapper;
 import org.apache.fineract.portfolio.calendar.domain.Calendar;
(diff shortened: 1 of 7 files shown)


fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/DefaultScheduledDateGenerator.java
```diff
@@ -57,10 +57,10 @@ public LocalDate generateNextRepaymentDate(final LocalDate lastRepaymentDate, fi
             boolean isFirstRepayment, final HolidayDetailDTO holidayDetailDTO) {
         final LocalDate firstRepaymentPeriodDate = loanApplicationTerms.getCalculatedRepaymentsStartingFromLocalDate();
         LocalDate dueRepaymentPeriodDate = null;
+        Calendar currentCalendar = loanApplicationTerms.getLoanCalendar();
         if (isFirstRepayment && firstRepaymentPeriodDate != null) {
             dueRepaymentPeriodDate = firstRepaymentPeriodDate;
         } else {
-            Calendar currentCalendar = loanApplicationTerms.getLoanCalendar();
             dueRepaymentPeriodDate = getRepaymentPeriodDate(loanApplicationTerms.getRepaymentPeriodFrequencyType(),
```
nazeer1100126: You have removed  Calendar currentCalendar = loanApplicationTerms.getLoanCalendar(); 
It is going to affect existing functionality

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/DefaultScheduledDateGenerator.java
```diff
@@ -96,10 +96,33 @@ public LocalDate generateNextRepaymentDate(final LocalDate lastRepaymentDate, fi
                         loanApplicationTerms.getNumberOfdays());
             }
         }
-        
+        if (currentCalendar == null && holidayDetailDTO.getWorkingDays().getExtendTermForRepaymentsOnHolidays()) {
```
nazeer1100126, line 99: if extendTermForRepaymentsOnHolidays is true, why can't we call adjustRepaymentDate method  and return that date?

nazeer1100126:
I believe we no need to have two boolean variables to extend the term.
1) extendTermForDailyRepayments
2) extendTermForRepaymentsOnHolidays 

We can have a single config extendTerm. If this config is enabled, generateNextRepaymentDate method should return adjusted date

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/DefaultScheduledDateGenerator.java
terencemo: Tried calling adjustRepaymentDate. It doesn't reschedule for holidays other than weekend. Went through the code of adjustRepaymentDate and found nothing that handles holidays. The code block after the if does something which I couldn't find in any existing function in the codebase.

terencemo:
extendTermForDailyRepayments flag is for a different purpose than extendTermForRepaymentsOnHolidays. There should be a way to specify which of the 2 is required and hence 2 bools. If a single bool is used, what is the behaviour if its set? Does it extend the term for both loans with daily repayment schedule and for holidays? What if the API user needs one of these 2 functionalities?

nazeer1100126:
I agree to your point :-)

PR #227, opened 2016-09-23
Fineract-225

RanjithKumar5550:


fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/data/DataValidatorBuilder.java
```diff
@@ -180,6 +180,28 @@ public DataValidatorBuilder notNull() {
         return this;
     }
 
+    public DataValidatorBuilder notNullCheckbox() {
```
nazeer1100126, line 183: This method is not required. The logic in LoanRescheduleRequestDataValidator is not correct. Please correct it.

PR #232, opened 2016-11-03
sql fix

nikpawar89:
Minor fix for sql


fineract-provider/src/main/resources/sql/migrations/core_db/V316__address_module_tables_metadat.sql
```diff
@@ -20,7 +20,7 @@
 
 
 -- code inserts
-INSERT INTO `m_code` (`code_name`, `is_system_defined`) VALUES ('STATE',1);
+INSERT INTO `m_code` (`code_name`, `is_system_defined`) VALUES ('STATES',1);
```
nazeer1100126: You need to add new migration to alter column name. This migration file is already deployed on servers, so it won't have any effect

PR #234, opened 2016-11-03
address sql fix

nikpawar89:
Update statement added for changing 'STATE' to 'STATES'


rajuan:
You will have to add new SQL file with latest version, we cannot modify the old SQL script..... deployments that have already gone past v316 will not get this change.

Also, you cannot just change the STATE code_name to STATES by default..... If someone is using STATE as their custom code name value, it will affect their usage in their deployments.

You will have to find some intelligent solution that solves your problem and doesn't create problem for other existing deployments

nikpawar89:
@rajuan Thank you for your thoughts Adi. I have found a workaround which would not involve change in SQL files.

PR #245, opened 2016-11-17
Fix for MIFOSX-2799

gkrishnan724:
Issue:
When a new user is created in the community-app, the E-mail received on registration was too complicated to understand
Fix:
Changed the E-mail in a more understandable way.


gkrishnan724:
Hope this one is fine :-P

PR #246, opened 2016-11-17
Put gradle version in build.gradle and avoid having to download the correct gradle version prior to building

maddingo:


nazeer1100126:
I tried your changes. But I got the following error.
Error: Could not find or load main class org.gradle.wrapper.GradleWrapperMain
It is not downloading the gradle wrapper jar file.

maddingo:
@nazeer1100126  Thanks for trying my PR. 
I can reproduce the steps I am running with the following Dockerfile, i.e. on a clean environment:
```
FROM ubuntu:16.04

RUN apt-get -y update && apt-get -y install git openjdk-8-jdk

RUN git clone https://github.com/maddingo/incubator-fineract && cd incubator-fineract/fineract-provider && ./gradlew war

```
Which tasks are you trying to run? Maybe you have a newer version?

nazeer1100126:
You are able to run as already gradle-wrapper jar is available in https://github.com/apache/incubator-fineract/tree/develop/fineract-provider/gradle/wrapper/ directory. The purpose of your PR is not to download the jar and place in this directory. So I have removed this jar from mentioned directory and tried to build gradlew clean build. But it didn't work.

maddingo:
I am not a Gradle wizzard. But here is how I understand it. Gradle needs the gradle jar file to work. But what the change does it ensures that the right gradle version is used and you don't require the developer to  download and set up the right gradle version manually. If there is a requirement for a specific gradle version, it should be put in the build tool.

nazeer1100126:
@maddingo Have you tested by removing existing gradle jar file? Is it downloading and build happening properly?

maddingo:
@nazeer1100126 Sorry for the late reply. I think you can close this issue. I found a way to set up my development environment without this change.

vorburger:
This has meanwhile been done by someone else, and one can now just use:

    ./gradlew build

PR #248, opened 2016-11-18
Fix for resolving scorecards, and replacing dead fk relation

mgeiss:


nazeer1100126:
Migration is failing with V323__spm_replace_dead_fk_with_exisiting_one.sql

mgeiss:
what is the issue? it works fine on my system ... but I use a MariaDB 10

PR #256, opened 2016-12-06
FINERACT-7: Sort loanInstallmentCharges before updating

abdulazizali77:
Less than ideal fix (sorting overhead), but most topically contained. Potentially we should change the Sets to Lists, and or redefine the Entities, but that would affect too many things.

Have left a potentially iffy corner case in updateInstallmentCharges as is.

Additionally there is a pre-existing defect where a PersistenceException caused by an FK constraint is raised when all of the charges are waived. Have not investigated, potentially there is an existing bug.

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanCharge.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanCharge.java
@@ -22,6 +22,7 @@ import java.math.BigDecimal;
 import java.math.MathContext;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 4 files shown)


fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanCharge.java
```diff
@@ -494,17 +494,29 @@ private void updateInstallmentCharges() {
...
-            for (final LoanInstallmentCharge chargePerInstallment : this.loanInstallmentCharge) {
+            //FINERACT-7: NB: this is really overkill and slow, without changing the Sets and or Entities this is the most
+            //locally safe way to fix this issue
+            ArrayList<LoanInstallmentCharge> oldChargeInstallments = new ArrayList<LoanInstallmentCharge>();
+            ArrayList<LoanInstallmentCharge> newChargeInstallments = new ArrayList<LoanInstallmentCharge>();
+
+            oldChargeInstallments.addAll(this.loanInstallmentCharge);
+            newChargeInstallments.addAll(chargePerInstallments);
+            Collections.sort(oldChargeInstallments);
+            Collections.sort(newChargeInstallments);
```
nazeer1100126: After sorting newChargeInstallments, it is never used

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanCharge.java
```diff
@@ -494,17 +494,29 @@ private void updateInstallmentCharges() {
...
+            final LoanInstallmentCharge[] loanChargePerInstallmentArray = chargePerInstallments.toArray(new LoanInstallmentCharge[chargePerInstallments.size()]);
```
nazeer1100126: I believe this array object is unnecessary. You can modify Loan.generateInstallmentLoanCharges(final LoanCharge loanCharge) method to return List object. Please use this list object

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanCharge.java
abdulazizali77: fixed. (was a mistake in patch cleaning)

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanCharge.java
abdulazizali77: Youre right, have modified the interface accordingly

abdulazizali77: Please rereview, thank you.

nazeer1100126:
@abdulazizali77 we will accept PR(s) with a single commit to maintain cleaner history. Can you please send PR with a single commit.

abdulazizali77:
@nazeer1100126 squashed

nazeer1100126:
@abdulazizali77 I am seeing following integration test failure with these changes ClientLoanIntegrationTest.testLoanScheduleWithInterestRecalculation_WITH_REST_SAME_AS_REPAYMENT_INTEREST_COMPOUND_NONE_STRATEGY_REDUCE_EMI_WITH_INSTALLMENT_CHARGE()
Have you run integration tests before sending this PR?

abdulazizali77:
@nazeer1100126 Hi, sorry i didnt run the integration test before. I did give the integration tests a run with and without the changes and it seems i get identical failures on ClientLoanIntegrationTest (and other tests).
I will ascertain whether the two modified methods in the PR are called during any of the integration tests. From first glance however testLoanScheduleWithInterestRecalculation_WITH_REST_SAME_AS_REPAYMENT_INTEREST_COMPOUND_NONE_STRATEGY_REDUCE_EMI_WITH_INSTALLMENT_CHARGE doesnt seem to be calling Loan.generateInstallmentLoanCharges nor LoanCharge.updateInstallmentCharges. (possibly i am wrong)

avikganguly01:
@abdulazizali77 : I know this is a bit late but if you can confirm that the integration tests are not failing with your code after rebasing, then it will be great.

abdulazizali77:
@avikganguly01 Apologies, would take a bit of effort to reverify this, potentially code has diverged a lot in the time when this PR had been submitted

abdulazizali77:
@avikganguly01 Looking at the conflicts, it looks like the issue could have already been fixed, as i remember it all i did was make the list comparable and then sort it, it seems like `develop` already has such code and typed properly too ! I think your effort is just to reverify the issue, because it looks likely fixed.
However, the issue of failing/succeeding/ integration tests is something that i cannot verify.

PR #257, opened 2016-12-08
Created a better e-mail message for a new user

mdew192837:
* Subject simplified to 'New User Information'
* Simplified the body to only include login credentials
* Reworded and clarified the password change required when a user first logs in.


nazeer1100126:
This issue is already resolved. Please check https://github.com/apache/incubator-fineract/commit/8cc4b6e4f1aa3a38a153e484b2d4003118d7a6ce

mdew192837:
@nazeer1100126 I made improvements to the message per the request of Santosh. I think the proposed change will simplify it even more. An e-mail message does not need to be that cluttered, especially for a login e-mail. Let me know what you think.

nazeer1100126:
@mdew192837 Previous commit on this issue looks OK to me.

PR #266, opened 2016-12-25
Fix to charge fees when total amount changed during approval.

tangemicioglu:
The issue is currently at https://github.com/openMF/community-app/issues/1719 but the problem was in Fineract instead of the community-app.


nazeer1100126:
If the loan is enabled with Multi disbursements, recalculateAllCharges() method will be called twice while approving a loan with this change.

tangemicioglu:
Would it be a better alternative to put it between the multi disbursement and approved loan amount blocks? It would run every time the approve loan function is called but it would run once.

nazeer1100126:
@Eiliries Yes it is good to moving existing call from multidisbursement if condition.

tangemicioglu:
@nazeer1100126 I've changed it so that it's called once and it works without any problems.

PR #271, opened 2017-01-04
 Fineract-265:addressModuleNullPointerFix

nikpawar89:
This is a fix for a bug in address module. The address module throws a null pointer exception when the user creates a new client and puts in the address without activating it ( keeping the enable check box as unchecked). Now the code assumes that the address is not active by default and only checks for value if the parameter has one.


nazeer1100126:
@nikpawar89 Changes looks OK.

PR #273, opened 2017-01-09
Added recurringFrequency to recurringDeposit files.

tangemicioglu:
This is the back-end part of the fix for https://github.com/openMF/community-app/issues/1682. The API docs have also been updated as the recurring deposit product now also uses recurringFrequency instead of recurringDepositFrequency for internal consistency.

### DIFF
--- a/api-docs/apiLive.htm
+++ b/api-docs/apiLive.htm
@@ -23609,12 +23609,12 @@ Content-Type: application/json
 	            <tr class=alt><td>lockinPeriodFrequencyType</td></tr>
 	            e.g. 2 <b>Months</b></td></tr>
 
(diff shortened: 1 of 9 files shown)


nazeer1100126:
Changes looks OK. Basically you are capturing recurringFrequencyType
recurringFrequency data. But what is the use of these values?

tangemicioglu:
For this PR, I just made them stored in the product and account so that they can be used. I think it should have an effect on the interest chart but I'm not sure how (especially considering there is a depositPeriodFrequency variable as well). The documentation specifies that these values should have been captured anyways, but they weren't working properly for a while.

PR #274, opened 2017-01-12
FINERACT-352 fix showing penalties after last loan's installment

lionelraymundi:
The new installment created in which penalties should be applied after loan's maturity date, was created twice. This messed up with the charges recalculation.


nazeer1100126:
@lionelraymundi Changes Looks OK.

PR #298, opened 2017-02-24
FINERACT-406 allow disbursal of topup loan on same date of transaction of previous loan

lionelraymundi:


nazeer1100126:
Change looks good.

PR #299, opened 2017-03-01
FINERACT-412 - Add Share account in Global Search

ippezshelby:
Commit for FINERACT-412 - Add Share account in Global Search

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/search/data/SearchConditions.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/search/data/SearchConditions.java
@@ -28,6 +28,7 @@ public class SearchConditions {
     private final Boolean groupSearch;
     private final Boolean loanSeach;
 	private final Boolean savingSeach;
+	private final Boolean shareSeach;
     private final Boolean clientIdentifierSearch;
     private  Boolean exactMatch;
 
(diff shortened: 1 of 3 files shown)


fineract-provider/src/main/java/org/apache/fineract/portfolio/search/data/SearchConditions.java
```diff
@@ -43,18 +44,21 @@ public SearchConditions(final String searchQueryParam, final String searchResour
                 SEARCH_SUPPORTED_RESOURCES.LOANS.name().toLowerCase())) ? true : false;
         this.savingSeach = (null == searchResource || searchResource.toLowerCase().contains(
                 SEARCH_SUPPORTED_RESOURCES.SAVINGS.name().toLowerCase())) ? true : false;
+		this.shareSeach = (null == searchResource || searchResource.toLowerCase().contains(
+                SEARCH_SUPPORTED_RESOURCES.SAVINGS.name().toLowerCase())) ? true : false;
```
nazeer1100126, line 48: shareSeach should look for SEARCH_SUPPORTED_RESOURCES.SHARES not SEARCH_SUPPORTED_RESOURCES.SAVINGS

nazeer1100126 (approved):

fineract-provider/src/main/java/org/apache/fineract/portfolio/search/data/SearchConditions.java
ippezshelby: I will fix that and send new PR

PR #315, opened 2017-03-16
fix for FINERACT-243 : Constitution type is not required field

tonic889:
modified:   fineract-provider/src/main/java/org/apache/fineract/portfolio/client/data/ClientDataValidator.java
	modified:   fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/ClientNonPerson.java
	new file:   fineract-provider/src/main/resources/sql/migrations/core_db/V326__nonpersonclient_constitution_optional.sql

fix for FINERACT-243 : Constitution type is not required field


nazeer1100126:
@tonic889 I have already done the changes and sent a PR  on the same. Please look at https://github.com/apache/incubator-fineract/pull/313

tonic889:
So what is the right way to avoid this duplication of effort again in the future?

avikganguly01:
@tonic889 - The right way to avoid this duplication of effort would be to assign the issue to yourself and mark it as in progress so that someone else does not pick it up. Apologies for the inconvenience cause.

PR #327, opened 2017-04-06

nazeer1100126:


mgeiss:
This is a nice start, but we should also make sure we are not getting statements as mentioned here ... https://en.wikipedia.org/wiki/SQL_injection#Technical_implementations

PR #328, opened 2017-04-07

nazeer1100126:


fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/SQLInjectionValidator.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class SQLInjectionValidator {
+
+	private final static String[] DDL_COMMANDS = { "create", "drop", "alter", "truncate", "comment" };
+
+	private final static String[] DML_COMMANDS = { "select", "insert", "update", "delete", "merge", "upsert", "call" };
+
+	private final static String[] COMMENTS = { "--", "({", "/*" };
```
alexivanov, line 31: **#** could also be used for commenting out queries

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/SQLInjectionValidator.java
```diff
@@ -0,0 +1,116 @@
...
+
+	private final static String SQL_PATTERN = "[a-zA-Z_, ()0-9]*";
```
alexivanov, line 33: **=** , **'** / **"** should be matched. **%** could also be added to the list.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/SQLInjectionValidator.java
```diff
@@ -0,0 +1,116 @@
...
+		}
+
+		while (inputSqlString.indexOf("= ") > 0) { //Don't remove space after = operator
+			inputSqlString = inputSqlString.replaceAll("= ", "=");
+		}
+
+		StringTokenizer tokenizer = new StringTokenizer(inputSqlString, " ");
+		while (tokenizer.hasMoreTokens()) {
+			String token = tokenizer.nextToken().trim();
+			if (token.equals("'")) {
```
alexivanov, line 69: Spaces before and after **"** / **'** would be nice to be added as they can be in the middle or prepended/appended to a token. e.g `column = "foo"`

alexivanov (commented):
Overall a good approach :+1:

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/SQLInjectionValidator.java
nazeer1100126: @MegaAlex are you taking about escape char \ or / ?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/SQLInjectionValidator.java
nazeer1100126: @MegaAlex I am removing the spaces before and after of '='. I believe above case won't appear.

PR #331, opened 2017-04-13

nazeer1100126:


mgeiss (commented):
Looks good to me. Maybe another commiter can look at it too.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/SQLInjectionValidator.java
```diff
@@ -0,0 +1,117 @@
...
+
+public class SQLInjectionValidator {
+
+	private final static String[] DDL_COMMANDS = { "create", "drop", "alter", "truncate", "comment" };
+
+	private final static String[] DML_COMMANDS = { "select", "insert", "update", "delete", "merge", "upsert", "call" };
+
+	private final static String[] COMMENTS = { "--", "({", "/*", "#" };
+
+	private final static String SQL_PATTERN = "[a-zA-Z_=,.'\"`% ()0-9]*";
```
rajuan, line 33: ! or <> is not allowed.... inequality or comparison cannot be done as part of sqlSearch

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/SQLInjectionValidator.java
```diff
@@ -0,0 +1,117 @@
...
+			}
+		}
+
+		for (String comments : COMMENTS) {
+			if (lowerCaseSQL.contains(comments)) {
+				throw new SQLInjectionException();
+			}
+		}
+
+		//Removing the space before and after '=' operator
```
rajuan, line 55: Code between line 55 and 109 is futile in preventing unauthorised data access (office hierarchy restriction)....
For eg... it will throw an exception for string "' OR '1' = '1'"
but not for 
a. "a' OR '1' = '1" 
b. "a' OR '2' != '1'"

Basically we can retrieve all the data irrespective of user hierarchy restriction.

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/service/ClientReadPlatformServiceImpl.java
```diff
@@ -225,6 +228,7 @@ private String buildSqlStringFromClientCriteria(final SearchParameters searchPar
 
         String extraCriteria = "";
         if (sqlSearch != null) {
+        	SQLInjectionValidator.validateSQLInput(sqlSearch);
```
rajuan, line 231: This check should done for all the string params that are received for eg. in this class it should be checked for externalId, name and orderBy etc......

Also for completeness these checks needs to be done across all read platform services

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/SQLInjectionValidator.java
nazeer1100126: The main purpose of the fix is not to return un authorized data. In above cases, it won't return any data as where condition fails.

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/service/ClientReadPlatformServiceImpl.java
nazeer1100126: Instead of we are checking whether operand and value are same. This case will be considered as invalid sql

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/service/ClientReadPlatformServiceImpl.java
```diff
@@ -225,6 +228,7 @@ private String buildSqlStringFromClientCriteria(final SearchParameters searchPar
 
         String extraCriteria = "";
```
rajuan, line 229: I feel supporting sqlSearch is futile..... based on the validation code, we can always find a workaround injection....
I suggest getting rid of this support....

nazeer1100126: This use case is legacy and used by some customers. So we can't get rid of this fix. 
So we have discussed internally before suggesting this fix.

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/service/ClientReadPlatformServiceImpl.java
rajuan, line 231: Instead of validating input param string.... I would suggest rewriting the query as prepared statements..... Have a list to add/track a value when a sql is appended.....

For eg.... replace extraCriteria.append(" and g.external_id = ").append(ApiParameterHelper.sqlEncodeString(externalId));

with 
extraCriteria.append(" and g.external_id = ? ")
paramList.add(externalId)

convert the paramList to array and use it in the jdbcTemplate in the end while executing the query.....

I'm not sure if this approach could be used for order by clause

nazeer1100126: My approach is simple. I am accepting any param. But both operand and value should not be equal which is solving the actual purpose of this fix.

mgeiss: I'd suggest to parse the params before we hit any depper layer ... so we can provide an early exit instead of relying to a depper level ...

PR #335, opened 2017-04-20
teller cashier issues #334 #432 #433 #434 #435

nazeer1100126:


nazeer1100126:
This is merged. Due to sync issue it is not reflecting here.

PR #344, opened 2017-05-03
bulk jlg loan bug fixes

nazeer1100126:


nazeer1100126:
issue list got fixed 333,443,444,445,446,447,448,449,450,451

PR #351, opened 2017-05-10
MIFOSX-2149-Surplus Payment Treatment for Loan

ippezshelby:


fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -1789,6 +1791,45 @@ private LoanCharge retrieveLoanChargeBy(final Long loanId, final Long loanCharge
...
+    @CronTarget(jobName = JobName.ALLOCATE_OVERPAYMENTS_TO_SAVINGS)
+    public void allocateOverpayments() throws JobExecutionException {
+        final Collection<Long> loanIds = this.loanReadPlatformService.fetchOverpayedLoansForAllocation();
+        final StringBuilder sb = new StringBuilder();
+
+        if(loanIds != null){
+            for(final Long loanId : loanIds){
+
+                try{
+                    final AccountAssociations accountAssociations = this.accountAssociationRepository.findByLoanIdAndType(loanId,AccountAssociationType.LINKED_ACCOUNT_ASSOCIATION.getValue());
```
nazeer1100126: What happens when linked savings account is not available? There might be a null pointer exception here.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanproduct/data/LoanProductData.java
```diff
@@ -117,6 +117,7 @@
     private final Integer minimumDaysBetweenDisbursalAndFirstRepayment;
     private final boolean canDefineInstallmentAmount;
     private final Integer installmentAmountInMultiplesOf;
+	private final boolean canAutoAllocateOverpayments;
```
nazeer1100126, line 120: I believe, when canAutoAllocateOverpayments is enabled from loan product definition, then we need to mandate linked savings account for the loans created from this product

nazeer1100126 (changes requested):

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
ippezshelby: I thought 
if(loanIds != null){ 
Will handle this exception

nazeer1100126: Assume a case with over paid loans which doesn't have linked savings account. That time accountAssociations will be null right?

PR #354, opened 2017-05-18
FINRACT-456 Javadoc generation possible now

raj140916:
With these changes, it would be possible to generate javadocs for the project using gradle javadoc task


nazeer1100126:
@raj140916 thank you

PR #362, opened 2017-06-07
Initial Version of AdHocQuery

iamrupok:
Please Review the code for initial version of the API AdHocQuery

### DIFF
--- a/fineract-db/adhoq.sql
+++ b/fineract-db/adhoq.sql
@@ -1,0 +1,15 @@
+CREATE TABLE `m_adhoc` (
+	`id` BIGINT(20) NOT NULL AUTO_INCREMENT,
+	`name` VARCHAR(100) NULL DEFAULT NULL,
+	`query` VARCHAR(2000) NULL DEFAULT NULL,
+	`table_name` VARCHAR(100) NULL DEFAULT NULL,
+	`table_fields` VARCHAR(1000) NULL DEFAULT NULL,
+	`IsActive` TINYINT(1) NOT NULL DEFAULT '0',
+	`created_date` DATETIME NULL DEFAULT NULL,
+	`createdby_id` BIGINT NOT NULL,
+	PRIMARY KEY (`id`),
+	CONSTRAINT `Created_by` FOREIGN KEY (`Created_by`) REFERENCES `m_appuser` (`id`)
+)
+COLLATE='latin1_swedish_ci'
+ENGINE=InnoDB
+;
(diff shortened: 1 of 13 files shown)


fineract-db/adhoq.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE `m_adhoc` (
```
nazeer1100126, line 1: This migration script should be added under fineract-provider\src\main\resources\sql\migrations\core_db. and check the existing migration scripts for filename format

fineract-provider/src/main/java/org/apache/fineract/commands/service/CommandWrapperBuilder.java
```diff
@@ -2868,4 +2868,43 @@ public CommandWrapperBuilder deleteSmsCampaign(final Long resourceId) {
...
+        this.actionName = "CREATE";
+        this.entityName = "ADHOC";
+        this.href = "/adhoc/template";
+        return this;
+    }
+    public CommandWrapperBuilder updateAdHoc(final Long adHocId) {
+        this.actionName = "UPDATE";
+        this.entityName = "ADHOC";
+        this.entityId = adHocId;
+        this.href = "/roles/" + adHocId;
```
nazeer1100126, line 2898: why roles here? it should be adhoc?

fineract-provider/src/main/java/org/apache/fineract/adhocquery/api/AdHocApiResource.java
```diff
@@ -0,0 +1,194 @@
...
+
+@Path("/AdhocQUery")
+@Component
+@Scope("singleton")
+public class AdHocApiResource {
+
+    /**
+     * The set of parameters that are supported in response for {@link AdhocData}
+     */
+    private final Set<String> RESPONSE_DATA_PARAMETERS = new HashSet<>(Arrays.asList("id", "name", "query", "table_name","table_field"));
```
nazeer1100126, line 62: RESPONSE_DATA_PARAMETERS should be id, name, query, tableName, tableFields, isActive

nazeer1100126 (changes requested):

iamrupok:
Thank you for you feedback i will update accordingly.by the way
i have added my beans appContext.xml files no error during tomcat start up
but when i try to call the api with rest client it says
not found.but other serivce are working .Any suggestion will be great help
for me

nazeer1100126:
You need to modify appContext.xml. You need to add this new package to bean scan and you need to add repository package also. Please include this xml as part of next PR

nazeer1100126:
@iamrupok close this PR

PR #366, opened 2017-06-12
Allow predefined SMS messages and Emails to be sent via existing SMS/Email implementations

alexivanov:
Allow SMS messages without SMS campaign to be saved

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/GmailBackedPlatformEmailService.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/GmailBackedPlatformEmailService.java
@@ -39,7 +39,25 @@ public class GmailBackedPlatformEmailService implements PlatformEmailService {
 	}
 
     @Override
-    public void sendToUserAccount(final EmailDetail emailDetail, final String unencodedPassword) {
+    public void sendToUserAccount(String organisationName, String contactName,
+                                  String address, String username, String unencodedPassword) {
+
+	    final String subject = "Welcome " + contactName + " to " + organisationName;
+	    final String body = "You are receiving this email as your email account: " +
+                address + " has being used to create a user account for an organisation named [" +
+                organisationName + "] on Mifos.\n" +
+                "You can login using the following credentials:\nusername: " + username + "\n" +
+                "password: " + unencodedPassword + "\n" +
+                "You must change this password upon first log in using Uppercase, Lowercase, number and character.\n" +
+                "Thank you and welcome to the organisation.";
+
+	    final EmailDetail emailDetail = new EmailDetail(subject, body, address, contactName);
+	    sendDefinedEmail(emailDetail);
+
+    }
+
+    @Override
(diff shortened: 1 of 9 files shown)


avikganguly01 (approved):
Code looks good. 

@Project Maintainers,
Can this be merged on priority?

fineract-provider/src/main/resources/sql/migrations/core_db/V327__sms-messages-without-campaign.sql
```diff
@@ -0,0 +1,20 @@
+--
+-- Licensed to the Apache Software Foundation (ASF) under one
```
nazeer1100126: Change this migration script number to 329

nazeer1100126 (commented):
Can you share some details on the use case without having campaign definition please.

alexivanov:
One use case that I see is sending one time passwords for two-factor authentication via SMS/Email.

nazeer1100126:
@MegaAlex Currently provider id is fetched from message-gateway component while creating a sms campaign. So how does this provider id is configured with respect to the use case mentioned by you.

alexivanov:
@nazeer1100126 In regards to two-factor authentication, I was thinking something along the lines of using global configurations or a configuration endpoint just for two-factor. Ultimately the goal of this PR is to make the SMS & Email more generic so they can be used not only for login email and sms campaigns.

As for the SMS service, setup of tenant & sms bridges in the message gateway directly from Fineract is part of my GSoC proposal but it should be discussed further.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/data/EmailData.java
```diff
@@ -0,0 +1,51 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.infrastructure.core.data;
+
+public class EmailData {
```
nazeer1100126: Do we require this? I see EmailDetail present. Refactor existing class to support subject, body.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/GmailBackedPlatformEmailService.java
```diff
@@ -80,4 +81,34 @@ public void sendToUserAccount(final EmailDetail emailDetail, final String unenco
             throw new PlatformEmailSendException(e);
         }
     }
+
+    @Override
+    public void sendDefinedEmail(EmailData emailDetails) {
```
nazeer1100126: Modify the existing method sendToUserAccount to make use of this method instead of having duplicate implementation.

nazeer1100126 (changes requested):

alexivanov:
PR is updated.

PR #370, opened 2017-06-15
adhoc query

iamrupok:


fineract-provider/src/main/java/org/apache/fineract/adhocquery/domain/AdHoc.java
```diff
@@ -0,0 +1,148 @@
...
+        this.query=StringUtils.defaultIfEmpty(query,null);
+        this.tableName=StringUtils.defaultIfEmpty(tableName,null);
+        this.tableFields=StringUtils.defaultIfEmpty(tableFields,null);
+        this.email=StringUtils.defaultIfEmpty(email,null);
+        this.isActive = BooleanUtils.toBooleanDefaultIfNull(isActive, false);
+       
+    }
+    public static AdHoc fromJson(final JsonCommand command) {
+        final String name = command.stringValueOfParameterNamed(AdHocJsonInputParams.NAME.getValue());
+        final String query = command.stringValueOfParameterNamed(AdHocJsonInputParams.QUERY.getValue());
```
nazeer1100126, line 71: Do we need to validate this query string not to have cases where authorized users sending malicious SQL statements?

iamrupok:
Do you suggest adding an enum for DDL and then do the checking  using enum as we need DML
support only

planning to add
final String sqlQuery = this.fromApiJsonHelper.extractStringNamed("query",
element);

if(sqlQuery.toLowerCase().contains("drop")||sqlQuery.toLowerCase().contains("delete")){
            final ApiParameterError error =
ApiParameterError.parameterError("validation.msg.unsupported.sql", "Delete
and Drop Not Supported",
                      "query", sqlQuery);
            dataValidationErrors.add(error);
        }
in AdHocDataValidator class for create and update do you think this would
be suffiient

nazeer1100126:
@iamrupok You need to check whether DDL and DML commands are allowed are not. I don't know the real use case how are you going to use, it's up to you to decide. Please find the DDL and DML commands below.
DDL_COMMANDS = { "create", "drop", "alter", "truncate", "comment" },
DML_COMMANDS = { "select", "insert", "update", "delete", "merge", "upsert", "call" };

nazeer1100126:
@iamrupok please let me know if you need more details.

iamrupok:
created a new PR 395

PR #374, opened 2017-06-21
Two-Factor Authentication

alexivanov:
This PR was done as part of GSoC 2017 [project](https://summerofcode.withgoogle.com/projects/#5967202743222272) - Two-Factor Authentication.

Project spec located [here](https://mifosforge.jira.com/wiki/spaces/projects/pages/185277689/GSoC+2017+-+Two-Factor+Authentication).

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/commands/service/CommandWrapperBuilder.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/commands/service/CommandWrapperBuilder.java
@@ -2929,7 +2929,7 @@ public class CommandWrapperBuilder {
         this.href = "/smscampaigns/"+resourceId;
         return this;
     }
-    
+
     public CommandWrapperBuilder holdAmount(final Long accountId) {
         this.actionName = "HOLDAMOUNT";
         this.entityName = "SAVINGSACCOUNT";
@@ -3039,6 +3039,20 @@ public class CommandWrapperBuilder {
         this.entityId = adHocId;
         this.href = "/adhocquery/" + adHocId;
         this.json = "{}";
+        return this;
+    }
+
+    public CommandWrapperBuilder invalidateTwoFactorAccessToken() {
+        this.actionName = "INVALIDATE";
+        this.entityName = "TWOFACTOR_ACCESSTOKEN";
+        this.href = "/twofactor/invalidate";
+        return this;
+    }
+
+    public CommandWrapperBuilder updateTwoFactorConfiguration() {
+        this.actionName = "UPDATE";
+        this.entityName = "TWOFACTOR_CONFIGURATION";
+        this.href = "/twofactor/configure";
         return this;
     }
 }
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/data/TwoFactorConfigurationValidator.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/data/TwoFactorConfigurationValidator.java
@@ -1,0 +1,120 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.infrastructure.security.data;
+
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.commons.lang.StringUtils;
+import org.apache.fineract.infrastructure.core.data.ApiParameterError;
+import org.apache.fineract.infrastructure.core.data.DataValidatorBuilder;
+import org.apache.fineract.infrastructure.core.exception.InvalidJsonException;
+import org.apache.fineract.infrastructure.core.exception.PlatformApiDataValidationException;
+import org.apache.fineract.infrastructure.core.serialization.FromJsonHelper;
+import org.apache.fineract.infrastructure.security.constants.TwoFactorConfigurationConstants;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Profile;
+import org.springframework.stereotype.Component;
+
+import com.google.gson.JsonElement;
+import com.google.gson.reflect.TypeToken;
+
+@Component
+@Profile("twofactor")
+public class TwoFactorConfigurationValidator {
+
+    private final FromJsonHelper fromJsonHelper;
+
+    @Autowired
+    public TwoFactorConfigurationValidator(FromJsonHelper fromJsonHelper) {
+        this.fromJsonHelper = fromJsonHelper;
+    }
+
+    public void validateForUpdate(final String json) {
+        if (StringUtils.isBlank(json)) {
+            throw new InvalidJsonException();
+        }
+
+        boolean atLeastOneParameterPassedForUpdate = false;
+        final Type typeOfMap = new TypeToken<Map<String, Object>>() {}.getType();
+        this.fromJsonHelper.checkForUnsupportedParameters(typeOfMap, json,
+                TwoFactorConfigurationConstants.REQUEST_DATA_PARAMETERS);
+        final JsonElement element = this.fromJsonHelper.parse(json);
+
+        final List<ApiParameterError> dataValidationErrors = new ArrayList<>();
+        final DataValidatorBuilder baseDataValidator = new DataValidatorBuilder(dataValidationErrors)
+                .resource(TwoFactorConfigurationConstants.RESOURCE_NAME);
+
+
+        for(String parameterName : TwoFactorConfigurationConstants.BOOLEAN_PARAMETERS) {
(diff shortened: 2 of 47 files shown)


nazeer1100126:
@MegaAlex Send PR with single commit and please resolve the conflicts

alexivanov:
@nazeer1100126 ready for review

fineract-provider/src/main/java/org/apache/fineract/commands/service/CommandWrapperBuilder.java
```diff
@@ -3041,4 +3041,18 @@ public CommandWrapperBuilder deleteAdHoc(Long adHocId) {
         this.json = "{}";
         return this;
     }
+
+    public CommandWrapperBuilder invaldiateTwoFactorAccessToken() {
```
nazeer1100126: invaldiateTwoFactorAccessToken spelling mistake.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/PlatformEmailService.java
```diff
@@ -26,5 +26,4 @@ void sendToUserAccount(String organisationName,String contactName,
                            String address, String username, String unencodedPassword);
 
     void sendDefinedEmail(EmailDetail emailDetails);
```
nazeer1100126: there is no change in this file , please remove this file

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/domain/TFAccessToken.java
```diff
@@ -0,0 +1,137 @@
...
+    private Date validTo;
+
+    @Column(name = "enabled", nullable = false)
+    private boolean enabled;
+
+    public TFAccessToken() {
+    }
+
+    public static TFAccessToken create(String token, AppUser user, int tokenLiveTimeInSec) {
+        DateTime validFrom = new DateTime();
```
nazeer1100126: make use of DateUtils to get tenant date & time.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/service/TwoFactorConfigurationServiceImpl.java
```diff
@@ -0,0 +1,293 @@
...
+        final Map<String, Object> templateData = processTemplateDataFor(user, request);
+        return compileTextTemplate(getSmsText(), TwoFactorConstants.SMS_DELIVERY_METHOD_NAME, templateData);
+    }
+
+    @Override
+    public Integer getOTPTokenLength() {
+        Integer defaultValue = 5;
+        Integer value = getIntegerConfig(TwoFactorConfigurationConstants.OTP_TOKEN_LENGTH,
+                defaultValue);
+        if(value < 1) {
```
nazeer1100126: instead of checking here for value less than 1 , from getInteger method only return default value

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/service/TwoFactorServiceImpl.java
```diff
@@ -0,0 +1,221 @@
...
+            }
+            final OTPRequest request = generateNewToken(emailDelivery, extendedAccessToken);
+            final String emailSubject = configurationService.getFormattedEmailSubjectFor(user, request);
+            final String emailBody = configurationService.getFormattedEmailBodyFor(user, request);
+            final EmailDetail emailData = new EmailDetail(emailSubject, emailBody, user.getEmail(),
+                    user.getFirstname() + " " + user.getLastname());
+            emailService.sendDefinedEmail(emailData);
+            otpRequestRepository.addOTPRequest(user, request);
+            return request;
+        } else {
```
nazeer1100126: unnecessary else block.

fineract-provider/src/main/resources/META-INF/spring/securityContext.xml
```diff
@@ -42,19 +42,26 @@
...
+			<intercept-url pattern="/api/*/twofactor" access="isFullyAuthenticated()"
+						   method="GET" requires-channel="https" />
+			<intercept-url pattern="/api/*/twofactor" access="isFullyAuthenticated()"
+						   method="POST" requires-channel="https" />
+			<intercept-url pattern="/api/*/twofactor/validate" access="isFullyAuthenticated()"
+						   method="POST" requires-channel="https" />
+			<intercept-url pattern="/api/**" access="isFullyAuthenticated() and hasAuthority('TWOFACTOR_AUTHENTICATED')"
 				method="GET" requires-channel="https" />
-			<intercept-url pattern="/api/**" access="isFullyAuthenticated()"
+			<intercept-url pattern="/api/**" access="isFullyAuthenticated() and hasAuthority('TWOFACTOR_AUTHENTICATED')"
```
nazeer1100126, line 53: for pattern="/api/*/twofactor , why we have only isFullyAuthenticated() access?
 for other pattern like  pattern="/api/** we have access access="isFullyAuthenticated() and hasAuthority('TWOFACTOR_AUTHENTICATED') ?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/data/TwoFactorConfigurationValidator.java
```diff
@@ -0,0 +1,120 @@
...
+        }
+    }
+
+    private void validateBooleanParameter(final String name, final JsonElement element,
+                  final DataValidatorBuilder baseDataValidator) {
+        final String value = this.fromJsonHelper.extractStringNamed(name, element);
+        baseDataValidator.reset().parameter(name).value(value).notNull().trueOrFalseRequired(value);
+    }
+
+    private void validateStringParameter(final String name, final JsonElement element,
```
nazeer1100126, line 109: with this approach of validation, we can not validate on length property.
if for any parameter if have max length 50 , then it will hit to database.

fineract-provider/src/main/resources/META-INF/spring/securityContext.xml
alexivanov: In the implementation I am proposing, I look at 2fa as an authorisation problem. Users that are two-factor authenticated are granted the _TWOFACTOR_AUTHENTICATED_ authority. Endpoints under _/twofactor_ and _/twofactor/validate_ don't require fully authenticated uses, thus no need to check for TWOFACTOR_AUTHENTICATED.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/data/TwoFactorConfigurationValidator.java
alexivanov: I am not sure that I understand you completely. AFAICS the values are correctly verified for length > 1000. Do we need to verify the keys also as they are hard-coded?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/service/TwoFactorServiceImpl.java
alexivanov: Without the else, no error would be returned if a request for OTP is sent for undefined delivery method. Would you like me to do the validation somewhere else?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/data/TwoFactorConfigurationValidator.java
nazeer1100126: I mean to say, if i have two string parameters and 100 & 200 is max length allowed for them, it is not possible with this approach right?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/service/TwoFactorServiceImpl.java
nazeer1100126: I mean to say, you no need to have else block. Directly throw new OTPDeliveryMethodInvalidException();

alexivanov: Oh, makes perfect sense, yes :+1:

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/data/TwoFactorConfigurationValidator.java
alexivanov: I validate string parameters here because of the 1000 char limit of the column in the 'key-value' table that is used to store the configs. I haven't really implemented per-key length limits because that would be kind of redundant for the set of this config parameters. I validate them only for the 1000 hard column limit. Sorry if that doesn't make sense.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/domain/TFAccessToken.java
alexivanov: I opted for using timestamps for the expiry times for connivance as the period is not that long. Is using date time more feasible in this case also?

api-docs/api-docs.iml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
avikganguly01: Add iml files to your .gitignore

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/domain/TFAccessToken.java
avikganguly01: @MegaAlex : Point is tenant timezone and hosted server timezone can be different, so use DateUtils.

nazeer1100126:
@MegaAlex does all review comments taken care? If yes, can you send the PR with single commit please.

alexivanov:
@nazeer1100126 comments are addressed, commits are squashed.

Ezcred:
Hi @MegaAlex @nazeer1100126 ,

By when do you think this pull request will be merged ? Any plans of getting otp based authentication ( without using username/password) similar to what uber, whatsapp do.

-
Thanks

PR #375, opened 2017-06-25
FINERACT-437-Remove generic exceptions

ThisuraThejith:
Add well defined exception for exceptions thrown by the fineract to resolve error MITRE, CWE-397 - Declaration of Throws for Generic Exception.


nazeer1100126:
@ThisuraThejith I suggest to have separate exception classes or generic type exceptions (for each sms and journaly entry) which takes errorcode, user message in the constructor.

ThisuraThejith:
@nazeer1100126  Sure will do that. So instead of enums we'll pass the error codes and the message in the constructor when we are throwing them, right?

ThisuraThejith:
@nazeer1100126 I have updated the PR. Please review and let me know if there are any issues.

PR #382, opened 2017-07-12
Notification module

cnangeh:
Work done:

- Work done by Adhyan last year
- User subscription to topic on user creation
- Topic creation on entity creation
- Upgrade process of user notification
- Migration script - topic, topic subscription models for new and already existing users and entities


pranjal-goswami:
Hi @Anh3h 
Can you squash the commits into a single commit for review?

cnangeh:
ok.Will do that right away

PR #398, opened 2017-08-03
family members migration fix

nikpawar89:
Migration fix for family members module.


nazeer1100126:
@nikpawar89 by default AUTO_INCREMENT is 1. So we no need to mention in migration script. 
I will modify this after merging this 👍

PR #436, opened 2017-12-29
FINERACT-416 Interest of the whole term

MexinaD:
@avikganguly01  Can you please preview this PR so that it can be in upcoming release, it is  a long waited feature here in Tanzania.
Its jira ticket is https://issues.apache.org/jira/browse/FINERACT-416

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/AprCalculator.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/AprCalculator.java
@@ -19,14 +19,14 @@
 package org.apache.fineract.portfolio.loanaccount.loanschedule.domain;
 
 import java.math.BigDecimal;
-
+import java.math.RoundingMode;
 import org.apache.fineract.portfolio.common.domain.PeriodFrequencyType;
 import org.springframework.stereotype.Component;
 
 @Component
 public class AprCalculator {
 
-    public BigDecimal calculateFrom(final PeriodFrequencyType interestPeriodFrequencyType, final BigDecimal interestRatePerPeriod) {
+    public BigDecimal calculateFrom(final PeriodFrequencyType interestPeriodFrequencyType, final BigDecimal interestRatePerPeriod,  final Integer numberOfRepayments, final Integer repaymentEvery, final PeriodFrequencyType repaymentPeriodFrequencyType) {
         BigDecimal defaultAnnualNominalInterestRate = BigDecimal.ZERO;
         switch (interestPeriodFrequencyType) {
             case DAYS:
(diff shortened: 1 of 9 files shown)


MexinaD:
@nazeer1100126  Can you check this

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/AprCalculator.java
```diff
@@ -41,10 +41,29 @@ public BigDecimal calculateFrom(final PeriodFrequencyType interestPeriodFrequenc
             case YEARS:
                 defaultAnnualNominalInterestRate = interestRatePerPeriod.multiply(BigDecimal.valueOf(1));
             break;
+            case WHOLE_TERM:
+                        final BigDecimal ratePerPeriod = interestRatePerPeriod.divide(BigDecimal.valueOf(numberOfRepayments*repaymentEvery), 8, RoundingMode.HALF_UP);
+                         
+                         switch (repaymentPeriodFrequencyType) {
+                             case DAYS:
+                                 defaultAnnualNominalInterestRate = ratePerPeriod.multiply(BigDecimal.valueOf(365));
```
avikganguly01, line 49: Shoudn't this consider the no. of days in a year as per product configuration instead of 365?

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/AprCalculator.java
```diff
@@ -41,10 +41,29 @@ public BigDecimal calculateFrom(final PeriodFrequencyType interestPeriodFrequenc
             case YEARS:
                 defaultAnnualNominalInterestRate = interestRatePerPeriod.multiply(BigDecimal.valueOf(1));
             break;
+            case WHOLE_TERM:
+                        final BigDecimal ratePerPeriod = interestRatePerPeriod.divide(BigDecimal.valueOf(numberOfRepayments*repaymentEvery), 8, RoundingMode.HALF_UP);
```
avikganguly01, line 45: Why divide by 8?

avikganguly01 (commented):
@MexinaD : Hey Mexina, sorry for the delay.

Can you confirm couple of things :- 

1. The loan terms can be overriden at the loan account level right? If that is the case, corresponding changes need to be made at the loan account level to allow to override the product term settings. 

2. Have you checked the usage of PeriodFrequencyType in the codebase and found no obvious cases where usage of this new enum value won't work with other product settings? If so, appropriate exceptions need to be thrown for such combinations.

3. Are any of the LMS integration tests failing when you run ./gradlew integrationTest? 

Once you have addressed all the comments, can you enhance the example you provided at https://issues-test.apache.org/jira/browse/FINERACT-416 with expected results?

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/AprCalculator.java
MexinaD: Its not divided by 8 but i rounded off the result of InterestRatePerPeriod divide by (numberOfRepayment*repayedEvery) to 8 us it was applied in another code in Loan.java "final MathContext mc = new MathContext(8, roundingMode);" So i assumed the roundoff that the system is using is 8.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/AprCalculator.java
MexinaD: I followed what was already implemented for "DAYS" frequency from the first switch loop.

MexinaD:
@avikganguly01  Regarding the above cases
1. The feature is to add another interest method, and since the interest method is primary configured in loan product and that's why the changes are made in loan product level, like any other interest methods 

2. I didn't check the use of PeriodFrequencyType to another product settings at the beginning thinking it wasn't necessary as the feature is only applicable to loan product. if the new change will not apply to other product, is there need to deal with other combinations?

3. I don't have a knowledge of LMS integrations but when i run that command there are some tests failing and other skipped.

MexinaD:
@avikganguly01 Note: This is an improvement

MexinaD:
@avikganguly01 Is there anything to modify?

MexinaD:
Hello @ShruthiRajaram Can you help to review this too

PR #441, opened 2018-02-08
FINERACT-574 

Nenge1:
This pull request fix the bug of duplicate posting in journal entries when share account is rejected


ShruthiRajaram:
@Nenge1 While reviewing the PR, I observed few other scenarios which I mentioned in the JIRA issue.
Those needs to be addressed before fixing the current issue

PR #442, opened 2018-02-16
Fix for In-App Triggered Notifications

conradsp:
For a triggered SMS message, check if notification flag is set. If so, send to GCM service.


nazeer1100126:
@conradsp Changes Looks OK. Please send PR with single commit, so that any committer can merge this code.

conradsp:
Will re-submit PR with single change

PR #446, opened 2018-02-25
FINERACT-546 - remove group from center fix

viswaramamoorthy:
1. Center from which group being disassociated is really stored as parent of the group. Introduced a parent check method to fix disassociation of groups

2. Originally disassociateGroups command was returning result as 'clientMembers' - assuming this is typo. Fixed to return result as groupMembers
3. Updated test case CenterIntegrationTest to validate


ShruthiRajaram (changes requested):
@viswaramamoorthy  you need to remove group from a center. And newly added method 'hasGroupAsParent ' is not a valid solution. I strongly believe there is a JPA annotation problem. Can you look into that?

PR #453, opened 2018-05-11
FINERACT-610 - Post Interest As On Permision

MexinaD:


ShruthiRajaram:
As it is related to front end, closing the PR

MexinaD:
Hello @ShruthiRajaram 
The issue is in back end and not front end, since the permissions are created by schema and not through front end.

ShruthiRajaram:
Hi @MexinaD , I understand that permissions are stored in db. But the logic for the post as on date is present under a different class with different permission. 
And the permission which were referred in the front end was invalid as there is no handler for that permission. So we corrected the permission in the front end to work properly.

MexinaD:
Hello @ShruthiRajaram 
Thank you for that clarifications

PR #454, opened 2018-05-25
FINERACT-623 - Restriction of product mix fix

MexinaD:


ShruthiRajaram (changes requested):
Instead of having duplicate code in "LoanApplicationWritePlatformServiceJpaRepositoryImpl"  from "LoanWritePlatformServiceJpaRepositoryImpl". I suggest you to change the method in LoanWritePlatformServiceJpaRepositoryImpl to public static and use the same.  Reason being, if a developer changes one of the methods and forgets to do the same change in other, there will be code inconsistency. so..
And as far as the Exception class is considered, use" GeneralPlatformDomainRuleException" and throw a proper error message.

MexinaD:
Hello @ShruthiRajaram , thank you for your comments, am working on them

MexinaD:
Hello @ShruthiRajaram 
I have already updated the PR with change of implementation, if the loan products should not co-exist, its enough to be checked on the time of application, please review

ShruthiRajaram:
@MexinaD  I think there is a little confusion over my last comment. I was completely fine with validation during disbursal and submission. All I wanted to convey was, change the method signature of "checkForProductMixRestrictions" and "checkForProductMixRestrictions" present in LoanWritePlatformServiceJpaRepositoryImpl, so that you can have a common method and call it from  "LoanApplicationWritePlatformServiceJpaRepositoryImpl" and "LoanWritePlatformServiceJpaRepositoryImpl". So that you handle all the scenarios you described previously. I believe the current change checks only during submission. If you want to check only during the submission and not during disbursal for existing loans, let me know I will approve the change.

MexinaD:
Hello @ShruthiRajaram I think if the loan has already passed the state of approval that means the user was okay with their co-existence at the time he/she was approving the loan, So i want the check to be on submission

ShruthiRajaram (approved):

PR #466, opened 2018-07-31
FINERACT:628 API support for self share accounts

ShruthiRajaram:


ShruthiRajaram:
@conradsp kindly review this PR. My other ticket is dependent on the newly added file in this PR.  And there will be conflict if I create same files again. Thank you

PR #482, opened 2018-10-10
Fix Addresses Path

drarko:
Path for addresses resource is inconsistent
/client/{clientId}/addresses

Other resources are plurals:
/clients/{clientId}/charges
/clients/{clientId}/identifiers
/clients/{clientId}/accounts

Change addresses path to:
/clients/{clientId}/addresses

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/client/api/ClientAddressApiResources.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/client/api/ClientAddressApiResources.java
@@ -49,7 +49,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Scope;
 import org.springframework.stereotype.Component;
 
-@Path("/client")
+@Path("/clients")
 @Component
 @Scope("singleton")
 public class ClientAddressApiResources {


alexivanov (changes requested):
Nice catch :+1: 
This is a breaking API change for a feature added long time ago requiring a wider discussion. If it is decided to go with your proposal, we need to make sure we allow enough time for users to migrate(e.g deprecate first).

vorburger:
@drarko thank you very much for your interest in and contribution to Fineract/Mifos!  I'm an (old) volunteer who got reminded of Mifos/Fineract this week-end at the FOSDEM conference, and I'm taking a moment to help the project clean up some of it's old issues and pull requests etc.  

We were wondering whether you are still alive, and what you thought about @alexivanov review comment... perhaps it's too late to change this public API, and this will create a lot more confusion and break external REST clients - would you be willing to abandon this PR yourself and just close it? Please don't let this discourage you - we look very much forward to more contributions from you!

If we don't hear from you within 2 weeks, we will take the liberty to close this PR without merging it, to avoid confusion for future new contributors looking at this repository - hope you understand and that's fair.

/Cc FYI @apache/fineract-committers

drarko:
Totally understand :+1:

PR #487, opened 2018-11-02
FINERACT-656 unable to import client via bulk import when address is …

ShruthiRajaram:
…enabled

When multiple address addition  during client creation was introduced, client creation using excel sheet was not handled


raakeshtomarnew:
When i import client data its show sucess but no client imported.

ShruthiRajaram:
@raakeshtomarnew kindly refer https://github.com/apache/fineract/pull/488 PR and when you try to reproduce the scenario kindly enable address configuration.

raakeshtomarnew:
Not Solved same problem

PR #488, opened 2018-11-02
FINERACT-656 and FINERACT-655 unable to import client via bulk import…

ShruthiRajaram:
… when address is enabled


vinay002-dotcom (commented):
after merged this code i'm not able to get drop down list for address data

PR #489, opened 2018-11-05
FINERACT-659 Account Id issue while importing loan transaction

ShruthiRajaram:


MexinaD:
Hello @ShruthiRajaram 
It seems you have handled during uploading the system to allow the Account id with Long datatype so as to accommodate the big number.

But i have see other case where if the loan accounts have account preferences the system don't allow to download since they are in a format like "MFL000000021"  which is not a number.

I think its better to handle these issues together.

ShruthiRajaram:
@MexinaD thanks for the input. I will try to include that modification as well

ShruthiRajaram:
Hi @MexinaD I have made changes that fixes both download and upload of bulk loan repayment transaction (with/without account number preferences enabled) . Kindly review.

MexinaD:
Hello @ShruthiRajaram
I tested and it worked fine.

ShruthiRajaram:
@MexinaD Can you review and provide feedback/approve the PR?

MexinaD:
Reviewed and can be merged

PR #510, opened 2019-01-17
Thread.run()->Thread.start()

SunnyDay74:
Hello,
To utilize the multiple-thread of Java, is it good to use Thread.start() instead of Thread.run()?

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/jobs/service/SchedulerStopListener.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/jobs/service/SchedulerStopListener.java
@@ -72,7 +72,7 @@ public class SchedulerStopListener implements JobListener {
                     SchedulerStopListener.this.jobRegisterService.stopScheduler(schedulerName);
                 }
             });
-            newThread.run();
+            newThread.start();
         }
     }
 


vorburger:
@SunnyDay74 thank you very much for your interest in and contribution to Fineract/Mifos!  I'm an (old) volunteer who got reminded of Mifos/Fineract this week-end at the FOSDEM conference, and I'm taking a moment to help the project clean up some of it's old issues and pull requests etc.  

>  To utilize the multiple-thread of Java, is it good to use Thread.start() instead of Thread.run()?

They don't do the same thing, see the JavaDoc of [start](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Thread.html#start())() VS [run](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Thread.html#run())() .. I think you may have stumbled upon a small real issue here: Given that how it is now, using run() it doesn't actually really start a thread anyway, these few lines are unnecessarily complex - I think you could just remove the entire `new Thread(new Runnable() {` and simply do that line which is inside `run()` right after the if (without Thread), do you see why, and would you be willing to update this PR with that? We would love to merge a contribution from you as a First-time contributor - and hope to see more from you!

If we don't hear from you within 2 weeks, we will take the liberty to close this PR without merging it, to avoid confusion for future new contributors looking at this repository - hope you understand and that's fair.

vorburger (changes requested):
see comment in main conversation

vishwasbabu:
@SunnyDay74 : Closing this pr as per @vorburger's comments.

PR #511, opened 2019-01-21
Fix bug in generating notification.

cnangeh:
When querying the users required to get a notification, add new topic subscribers
to the current set of subscribers, instead of replacing the currents set of subscribers
with the new topic subscribers.


santoshconflux:
Related ticket: https://issues.apache.org/jira/browse/FINERACT-695

rahulg963 (approved):

santoshconflux:
@rahulg963 , Can you merge it?

PR #275, opened 2017-01-16
[MIFOSX-2114] Change assignment staff to group and clients in center.

k4wel:

PR #514, opened 2019-02-08
[MIFOSX-2114] Change assignment staff to groups and clients in center.

k4wel:
Recreated from #275.

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/client/service/ClientWritePlatformServiceJpaRepositoryImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/client/service/ClientWritePlatformServiceJpaRepositoryImpl.java
@@ -247,9 +247,15 @@ public class ClientWritePlatformServiceJpaRepositoryImpl implements ClientWriteP
             }
 
             Staff staff = null;
-            final Long staffId = command.longValueOfParameterNamed(ClientApiConstants.staffIdParamName);
-            if (staffId != null) {
-                staff = this.staffRepository.findByOfficeHierarchyWithNotFoundDetection(staffId, clientOffice.getHierarchy());
+            if (clientParentGroup != null) {
+                if (clientParentGroup.getParent() != null) {
+                    staff = clientParentGroup.getParent().getStaff();
+                }
+            } else {
+                final Long staffId = command.longValueOfParameterNamed(ClientApiConstants.staffIdParamName);
+                if (staffId != null) {
+                    staff = this.staffRepository.findByOfficeHierarchyWithNotFoundDetection(staffId, clientOffice.getHierarchy());
+                }
             }
 
             CodeValue gender = null;
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/group/service/GroupingTypesWritePlatformServiceJpaRepositoryImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/group/service/GroupingTypesWritePlatformServiceJpaRepositoryImpl.java
@@ -158,9 +158,15 @@ public class GroupingTypesWritePlatformServiceJpaRepositoryImpl implements Group
             validateOfficeOpeningDateisAfterGroupOrCenterOpeningDate(groupOffice, groupLevel, activationDate);
 
             Staff staff = null;
-            final Long staffId = command.longValueOfParameterNamed(GroupingTypesApiConstants.staffIdParamName);
-            if (staffId != null) {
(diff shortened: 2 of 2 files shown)


vorburger:
@TypicalGitHubUser would you mind to re-rebase this, so that Travis CI will run on it?

@nazeer1100126 will you re-review this as requested in #275 ?

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/service/ClientWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -247,9 +247,15 @@ public CommandProcessingResult createClient(final JsonCommand command) {
...
             Staff staff = null;
-            final Long staffId = command.longValueOfParameterNamed(ClientApiConstants.staffIdParamName);
-            if (staffId != null) {
-                staff = this.staffRepository.findByOfficeHierarchyWithNotFoundDetection(staffId, clientOffice.getHierarchy());
+            if (clientParentGroup != null) {
+                if (clientParentGroup.getParent() != null) {
+                    staff = clientParentGroup.getParent().getStaff();
+                }
+            } else {
+                final Long staffId = command.longValueOfParameterNamed(ClientApiConstants.staffIdParamName);
```
vishwasbabu, line 255: If the Client has a parent group, but the parent group does not have a center, the else condition would not execute and staffId passed in through the API call would be lost. You would now end up with a Client with no staff assigned when the API caller had explicitly passed in a staffId

Ideally, the behavior should be such that.

1) If the staffId is passed in through the API call, set the same to the client
2) If not, inherit the staffId from the client's parent (if the client has a parent)

fineract-provider/src/main/java/org/apache/fineract/portfolio/group/service/GroupingTypesWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -158,9 +158,15 @@ private CommandProcessingResult createGroupingType(final JsonCommand command, fi
             validateOfficeOpeningDateisAfterGroupOrCenterOpeningDate(groupOffice, groupLevel, activationDate);
 
             Staff staff = null;
-            final Long staffId = command.longValueOfParameterNamed(GroupingTypesApiConstants.staffIdParamName);
-            if (staffId != null) {
-                staff = this.staffRepository.findByOfficeHierarchyWithNotFoundDetection(staffId, groupOffice.getHierarchy());
+            if ((GroupTypes.GROUP).equals(groupingType)) {
+                if (parentGroup != null) {
+                    staff = parentGroup.getStaff();
```
vishwasbabu, line 163: Same issue as the previous comment

fineract-provider/src/main/java/org/apache/fineract/portfolio/group/service/GroupingTypesWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -388,11 +394,16 @@ private CommandProcessingResult updateGroupingType(final Long groupId, final Jso
...
 
             if (actualChanges.containsKey(GroupingTypesApiConstants.staffIdParamName)) {
-                final Long newValue = command.longValueOfParameterNamed(GroupingTypesApiConstants.staffIdParamName);
-
                 Staff newStaff = null;
-                if (newValue != null) {
-                    newStaff = this.staffRepository.findByOfficeHierarchyWithNotFoundDetection(newValue, groupHierarchy);
+                if ((GroupTypes.GROUP).equals(groupingType)) {
+                    if (groupForUpdate.getParent() != null) {
+                        newStaff = groupForUpdate.getParent().getStaff();
```
vishwasbabu, line 400: Here too

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/service/ClientWritePlatformServiceJpaRepositoryImpl.java
vishwasbabu, line 255: Please add an integration test case which validates the functional changes made

vishwasbabu (changes requested):

vishwasbabu:
@TypicalGitHubUser : Please let us know if you have the time to address these review comments anytime in the near future. If not, we shall be closing this PR on the 15th of March in an effort to help committers have a clearer view of pending pull requests. Hope you understand and think this is fair.

k4wel:
@vishwasbabu No, I won't be able to do this in near future.

TypicalGitHubUser:
@vishwasbabu No, I won't be able to do this in near future.

vishwasbabu:
Closing this PR

PR #520, opened 2019-02-13
fix missing & bad license headers and config (req. for TravisCI FINERACT-697)

vorburger:
https://issues.apache.org/jira/browse/FINERACT-697


vorburger:
@avikganguly01 @vishwasbabu hope it's OK if I "self merge" this trivial change... req. for FINERACT-697.

PR #524, opened 2019-02-15
FINERACT-428 Parallelization of Jobs : Parallelizing and paging of recalculate interest for loans

wkk91193:
Opening a new PR, closing #469 after Squashing the commits


fineract-provider/src/main/java/org/apache/fineract/infrastructure/jobs/domain/JobParameter.java
```diff
@@ -0,0 +1,84 @@
...
+    private String parameterValue;
+
+
+    public JobParameter(final Long jobId, final String parameterName, final String parameterValue) {
+        this.jobId = jobId;
+        this.parameterName = parameterName;
+        this.parameterValue = parameterValue;
+    }
+
+    public static JobParameter getInstance(final Long jobId, final String parameterName, final String parameterValue){
```
vorburger, line 48: this seems redundant with the public constructor? Not a huge deal, and I don't mind merging it even with this, but just thought I would let you know anyway.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/jobs/domain/JobParameter.java
```diff
@@ -0,0 +1,84 @@
...
+        return parameterValue;
+    }
+
+    public void setParameterValue(final String parameterValue) {
+        this.parameterValue = parameterValue;
+    }
+
+
+    @Override
+    public boolean equals(Object obj) {
```
vorburger, line 78: it is almost always a bug to only override `equals` and not `hashCode` ... Google it for more background. May I suggest that you either remove this `equals` method or also implement `hashCode`? I would prefer this not getting merged before that is addressed. Sorry I did not notice this before. PS: https://issues.apache.org/jira/browse/FINERACT-702

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/RecalculateInterestPoster.java
```diff
@@ -0,0 +1,152 @@
...
+                            sb.append("Interest recalculation for loans failed " + exception.getMessage()) ;
+                            break;
+                        }
+                    } catch (Exception e) {
+                        Throwable realCause = e;
+                        if (e.getCause() != null) {
+                            realCause = e.getCause();
+                        }
+                        logger.error("Interest recalculation for loans failed for account:"	+ loanId + " with message "
+                                + realCause.getMessage());
```
vorburger: add `, e` here, so that the full stack trace of the problem is logged, so:

    logger.error("Interest recalculation for loans failed for account:"	+ loanId + " with message " + realCause.getMessage(), e);

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/RecalculateInterestPoster.java
```diff
@@ -0,0 +1,152 @@
...
+                    }
+                    i++;
+                }
+                logger.info("Loans count " + i);
+            }
+            if (sb.length() > 0) {
+                try {
+                    throw new JobExecutionException(sb.toString());
+                } catch (JobExecutionException e) {
+                    logger.info("JobExecutionException occured :"+e.toString());
```
vorburger: use `,e` instead of `e.toString()` here, so:

    logger.info("JobExecutionException occured :", e);

also, unless there is a good reason which I'm missing, don't you want to use `logger.error` instead of `logger.info` in this case?

vorburger (changes requested):

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanSchedularServiceImpl.java
```diff
@@ -184,4 +191,133 @@ public void recalculateInterest() throws JobExecutionException {
...
+		// gets the loanIds data set iteratively and call addAccuruals for that paginated dataset
+		do {
+			int totalFilteredRecords = loanIds.size();
+			logger.info("Starting accrual - total filtered records - " + totalFilteredRecords);
+			recalculateInterest(loanIds, threadPoolSize, batchSize,
+					executorService);
+			maxLoanIdInList+=pageSize+1;
+			loanIds = this.loanReadPlatformService
+					.fetchLoansForInterestRecalculation(pageSize, maxLoanIdInList, officeHierarchy);
+		} while (!CollectionUtils.isEmpty(loanIds));
```
vorburger, line 235: are you sure that this is properly thread safe? Hopefully the `List<Long> loanIds` is not just a normal `ArrayList` or so, because then it's not.. it would have to be a something like a `CopyOnWriteArrayList` or `ConcurrentSkipListSet` or so for this to be thread safe. Ihave mot looked more into it, just wanted to point it out for you.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/jobs/domain/JobParameter.java
wkk91193: @vorburger  using getInstance() method I'm reusing the logic of instantiating job parameter instance. Shall we keep this?

vorburger: @wkk91193 sure sure, this point is the most minor of my feedback - you can ignore this point.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanSchedularServiceImpl.java
vorburger: make sure that `List<Long> loanIds` is something like `Collections.synchronizedList(new ArrayList())` should be fine as well.

wkk91193:
@vorburger @vishwasbabu  Requested changes made.Let me know if further changes are neccessary

vishwasbabu:
@wkk91193 : The travis build is failing.

vorburger:
> @wkk91193 : The travis build is failing.

Two separate problems here, I think:

1. org.apache.fineract.integrationtests.SchedulerJobsTestResults > testUpdateLoanArrearsAgingJobOutcome  seems to fail with this change?

2. it's confusing, due to a more "infra" issue, see https://jira.apache.org/jira/browse/FINERACT-732 and note #547 and rebase this when that is on master to see more clearly...

vorburger:
Let's close this now, just to have less open PRs; you can open a new PR with this when you resolved the feedback and conflicts.

vorburger:
@wkk91193 I just noticed, after closing it, that you pushed updates to this just 2 days ago, sorry I hadn't seen this before, so it look like you are actually working on this? Then we'll of course immediately re-open it. FYI the build currently fails again, but #657 will fix that ASAP. If you could rebase it after that, we can see if this passes.

awasum:
@wkk91193 , Please check the Travis CI logs. there is an error somewhere which can be fixed.

```provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanSchedularServiceImpl.java:202: error: no suitable method found for findOne(Long)
                final Office office = this.officeRepository.findOne(officeIdLong);
                                                           ^
    method QueryByExampleExecutor.<S>findOne(Example<S>) is not applicable
      (cannot infer type-variable(s) S
        (argument mismatch; Long cannot be converted to Example<S>))
    method JpaSpecificationExecutor.findOne(Specification<Office>) is not applicable
      (argument mismatch; Long cannot be converted to Specification<Office>)
  where S is a type-variable:
    S extends Office declared in method <S>findOne(Example<S>)
/home/travis/build/apache/fineract/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanSchedularServiceImpl.java:234: error: cannot find symbol
                } while (!CollectionUtils.isEmpty(loanIds));
                          ^
  symbol:   variable CollectionUtils
  location: class LoanSchedularServiceImpl
````

See more here: https://travis-ci.org/apache/fineract/builds/616540089?utm_source=github_status&utm_medium=notification

awasum:
@wkk91193 Also rebase this with the upstream(apache fineract) develop branch as we have merged in the spring boot upgrade and fixed some travis failures...

wkk91193:
@awasum @vorburger Build is successful now

awasum:
> @awasum @vorburger Build is successful now

Good progress. Please check the changes requested by @vorburger  and the other comments not yet resolved. We are close to merging this.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/jobs/domain/JobParameter.java
vorburger: @wkk91193 I still don't really understand why you have both a public constructor and a public static factory method here.. this is un-usual, usually it's either or the other.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/RecalculateInterestPoster.java
vorburger: @wkk91193 could you _Resolve conversation_ yourself whenever you amend PRs and have taken review feedback into account - this makes it easier for reviewers? You appear to have addressed this point - will you click _Resolve conversation_ here?

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanSchedularServiceImpl.java
vorburger: Any comment re. this? Do you think this code is thread safe as it is?

wkk91193: @vorburger I just made them thread-safe. Thanks for your suggestions

wkk91193:
@vorburger your requested changes are now resolved

vorburger (approved):
Let's get this merged... it can always be improved if it's not yet perfect.

PR #542, opened 2019-03-08
add new Governance and Policies section to Readme

vorburger:
With wording prohibiting PRs of thousands of lines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,12 @@ Video Demonstration
 
 Apache Fineract / Mifos X Demo (November 2016) - <https://www.youtube.com/watch?v=h61g9TptMBo>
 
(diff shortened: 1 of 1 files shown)


vishwasbabu (commented):
Short and to the point. Looks good to me.

awasum (approved):
This looks great. I think it will pass across the right message.

vorburger:
As per @myrle-krantz feedback on PMC private list, instead of having this wording here, it's now linking to a (new) page on the Wiki seeded with the same phrasing which @vishwasbabu and @awasum +1 here.. I generally prefer not to "self merge" if possible, so can I let one of you merge this one as well as #543 ?

PR #548, opened 2019-03-11
Use gradlew --console=plain, hoping to reduce output (FINERACT-732)

vorburger:


vorburger:
this seems to be a Good Idea, see comments in https://jira.apache.org/jira/browse/FINERACT-732 for details.

PR #581, opened 2019-06-12
remove Instructions to download gradle wrapper from README

vorburger:
because the Gradle Wrapper is available in fineract-provider/gradle/wrapper and therefore not a pre-requisite to contribute

> as already removed from Requirements in previous commit 2f80674a2c08a91faf45ce25ba435cfcb2211da8

BTW sorry I force commited that, I meant to raise a PR but clicked wrong)

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,16 +7,6 @@ Requirements
 * Java >= 1.8 (Oracle JVMs have been tested)
 * MySQL 5.5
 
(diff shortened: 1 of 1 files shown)


awasum (approved):
Good catch. @vorburger

vorburger:
> Good catch. @vorburger

so merge it? :smile_cat:

PR #550, opened 2019-03-15
Fineract 700 upgrade gradle - Flyway plugin Upgrade

muarachmann:
## Description
Upgraded flyway plugin to the latest as per #525 and added migration for `migrateTenantListDB`

## Checklist
Please make sure these boxes are checked before submitting your pull request - thanks!

- [ ] Commit message starts with the issue number from https://issues.apache.org/jira/projects/FINERACT/. Ex: FINERACT-646 Pockets API.

- [ ] Coding conventions at https://cwiki.apache.org/confluence/display/FINERACT/Coding+Conventions have been followed.

- [ ] API documentation at https://github.com/apache/fineract/blob/develop/api-docs/apiLive.htm has been updated with details of any API changes.

- [ ] Integration tests have been created/updated for verifying the changes made.

- [ ] All Integrations tests are passing with the new commit.

- [ ] PR contains a single commit (If you have multiple commits, please squash them into a single commit).

Our guidelines for code reviews is at https://cwiki.apache.org/confluence/display/FINERACT/Code+Review+Guide

PR #525, opened 2019-02-17
FINERACT-700: upgrade Gradle from 2.10 to 5.2.1

vorburger:
https://issues.apache.org/jira/browse/FINERACT-700

PR #582, opened 2019-06-12
use MariaDB ConnectorJ instead of Drizzle MySQL JDBC driver -only- during build for Flyway (fixes FINERACT-761)

vorburger:


vorburger:
I'm not sure I understand the TravisCI build failure here... don't see an obvious error [in its raw build log](https://api.travis-ci.org/v3/job/544611583/log.txt).

@michaelbukachi @muarachmann @awasum given your work over in #550, do you see what's going on here?

muarachmann:
@vorburger absolutely. I see what is ongoing i too switched to mysql-connector which worked but due to the license stuff and legal stuff jdbc had to be patched somehow. If the LEGAL stuff can be handle by the legal team as you mentioned on JIRA it would be cool

michaelbukachi:
@vorburger reference to the drizzle driver is thrown quite a number of places beyond `build.gradle`. I found it in `xml` and `.java` files. Had to do quite a bit of find and replace to get rid of it completely.

vorburger:
> I see what is ongoing i too switched to mysql-connector which worked but due to the license stuff and legal stuff jdbc had to be patched somehow. 

If you have the time, I'd love to learn more and hear details... don't see an obvious error [in its raw build log](https://api.travis-ci.org/v3/job/544611583/log.txt) - if you do, and can point to (copy/paste) it, and how you think it can be solved, that would be very interesting.

> If the LEGAL stuff can be handle by the legal team as you mentioned on JIRA it would be cool

I think in https://issues.apache.org/jira/browse/LEGAL-462 we are converging on that it's OK to use the the MariaDB.org (not Oracle's MySQL) Connector/J JDBC driver instead of Drizzle at build time for Flyway, so let's assume that's a Go and get this PR to pass?

Alternatively you could also pick up your work in #550 if you want to rebase it on my rebased #525. But I think it may be easier to do this as two completely separate steps - let us FIRST have a PR, this or one from you, which ONLY switches to use MariaDB ConnectorJ instead of Drizzle MySQL JDBC driver -only- during build for Flyway (fixes FINERACT-761), but does not do the Gradle etc. upgrade, yet. Once that works, let us then rebase #525 and finish that off. Sounds like a plan? Can you help with the first step?

> @vorburger reference to the drizzle driver is thrown quite a number of places beyond build.gradle. I found it in xml and .java files. Had to do quite a bit of find and replace to get rid of it completely.

Careful - my understanding of the legal situation is that it IS OK to switch to use MariaDB ConnectorJ instead of Drizzle MySQL JDBC driver during build for Flyway, but that we do not want do, have to or legally easily can find and replace to get rid of it completely elsewhere. Having this mix should be fine, I think.

michaelbukachi:
If it's okay to go ahead with it. I could take a look at your code later tonight and try to solve failing builds.

vorburger:
> If it's okay to go ahead with it. 

[LEGAL-462](https://issues.apache.org/jira/browse/LEGAL-462) just concluded! :smile: 

> I could take a look at your code later tonight and try to solve failing builds.

That would be really really cool! Thanks a ton in advance. Let's do it step by step - minimal PRs.

vorburger:
> I'm not sure I understand the TravisCI build failure here... don't see an obvious error

I've had a quick closer look, and wonder if these may just be red herring false hits?!  Neither of these problems have anything at all to do with this PR...

... the first time, in Build #162, it seems to have failed in the license test build step just due to `Could not GET 'https://jcenter.bintray.com/junit/junit/4.12/junit-4.12.jar'. Received status code 504 from server: Gateway Time-out`.  The second time (after I added the amended README to this PR), in Build #167 it fails early when it tries to set the mysql password; watch [FINERACT-763](https://issues.apache.org/jira/browse/FINERACT-763).

@michaelbukachi I'm curious if this PR locally builds fine for you? (Using all steps from the [.travis.yml](https://github.com/apache/fineract/blob/develop/.travis.yml)).

vorburger:
Voilà, build is green now! 

@awasum @angelboxes @muarachmann @vishwasbabu plz review and merge this? Tx!

michaelbukachi:
Awesome stuff! This is truly a breakthrough!

vishwasbabu:
@vorburger : Thinking back, we could probably have simply removed all the flyway tasks from gradle. I am not sure if anybody was actually using them (the gradle tasks for  migrating DB's listed in the readme are not really required as flyway would run anyway on application startup)

vorburger:
> @vorburger : Thinking back, we could probably have simply removed all the flyway tasks from gradle. I am not sure if anybody was actually using them (the gradle tasks for migrating DB's listed in the readme are not really required as flyway would run anyway on application startup)

https://issues.apache.org/jira/browse/FINERACT-765

PR #587, opened 2019-06-13
minor README clean up to remove more left over Gradle wrapper mentions

vorburger:
as in #581 (the last mention is intentionally left, because #586 removes that one)

@vishwasbabu @awasum @angelboxes review and merge?

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,14 +10,12 @@ Requirements
 Instructions to run Apache RAT (Release Audit Tool)
 ============
 1. Extract the archive file to your local directory.
-2. Download gradle-wrapper.jar version 2.10 and place it in the fineract-provider/gradle/wrapper folder. See 'Instructions to download gradle wrapper' above.
-3. Run `./gradlew rat`. A report will be generated under build/reports/rat/rat-report.txt
+2. Run `./gradlew rat`. A report will be generated under build/reports/rat/rat-report.txt
 
 Instructions to build a war file
 ============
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,14 +10,12 @@ Requirements
 Instructions to run Apache RAT (Release Audit Tool)
 ============
 1. Extract the archive file to your local directory.
-2. Download gradle-wrapper.jar version 2.10 and place it in the fineract-provider/gradle/wrapper folder. See 'Instructions to download gradle wrapper' above.
-3. Run `./gradlew rat`. A report will be generated under build/reports/rat/rat-report.txt
```
vishwasbabu, line 14: Apache releases prohibit shipping any binaries. Since the README is also shipped with releases, it provides details for downloading Gradle wrapper which would have been deleted as a part of the release process. Looks like the same was removed with https://github.com/apache/fineract/commit/92bdd539e500e6148d04577dc4aef2596195a203#diff-04c6e90faac2675aa89e2176d2eec7d8

vorburger: Oh, I see... yeah, that was done in #581 (FYI @awasum), sorry about that. I've partially reverted it, and amended the text so that it's clearer - how about #591 ? With that, perhaps we don't need to repeat it N times in the README (and have old 2.10 version numbers which are wrong and won't be maintained..), what do you think? If that makes sense to you, we could still merge this one?

PR #589, opened 2019-06-14
rename SQL Query Parameters (fixes FINERACT-766)

vorburger:


vorburger:
@vishwasbabu @awasum @angelboxes is this safe to change?

see https://issues.apache.org/jira/browse/FINERACT-766

vishwasbabu:
AFAIk , spring parses the method name to derive the query param names, so this change should be safe to merge.

vorburger:
> AFAIk , spring parses the method name to derive the query param names, so this change should be safe to merge.

That's what I also thought at first, but then I hesitated, because why do we get the error in FINERACT-766 if it's using the method name? Clearly it's using the argument name for... something, which is causing the reserved name clash. But I have to admit I'm a bit rusty on the details of what exactly Spring Data does here (creates a JPA query?) .. let me post to the list asking if others can confirm that this is perfetly safe.

angelboxes:
It seems to be right, unless you change how the entity is mapped or the name of the function findByKey,  I don't see how this change could give you a different result.

vorburger:
Ok, let's merge it then.. I'll (exceptionally) self merge this one, based on the discussion and confidence that it shouldn't break queries expressed here.

PR #593, opened 2019-06-20
Create one-touch deploy with docker

conradsp:
## Description
Allow users to launch a new Fineract instance easily with Docker and docker-compose.  Also allow local devs to run in a docker environment.

## Checklist


docker/build.gradle
```diff
@@ -0,0 +1,553 @@
+/**
```
vorburger: @conradsp this appears to be a verbatim copy/paste of the `build.gradle` at the root of the project? We should avoid this - you know how these things go, I guarantee that this won't be kept up to date... if the reason that you to do this was because you couldn't get Docker to access files outside of the `docker/` sub-directory you're introducing here, then if I were you I would probably just put the `Dockerfile` at the root - not the end of the world, IMHO.

docker/README.md
```diff
@@ -0,0 +1,25 @@
+Installing Fineract with Docker and docker-compose
```
vorburger: can I suggest that we simply add this to the root README? It will be much easier for people to see, IMHO.

docker/Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM bitnami/tomcat:7.0.94 as fineract
+
+USER root
+RUN apt-get update -qq && apt-get install -y git openjdk-8-jdk wget
+ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64/jre
+
+RUN git clone https://github.com/apache/fineract.git
```
vorburger: hang on - why `git clone` here, instead of just what you have commented out, below? May be I am missing something about the intended "flow" of things here, but your doc says that as a user you have to "Clone the Fineract Github repository", so then you can just use the local sources directly (as you are in the commented out line 17), no?

vorburger (changes requested):
@conradsp cool! I have some review feedback...

PS: Once I am fully done with https://issues.apache.org/jira/browse/FINERACT-730, we'll be able to simplify this, and ditch Tomcat. I don't want to hold this up with that though, so I'm happy to Approve this for merge as soon as you could address the review feedback.

README.md
```diff
@@ -49,6 +49,26 @@ Run the following commands, very similarly to how [.travis.yml](.travis.yml) doe
...
+It is possible to do a 'one-touch' installation of Fineract using docker-compose
+
+  Prerequisites:
+  * docker and docker-compose installed on your machine
+
+
+  Installing a new Fineract instance:
+
+  * Clone the Fineract Github repository
+  * Navigate to the docker directory
```
vorburger: This is no longer needed now, right? Remove?

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+USER root
+RUN apt-get update -qq && apt-get install -y git openjdk-8-jdk wget
+ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64/jre
+
+RUN mkdir fineract
+COPY . fineract
+
+WORKDIR fineract
+RUN sed -i 's/localhost/fineractmysql/' ./fineract-provider/build.gradle
```
vorburger, line 29: Is this for the mysql url? Why is it needed? Given how "./gradlew clean war" doesn't run any of the SQL tasks, can this be removed? If not, perhaps an inline comment to document what it's needed for? (FYI I am hoping to remove the Flyway related Gradle tasks, see https://issues.apache.org/jira/plugins/servlet/mobile#issue/FINERACT-765)

vorburger (changes requested):

README.md
conradsp: Yes, will remove the instruction to navigate to the docker directory

Dockerfile
conradsp: In build.gradle, we expect mysql to be running on localhost (I am glad that this is going away with your changes). In the docker-compose environment, the mysql instance can be referenced using the hostname 'fineractmysql'. So, simply replacing that so Flyway can find the DB server.

vorburger: Right, but you (in this Dockerfile) don't even actually run those Flyway tasks (which are hopefully going to go away some time) - you, correctly only `RUN ./gradlew clean war` - and that doesn't actually require _mysql to be running on localhost_ agreed?

fineract-provider/build.gradle
```diff
@@ -125,6 +125,10 @@ rat {
     '**/native/**',
     '**/wrapper/**',
 	'**/build/**',
+    // docker
+    '**/Dockerfile/**',
+    '**/docker-compose.yml/**',
+    '**/initdb/**',
```
vorburger: it just occured to me that it would probably be better to add the ASF required license headers instead of suprressing this here like this... at least for YAML files (like docker-compose.yml) it's possible, see the https://github.com/apache/fineract/blob/develop/.travis.yml, and for 01-databases.sql it's possible as well, see e.g. the https://github.com/apache/fineract/blob/develop/fineract-db/mifospltaform-tenants-first-time-install.sql et all. Not 100% sure re. the Dockerfile, but just try adding a license header to it, and see if rat (the license check tool, which you can run locally via `./gradlew licenseMain licenseTest licenseIntegrationTest`) is happy, without having to add these ignores?

initdb/01-databases.sql
```diff
@@ -0,0 +1,6 @@
+# create databases
```
vorburger: instead of putting this into the new `initdb/` directory, how about just sticking this into the existing `fineract-db/` ?

vorburger (changes requested):
Sorry for my delay in reviewing this (and being a PITA and asking for more changes!) ..

vorburger:
@conradsp FYI https://issues.apache.org/jira/browse/FINERACT-760 for what we can do next, once you have had a chance to address my (last, I promise, sorry) review feedback and I'll have merged this!

initdb/01-databases.sql
conradsp: I moved this file to fineract-db/docker.

fineract-provider/build.gradle
conradsp: Thanks for your feedback. I updated all 3 of those files and added the license header. RAT seems fine with it. As far as the changes to build.gradle, if I leave the DB host as localhost, the docker images crashes on startup when Fineract runs the Flyway migration:

FlywayException: Unable to obtain Jdbc connection from DataSource; for Tenant DB URL: jdbc:mysql:thin://localhost:3306/mifostenant-default, username: root

I can dig into it a bit more, but appears that some part of the gradle build configuration is carried into the war file...

awasum: Flyway will pick the tenant connection info from the mifosplatform-tenant DB and use that to try and connect to the different tenants. By default, the mysql host value is localhost. This is what causes the problem in the container at startup.
What you can do is to copy and modify the list-db SQL scripts to change the mysql host to the dockerhost of mysql. Its a dirty fix.

vorburger: Duh! OK.. I still don't entirely get what's going on here, but don't want to hold this up any further, and do trust you (we can always get back to fully resolving this mystery in [FINERACT-765](https://issues.apache.org/jira/browse/FINERACT-765)). Thanks for your patience with me re. this point.

vorburger:
LGTM now, merging.  (just FYI https://issues.apache.org/jira/browse/FINERACT-772)

vorburger (approved):

fineract-provider/build.gradle
vorburger: see #613

PR #595, opened 2019-07-01
One-touch deploy: Set host to fineractmysql when running in docker

conradsp:
## Description
This is a quick update to the One-touch deploy process. When we seed the tenant database, we want to default the hostname to fineractmysql rather than localhost. Eventually, this will be all that we need to do, rather than updating the build.gradle file. 

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,6 +27,7 @@ COPY . fineract
 
 WORKDIR fineract
 RUN sed -i 's/localhost/fineractmysql/' ./fineract-provider/build.gradle
+RUN sed -i 's/localhost/fineractmysql/' ./fineract-provider/src/main/resources/sql/migrations/list_db/V1__mifos-platform-shared-tenants.sql
 RUN ./gradlew clean war
 RUN mv build/libs/fineract-provider.war /opt/bitnami/tomcat/webapps
 


Dockerfile
```diff
@@ -27,6 +27,7 @@ COPY . fineract
 
 WORKDIR fineract
 RUN sed -i 's/localhost/fineractmysql/' ./fineract-provider/build.gradle
```
vorburger, line 29: > Eventually, this will be all that we need to do, rather than updating the build.gradle file.

only after https://issues.apache.org/jira/browse/FINERACT-765, or could this be done now already?

vorburger:
@conradsp what bothers me slightly (but not enough to not merge this, after we clarify the small question, above, so this just FYI) is that as we get this on Docker Hub (https://issues.apache.org/jira/browse/INFRA-18673), we're going to be pushing a container image with a hard-ocded DB URL that is specific to `docker compose` .. I'd like to eventually contribute e.g. Kubernetes deployment steps, where the DB would be differerent. I can't really think of a better solution for this problem right now either though.

conradsp:
Thanks Michael - I agree with your comments. As soon as
https://issues.apache.org/jira/browse/FINERACT-765 is merged, I will remove
the build.gradle changes in the Dockerfile.

Will continue to think about a more robust solution. Ideally the db seeding
scripts will not have a db hostname encoded at all and would pull that from
an environment variable (defaulting to localhost or some sane default).

Thanks,
Steve

vorburger:
> As soon as FINERACT-765 is merged, I will remove the build.gradle changes in the Dockerfile.

OK, so for now we do need both (that wasn't clear to me; I still don't really fully understand it).

> Will continue to think about a more robust solution. Ideally the db seeding scripts will not have a db hostname encoded at all and would pull that from an environment variable (defaulting to localhost or some sane default).

+1 (I've created just https://issues.apache.org/jira/browse/FINERACT-773 re. this, for some day.)

vorburger (approved):

vorburger:
see #613

PR #596, opened 2019-07-01
re-enable commented out TemplateMergeServiceTest.compileHelloTemplate (which actually passes)

vorburger:


vorburger:
FTR: I lococally tested that `./gradlew test` still passes on `develop` with this merged after #602 (because I probably had raised this PR before noticing the bigger problem in #602), and wanted to make sure we didn't accidentially break thing - it's fine, this test still passes, even now that we are actually really running unit tests again.

PR #598, opened 2019-07-01
add .gitpod.yml for http://gitpod.io

vorburger:
try https://gitpod.io#github.com/apache/fineract/pull/598 to check this out...

based on https://github.com/svenefftinge/fineract/commit/da888d808d37abcc10f7bd3548c6283965bff244

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,5 +1,42 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements. See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership. The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License. You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+# https://www.gitpod.io
+# https://www.gitpod.io/docs/41_Config_Gitpod_File/
+# https://github.com/gitpod-io/workspace-images/tree/master/mysql
+
 image: gitpod/workspace-mysql
 ports:
+- port: 3306
+  onOpen: ignore
 - port: 8005
+  onOpen: ignore
+- port: 8009
+  onOpen: ignore
+- port: 8080
+- port: 8081
+  onOpen: ignore
+- port: 8443
 tasks:
-- command: ./gradlew install
+- command: |
+      # NB MySQL on GitPod.io is currently quite slow.
+      # similar to .travis.yml, but using authentication_string= instead of password= due to mysql major version difference
+      mysql -e "USE mysql;\nUPDATE user SET authentication_string=PASSWORD('mysql') WHERE user='root';\nFLUSH PRIVILEGES;\n"
+      ./gradlew createDB -PdbName=mifosplatform-tenants
+      ./gradlew createDB -PdbName=mifostenant-default
+      ./gradlew build -x test
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,5 @@
+image: gitpod/workspace-mysql
+ports:
+- port: 8005
+tasks:
+- command: ./gradlew install
--- a/README.md
+++ b/README.md
@@ -1,8 +1,15 @@
-Apache Fineract: A Platform for Microfinance [![Build Status](https://travis-ci.org/apache/fineract.svg?branch=develop)](https://travis-ci.org/apache/fineract)  [![Docker Hub](https://img.shields.io/docker/pulls/apache/fineract.svg)](https://hub.docker.com/r/apache/fineract)  [![Docker Build](https://img.shields.io/docker/cloud/build/apache/fineract.svg)](https://hub.docker.com/r/apache/fineract/builds)
+Apache Fineract: A Platform for Microfinance  [![Build Status](https://travis-ci.org/apache/fineract.svg?branch=develop)](https://travis-ci.org/apache/fineract)  [![Docker Hub](https://img.shields.io/docker/pulls/apache/fineract.svg)](https://hub.docker.com/r/apache/fineract)  [![Docker Build](https://img.shields.io/docker/cloud/build/apache/fineract.svg)](https://hub.docker.com/r/apache/fineract/builds)
 ============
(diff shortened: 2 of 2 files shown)


.gitpod.yml
```diff
@@ -0,0 +1,30 @@
...
+# under the License.
+#
+
+# https://www.gitpod.io
+# https://www.gitpod.io/docs/41_Config_Gitpod_File/
+# https://github.com/gitpod-io/workspace-images/tree/master/mysql
+
+image: gitpod/workspace-mysql
+ports:
+- port: 8080
```
vorburger, line 31: @svenefftinge I noticed you had used port 8005, what's that for? Just an arbitrary example, or something GitPod requires? How to specify several ports here, like 8080 and 8443? https://www.gitpod.io/docs/43_Config_Ports/ only mentions ranges.

vorburger: Talking about ports... how do you suppress messages like these in GitPod:

> A service is listening on port 8005 but is not exposed
> A service is listening on port 3306 but is not exposed

I suspect these are perhaps ports that the Gradle build opens, but which it doesn't need to expose - nor ask about every time, that's confusing / noisy.

vorburger: @svenefftinge oh I'm dumb, I think I just figured (both of) these out, see 723ca56c8b754a19cda8f202d8d13f04715d6bc5

vorburger:
@meysholdt (/Cc @svenefftinge) FYI the mysql in the "worker container" (or whatever you call it) is very slow, noticeable slower than on a developer workstation, and it takes a non-negligable amount of time to start this project, because it runs (a lot of) DB Schema mgiration scripts (using Flyway) at boot... I don't suppose there is some way to grant the "Terminal container" (?) a higher priority? Just asking.

vorburger:
https://github.com/gitpod-io/gitpod/issues/648

.gitpod.yml
```diff
@@ -0,0 +1,41 @@
...
+- port: 8081
+  onOpen: ignore
+- port: 8443
+tasks:
+- command: >
+      # similar to .travis.yml, but using authentication_string= instead of password= due to mysql major version difference
+      mysql -e "USE mysql;\nUPDATE user SET authentication_string=PASSWORD('mysql') WHERE user='root';\nFLUSH PRIVILEGES;\n";
+      ./gradlew createDB -PdbName=mifosplatform-tenants;
+      ./gradlew createDB -PdbName=mifostenant-default;
+      ./gradlew test
```
vorburger: @meysholdt do you happen to be able to spot what I'missing here, it's NOK and causes:

```
gitpod /workspace/fineract $ {
ifostenant-default; ./gradlew testlatform-tenants; ./gradlew createDB -PdbName=mi
> 
> }; HISTFILE=/workspace/.cmd-0 history -r
bash: syntax error near unexpected token `}'
gitpod /workspace/fineract $ 
```

.gitpod.yml
svenefftinge: Yes, that's how you do it :)

svenefftinge:
Would be interesting to learn what makes mysql slow in the container. The nodes are pretty powerful and the database sits on an SSD. @meysholdt any idea? (cc @jankeromnes)

.gitpod.yml
meysholdt: you may want to try `command: |` instead of `command: >`. The `|` causes the following string to be treated as multi-line. The `>` makes it single-line. 

With `command: |` the `;` at the end of the `bash` lines are no longer needed, e.g. in `./gradlew ... ;`

meysholdt:
There is no idea on the top of my head. As you said, the hardware is good :) 

MySQL runs with this config:
https://github.com/gitpod-io/workspace-images/blob/master/mysql/mysql.cnf

Maybe the config holds some flags which are not optimal?

@vorburger how dos your MySQL config look like when MySQL performs well?

.gitpod.yml
jankeromnes: I think this error is caused by the `#` comment inside the `command` (because Gitpod folds commands into one line, and doesn't currently support/remove comments).

I tried this workaround: https://github.com/jankeromnes/fineract/commit/0830e34e4939df5d6c45bf46ea2a2b22355f24ba which seems to make your `command` work in Gitpod.

EDIT: Ah, I hadn't seen @meysholdt's reply, sorry. I didn't know about `command: |`, that's interesting.

meysholdt: https://yaml-multiline.info :-D

.gitpod.yml
awasum: What version of MySQL is this?
Could not get alot of info here: https://hub.docker.com/r/gitpod/workspace-mysql

meysholdt: After opening https://gitpod.io/#github.com/apache/fineract/pull/598 I ran in the terminal:

```
$ mysqld --version
mysqld  Ver 5.7.26-0ubuntu0.18.10.1 for Linux on x86_64 ((Ubuntu))
```

Find the `Dockerfile` here: https://github.com/gitpod-io/workspace-images/tree/master/mysql

The mysql-image inherits from [workspace-full](https://github.com/gitpod-io/workspace-images/blob/master/full/Dockerfile) which again inherits from `ubuntu:consmic`. 

The mysql-version is the latest available in `ubuntu:consmic`. Let me know if anther version would be better for some reason :)

awasum: Hmm,,, MySQL 5.7 is fine and supported.. I believe.

vorburger:
FTR: This is pending either #525 (Gradle upgrade in Fineract) or https://github.com/gitpod-io/gitpod/issues/648 (making it possible to use older Gradle versions in GitPod.io); without either of those solved, it does not actually work.

meysholdt:
Can you describe how the system's gradle is called when in fact you call `./gradew`? 

I opened https://gitpod.io/#github.com/apache/fineract/pull/598 and the following commands resulted in "build successful":
```
./gradlew createDB -PdbName=mifosplatform-tenants;
./gradlew createDB -PdbName=mifostenant-default;
./gradlew test
```

vorburger:
@meysholdt Yeah, typing commands into the Terminal works fine, the problem is what I raised in https://github.com/gitpod-io/gitpod/issues/648 - editing Java source files doesn't really work because the Language Server doesn't (properly, fully) start, due to the problem you'll see in the Problems view on GitPod.io which I reported in https://github.com/gitpod-io/gitpod/issues/648.

FYI I just had a breakthrough in #525 (Gradle upgrade in Fineract), and expect that to get merge soon, after which I may merge this, because then we'll happen to have alignment for the Gradle version this project needs to build and the one that GitPod happens to have packaged right now. But that just hides what IMHO is still a real problem in GitPod though (https://github.com/gitpod-io/gitpod/issues/648) - any project that needs a Gradle version that is not compatible with the one you packaged won't properly work. Gradle in practice is not  backwards compatible over major versions. 

Hope this helps and clarifies? Do shout if it's not and I can help to make this clearer in some way.

.gitpod.yml
vorburger: Thanks! Now done as you suggested in 529735b431b989d2acad2f46bd5983df7e72a360.

PR #599, opened 2019-07-02
add suitable .dockerignore

vorburger:
This reduces the `docker build` time spent in `STEP 6: COPY . fineract` from about 6m to 3m,
because it avoids copying locally built artifacts into the container (which we don't need there, we're just about to re-build them in the container anyway).

This also provides better isolation from the local dev environment and the should lead to more reproducible container builds.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+build/*
+.gradle/*


vorburger:
@conradsp how about this?  Plz merge if OK for you... thanks!

conradsp:
Thanks @vorburger - that's a good addition. Will merge.

vorburger:
@conradsp you merged this wihout noticing it broke the build... :smile: #600 fixes it.

conradsp:
Thanks @vorburger! Darn RAT checks! :)

PR #600, opened 2019-07-03
add missing license header to .dockerignore

vorburger:


conradsp:
@vorburger - I verified that it builds correctly. 😄 Merging now.

PR #601, opened 2019-07-03
{WIP, DO NOT MERGE} add classpath duplicates detection test

vorburger:


vorburger:
Please do not merge this yet, I just noticed that we actually have a bigger problem to fix first in #602, then rebase this on that, and then this will illustrate a number of problems we should fix first, and only then merge this.

vorburger:
Perhaps this PR (also..) should wait until we're fully done with the Gradle upgrade (although I initally built it to help me understand a problem I ran into there). This is because it points out `/home/vorburger/.gradle/wrapper/dists/gradle-2.10-bin/baigpnfu14tdk6ztbfwcl8275/gradle-2.10/lib/groovy-all-2.4.4.jar` as having CP dupes of groovy-2.3.6.jar - but the former should never be on the application's runtime classpath anyway.. not sure if this was a bug in Gradle 2.10 which is perhaps fixed in later versions?

vorburger:
This will definitely first also require #607 (and probably more).

awasum:
What problem is this PR trying to solve? Can we create a Jira issue for it. Please, I like to understand the context. Any background on this? @vorburger

vorburger:
#803 picks this up...

PR #603, opened 2019-07-03
make OpenJPA version easier to change

vorburger:
by making it only in 1 place instead of in 2

This does not actually change the version (but a follow-up will).


conradsp:
Looks fine to me. Builds successfully. Merging.

PR #604, opened 2019-07-03
use slf4j instead Apache Commons Logging in CustomAuthenticationFailureHandler

vorburger:


vorburger:
Given how we're using Slf4j everywhere else, and that this PR passed the TravisCI build, this seems very low risk to me, so I'm going to go ahead and exceptionally self merge this trivial change.

PR #605, opened 2019-07-03
ditch un-used import of (very old) serp.util.Strings

vorburger:


vorburger:
self merging trivial change which passed TravisCI build and has 0 risk...

PR #606, opened 2019-07-03
remove only use of Apache BVal @NotEmpty annotation (in InteropIdentifierRequest/Response)

vorburger:
I'm not sure if this even worked?  There's no test coverage for it.

This is required to be able to (easily) switch from openjpa-all to openjpa in upcoming #607

The alternative would be to add an explicit dependency to BVal; but that seems unnecessary, and avoiding it will likely reduce other dependency issues.

PR #607, opened 2019-07-03
use openjpa (with real dependencies) instead of shaded openjpa-all

vorburger:
Because the openjpa-all.jar is "shaded" and that causes classpath duplicate problems.

PR #608, opened 2019-07-03
bump OpenJPA version from 2.4.1 to 2.4.3 for Java 11 compliance

vorburger:
This will require # (which in turn first requires a few other minor PRs to be merged).

The upgrade is because 2.4.1 was not yet Java 11 compatible (due to an old ASM)
    
In the context of the Gradle upgrade work, I'd like to upgrade to Java 11 soon-ish, and am starting to experiment with it.

This first requires #604, #605, #606.

### DIFF
--- a/fineract-provider/build.gradle
+++ b/fineract-provider/build.gradle
@@ -29,7 +29,7 @@ project.ext.springDataJpaVersion = '1.7.0.RELEASE' // also change spring-boot-gr
 
 buildscript {
   ext {
-    openJPAVersion = '2.4.1'
+    openJPAVersion = '2.4.3'
   }
   repositories {
   	 jcenter()


vorburger:
This, ideally (to avoid doing required adjustment work twice) requires #607 as a pre-requisite, first.

vorburger:
@conradsp what do you think?  @lingocoder FYI

vorburger:
Actually, seeing how the merge of #607 apparently just broke #525 again, perhaps let's hold off merging this until affter #525 is merged, then rebase this on top of that, and only then merge this?

vorburger:
> rebase this on top of that, and only then merge this?

done, and it passes. @awasum @conradsp @vishwasbabu would you like to merge this?

vorburger:
FTR: https://issues.apache.org/jira/browse/OPENJPA-2730 has what we need OpenJPA v2.4.3. This alone doesn't quite do the trick for https://github.com/apache/fineract/pull/609, see https://github.com/radcortez/openjpa-gradle-plugin/issues/8 for why. IMHO it would still make sense to merge this already right now though; I'll deal with the follow-up in #609.

awasum (approved):

PR #610, opened 2019-07-03
remove Tomcat Jasper JSP Engine, as we don't use JSP

vorburger:
This probably also makes integration test start-up a (little) bit faster.


vorburger:
I may actually need Jasper later again when upgrading Tomcat 7.0.54 to 8.5.42 (where `com.radcortez.gradle:openjpa-gradle-plugin` does not start without Jasper..), so if this is causing any other problems, we can also just forget about it. I'll close it.

PR #611, opened 2019-07-04
upgrade Apache Tomcat from 7.0.54 to 7.0.94

vorburger:
This fixes the following weird problem I ran into during the Gradle & Spring Boot / OpenJPA Gradle plugins upgrades; it's likely some Tomcat bug that got fixed between .54 and .94:

    Unable to process resource element [jndi:/localhost/fineract-provider/WEB-INF/classes/org/apache/fineract/.../*.class] for annotations

### DIFF
--- a/fineract-provider/dependencies.gradle
+++ b/fineract-provider/dependencies.gradle
@@ -17,7 +17,7 @@
  * under the License.
  */
 dependencies {
-        def tomcatVersion = '7.0.54'
+        def tomcatVersion = '7.0.94'
         tomcat "org.apache.tomcat.embed:tomcat-embed-core:${tomcatVersion}",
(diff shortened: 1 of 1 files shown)


vorburger:
@conradsp review? Specifically re. container (Docker) - but if anything you should jump up :joy: in joy, because this aligns the version we use for `integrationTest` with the `bitnami/tomcat:7.0.94` you've used in the (new) `Dockerfile` ... so fine for you, I guess?

@vishwasbabu would this be OK for you, any objections, or could you merge it?

conradsp:
Yes, this is good. Now let's get to Tomcat 8! :) @vishwasbabu, please merge if OK with you.

vorburger:
Re. Tomcat 8, my preference would be to ditch pre-historical times (AKA WAR
files), and just do regular Spring Boot (with java -jar
fineract-provider.war via embedded Tomcat, whatever version, not our
problem- in theory).

>

awasum:
You are right @vorburger .. java -jar file.war(jar) works well with Docker too..as cloud native becomes even popular.

conradsp:
@vorburger - I agree 100% on getting rid of Tomcat dependency. As @awasum indicated, running as a straight jar would be the best option.

awasum:
Merging

PR #612, opened 2019-07-04
comment out PMD & FindBugs

vorburger:
Too slow, and nobody looked at it in years anyway.
The idea is to later re-activate, with enforcement.

see also https://issues.apache.org/jira/browse/FINERACT-712


awasum:
We will look at this after gradle upgrade

PR #613, opened 2019-07-04
remove Flyway completely from Gradle build (fixes FINERACT-765)

vorburger:
https://issues.apache.org/jira/browse/FINERACT-765

we (of course), still, but only, run it in-App at runtime boot

This is a clean-up which removes something we don't actually need.

Most importantly, it unblocks the Gradle upgrade in FINERACT-700, where this doesn't work anymore anyway (and while there's a fix making it work on one of my branches, it seems simpler to just remove it).


Dockerfile
```diff
@@ -26,7 +26,6 @@ RUN mkdir fineract
 COPY . fineract
 
 WORKDIR fineract
-RUN sed -i 's/localhost/fineractmysql/' ./fineract-provider/build.gradle
```
vorburger, line 29: @conradsp you'll probably want to (manually) test that this doesn't break the container? /Cc @awasum

michaelbukachi:
Awesome stuff! I too never use the flyway plugin to run any tasks. So I find no use for it.

conradsp:
Thanks @vorburger. I have validated that the docker build runs fine and we no longer need to modify build.gradle in the Dockerfile. @vishwasbabu - you can merge if all looks good to you.

PR #614, opened 2019-07-04
Stop creating unnecessary overdue charge when amount is 0.

mohitsinha:
## Description

## Checklist

### DIFF
--- a/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/SchedulerJobsTestResults.java
+++ b/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/SchedulerJobsTestResults.java
@@ -725,6 +725,70 @@ public class SchedulerJobsTestResults {
     }
 
     @Test
+    public void testAvoidUnncessaryPenaltyWhenAmountZeroForOverdueLoansJobOutcome() throws InterruptedException {
+        this.savingsAccountHelper = new SavingsAccountHelper(this.requestSpec, this.responseSpec);
+        this.schedulerJobHelper = new SchedulerJobHelper(this.requestSpec, this.responseSpec);
+        this.loanTransactionHelper = new LoanTransactionHelper(this.requestSpec, this.responseSpec);
+
+        final Integer clientID = ClientHelper.createClient(this.requestSpec, this.responseSpec);
+        Assert.assertNotNull(clientID);
+
+        Integer overdueFeeChargeId = ChargesHelper
+                .createCharges(this.requestSpec, this.responseSpec, ChargesHelper.getLoanOverdueFeeJSONWithCalculattionTypePercentage("0.000001"));
+        Assert.assertNotNull(overdueFeeChargeId);
+
+        final Integer loanProductID = createLoanProduct(overdueFeeChargeId.toString());
+        Assert.assertNotNull(loanProductID);
+
+        final Integer loanID = applyForLoanApplication(clientID.toString(), loanProductID.toString(), null);
+        Assert.assertNotNull(loanID);
+
+        HashMap loanStatusHashMap = LoanStatusChecker.getStatusOfLoan(this.requestSpec, this.responseSpec, loanID);
+        LoanStatusChecker.verifyLoanIsPending(loanStatusHashMap);
+
+        loanStatusHashMap = this.loanTransactionHelper.approveLoan(AccountTransferTest.LOAN_APPROVAL_DATE, loanID);
+        LoanStatusChecker.verifyLoanIsApproved(loanStatusHashMap);
+
+        loanStatusHashMap = this.loanTransactionHelper.disburseLoan(AccountTransferTest.LOAN_APPROVAL_DATE_PLUS_ONE, loanID);
+        LoanStatusChecker.verifyLoanIsActive(loanStatusHashMap);
+
+        ArrayList<HashMap> repaymentScheduleDataBefore = this.loanTransactionHelper.getLoanRepaymentSchedule(this.requestSpec,
+                this.responseSpec, loanID);
+
+        String JobName = "Apply penalty to overdue loans";
+        Integer jobId = 12;
+
+        this.schedulerJobHelper.executeJob(JobName);
+
+        HashMap schedulerJob = this.schedulerJobHelper.getSchedulerJobById(this.requestSpec, this.responseSpec, jobId.toString());
+
+        Assert.assertNotNull(schedulerJob);
+        while ((Boolean) schedulerJob.get("currentlyRunning") == true) {
+            Thread.sleep(15000);
+            schedulerJob = this.schedulerJobHelper.getSchedulerJobById(this.requestSpec, this.responseSpec, jobId.toString());
+            Assert.assertNotNull(schedulerJob);
+        }
+
+        final HashMap chargeData = ChargesHelper.getChargeById(this.requestSpec, this.responseSpec, overdueFeeChargeId);
+
+        ArrayList<HashMap> repaymentScheduleDataAfter = this.loanTransactionHelper.getLoanRepaymentSchedule(this.requestSpec,
+                this.responseSpec, loanID);
+
+        Assert.assertEquals("Verifying From Penalty Charges due fot first Repayment after Successful completion of Scheduler Job",
+                0, repaymentScheduleDataAfter.get(1).get("penaltyChargesDue"));
+
+        final ArrayList loanCharges = this.loanTransactionHelper.getLoanCharges(this.requestSpec,
+                this.responseSpec, loanID);
+
+        Assert.assertNull("Verifying that charge isn't created when the amount is 0", loanCharges);
+
+        loanStatusHashMap = this.loanTransactionHelper.undoDisbursal(loanID);
+        LoanStatusChecker.verifyLoanIsApproved(loanStatusHashMap);
+        LoanStatusChecker.verifyLoanIsWaitingForDisbursal(loanStatusHashMap);
+
+    }
+
+    @Test
     public void testUpdateOverdueDaysForNPA() throws InterruptedException {
         this.schedulerJobHelper = new SchedulerJobHelper(this.requestSpec, this.responseSpec);
         this.loanTransactionHelper = new LoanTransactionHelper(this.requestSpec, this.responseSpec);
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
@@ -2490,6 +2490,9 @@ public class LoanWritePlatformServiceJpaRepositoryImpl implements LoanWritePlatf
 
                 final LoanCharge loanCharge = LoanCharge.createNewFromJson(loan, chargeDefinition, command, entry.getValue());
 
+                if (BigDecimal.ZERO.compareTo(loanCharge.amount()) == 0) {
+                    continue;
+                }
                 LoanOverdueInstallmentCharge overdueInstallmentCharge = new LoanOverdueInstallmentCharge(loanCharge, installment,
                         entry.getKey());
                 loanCharge.updateOverdueInstallmentCharge(overdueInstallmentCharge);
--- a/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/ClientLoanIntegrationTest.java
+++ b/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/ClientLoanIntegrationTest.java
@@ -3787,7 +3787,7 @@ public class ClientLoanIntegrationTest {
         todaysDate = Calendar.getInstance(Utils.getTimeZoneOfTenant());
 
         Integer overdueFeeChargeId = ChargesHelper.createCharges(this.requestSpec, this.responseSpec,
-                ChargesHelper.getLoanOverdueFeeJSONWithCalculattionTypePercentage());
+                ChargesHelper.getLoanOverdueFeeJSONWithCalculattionTypePercentage("10"));
         Assert.assertNotNull(overdueFeeChargeId);
 
         final Integer clientID = ClientHelper.createClient(this.requestSpec, this.responseSpec);
--- a/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/charges/ChargesHelper.java
+++ b/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/charges/ChargesHelper.java
@@ -275,10 +275,10 @@ public class ChargesHelper {
         return chargesCreateJson;
     }
     
-    public static String getLoanOverdueFeeJSONWithCalculattionTypePercentage() {
+    public static String getLoanOverdueFeeJSONWithCalculattionTypePercentage(String penaltyPercentageAmount) {
         final HashMap<String, Object> map = populateDefaultsForLoan();
         map.put("penalty", ChargesHelper.penalty);
-        map.put("amount", "10");
+        map.put("amount", penaltyPercentageAmount);
         map.put("chargePaymentMode", ChargesHelper.CHARGE_PAYMENT_MODE_REGULAR);
         map.put("chargeTimeType", CHARGE_OVERDUE_INSTALLMENT_FEE);
         map.put("chargeCalculationType", ChargesHelper.CHARGE_CALCULATION_TYPE_PERCENTAGE_AMOUNT_AND_INTEREST);
--- a/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/loans/LoanTransactionHelper.java
+++ b/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/loans/LoanTransactionHelper.java
@@ -106,6 +106,12 @@ public class LoanTransactionHelper {
         return (ArrayList) response.get("periods");
     }
 
+    public ArrayList getLoanCharges(final RequestSpecification requestSpec, final ResponseSpecification responseSpec,
+            final Integer loanID) {
+        final String URL = "/fineract-provider/api/v1/loans/" + loanID + "?associations=charges&" + Utils.TENANT_IDENTIFIER;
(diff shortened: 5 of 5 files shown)


mohitsinha:
@conradsp @vishwasbabu @awasum Can anyone review this?

mohitsinha:
@vorburger

cnangeh:
@mohitsinha the code update looks good. Can you update the necessary unit test to reflect this new update?

mohitsinha:
@Anh3h Hey, thanks for the review. I didn't find any test for existing crons. So, it will be difficult for me to setup everything from scratch. If it's any helpful, I have done manual testing for this.

mohitsinha:
@vorburger Can this be merged without a test case?

awasum:
@mohitsinha It will be helpful to take sometime and add tests for the feature you provided. If there is no test for that module, then its recommended to invest some time and put tests for maintenance purposes.

Imagine the next person who works on this particular file or feature type also does not write a test, it means over time, we will keep merging without tests. This can create a situation where we are creating features without test all because.. "it works manually".  Set the foundation so your future should be easy to manage.

mohitsinha:
@Anh3h The tests are added.

cnangeh:
This looks to me.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -2359,6 +2359,9 @@ public void applyOverdueChargesForLoan(final Long loanId, Collection<OverdueLoan
                     null, null, null, loanId, null, null, null, null,null,null);
             LoanOverdueDTO overdueDTO = applyChargeToOverdueLoanInstallment(loanId, overdueInstallment.getChargeId(),
                     overdueInstallment.getPeriodNumber(), command, loan, existingTransactionIds, existingReversedTransactionIds);
+            if(overdueDTO == null){
```
vorburger: is there really a realistic case when `overdueDTO` can be `null` ? If not, then this could hide future bugs, and it would be better to remove this and let it explode with a NullPointerException to find such bugs, instead of silently ignoring a condition which shold never happen. If it can happen, then it's fine - but I don't see how.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -2490,6 +2493,9 @@ public LoanOverdueDTO applyChargeToOverdueLoanInstallment(final Long loanId, fin
 
                 final LoanCharge loanCharge = LoanCharge.createNewFromJson(loan, chargeDefinition, command, entry.getValue());
 
+                if(BigDecimal.ZERO.compareTo(loanCharge.amount()) >= 0){
```
vorburger: This would be also true not only if `loanCharge.amount()` is 0 but also if it's negative. Is that really intended?

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/SchedulerJobsTestResults.java
```diff
@@ -724,6 +724,59 @@ public void testApplyPenaltyForOverdueLoansJobOutcome() throws InterruptedExcept
 
     }
 
+    @Test
+    public void testAvoidUnncessaryPenaltyWhenAmountZeroForOverdueLoansJobOutcome() throws InterruptedException {
```
vorburger, line 728: I'm unable to say if this is functionally correct - would someone else be able to review and confirm it is?

vorburger: @awasum @Anh3h do you understand more about the functionality here than I do and can you confirm this new test is functionally correct?

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
mohitsinha: Not really. I did it just to be safe.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
mohitsinha: ```applyChargeToOverdueLoanInstallment``` won't ever return a null `overdueDTO` as the return itself is a constructor call. I can let it explode but the problem is this whole logic is inside a for loop for all loans that are overdue. Hence, if this explodes it will impact the normal flow for overdue penalty for other loans. Hence it had to be handled gracefully. I can let it explode with a `try catch block`

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/SchedulerJobsTestResults.java
mohitsinha: I can throw some light into this. 
It is creating a loan with a very small interest rate 0.000001%.
It is verifying that the `totalChargesDue` is 0 for the loan ( which is also the feature currently as 0+ 0 + 0 is 0).
I am then calling the api `/loans/{id}?associations=charges` and then verifying that there are no charges created for this loan. This actually tests correctly what I have done.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
vorburger: > `applyChargeToOverdueLoanInstallment` won't ever return a null `overdueDTO` as the return itself is a constructor call. 

OK but then you agree that this null check is unnecessary? If as you say it's impossible for `overdueDTO` to ever be null (because `applyChargeToOverdueLoanInstallment` ends with a `return new LoanOverdueDTO`) then why add this nul lcheck as part of this PR?

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
vorburger: would it be clearer for future readers of this code to check only for `equals(ZERO)` instead? Just a thought.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
mohitsinha: `applyChargeToOverdueLoanInstallment` is  creating a `LoanCharge` and returning `LoanOverdueDTO`. I want to avoid creating the `LoanOverdueInstallmentCharge` on certain conditions ( the amount is zero). As the function holds all the logic for loan charges and returns using a constructor, I had to `return null` to escape the logic of charge creation. 

Hence the null pointer check is due to my code changes inside `applyChargeToOverdueLoanInstallment`

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
mohitsinha: Sure. I will make the change

mohitsinha: `equals` won't work because it also takes into consideration the scale. I have changed it to use `compareTo == 0`

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
vorburger: oh right, now I get it. Sorry, I was a bit slow on this one! :smile_cat:

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
vorburger: OK, cool - resolves conversation for me; thanks!

vorburger:
@awasum @Anh3h this LGTM now, and I have no more objections. I'm hesitant to merge it myself, as I suspect that you understand more about the functionality here than I do and can you confirm that the implementation is fine and the new test is functionally correct... could I let you review and merge this? Tx.

vorburger (approved):

mohitsinha:
@awasum @Anh3h Do you feel anything is still pending on this one?

awasum:
I dont really understand this part of the software very well. I feel we can just merge and if we get any problems, we revert. 
For now, Please @vishwasbabu  will you have time to review this within the next few days after which you can merge or I will just do that after 48 hours.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -2490,6 +2493,9 @@ public LoanOverdueDTO applyChargeToOverdueLoanInstallment(final Long loanId, fin
 
                 final LoanCharge loanCharge = LoanCharge.createNewFromJson(loan, chargeDefinition, command, entry.getValue());
 
+                if (BigDecimal.ZERO.compareTo(loanCharge.amount()) == 0) {
+                    return null;
```
angelboxes: I don't think its okay to return null in this line, as you can see in this part of the code there is a for statement that iterates through a set of dates. If the amount for the charge in the first date entry were to be 0, the next dates wouldn't create any new charges even if the charge amount is suposed to be greater than 0.

I suggest to replace the return null statement with a continue statement, to continue iterating the entry set

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
angelboxes: I made another comment why null shouldn't be returned by the function applyChargeToOverdueLoanInstallment, so I suggest to also remove this if statement.

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/SchedulerJobsTestResults.java
```diff
@@ -724,6 +724,59 @@ public void testApplyPenaltyForOverdueLoansJobOutcome() throws InterruptedExcept
...
+
+        loanStatusHashMap = this.loanTransactionHelper.disburseLoan(AccountTransferTest.LOAN_APPROVAL_DATE_PLUS_ONE, loanID);
+        LoanStatusChecker.verifyLoanIsActive(loanStatusHashMap);
+
+        ArrayList<HashMap> repaymentScheduleDataBefore = this.loanTransactionHelper.getLoanRepaymentSchedule(this.requestSpec,
+                this.responseSpec, loanID);
+
+        String JobName = "Apply penalty to overdue loans";
+        this.schedulerJobHelper.executeJob(JobName);
+
```
angelboxes, line 762: Executing the "Apply penalties" job doesn't guarantee you that the created loan is processed by the time this test finishes its run. It will only start the job. Before checking the loan charges, you should check either if the job finished or if the loan was processed. This job will also check penalties for any other active loan with charges and depending on loan configuration and  the date it was created, this could take some time.

angelboxes (changes requested):
I added some comments about the changes made in the applyChargeToOverdueLoanInstallment function that should be addressed just in case.

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/SchedulerJobsTestResults.java
mohitsinha: Yes, I totally agree with this. The Job usually takes some time when the number of loans have increased.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
mohitsinha: Currently we have only these types of `ChargeCalculationType`.
```    FLAT(1, "chargeCalculationType.flat"), //
    PERCENT_OF_AMOUNT(2, "chargeCalculationType.percent.of.amount"), //
    PERCENT_OF_AMOUNT_AND_INTEREST(3, "chargeCalculationType.percent.of.amount.and.interest"), //
    PERCENT_OF_INTEREST(4, "chargeCalculationType.percent.of.interest"),
    PERCENT_OF_DISBURSEMENT_AMOUNT(5,"chargeCalculationType.percent.of.disbursement.amount");
```
Even though it seems, that the amount can be different for a different iteration, in fact it is always the same.

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/SchedulerJobsTestResults.java
mohitsinha: In fact, this check should be there in all the tests which rely on jobs to be completed.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
angelboxes: Ok that may be right I'm just saying it as I don't take anything for granted and also it will avoid returning a null value as the overdueDTO.

Also, something similiar it could happen in the applyOverdueChargesForLoan function with the if statement:

if(overdueDTO == null){
     return;
}

Because you are iterating a list of multiple overdue installments for the same loan. If the penalty charge is calculated based on the installment interest amount, that amount can be different for each one of those installments, for the first one the calculated amount for all the installment charges can be 0, but for the next ones it could be a value greater than zero. Making a return call would not calculate charges for the next overdue installments.

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/SchedulerJobsTestResults.java
angelboxes: > In fact, this check should be there in all the tests which rely on jobs to be completed.

I agree,, in this case this the only way that loan penalty charges can be created,, but the job running time is something we should be concerned, by personal experience I can tell you that a loan penalty charge with daily  frequency could take forever specially if the overdue date is very old.

mohitsinha: True. I made a change for that.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
mohitsinha: I have made those changes.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
mohitsinha: I have made those changes.

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/SchedulerJobsTestResults.java
angelboxes: Ok, it seems right.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
angelboxes: Ok, it seems right. 
You can be right about charge calculations as actually I'm not too concerned about how exactly charges amount are calculated if they always return the same amount, but skipping every other loan overdue installments could also ignore some that may have a different charge definition and this needed to be addressed at least in this function. Thanks.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
angelboxes: It seems right. Thanks

angelboxes (approved):

vorburger:
@angelboxes sorry I forgot if you are a committer? If yes, please do go ahead and merge PRs, such as this, and others, which LGTY... you don't have to wait e.g. for me; on the contrary, I very much hope for more active other committers who review and merge.

awasum:
Go for it @angelboxes .

angelboxes:
> @angelboxes sorry I forgot if you are a committer? If yes, please do go ahead and merge PRs, such as this, and others, which LGTY... you don't have to wait e.g. for me; on the contrary, I very much hope for more active other committers who review and merge.

Yes I'm a committer. 

Wow!, I feel hesitant as I always like to see if there's something left to be said by someone else, But if I have the go-ahead by you I will merge it right now.

awasum:
There is always revert...

PR #618, opened 2019-07-09
add spring-boot-starter-actuator dependency

vorburger:
see https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-endpoints.html

### DIFF
--- a/fineract-provider/dependencies.gradle
+++ b/fineract-provider/dependencies.gradle
@@ -37,6 +37,8 @@ dependencies {
                 [group: 'org.springframework.boot', name: 'spring-boot-starter-web', version: springBootVersion],
                 [group: 'org.springframework.boot', name: 'spring-boot-starter-data-jpa', version: springBootVersion],
                 [group: 'org.springframework.boot', name: 'spring-boot-starter-security', version: springBootVersion],
+                [group: 'org.springframework.boot', name: 'spring-boot-starter-actuator', version: springBootVersion],
+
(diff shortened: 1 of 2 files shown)


vorburger:
see #619 for one possible use of this - but I think it's handy more generally.

Would anyone have any concerns / objections to adding this?

mohitsinha:
The endpoints exposed by this particular version of actuator should be safe. Or, we can manually configure the endpoints that we need.

vorburger:
Travis CI Build Failure is totally unrelated to this change, let's discuss that in [FINERACT-779](https://jira.apache.org/jira/browse/FINERACT-779) ...

vorburger:
> Travis CI Build Failure is totally unrelated to this change, let's discuss that in [FINERACT-779](https://jira.apache.org/jira/browse/FINERACT-779) ...

No, it actually was due to this change, see details in [FINERACT-779](https://jira.apache.org/jira/browse/FINERACT-779); I've now amended this PR with a change to the logging configuration which should hopefully make this PR pass the Travis build.

vorburger:
@awasum @conradsp review and merge this one, whenver you have a moment? (Not urgent.)

conradsp:
Looks good - but wondering if we should be logging at debug level for the actuator? Should we change that to info?

vorburger:
That logs too much, and fails Integration Test on Travis .. (more recent
Spring versions changed that particular log from info to debug).

conradsp:
Thanks for the clarification. I will merge.

PR #619, opened 2019-07-09
make integrationTest await application health readyness

vorburger:
This was an idea I had when wondering if the problem we hit during the Gradle upgrade was just that Tomcat took longer to boot Fineract (it turned out that that wasn't the cause).
    
I'm not sure if gradle-tomcat-plugin already does the equivalent (but suspect it does not), but given that integrationTests work, not entirely clear how much value this really adds.
    
Given that it doesn't really hurt either, perhaps we can add it anyway.

### DIFF
--- a/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/Utils.java
+++ b/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/Utils.java
@@ -28,17 +28,24 @@ import java.io.File;
 import java.text.DateFormat;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
-import java.util.*;
+import java.util.Calendar;
+import java.util.Locale;
+import java.util.Random;
+import java.util.TimeZone;
+
+import com.jayway.restassured.RestAssured;
+import com.jayway.restassured.path.json.JsonPath;
+import com.jayway.restassured.response.Response;
+import com.jayway.restassured.specification.RequestSpecification;
+import com.jayway.restassured.specification.ResponseSpecification;
 
 import org.apache.commons.lang.StringUtils;
 import org.apache.http.conn.HttpHostConnectException;
 import org.joda.time.DateTimeZone;
 import org.joda.time.LocalDate;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-import com.jayway.restassured.RestAssured;
-import com.jayway.restassured.path.json.JsonPath;
-import com.jayway.restassured.specification.RequestSpecification;
-import com.jayway.restassured.specification.ResponseSpecification;
 /**
  * Util for RestAssured tests. This class here in src/integrationTest is
  * copy/pasted to src/test; please keep them in sync.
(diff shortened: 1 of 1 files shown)


mohitsinha:
I don't really know the endpoints exposed by actuator by default for this particular version.
Just to be sure, I think we should only enable those actuator endpoints that we need.

michaelbukachi:
@vorburger I noticed that you have introduced files from the `ch.vorburger.mariadb4j` package in `MariaDB4jDataSourceConfiguration` and  `MariaDB4jSetupService`. However, the dependencies have not been added to `dependencies.gradle` or `dev-dependencies.gradle`. Is this due to the licensing issues? I'm running try to run individual tests using IntelliJ but can't proceed. Running all the tests using `.\gradlew clean integrationTest` works fine though. This is from the latest commit on the **develop** branch.

vorburger:
> I don't really know the endpoints exposed by actuator by default for this particular version.
> Just to be sure, I think we should only enable those actuator endpoints that we need.

#618 addressed this concern, agreed? Do you have any objection to merging this PR as it is now?

> @vorburger I noticed that you have introduced files from the `ch.vorburger.mariadb4j` package

@michaelbukachi yeah that stuff could do with some clean-up, but... that's totally unrelated to this PR? :smile:

michaelbukachi:
Haha. Sorry about that. I just didn't know how to reach you.

vorburger:
> Haha. Sorry about that. I just didn't know how to reach you.

dev@fineract.apache.org ?  :smiley_cat:

mohitsinha:
@vorburger Should we restrict the endpoints exposed by the actuator to the ones that are safe (/health)?

vorburger:
> @vorburger Should we restrict the endpoints exposed by the actuator to the ones that are safe (/health)?

I haven't actually had a closer look.. I'm kind of going on the assumption (hoping) that Spring Boot does not expose anything un-safe out of the box... but you may know more about this - is there anything which could cause concern? I thought [your comment on Jul 9 over in #618](https://github.com/apache/fineract/pull/618#issuecomment-509774541) meant that we were OK re. spring-boot-starter-actuator? Anyway, this PR only makes the integrationTest await application health readyness  - the actuator is already active, since #618.

Do you have any objection to merging this PR as it is now?

vorburger:
@awasum @Anh3h @conradsp if @mohitsinha raises no objections by let's say tomorrow, wanna review and if OK merge this? Tx!

awasum:
Spring Boot Actuator 1.2: https://docs.spring.io/autorepo/docs/spring-boot/1.2.4.RELEASE/reference/htmlsingle/#production-ready

Latest default exposed endpoints: https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-endpoints.html

I dont really think this default will cause problems. I have not reviewed it closely. Please, Can someone find the time to dig deeper? @mohitsinha  @michaelbukachi

mohitsinha:
@awasum  Thanks.
I have no further concerns regarding this PR.

mohitsinha (approved):
Looks good.

cnangeh:
ok. Merging this now.

PR #625, opened 2019-07-29
Update Dockerfile to use user-provided MySQL IP

cnangeh:
- Update Dockerfile to use user-provided MySQL server IP.
- Update README.md to provide instructions to run Fineract container with remote MySQL server.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,13 +21,15 @@ RUN mkdir fineract
 COPY . fineract
 
 WORKDIR fineract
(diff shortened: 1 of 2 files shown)


conradsp:
This looks good to me. Thanks for your work on it.

vorburger:
This seems a useful great first step (thanks @Anh3h) for https://issues.apache.org/jira/browse/FINERACT-773 for which IMHO we should eventually find an even better and more permanent solution... in an ideal world, one should not have to rebuild the container to have it connect to a different DB.

PR #627, opened 2019-08-22
rm hardcoded system time from .travis.yml (fixes FINERACT-788)

vorburger:
see https://issues.apache.org/jira/browse/FINERACT-788

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,14 +27,10 @@ jdk:
 services:
   - mysql
 
-# Hardcoding system time as a temporary fix for running buggy test cases
 before_install:
   - echo "USE mysql;\nUPDATE user SET password=PASSWORD('mysql') WHERE user='root';\nFLUSH PRIVILEGES;\n" | mysql -u root
   - mysql -u root -pmysql -e 'CREATE DATABASE IF NOT EXISTS `mifosplatform-tenants`;'
   - mysql -u root -pmysql -e 'CREATE DATABASE IF NOT EXISTS `mifostenant-default`;'
-  - export TZ=Asia/Kolkata
-  - sudo service ntp stop
-  - sudo date --set="23 February 2019 01:02:03"
 
 # https://docs.travis-ci.com/user/languages/java/#caching
 before_cache:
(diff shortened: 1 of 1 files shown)


vishwasbabu:
@vorburger , the date was hardcoded due to https://issues.apache.org/jira/browse/FINERACT-722 and the timezone was hardcoded due to https://issues.apache.org/jira/browse/FINERACT-723 .

vorburger:
> @vorburger , the date was hardcoded due to https://issues.apache.org/jira/browse/FINERACT-722 and the timezone was hardcoded due to https://issues.apache.org/jira/browse/FINERACT-723 .

Hm, OK but between "Integration test cases fail when run on the 1st of a month" (FINERACT-722) and "Travis is just completely broken right now" (FINERACT-788) may be the former is the lesser of two evils? :smiling_imp: Or is there another better quick fix here that anyone can think of?

I guess we could keep (just) the `export TZ=Asia/Kolkata` - would that be better?

vishwasbabu:
> Hm, OK but between "Integration test cases fail when run on the 1st of a month" (FINERACT-722) and "Travis is just completely broken right now" (FINERACT-788) may be the former is the lesser of two evils? 😈 Or is there another better quick fix here that anyone can think of?
> 
> I guess we could keep (just) the `export TZ=Asia/Kolkata` - would that be better?

Yes of course :) . We could also hard-code the date to "21 August 2019 01:02:03" and hope FINERACT-722 is fixed within the next 5 months !

vorburger:
> > I guess we could keep (just) the `export TZ=Asia/Kolkata` - would that be better?

In a 2nd commit to this PR, I've just reverted that removal, so we keep `TZ` but rm `date`.

> We could also hard-code the date to "21 August 2019 01:02:03" 

Nah, that just postpones the problem - the same issue will just re-occur when services.gradle.org renews their cert the next time around... I really think we should avoid hard-coding a fixed time here (because of this problem).

> and hope FINERACT-722 is fixed within the next 5 months !

hope dies last! :smile_cat: 

I'll self merge this in 48h, unless someone else does before or suggests a better solution.

awasum:
See new PR: https://github.com/apache/fineract/pull/914 which may be related to this...

PR #633, opened 2019-09-04
fix .travis.yml so that integrationTest does not run if test already failed

vorburger:


awasum:
Makes sense. preventing unnecessary execution.

This will really be used once a test fails.

Merge this?

vorburger:
> Merge this?

yes, please! :smile_cat: Tx.

vorburger:
actually wait a second - have you noticed that Travis hasn't run on this PR? That's weird...

awasum:
Yeah... maybe leave this and instead merge #634  which contains this commit and one addition commit

vorburger:
>  have you noticed that Travis hasn't run on this PR? That's weird...

so Travis runs just fine on #635 .. I think this as is originally is wrong - the `\` line break probably doesn't work - let me change this to be all on a single line instead.

vorburger:
@awasum fixed it - looks good to you now - merge?

awasum:
Yes.

PR #634, opened 2019-09-04
remove clean in .travis.yml

vorburger:


awasum:
Is there a duplicate in the other #633 ? maybe we need to close that and instead just merge this?

Anyway..its trivial and you can handle it

vorburger:
> Is there a duplicate in the other #633 ? maybe we need to close that and instead just merge this?

Sorry for the confusion. This PR this removes the clean, in its 2nd commit, but it builds on top of #633 which is the first commit... I raise 2 separate PRs for separate review - but it's probably just confusing... :smile: GitHub, contrary to Gerrit, doesn't support this kind of workflow very well - if you want to make 2 changes which touch the same lines of code. (If you raise two separate PRs from master, then one will nessarily conflict with the other, which is just overhead.)

What I don't understand is why Travis CI doesn't run for these 2 PRs... or is it just overloaded? Let's wait a bit.

awasum:
@vorburger  with #633 merged, this seems to have conflicts, want to pull from develop and fix this?

vorburger:
> @vorburger with #633 merged, this seems to have conflicts, want to pull from develop and fix this?

@awasum done

awasum:
Thank @vorburger

PR #638, opened 2019-09-16
Added modifications to allow MySQL 8 compatibility (tested on MySQL…

nemux:
* Added modifications to allow MySQL 8 compatibility (tested on MySQL 8.0.17)


vorburger:
This PR as-is currently seems to fail a number of integration tests in the build. While that could perhaps be fixed, the bigger problem here is the switch from the `org.drizzle.jdbc.DrizzleDriver` to the `com.mysql.jdbc.Driver` ... while that would be nice, there has been pushback in the past, because ASF hosted projects can, apparently, not distribute the LGPL licensed `com.mysql.jdbc.Driver`... is switching this a must have to allow MySQL 8 compatibility? If that's just something you "also did" as part of this PR, it would be much better if you could remove those changes again, and only do the minimum necessary to make it work with MySQL 8 ...  the best way to "prove" that it does would be to include a change to the https://github.com/apache/fineract/blob/develop/.travis.yml to make it use a newer version of the mysql service.

vorburger (changes requested):
NOK as is, as per comment posted to Conversation above

vorburger:
Let's close this now, just to have less open PRs; you can open a new PR with this when you resolved the feedback and conflicts.

vorburger:
==> https://issues.apache.org/jira/browse/FINERACT-762 (incl. https://github.com/apache/fineract/pull/757)

awasum:
> ==> https://issues.apache.org/jira/browse/FINERACT-762 (incl. #757)

PR: https://github.com/apache/fineract/pull/757 now merged. Thanks for the hint @nemux

vorburger:
==> https://issues.apache.org/jira/browse/FINERACT-901

PR #516, opened 2019-02-11
FINERACT-699 - new configuration parameters 'fromEmail' and 'fromName'

aasaru:

PR #639, opened 2019-09-24
Enhancement for FINERACT #516

MexinaD:
## Description
https://issues.apache.org/jira/browse/FINERACT-516

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/useradministration/api/UsersApiResource.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/useradministration/api/UsersApiResource.java
@@ -64,7 +64,7 @@ public class UsersApiResource {
      * {@link AppUserData}.
      */
     private final Set<String> RESPONSE_DATA_PARAMETERS = new HashSet<>(Arrays.asList("id", "officeId", "officeName", "username",
-            "firstname", "lastname", "email", "allowedOffices", "availableRoles", "selectedRoles", "staff"));
+            "firstname", "lastname", "email", "allowedOffices", "availableRoles", "selectedRoles", "staff", "currentPassword"));
 
     private final String resourceNameForPermissions = "USER";
 
--- a/fineract-provider/src/main/java/org/apache/fineract/useradministration/domain/AppUser.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/useradministration/domain/AppUser.java
@@ -230,7 +230,7 @@ public class AppUser extends AbstractPersistableCustom<Long> implements Platform
         final String passwordEncodedParamName = "passwordEncoded";
         if (command.hasParameter(passwordParamName)) {
             if (command.isChangeInPasswordParameterNamed(passwordParamName, this.password, platformPasswordEncoder, getId())) {
-                final String passwordEncodedValue = command.passwordValueOfParameterNamed(passwordParamName, platformPasswordEncoder,
+                String passwordEncodedValue = command.passwordValueOfParameterNamed(passwordParamName, platformPasswordEncoder,
                         getId());
                 actualChanges.put(passwordEncodedParamName, passwordEncodedValue);
                 updatePassword(passwordEncodedValue);
@@ -329,7 +329,7 @@ public class AppUser extends AbstractPersistableCustom<Long> implements Platform
         return actualChanges;
     }
 
-    private String[] getRolesAsIdStringArray() {
+    public String[] getRolesAsIdStringArray() {
         final List<String> roleIds = new ArrayList<>();
 
         for (final Role role : this.roles) {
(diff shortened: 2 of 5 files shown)


fineract-provider/src/main/java/org/apache/fineract/useradministration/domain/AppUser.java
```diff
@@ -230,7 +230,7 @@ public void updateRoles(final Set<Role> allRoles) {
         final String passwordEncodedParamName = "passwordEncoded";
         if (command.hasParameter(passwordParamName)) {
             if (command.isChangeInPasswordParameterNamed(passwordParamName, this.password, platformPasswordEncoder, getId())) {
-                final String passwordEncodedValue = command.passwordValueOfParameterNamed(passwordParamName, platformPasswordEncoder,
```
vorburger, line 233: why remove this `final` ? It doesn't seem to do any harm, and prevents the variable from being changed again by mistake.

fineract-provider/src/main/java/org/apache/fineract/useradministration/service/AppUserWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -216,6 +217,28 @@ public CommandProcessingResult updateUser(final Long userId, final JsonCommand c
...
             }
 
+            final String passwordParamName = "password";
+            final String currentPasswordParamName = "currentPassword";
+            final String currentPasswordEncodedValue = command.passwordValueOfParameterNamed(currentPasswordParamName, platformPasswordEncoder,
+                        userToUpdate.getId());
+            AppUser currentUser = getAppUserIfPresent();
+
+            if (command.hasParameter(passwordParamName)) {
+            for(String id : currentUser.getRolesAsIdStringArray()) {
```
vorburger, line 227: nit pick: Seems to have wrong indention?

fineract-provider/src/main/java/org/apache/fineract/useradministration/service/AppUserWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -216,6 +217,28 @@ public CommandProcessingResult updateUser(final Long userId, final JsonCommand c
             	clients = this.clientRepositoryWrapper.findAll(clientIds);
             }
 
+            final String passwordParamName = "password";
+            final String currentPasswordParamName = "currentPassword";
+            final String currentPasswordEncodedValue = command.passwordValueOfParameterNamed(currentPasswordParamName, platformPasswordEncoder,
```
vorburger, line 222: would it make sense to rather do this only inside the `if (command.hasParameter(passwordParamName))` ?

fineract-provider/src/main/java/org/apache/fineract/useradministration/service/AppUserWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -216,6 +217,28 @@ public CommandProcessingResult updateUser(final Long userId, final JsonCommand c
...
+                        userToUpdate.getId());
+            AppUser currentUser = getAppUserIfPresent();
+
+            if (command.hasParameter(passwordParamName)) {
+            for(String id : currentUser.getRolesAsIdStringArray()) {
+                if (!id.equals("1")){
```
vorburger, line 228: I'm not familiar enough with the code to know what a Role as ID = "1" means... perhaps an inline comment could help to clarify this to future readers of this code?

vorburger (changes requested):

vorburger:
@MexinaD thanks for this! I have some (very minor) feedback, and would be happy to merge this when that's addressed. In an ideal world, it would be really really cool if there also was a new Integration Test to cover future non-regression for this Improvement ... do you think you could provide that as well?

fineract-provider/src/main/java/org/apache/fineract/useradministration/api/UsersApiResource.java
```diff
@@ -64,7 +64,7 @@
      * {@link AppUserData}.
      */
     private final Set<String> RESPONSE_DATA_PARAMETERS = new HashSet<>(Arrays.asList("id", "officeId", "officeName", "username",
-            "firstname", "lastname", "email", "allowedOffices", "availableRoles", "selectedRoles", "staff"));
+            "firstname", "lastname", "email", "allowedOffices", "availableRoles", "selectedRoles", "staff", "currentPassword"));
```
vorburger, line 67: is this secure and safe? From what (little) I understand, this will expose any user's current password in the API (to other accounts with admin permissions, I guess, but still) which perhaps doesn't seem like such a great idea? Passwords usually are "write only", and then verified, but never returned back out of a system.

@awasum @avikganguly01 @vishwasbabu or anyone else reading this, care to chime in if I misunderstand this and it's OK or if this is No Go?

vorburger (changes requested):

fineract-provider/src/main/java/org/apache/fineract/useradministration/api/UsersApiResource.java
mohitsinha: I think this should be avoided.

mohitsinha (commented):
I think we should avoid sending back sensitive data.

vorburger:
@MexinaD I was curious if you wanted to address the feedback, or abandon this?

I'll close this PR on 2020.01.31 if there are no updates until then; FYI @edcable.

PR #643, opened 2019-10-11
Add Z,ro flags to Docker Compose Volume

vorburger:
This is required on secure Linux distributions such as Fedora/CentOS/RHEL where SELinux is enabled, where without this the mysql container fails to start due to:

    ls: cannot open directory '/docker-entrypoint-initdb.d/': Permission denied

These Z,ro volume mount flags are (should be..) harmless and ignored by Docker on other platforms, such as insecure ;) Linux distributions without SELinux, or Windows and Mac.

PS: I personally don't actually use real Docker, but instead I use Podman (see https://github.com/containers/libpod; install via `dnf install podman-docker`) and https://github.com/containers/podman-compose/) via `pip3 install podman-compose`; it's more secure (because it doesn't need a daemon and root to run containers).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -21,7 +21,7 @@ services:
   fineractmysql:
     image: mysql:5.7
     volumes:
-      - ./fineract-db/docker:/docker-entrypoint-initdb.d
+      - ./fineract-db/docker:/docker-entrypoint-initdb.d:Z,ro
     restart: always
     environment:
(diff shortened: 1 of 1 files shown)


vorburger:
@conradsp @Anh3h wanna review and merge this? Tx!

conradsp:
> @conradsp @Anh3h wanna review and merge this? Tx!

Looks good - thanks!

PR #644, opened 2019-10-11
Add docker-compose non-regression test to Travis CI

vorburger:


vorburger:
@conradsp I have a feeling you'll like this? :smile:  (If it works, let's see how Travis likes it...)

conradsp:
> @conradsp I have a feeling you'll like this? 😄 (If it works, let's see how Travis likes it...)

It looks like port 3306 is already bound on the CI machine. I suppose we could use another port for running CI in docker.

vorburger:
> It looks like port 3306 is already bound on the CI machine. 
> I suppose we could use another port for running CI in docker.

Yeah.. but that's not completely trivial, it will need [FINERACT-773](https://issues.apache.org/jira/browse/FINERACT-773), which actually I've started looking into (for another reason). But In the meantime, let me try if we could just `sudo service mysql stop` before `docker-compose` ...

vorburger:
@conradsp there were actually a couple of more problems here, but I managed to get it all working nicely now... could you review and merge this one? Thanks!

vorburger:
@awasum if you have time & interest, perhaps you'd also like to review this one?

PR #648, opened 2019-10-15
Make it possible to customize JDBC URLs in Docker Image at run-time instead of at build (FINERACT-773)

vorburger:
see [FINERACT-773](https://issues.apache.org/jira/browse/FINERACT-773)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,4 +62,5 @@ script:
 # @see https://docs.travis-ci.com/user/job-lifecycle/#customizing-the-build-phase
   - ./gradlew --console=plain licenseMain licenseTest licenseIntegrationTest test  &&  ./gradlew --console=plain integrationTest
 # We must stop mysql the mysql system service to avoid port 3306 conflicts (unless we run the mysql in docker-compose on another port; req. FINERACT-773)
-  - sudo service mysql stop  &&  docker-compose build  &&  docker-compose up -d  &&  sleep 30s  &&  http --verify=no --timeout 240 --check-status get https://localhost:8443/fineract-provider/actuator/health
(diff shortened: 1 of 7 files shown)


vorburger:
@Anh3h @conradsp please review (once the build passes) and merge if OK for you... this is the follow-up to the earlier #593, #595 and #625.

I've locally tested this, but if you want to be sure that `docker compose` still works, we can first get #644 merged and then rebase this one on that, to "prove" it does (because Travis CI will verify these changes, after #644).

vorburger:
@awasum if you have time & interest, perhaps you'd also like to review this one?

edcable:
@awasum Are you able to review this PR?

vorburger:
@awasum @Anh3h @conradsp would you like to review this?  I'd love to still see this make it into 1.4.0.

awasum:
> @awasum @Anh3h @conradsp would you like to review this? I'd love to still see this make it into 1.4.0.

On..it...let Travis run.

awasum:
Why is this failing? Seems the actuator health endpoint is not found?

vorburger:
> Why is this failing? Seems the actuator health endpoint is not found?

very curious, because locally it works (I've just re-tested it).  But I just had a "fun" idea how we'll (hopefully) be able to see the real reason for such failures much more easily (by adding `|| docker logs`) - let's see if that helps to clarify why this fails on Travis - build is currently running...

vorburger:
@awasum huh, now it passed... review? :smile: May be that 30s isn't (always) enough.. if this happens more often, then we should increase that. (But the new `|| docker logs` still seems like a neat idea, to help understand future problems, so I'm leaving that in.)

awasum:
Thanks for fixing this Mike, Merging now.

vorburger:
This actually broke more than it improved but #690 fixes it again.

PR #657, opened 2019-11-25
reduce excessive logging causing Travis CI problems (FINERACT-800)

vorburger:

PR #658, opened 2019-11-25
add .theia/settings.json

vorburger:
This avoids a prompt on http://gitpod.io.

PR includes #657, which is of course completely separate and unrelated, just so that this doesn't fail the build.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ build
 *.DS_Store
 .idea
 gradle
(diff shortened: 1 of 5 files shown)


vorburger:
@awasum would you merge this for me when the build passed?

awasum:
> @awasum would you merge this for me when the build passed?

Yes..I will do that...Thanks.

awasum (approved):
Looks good. Merging.

PR #660, opened 2019-11-27
FINERACT-803: validate username uniqueness

jamesidw:
## Description
Adds validation for uniqueness to the submitted username when creating/updating a user. Non-unique username was causing abrupt crash [see https://issues.apache.org/jira/browse/FINERACT-803?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel]

## Checklist

### DIFF
--- a/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/useradministration/users/UserHelper.java
+++ b/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/useradministration/users/UserHelper.java
@@ -32,19 +32,38 @@ public class UserHelper {
         return Utils.performServerPost(requestSpec, responseSpec, CREATE_USER_URL, getTestCreateUserAsJSON(roleId, staffId), "resourceId");
     }
 
+    public static Object createUser(final RequestSpecification requestSpec, final ResponseSpecification responseSpec, int roleId, int staffId, String username, String attribute) {
+        return Utils.performServerPost(requestSpec, responseSpec, CREATE_USER_URL, getTestCreateUserAsJSON(roleId, staffId, username), attribute);
+    }
+
     public static String getTestCreateUserAsJSON(int roleId, int staffId) {
-        String json = "{ \"username\": \"" + Utils.randomNameGenerator("User_Name_", 3)
+        return "{ \"username\": \"" + Utils.randomNameGenerator("User_Name_", 3)
                 + "\", \"firstname\": \"Test\", \"lastname\": \"User\", \"email\": \"whatever@mifos.org\","
                 + " \"officeId\": \"1\", \"staffId\": " + "\""
-                + Integer.toString(staffId)+"\",\"roles\": [\""
-                + Integer.toString(roleId) + "\"], \"sendPasswordToEmail\": false}";
-        System.out.println(json);
-        return json;
+                + staffId +"\",\"roles\": [\""
+                + roleId + "\"], \"sendPasswordToEmail\": false}";
+    }
 
+    private static String getTestCreateUserAsJSON(int roleId, int staffId, String username) {
+        return "{ \"username\": \"" + username
+            + "\", \"firstname\": \"Test\", \"lastname\": \"User\", \"email\": \"whatever@mifos.org\","
+            + " \"officeId\": \"1\", \"staffId\": " + "\""
+            + staffId +"\",\"roles\": [\""
+            + roleId + "\"], \"sendPasswordToEmail\": false}";
+    }
+
+    private static String getTestUpdateUserAsJSON(String username) {
+        return "{ \"username\": \"" + username
+            + "\", \"firstname\": \"Test\", \"lastname\": \"User\", \"email\": \"whatever@mifos.org\","
+            + " \"officeId\": \"1\"}";
     }
 
     public static Integer deleteUser(final RequestSpecification requestSpec, final ResponseSpecification responseSpec, final Integer userId) {
         return Utils.performServerDelete(requestSpec, responseSpec, createRoleOperationURL(userId), "resourceId");
+    }
+
+    public static Object updateUser(final RequestSpecification requestSpec, final ResponseSpecification responseSpec, int userId, String username, String attribute) {
+        return Utils.performServerPut(requestSpec, responseSpec, createRoleOperationURL(userId), getTestUpdateUserAsJSON(username), attribute);
     }
 
     private static String createRoleOperationURL(final Integer userId) {
--- a/fineract-provider/src/main/java/org/apache/fineract/useradministration/service/UserDataValidator.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/useradministration/service/UserDataValidator.java
@@ -32,6 +32,8 @@ import org.apache.fineract.infrastructure.core.data.DataValidatorBuilder;
 import org.apache.fineract.infrastructure.core.exception.InvalidJsonException;
 import org.apache.fineract.infrastructure.core.exception.PlatformApiDataValidationException;
 import org.apache.fineract.infrastructure.core.serialization.FromJsonHelper;
+import org.apache.fineract.useradministration.domain.AppUser;
+import org.apache.fineract.useradministration.domain.AppUserRepository;
 import org.apache.fineract.useradministration.domain.PasswordValidationPolicy;
 import org.apache.fineract.useradministration.domain.PasswordValidationPolicyRepository;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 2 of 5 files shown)


santoshconflux:
@jamesidw , Thanks for the PR.  @awasum or @vorburger , Could you kindly review this PR?

fineract-provider/src/main/java/org/apache/fineract/useradministration/service/UserDataValidator.java
```diff
@@ -173,6 +185,11 @@ public void validateForUpdate(final String json) {
         if (this.fromApiJsonHelper.parameterExists("username", element)) {
             final String username = this.fromApiJsonHelper.extractStringNamed("username", element);
             baseDataValidator.reset().parameter("username").value(username).notBlank().notExceedingLengthOf(100);
+            // this will be cumbersome if the username remains the same. Since the parameter seems to be optional, the client can work around it...[I hope]
+            AppUser exists = appUserRepository.findAppUserByName(username);
+            if (exists != null) {
```
awasum: This assumes the username will always be updated and will be different from the current one. In real life, the username usually stays the same when updating a user. Do the validation just for creation first and lets see if this resolves: https://issues.apache.org/jira/browse/FINERACT-803 as the issue occurs during creation.

awasum (changes requested):

fineract-provider/src/main/java/org/apache/fineract/useradministration/service/UserDataValidator.java
jamesidw: @awasum I'd changed the approach so we ignore the repeated username if it is the same as the current one but still flag the validation failure if the request would genuinely violate the unique constraint.

Let me just drop the validation in the update function as you've suggested.

awasum:
@santoshconflux  this requires functional testing. Can you pull this and test? TEst the edit user (change the username of an existing user) and let us know.

vorburger (approved):
LGTM, but I'll let @awasum merge it after @santoshconflux tested... 

@jamesidw in an ideal world you'd also contribute an integration test case for this, "proving" that changing existing users still works, and that new users with duplicate usernames cannot be created... is that something you'd be interested in contributing?

jamesidw:
@vorburger Sure thing. I'll give that a shot and update the PR.

santoshconflux:
@awasum , @vorburger , I tested locally pulling the code: 
We are getting the validation like this in create self service user page :
create self service user

Can  we get the meaningful validation like this which is shown while creating user by inputting duplicate username(Admin>Users>Create User or edit existing user with duplicate username)

edit user in Users

awasum:
> @awasum , @vorburger , I tested locally pulling the code:
> We are getting the validation like this in create self service user page :
> create self service user
> 
> Can we get the meaningful validation like this which is shown while creating user by inputting duplicate username(Admin>Users>Create User or edit existing user with duplicate username)
> 
> edit user in Users

@jamesidw  would you like to also look at this?

santoshconflux:
@jamesidw , In the ticket https://issues.apache.org/jira/browse/FINERACT-803, I mentioned that the validation should be "Username is already taken". That is also good. 
But let it be "User with username 'username' already exists" as in screenshot,  in order to maintain the uniformity.

awasum (approved):

jamesidw:
@awasum The error message should now match what @santoshconflux shared. 

I have a question though: what is being displayed in the screenshots is actually `userMessageGlobalisationCode` but it should be `defaultUserMessage` (at least that's what the other error messages seem to use). Is that a mistake on the frontend?

fineract-provider/src/main/java/org/apache/fineract/useradministration/service/UserDataValidator.java
```diff
@@ -75,6 +80,14 @@ public void validateForCreate(final String json) {
         final String username = this.fromApiJsonHelper.extractStringNamed("username", element);
         baseDataValidator.reset().parameter("username").value(username).notBlank().notExceedingLengthOf(100);
 
+        if (username != null) {
```
mohitsinha: I believe there is no point going inside this code block, if the username is blank/empty.
```suggestion
        if (StringUtils.isEmpty(username)) {
```
```suggestion
        if (StringUtils.isBlank(username)) {
```
@vorburger  @awasum What's your opinion on this?

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/useradministration/users/UserHelper.java
```diff
@@ -32,21 +32,46 @@ public static Integer createUser(final RequestSpecification requestSpec, final R
...
+
     public static String getTestCreateUserAsJSON(int roleId, int staffId) {
         String json = "{ \"username\": \"" + Utils.randomNameGenerator("User_Name_", 3)
                 + "\", \"firstname\": \"Test\", \"lastname\": \"User\", \"email\": \"whatever@mifos.org\","
                 + " \"officeId\": \"1\", \"staffId\": " + "\""
-                + Integer.toString(staffId)+"\",\"roles\": [\""
-                + Integer.toString(roleId) + "\"], \"sendPasswordToEmail\": false}";
+                + staffId +"\",\"roles\": [\""
+                + roleId + "\"], \"sendPasswordToEmail\": false}";
         System.out.println(json);
```
mohitsinha: Can we remove these print statements?

mohitsinha (changes requested):
Looks good to me. 
Some minor suggestions from my side.

fineract-provider/src/main/java/org/apache/fineract/useradministration/service/UserDataValidator.java
jamesidw: That makes sense. I've applied the optimization suggestion @mohitsinha

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/useradministration/users/UserHelper.java
jamesidw: Sure. I've removed them

vorburger:
I was about to merge this one, but then noticed that you @mohitsinha now also are a commiter, or are just about to become one?  Let's use this PR as the first opportunity for you to actually merge a contribution from someone yourself! @awasum I'll hold off from merging this to give @mohitsinha a chance to do this, perhaps you'd like to do the same.

awasum:
> I was about to merge this one, but then noticed that you @mohitsinha now also are a commiter, or are just about to become one? Let's use this PR as the first opportunity for you to actually merge a contribution from someone yourself! @awasum I'll hold off from merging this to give @mohitsinha a chance to do this, perhaps you'd like to do the same.

Yes. @mohitsinha go ahead, reviews this and merge.

mohitsinha (approved):

mohitsinha:
I will merge this, once I am done with the onboarding process

mohitsinha:
@jamesidw Can you combine all the 7 commits into 1?

jamesidw:
@mohitsinha I've squashed the commits

santoshconflux:
@mohitsinha , @jamesidw , @awasum . I just fetched the latest code to locally from develop branch. I am still getting the error: 
 "field is required
validation.msg.user.username.constraint.violation.duplicate.username -" 
Tested in incognito mode as well, but getting the above error.

mohitsinha:
Should we revert this change

jamesidw:
@santoshconflux please share the frontend repository you are using, I'd like to see what field is being displayed from the response

santoshconflux:
@jamesidw , I am using this one.
https://github.com/openMF/community-app from develop branch.

jamesidw:
Thanks @santoshconflux from this file [notificationWidget.js](https://github.com/openMF/community-app/blob/develop/app/scripts/modules/notificationWidget.js) it seems the correct approach is to use the `userMessageGlobalisationCode` instead. There is [already a code](https://github.com/openMF/community-app/blob/c48f89dc3397aaa16cad1cabfd6324e802e12fcd/app/global-translations/locale-en.json#L2372) that can work for this on the frontend so the fix would entail changing the `userMessageGlobalisationCode` being returned to match that.

@mohitsinha not sure what the correct approach is:

* Revert the changes and create a new PR?
* Create a new PR to fix what's already been merged?

Please advise

mohitsinha:
@jamesidw Let's first find the issue. 
Reverting the change should be more advisable as other users won't be affected.

I don't really have a problem if we add a new PR on top of the existing changes. If the bug is minor and won't affect other users, we can consider this option.

jamesidw:
@mohitsinha The issue as I understand it is: 

- The frontend uses the `userMessageGlobalisationCode` to create the message to be displayed. 
- I created a new `userMessageGlobalisationCode` in this PR and the frontend doesn't recognise it so it displays the `userMessageGlobalisationCode` text directly.

While looking through the frontend repository that @santoshconflux shared, I realised that the message he asked me to create already existed on the frontend...which meant it already existed on the backend as well. Indeed [this function](https://github.com/apache/fineract/blob/fb1e3d7c18fdcf208654f974dcb16f3eb34dad33/fineract-provider/src/main/java/org/apache/fineract/useradministration/service/AppUserWritePlatformServiceJpaRepositoryImpl.java#L343) was supposed to handle the exception and set the correct error response.

Let's revert the changes

mohitsinha:
@awasum @vorburger What's the standard way to revert a commit. 
Do I need to create a new PR after reverting the commit?
Or can this be reverted directly?

awasum:
> @awasum @vorburger What's the standard way to revert a commit.
> Do I need to create a new PR after reverting the commit?
> Or can this be reverted directly?

There should be a direct link or button somewhere on github to revert the PR.

awasum:
There is a revert button or link where you merged the PR...See
image

https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/reverting-a-pull-request

@santoshconflux  should describe exactly the test criteria so we all no when this will pass QA. Looks to me like the problem is understood but the QA criteria for a pass is not... I dont know if am making sense to you all.

PR #661, opened 2019-11-30
FINERACT-777 api / implementation scoped dependencies

percyashu:
https://issues.apache.org/jira/browse/FINERACT-777

## Description

## Checklist


percyashu:
@vorburger while changing compile to implement.
I tried testImplement and I got build errors of missing dependencies for some test. So I add  the require dependencies to testImplement and it stiil didnt build.
I had to add the dependencies to testCompile. 
Was this the right thing to do?

percyashu:
@vorburger can you review this?

fineract-provider/build.gradle
```diff
@@ -57,7 +57,7 @@ apply plugin: 'eclipse'
 apply plugin: 'idea'
 apply plugin: 'com.bmuschko.tomcat'
 apply plugin: 'project-report'
-apply plugin: 'java'
+apply plugin: 'java-library'
```
vorburger, line 60: @percyashu could you reply with a note for future code archeologists here why this change was needed?  As it seems to still work anyway, I'm inclined to merge this even without fully understanding what exactly this changes, if anything.

fineract-provider/dependencies.gradle
```diff
@@ -31,83 +31,103 @@ dependencies {
...
+
+    testCompile 'junit:junit:4.11',
+            'junit:junit-dep:4.11',
+            'org.mockito:mockito-core:1.9.5',
+            'com.jayway.restassured:rest-assured:2.3.3',
+            [group: 'com.mockrunner', name: 'mockrunner-jms', version: '1.0.6'],
+            [group: 'com.google.code.gson', name: 'gson', version: '2.2.4'],
+            [group: 'org.springframework', name:'spring-jms', version: '4.0.7.RELEASE'],
+            [group: 'joda-time', name: 'joda-time', version: '2.4'],
+            [  group: 'org.mnode.ical4j', name: 'ical4j', version: '1.0.4'],
```
vorburger, line 123: nit pick: wrong indention... but too minor to prevent merging this

fineract-provider/dependencies.gradle
```diff
@@ -31,83 +31,103 @@ dependencies {
...
+    )
+
+    implementation 'org.apache.bval:org.apache.bval.bundle:2.0.2'
+
+    testCompile 'junit:junit:4.11',
+            'junit:junit-dep:4.11',
+            'org.mockito:mockito-core:1.9.5',
+            'com.jayway.restassured:rest-assured:2.3.3',
+            [group: 'com.mockrunner', name: 'mockrunner-jms', version: '1.0.6'],
+            [group: 'com.google.code.gson', name: 'gson', version: '2.2.4'],
```
vorburger, line 120: I found it slightly surprising that we have to repeat the `api` dependencies here in `testCompile` ... there's probably a better way to do this with Gradle - does anyone know? I'll still merge this though - this always could (should) be cleaned up in a follow-up PR.

vorburger (approved):
LGTM - @percyashu Thank You for this (your first!) contribution!  More to come? :smiling_imp:

percyashu:
This change was needed because the old compile configuration now does not offer the guarantees that the api and implementation configurations provide.

Api configuration is used  to declare dependencies which will be transitively exposed to consumers of the library, and as such will appear on the compile classpath of consumers.

Implementation configuration should be used to declare dependencies which are internal to the component which will not be exposed to consumers, and therefore not leak into the consumers' compile classpath. This will also lead to faster compilation thanks to reduced classpath size.

For more info see https://docs.gradle.org/current/userguide/java_library_plugin.html

PR #642, opened 2019-10-11
FINERACT-724 Upgrade Spring Boot, Spring and Spring Security to their latest stable version

ivbond:
This is upgrade of Spring Boot, Spring, Spring Security, OpenJpa and other related libraries to their latest versions. 
All Integrations tests are passing with the new commits. But I would like to describe issues I have encountered during development:
1) Spring security does not support custom SaltSource. it is now random generated. So I used MessageDigestPasswordEncoder and V353__migrate_passwords_to_ss_5.sql for password migration.
2) There is huge trobule with OpenJpa and Spring5 compatibility. OpenJpaVendorAdapter was removed from Spring5. (See:  https://jira.spring.io/browse/SPR-16035?redirect=false&focusedCommentId=160186&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel) So I have upgraded OpenJpa and moved some code from spring 4 to fineract project. This solution has to be reviewed twice. I do not like it but this is the only way I have made it works. I have tried without jpaVendorAdapter but it leads to:
`"SQLException: Lock wait timeout exceeded; try restarting transaction"`

3) Alot of issues I had with dependencies.gradle and dev-dependencies.gradle they have alot of duplicates and dev is out of date. I think after Tomcat removal we could will leave only dependencies.gradle.

PR #662, opened 2019-11-30
FINERACT-797: Use Spring Boot BOM to avoid maintaining version numbers in dependencies.gradle

xurror:
FINERACT-797 - Spring Dependency Management
 - Update mockito to latest version
 - Move dependencies from dependencies.gradle to build.gradle
 - Removed implemented TODOs from XBRBuilderTest.java
https://issues.apache.org/jira/browse/FINERACT-797

## Description
In an ideal world, in a future PR, it would be really really cool if we could just avoid having to repeat using fixed version, and could use the implicit versions via Spring BOM dependency management, see https://docs.spring.io/spring-boot/docs/current/gradle-plugin/reference/html/#managing-dependencies ...

## Checklist

### DIFF
--- a/fineract-provider/build.gradle
+++ b/fineract-provider/build.gradle
@@ -21,11 +21,7 @@ Run as:
 gradle clean tomcatrunwar
 '''
 
-// project.ext.springBootVersion = '2.1.9.RELEASE'
-// project.ext.springVersion = '5.1.10.RELEASE'
-// project.ext.springOauthVersion = '2.3.6.RELEASE'
-// project.ext.jerseyVersion = '1.17'
-// project.ext.springDataJpaVersion = '2.2.0.RELEASE' // also change spring-boot-gradle-plugin version below
+project.ext.jerseyVersion = '1.17'
 
 buildscript {
   ext {
@@ -111,13 +107,13 @@ dependencyManagement {
         entry 'spring-boot-starter-actuator'
         entry 'spring-boot-starter-tomcat'
      }
-     dependencySet(group: 'com.sun.jersey', version: '1.17') {
+     dependencySet(group: 'com.sun.jersey', version: jerseyVersion) {
          entry 'jersey-core'
          entry 'jersey-servlet'
          entry 'jersey-server'
          entry 'jersey-json'
      }
-     dependencySet(group: 'com.sun.jersey.contribs', version: '1.17') {
+     dependencySet(group: 'com.sun.jersey.contribs', version: jerseyVersion) {
          entry 'jersey-spring'
          entry 'jersey-multipart'
      }
--- a/fineract-provider/build.gradle
+++ b/fineract-provider/build.gradle
@@ -21,11 +21,11 @@ Run as:
 gradle clean tomcatrunwar
 '''
 
-project.ext.springBootVersion = '2.1.9.RELEASE'
-project.ext.springVersion = '5.1.10.RELEASE'
-project.ext.springOauthVersion = '2.3.6.RELEASE'
-project.ext.jerseyVersion = '1.17'
-project.ext.springDataJpaVersion = '2.2.0.RELEASE' // also change spring-boot-gradle-plugin version below
+// project.ext.springBootVersion = '2.1.9.RELEASE'
+// project.ext.springVersion = '5.1.10.RELEASE'
+// project.ext.springOauthVersion = '2.3.6.RELEASE'
+// project.ext.jerseyVersion = '1.17'
+// project.ext.springDataJpaVersion = '2.2.0.RELEASE' // also change spring-boot-gradle-plugin version below
 
 buildscript {
   ext {
@@ -38,6 +38,7 @@ buildscript {
   }
 
   dependencies {
+     classpath 'io.spring.gradle:dependency-management-plugin:1.0.8.RELEASE'
      classpath 'com.bmuschko:gradle-tomcat-plugin:2.5',
                'nl.javadude.gradle.plugins:license-gradle-plugin:0.11.0',
                'org.zeroturnaround:gradle-jrebel-plugin:1.1.2',
@@ -48,6 +49,7 @@ buildscript {
   }
 }
 
+apply plugin: 'io.spring.dependency-management'
 apply plugin: "org.nosphere.apache.rat"
 apply plugin: 'rebel'
 apply plugin: 'license'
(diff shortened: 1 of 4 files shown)


xurror:
@vorburger please can you review this?

fineract-provider/build.gradle
```diff
@@ -21,11 +21,11 @@ Run as:
...
 '''
 
-project.ext.springBootVersion = '2.1.9.RELEASE'
-project.ext.springVersion = '5.1.10.RELEASE'
-project.ext.springOauthVersion = '2.3.6.RELEASE'
-project.ext.jerseyVersion = '1.17'
-project.ext.springDataJpaVersion = '2.2.0.RELEASE' // also change spring-boot-gradle-plugin version below
+// project.ext.springBootVersion = '2.1.9.RELEASE'
+// project.ext.springVersion = '5.1.10.RELEASE'
+// project.ext.springOauthVersion = '2.3.6.RELEASE'
```
vorburger: Let's just completely remove these 3 lines of Spring versions? (They are only used once further down.)

vorburger:
@ivan333m do you have any interest in helping to review this, given that this idea originally came up when I reviewed your PR #642? (I'll go over it as well and post some feedback from my side ASAP.)

fineract-provider/build.gradle
```diff
@@ -21,11 +21,11 @@ Run as:
...
+// project.ext.jerseyVersion = '1.17'
```
vorburger: Given that this Jersey version is repeated twice, how about keeping this single declaration, and using it in the 2 places below?

fineract-provider/build.gradle
```diff
@@ -21,11 +21,11 @@ Run as:
...
+// project.ext.springDataJpaVersion = '2.2.0.RELEASE' // also change spring-boot-gradle-plugin version below
```
vorburger: also just remove

fineract-provider/build.gradle
```diff
@@ -59,9 +61,83 @@ apply plugin: 'com.bmuschko.tomcat'
...
 apply plugin: 'openjpa'
+
 // apply plugin: 'pmd'
 // apply plugin: 'findbugs'
 
+dependencyManagement {
+  imports {
+    mavenBom 'org.springframework:spring-framework-bom:5.1.10.RELEASE'
+  }
+  dependencies {
```
vorburger, line 69: I'm probably misunderstanding the Spring BOM, but does it not also define (some of) these versions? Or is this list just the versions of all Fineract dependencies which the Spring BOM does not fix?

fineract-provider/src/test/java/org/apache/fineract/mix/report/service/XBRLBuilderTest.java
```diff
@@ -79,13 +79,10 @@ public void shouldCorrectlyBuildMap() {
             nodes = DocumentBuilderFactory.newInstance().newDocumentBuilder().parse(new ByteArrayInputStream(result.getBytes()))
                     .getElementsByTagName("Assets");
         } catch (final SAXException e) {
-            // TODO Auto-generated catch block
```
vorburger: This would ideally be raised as a separate PR (because it has nothing to do with the other changes proposed here), but it's so minor that it doesn't matter/ it's OK for me.

vorburger (changes requested):
Some minor feedback, and a question:

Would you be able to post a diff of a before/after dependency tree, so that we can see what versions, if any, will effectively change with this PR?

fineract-provider/build.gradle
xurror: Well, the spring framework-bom resolves all spring artifacts(like spring-core etc) without explicitly declaring them since they all use thesame version

It seems there is no spring-boot-bom to handle the spring boot artifacts and dependencies

xurror: My primary idea was to use generalized bom that handles dependencies for most of the packages we use and even more but the version conflicts were too big breaking builds

ivbond:
@xurror Nice changes, thanks. 

All points described bellow are OPTIONAL! 

 I would propose following:
1) According to the spring doc [spring doc](https://docs.spring.io/spring-boot/docs/current/gradle-plugin/reference/html/)
`When you apply the io.spring.dependency-management plugin, Spring Boot’s plugin will automatically import the spring-boot-dependencies bom from the version of Spring Boot that you are using....`. So you don't need to specify version on  spring boot `dependencySet`

2) As I know when you apply bom it controls all third-party dependencies that declared it the bom. Here is a [spring boot bom](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-dependencies/pom.xml). You can find spring and all related third-party dependencies there. Therefore we could exclude spring framework version and alot of third-party. 

You could generate spring boot application on Spring IO initializer and use it as an example.

fineract-provider/src/test/java/org/apache/fineract/mix/report/service/XBRLBuilderTest.java
xurror: True, I'll make a new PR for that

xurror:
> @xurror Nice changes, thanks.
> 
> All points described bellow are OPTIONAL!
> 
> I would propose following:
> 
> 1. According to the spring doc [spring doc](https://docs.spring.io/spring-boot/docs/current/gradle-plugin/reference/html/)
>    `When you apply the io.spring.dependency-management plugin, Spring Boot’s plugin will automatically import the spring-boot-dependencies bom from the version of Spring Boot that you are using....`. So you don't need to specify version on  spring boot `dependencySet`
> 2. As I know when you apply bom it controls all third-party dependencies that declared it the bom. Here is a [spring boot bom](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-dependencies/pom.xml). You can find spring and all related third-party dependencies there. Therefore we could exclude spring framework version and alot of third-party.
> 
> You could generate spring boot application on Spring IO initializer and use it as an example.

Thanks for this review, the new updates I have made should fix that

vorburger:
@xurror this now has conflicts, because I just merged #661 ... sorry about that. Would you mind to (manually...) resolve these conflicts?  There's really no way to avoid this, so I usually just merge cases like this in the order that such conflicting changes are proposed (e.g. in this csae, @percyashu #661 was before your #662 here). Thank You!

xurror:
Resolved conflicts with #661

fineract-provider/src/test/java/org/apache/fineract/mix/report/service/XBRLBuilderTest.java
```diff
@@ -79,10 +79,13 @@ public void shouldCorrectlyBuildMap() {
             nodes = DocumentBuilderFactory.newInstance().newDocumentBuilder().parse(new ByteArrayInputStream(result.getBytes()))
                     .getElementsByTagName("Assets");
         } catch (final SAXException e) {
+            // TODO Auto-generated catch block
```
vorburger: @xurror this looks weird and given (your) #663 I'm assuming is a mistake - remove?

```suggestion
```

fineract-provider/src/test/java/org/apache/fineract/mix/report/service/XBRLBuilderTest.java
```diff
@@ -79,10 +79,13 @@ public void shouldCorrectlyBuildMap() {
...
             e.printStackTrace();
         } catch (final IOException e) {
+            // TODO Auto-generated catch block
```
vorburger:
```suggestion
```

fineract-provider/src/test/java/org/apache/fineract/mix/report/service/XBRLBuilderTest.java
```diff
@@ -79,10 +79,13 @@ public void shouldCorrectlyBuildMap() {
...
             e.printStackTrace();
         } catch (final ParserConfigurationException e) {
+            // TODO Auto-generated catch block
```
vorburger:
```suggestion
```

fineract-provider/src/test/java/org/apache/fineract/mix/report/service/XBRLBuilderTest.java
vorburger: #663 did this, but in resolving the conflicts it looks like somehow this came back?

fineract-provider/build.gradle
```diff
@@ -59,9 +57,77 @@ apply plugin: 'com.bmuschko.tomcat'
...
+  imports {
+    mavenBom 'org.springframework:spring-framework-bom:5.1.10.RELEASE'
+  }
+  
+  dependencies {
+     dependency 'org.springframework.security.oauth:spring-security-oauth2:2.3.6.RELEASE'
+     dependency 'org.apache.openjpa:openjpa:3.1.0'
+     dependency 'com.squareup.retrofit:retrofit:1.6.1'
+     dependency 'com.squareup.okhttp:okhttp:2.0.0'
+     dependency 'com.squareup.okhttp:okhttp-urlconnection:2.0.0'
```
vorburger, line 74: see https://issues.apache.org/jira/browse/FINERACT-804 ...

fineract-provider/build.gradle
vorburger: > but the version conflicts were too big breaking builds

@xurror OK, then let's do that (you or someone else later) separately; I've just created new https://issues.apache.org/jira/browse/FINERACT-805 so that we don't forget to follow-up on that... (It's surely best to raise small PRs for each one of these, instead of one big one for everything together.)

vorburger (changes requested):
This would LGTM for me if you could just remove those re-introduced TODOs in `XBRLBuilderTest.java`?  Then I'm happy to merge this.  Note to myself, or anyone else (e.g. @awasum) who will merge this: We should "squash" the commits here into 1 via that GitHub UI drop-down merge button.

vorburger (approved):

PR #670, opened 2019-12-10
FINERACT-702: Integrate SpotBugs and fix over 300 high priority issues

awasum:
FINERACT-702: Integrate SpotBugs and fix over 300 high priority rule Violations

This PR also serves as a base solution for FINERACT-712 and FINERACT-696

Review thoroughly

## Description
https://issues.apache.org/jira/browse/FINERACT-702

## Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,7 +60,7 @@ script:
   - date
 # using "&&" instead of x2 "-" means that integrationTest does not run if test fails, which makes sense.
 # @see https://docs.travis-ci.com/user/job-lifecycle/#customizing-the-build-phase
-  - ./gradlew --console=plain licenseMain licenseTest licenseIntegrationTest test  &&  ./gradlew --console=plain integrationTest
+  - ./gradlew --console=plain licenseMain licenseTest licenseIntegrationTest check  &&  ./gradlew --console=plain integrationTest
 # We must stop mysql the mysql system service to avoid port 3306 conflicts (unless we run the mysql in docker-compose on another port; req. FINERACT-773)
   - sudo service mysql stop  &&  docker-compose build  &&  docker-compose up -d  &&  sleep 30s  &&  http --verify=no --timeout 240 --check-status get https://localhost:8443/fineract-provider/actuator/health || docker logs fineract_fineract-server_1
 # The || docker logs lets use see the root cause in case of failures
--- a/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/ClientLoanIntegrationTest.java
+++ b/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/ClientLoanIntegrationTest.java
@@ -18,27 +18,17 @@
  */
 package org.apache.fineract.integrationtests;
 
-import static org.junit.Assert.assertEquals;
-
-import java.math.BigDecimal;
-import java.text.DateFormat;
-import java.text.DecimalFormat;
-import java.text.SimpleDateFormat;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Calendar;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-
+import com.google.gson.JsonObject;
+import com.jayway.restassured.builder.RequestSpecBuilder;
+import com.jayway.restassured.builder.ResponseSpecBuilder;
+import com.jayway.restassured.http.ContentType;
+import com.jayway.restassured.path.json.JsonPath;
+import com.jayway.restassured.specification.RequestSpecification;
+import com.jayway.restassured.specification.ResponseSpecification;
 import org.apache.fineract.integrationtests.common.ClientHelper;
 import org.apache.fineract.integrationtests.common.SchedulerJobHelper;
 import org.apache.fineract.integrationtests.common.Utils;
-import org.apache.fineract.integrationtests.common.accounting.Account;
-import org.apache.fineract.integrationtests.common.accounting.AccountHelper;
-import org.apache.fineract.integrationtests.common.accounting.JournalEntry;
-import org.apache.fineract.integrationtests.common.accounting.JournalEntryHelper;
-import org.apache.fineract.integrationtests.common.accounting.PeriodicAccrualAccountingHelper;
+import org.apache.fineract.integrationtests.common.accounting.*;
 import org.apache.fineract.integrationtests.common.charges.ChargesHelper;
 import org.apache.fineract.integrationtests.common.loans.LoanApplicationTestBuilder;
 import org.apache.fineract.integrationtests.common.loans.LoanProductTestBuilder;
@@ -53,13 +43,13 @@ import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 
-import com.google.gson.JsonObject;
-import com.jayway.restassured.builder.RequestSpecBuilder;
-import com.jayway.restassured.builder.ResponseSpecBuilder;
-import com.jayway.restassured.http.ContentType;
-import com.jayway.restassured.path.json.JsonPath;
-import com.jayway.restassured.specification.RequestSpecification;
-import com.jayway.restassured.specification.ResponseSpecification;
+import java.math.BigDecimal;
+import java.text.DateFormat;
+import java.text.DecimalFormat;
+import java.text.SimpleDateFormat;
+import java.util.*;
+
+import static org.junit.Assert.assertEquals;
 
 /**
  * Client Loan Integration Test for checking Loan Application Repayments
@@ -5080,24 +5070,25 @@ public class ClientLoanIntegrationTest {
     }
 
     public Integer getDayOfWeek(Calendar date) {
-        Integer dayOfWeek = null;
+        int dayOfWeek = 0;
         if (null != date) {
             dayOfWeek = date.get(Calendar.DAY_OF_WEEK) - 1;
-            if (dayOfWeek.compareTo(0) == 0) {
+            if (dayOfWeek == 0) {
                 dayOfWeek = 7;
             }
         }
-        return dayOfWeek;
+        return Integer.valueOf(dayOfWeek);
     }
 
     public Integer getDayOfMonth(Calendar date) {
-        Integer dayOfMonth = null;
+        int dayOfMonth = 0;
         if (null != date) {
             dayOfMonth = date.get(Calendar.DAY_OF_MONTH);
-            if (dayOfMonth.compareTo(28) > 0) {
+            if (dayOfMonth == 28) {
                 dayOfMonth = 28;
             }
         }
-        return dayOfMonth;
+
+        return Integer.valueOf(dayOfMonth);
     }
 }
--- a/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/LoanReschedulingWithinCenterTest.java
+++ b/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/LoanReschedulingWithinCenterTest.java
@@ -151,7 +151,7 @@ public class LoanReschedulingWithinCenterTest {
 
         // VERIFY THE INTEREST
         Float interestDue = (Float) ((HashMap) loanRepaymnetSchedule.get(2)).get("interestDue");
-        assertEquals(String.valueOf(interestDue), "90.82");
+        assertEquals(String.valueOf(interestDue), "42.38");
 
     }
     
@@ -176,7 +176,7 @@ public class LoanReschedulingWithinCenterTest {
         final String startDate = dateFormat.format(today.getTime());
         final String frequency = "2"; // 2:Weekly
         final String interval = "2"; // Every one week
-        final Integer repeatsOnDay = today.get(Calendar.DAY_OF_WEEK) - 1;
+        final Integer repeatsOnDay = today.get(Calendar.DAY_OF_WEEK);
 
         Integer calendarId = CalendarHelper.createMeetingForGroup(this.requestSpec, this.responseSpec, centerId, startDate, frequency,
                 interval, repeatsOnDay.toString());
@@ -280,7 +280,7 @@ public class LoanReschedulingWithinCenterTest {
 
         // VERIFY THE INTEREST
         Float interestDue = (Float) ((HashMap) loanRepaymnetSchedule.get(2)).get("interestDue");
-        assertEquals(String.valueOf(interestDue), "41.05");
+        assertEquals(String.valueOf(interestDue), "19.16");
 
     }
 
--- a/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/CenterDomain.java
+++ b/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/CenterDomain.java
@@ -18,12 +18,11 @@
  */
 package org.apache.fineract.integrationtests.common;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-
+import com.google.gson.Gson;
 import org.apache.fineract.infrastructure.core.service.DateUtils;
 
-import com.google.gson.Gson;
+import java.util.ArrayList;
+import java.util.HashMap;
 
 public class CenterDomain implements Comparable<CenterDomain> {
 
@@ -45,7 +44,6 @@ public class CenterDomain implements Comparable<CenterDomain> {
                 final String name, final String externalId, final Integer staffId, final int officeID, final String officeName,
                 final String hierarchy, final ArrayList<HashMap> groupMembers) {
             this.id = id;
-            this.accountNo = accountNo;
             this.status = new HashMap();
             this.status.put("id", statusid);
             this.status.put("code", statuscode);
--- a/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/accounting/GLAccountBuilder.java
+++ b/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/accounting/GLAccountBuilder.java
@@ -18,12 +18,11 @@
  */
 package org.apache.fineract.integrationtests.common.accounting;
 
-import java.util.Calendar;
-import java.util.HashMap;
-
+import com.google.gson.Gson;
 import org.apache.fineract.integrationtests.common.Utils;
 
-import com.google.gson.Gson;
+import java.util.Calendar;
+import java.util.HashMap;
 
 public class GLAccountBuilder {
 
@@ -41,24 +40,24 @@ public class GLAccountBuilder {
     private static String name = Utils.randomStringGenerator("ACCOUNT_NAME_", 5);
 
     private static String GLCode = "";
-    private static String accountType = "";
-    private static String accountUsage = ACCOUNT_USAGE_DETAIL;
-    private static String manualEntriesAllowed = MANUAL_ENTRIES_ALLOW;
+    private String accountType = "";
+    private String accountUsage = ACCOUNT_USAGE_DETAIL;
+    private String manualEntriesAllowed = MANUAL_ENTRIES_ALLOW;
     private static String description = "DEFAULT_DESCRIPTION";
 
     public String build() {
         final HashMap<String, String> map = new HashMap<>();
         map.put("name", GLAccountBuilder.name);
         map.put("glCode", GLAccountBuilder.GLCode);
-        map.put("manualEntriesAllowed", GLAccountBuilder.manualEntriesAllowed);
-        map.put("type", GLAccountBuilder.accountType);
-        map.put("usage", GLAccountBuilder.accountUsage);
+        map.put("manualEntriesAllowed", manualEntriesAllowed);
+        map.put("type", accountType);
+        map.put("usage", accountUsage);
         map.put("description", GLAccountBuilder.description);
         return new Gson().toJson(map);
     }
 
     public GLAccountBuilder withAccountTypeAsAsset() {
-        GLAccountBuilder.accountType = ASSET_ACCOUNT;
+        accountType = ASSET_ACCOUNT;
         GLAccountBuilder.GLCode = Utils.randomStringGenerator("ASSET_", 2);
         GLAccountBuilder.GLCode += Calendar.getInstance().getTimeInMillis() + ""; // Added
         // unique
@@ -71,40 +70,40 @@ public class GLAccountBuilder {
     }
 
     public GLAccountBuilder withAccountTypeAsLiability() {
-        GLAccountBuilder.accountType = LIABILITY_ACCOUNT;
+        accountType = LIABILITY_ACCOUNT;
         GLAccountBuilder.GLCode = Utils.randomStringGenerator("LIABILITY_", 2);
         GLAccountBuilder.GLCode += Calendar.getInstance().getTimeInMillis() + "";
         return this;
     }
 
     public GLAccountBuilder withAccountTypeAsAsEquity() {
-        GLAccountBuilder.accountType = EQUITY_ACCOUNT;
+        accountType = EQUITY_ACCOUNT;
         GLAccountBuilder.GLCode = Utils.randomStringGenerator("EQUITY_", 2);
         GLAccountBuilder.GLCode += Calendar.getInstance().getTimeInMillis() + "";
         return this;
     }
 
     public GLAccountBuilder withAccountTypeAsIncome() {
-        GLAccountBuilder.accountType = INCOME_ACCOUNT;
+        accountType = INCOME_ACCOUNT;
         GLAccountBuilder.GLCode = Utils.randomStringGenerator("INCOME_", 2);
         GLAccountBuilder.GLCode += Calendar.getInstance().getTimeInMillis() + "";
         return this;
     }
 
     public GLAccountBuilder withAccountTypeAsExpense() {
-        GLAccountBuilder.accountType = EXPENSE_ACCOUNT;
+        accountType = EXPENSE_ACCOUNT;
         GLAccountBuilder.GLCode = Utils.randomStringGenerator("EXPENSE_", 2);
         GLAccountBuilder.GLCode += Calendar.getInstance().getTimeInMillis() + "";
         return this;
     }
 
     public GLAccountBuilder withAccountUsageAsHeader() {
-        GLAccountBuilder.accountUsage = ACCOUNT_USAGE_HEADER;
+        accountUsage = ACCOUNT_USAGE_HEADER;
         return this;
     }
 
     public GLAccountBuilder withMaualEntriesNotAllowed() {
-        GLAccountBuilder.manualEntriesAllowed = MANUAL_ENTRIES_NOT_ALLOW;
+        manualEntriesAllowed = MANUAL_ENTRIES_NOT_ALLOW;
         return this;
     }
 }
--- a/fineract-provider/src/main/java/org/apache/fineract/accounting/common/AccountingConstants.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/accounting/common/AccountingConstants.java
@@ -18,13 +18,13 @@
  */
 package org.apache.fineract.accounting.common;
 
+import org.apache.fineract.accounting.financialactivityaccount.data.FinancialActivityData;
+import org.apache.fineract.accounting.glaccount.domain.GLAccountType;
+
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-
-import org.apache.fineract.accounting.financialactivityaccount.data.FinancialActivityData;
-import org.apache.fineract.accounting.glaccount.domain.GLAccountType;
 
 public class AccountingConstants {
 
@@ -243,7 +243,7 @@ public class AccountingConstants {
         private final Integer value;
         private final String code;
         private final GLAccountType mappedGLAccountType;
-        private static List<FinancialActivityData> financialActivities;
+        private static List<FinancialActivityData> financialActivities = new ArrayList<>();
 
         private FINANCIAL_ACTIVITY(final Integer value, final String code, final GLAccountType mappedGLAccountType) {
             this.value = value;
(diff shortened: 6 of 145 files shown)


vorburger:
WOW! Nice.. really nice. Please ping me back here when the build is passing (I haven't even looked at why it's failing), I'd love to review and merge this PR!

fineract-provider/src/main/java/org/apache/fineract/organisation/monetary/data/CurrencyData.java
```diff
@@ -97,4 +89,23 @@ private String generateDisplayLabel() {
...
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (!(o instanceof CurrencyData)) return false;
+        CurrencyData that = (CurrencyData) o;
+        return decimalPlaces == that.decimalPlaces &&
+                code.equals(that.code) &&
+                name.equals(that.name) &&
+                inMultiplesOf.equals(that.inMultiplesOf) &&
```
vorburger: @awasum this is missing `null` checks, and it's also why the build is failing for you on this PR... just use [`Objects.equals()`](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#equals-java.lang.Object-java.lang.Object-) instead!

vorburger (changes requested):

fineract-provider/src/main/java/org/apache/fineract/organisation/monetary/data/CurrencyData.java
awasum: Thanks..Will handle this later on Friday.

vorburger:
@awasum I was curious why this still failed, so I pulled it locally, and the `docker compose` did still work for me locally... not entirely sure what's going on here, but looks like perhaps it was just way too slow - Travis may have been really busy?  I've raised #673, same thing (just squashed) plus slightly increased timeout - let's see if that one passes...

awasum:
> @awasum I was curious why this still failed, so I pulled it locally, and the `docker compose` did still work for me locally... not entirely sure what's going on here, but looks like perhaps it was just way too slow - Travis may have been really busy? I've raised #673, same thing (just squashed) plus slightly increased timeout - let's see if that one passes...

Looks..like the ShareAccount Integration tests are failing(locally) due to some equals or change I made while doing the spotbugs integration...I have not been able to pinpoint the problem further...But I will have a go at it again this evening. 

Anyone reading this who could help pinpoint the problem before me..that will be great. 

Last resort, I will revert the changes in the share account module and add one change at a time while run the integration tests till we get it.

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/ClientLoanIntegrationTest.java
```diff
@@ -5080,24 +5070,25 @@ private Integer applyForLoanApplicationWithProductConfigurationAsFalse(final Int
...
+        return Integer.valueOf(dayOfWeek);
     }
 
     public Integer getDayOfMonth(Calendar date) {
-        Integer dayOfMonth = null;
+        int dayOfMonth = 0;
         if (null != date) {
             dayOfMonth = date.get(Calendar.DAY_OF_MONTH);
-            if (dayOfMonth.compareTo(28) > 0) {
+            if (dayOfMonth == 28) {
```
vorburger:
```suggestion
            if (dayOfMonth > 28) {
```

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/CenterDomain.java
```diff
@@ -45,7 +44,6 @@ private Builder(final Integer id, final Integer statusid, final String statuscod
                 final String name, final String externalId, final Integer staffId, final int officeID, final String officeName,
                 final String hierarchy, final ArrayList<HashMap> groupMembers) {
             this.id = id;
-            this.accountNo = accountNo;
```
vorburger, line 48: this looks suspicious, and worth of a closer investigation..

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/accounting/GLAccountBuilder.java
```diff
@@ -41,24 +40,24 @@
     private static String name = Utils.randomStringGenerator("ACCOUNT_NAME_", 5);
 
     private static String GLCode = "";
-    private static String accountType = "";
-    private static String accountUsage = ACCOUNT_USAGE_DETAIL;
-    private static String manualEntriesAllowed = MANUAL_ENTRIES_ALLOW;
+    private String accountType = "";
```
vorburger, line 43: This change seems strange - why are some made to be non static, but others left static?

fineract-provider/src/main/java/org/apache/fineract/accounting/common/AccountingConstants.java
```diff
@@ -290,13 +290,11 @@ public static FinancialActivityData toFinancialActivityData(final int financialA
         }
 
         public static List<FinancialActivityData> getAllFinancialActivities() {
-            if (financialActivities == null) {
```
vorburger, line 293: Technically this changes the behaviour, agreed? Before, if it were to be called twice, it would do the conert and add only once (the first time when it was null), whereas as it's now it would re-add it. I don't understand enough of the functionality here, but do you think this is a safe change? Should we be extra careful and avoid touching this?

fineract-provider/src/main/java/org/apache/fineract/accounting/glaccount/domain/TrialBalance.java
```diff
@@ -103,6 +104,8 @@ public boolean equals(Object obj) {
                 && trialBalance.getTransactionDate().equals(this.getTransactionDate());
     }
 
-
-
+    @Override
```
vorburger, line 106: when I wanted to double check if this new hashCode() here did not forget any fields (you did not), I noticed that the existing equals() method does not include amount and closingBalance... doesn't that look (VERY) suspicious? Shall we fix that as well here while we're at it?

fineract-provider/src/main/java/org/apache/fineract/accounting/provisioning/domain/LoanProductProvisioningEntry.java
```diff
@@ -125,4 +126,9 @@ public boolean equals(Object obj) {
                 && entry.office.getId().equals(this.office.getId())
                 && entry.getCurrencyCode().equals(this.getCurrencyCode());
     }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(entry, criteriaId, office, currencyCode, loanProduct, provisioningCategory, overdueInDays, reservedAmount, liabilityAccount, expenseAccount);
```
vorburger: This again looks like it does not "match" the equals() method, which, to me, seems wrong... or am I not getting something JPA/OpenJPA (Hibernate) specific here? (I forgot what the deal was re. equals and hashCode for `@Entity` classes.)

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/AbandonedConnectionCleanupShutdownListener.java
```diff
@@ -66,6 +66,8 @@ private void shutDowncleanUpThreadAndDeregisterJDBCDriver() {
         }
         try {
             Thread.sleep(2000L);
-        } catch (Exception e) {}
+        } catch (Exception e) {
+            e.printStackTrace();
```
vorburger: use (or add) a Logger and log.error, instead of just e.printStackTrace();

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/data/EntityDataTableChecksTemplateData.java
```diff
@@ -18,13 +18,15 @@
  */
 package org.apache.fineract.infrastructure.dataqueries.data;
 
+import org.apache.commons.lang.builder.EqualsBuilder;
```
vorburger: why use Commons Lang EqualsBuilder and HashCodeBuilder here when everywhere else we just use java.lang.Objects's helpers? Perhaps better to stay consistent?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/gcm/domain/Sender.java
```diff
@@ -18,73 +18,29 @@
...
-import java.util.Map;
-import java.util.Random;
+import java.nio.charset.Charset;
+import java.util.*;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import static org.apache.fineract.infrastructure.gcm.GcmConstants.*;
+
+/*import org.json.simple.JSONArray;
```
vorburger: just completely remove this commented out super old import block

fineract-provider/src/main/java/org/apache/fineract/infrastructure/gcm/domain/Sender.java
```diff
@@ -789,7 +745,7 @@ protected static String getString(InputStream stream) throws IOException {
 			return "";
 		}
 		BufferedReader reader = new BufferedReader(
-				new InputStreamReader(stream));
+				new InputStreamReader(stream, Charset.defaultCharset()));
```
vorburger: hard-code this to use UTF-8 instead of using platform default charset is better here, as it make it more repeatable (otherwise it depends on OS settings, which can lead to hard to understand problems)

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/api/AuthenticationApiResource.java
```diff
@@ -112,12 +113,12 @@ public String authenticate(@QueryParam("username") @ApiParam(value = "username")
                     principal.hasSpecificPermissionTo(TwoFactorConstants.BYPASS_TWO_FACTOR_PERMISSION);
             if (this.springSecurityPlatformSecurityContext.doesPasswordHasToBeRenewed(principal)) {
                 authenticatedUserData = new AuthenticatedUserData(username, principal.getId(),
-                        new String(base64EncodedAuthenticationKey), isTwoFactorRequired);
+                        new String(base64EncodedAuthenticationKey, Charset.defaultCharset()), isTwoFactorRequired);
```
vorburger: use UTF-8 instead of platform default charset

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/api/AuthenticationApiResource.java
```diff
@@ -112,12 +113,12 @@ public String authenticate(@QueryParam("username") @ApiParam(value = "username")
...
             } else {
 
                 authenticatedUserData = new AuthenticatedUserData(username, officeId, officeName, staffId, staffDisplayName,
                         organisationalRole, roles, permissions, principal.getId(),
-                        new String(base64EncodedAuthenticationKey), isTwoFactorRequired);
+                        new String(base64EncodedAuthenticationKey, Charset.defaultCharset()), isTwoFactorRequired);
```
vorburger: also just use fixed UTF-8

fineract-provider/src/main/java/org/apache/fineract/portfolio/charge/domain/Charge.java
```diff
@@ -621,4 +600,34 @@ public TaxGroup getTaxGroup() {
...
+        return penalty == charge.penalty &&
+                active == charge.active &&
+                deleted == charge.deleted &&
+                name.equals(charge.name) &&
+                amount.equals(charge.amount) &&
+                currencyCode.equals(charge.currencyCode) &&
+                chargeAppliesTo.equals(charge.chargeAppliesTo) &&
+                chargeTimeType.equals(charge.chargeTimeType) &&
+                chargeCalculation.equals(charge.chargeCalculation) &&
+                chargePaymentMode.equals(charge.chargePaymentMode) &&
```
vorburger: use Objects.equals helper here, as some of these may well be null, and would cause NPE as is.

fineract-provider/src/main/java/org/apache/fineract/portfolio/charge/service/ChargeReadPlatformServiceImpl.java
```diff
@@ -225,6 +225,7 @@ private void processChargeExclusionsForLoans(ChargeTimeType[] excludeChargeTimes
                 sb.append(excludeChargeTimes[i].getValue());
             }
             excludeClause = excludeClause.append(" and c.charge_time_enum not in(" + sb.toString() + ") ");
+            excludeClause.append(" ");
```
vorburger, line 228: I'm guessing you had a good reason to add this, but can I ask why? Did SpotBugs suggest this? Why?

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/data/LoanTransactionData.java
```diff
@@ -319,7 +319,7 @@ public LocalDate dateOf() {
     }
 
     public boolean isNotDisbursement() {
-        return Integer.valueOf(1).equals(this.type.id());
+        return Integer.valueOf(1).equals(this.type.id().intValue());
```
vorburger:
```suggestion
        return type.id() == 1
```

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
```diff
@@ -790,21 +736,20 @@ private void removeOrModifyTransactionAssociatedWithLoanChargeIfDueAtDisbursemen
             LoanTransaction transactionToRemove = null;
             List<LoanTransaction> transactions = getLoanTransactions() ;
             for (final LoanTransaction transaction : transactions) {
-                if (transaction.isRepaymentAtDisbursement() && transaction.getLoanChargesPaid().contains(loanCharge)) {
```
vorburger, line 793: why remove this??

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
```diff
@@ -2787,7 +2732,8 @@ public LoanTransaction handlePayDisbursementTransaction(final Long chargeId, fin
             final boolean isScheduleRegenerateRequired = isRepaymentScheduleRegenerationRequiredForDisbursement(actualDisbursementDate);
             this.actualDisbursementDate = null;
             this.disbursedBy = null;
-            boolean isDisbursedAmountChanged = !this.approvedPrincipal.equals(this.loanRepaymentScheduleDetail.getPrincipal());
+            boolean isDisbursedAmountChanged =
+                    !this.approvedPrincipal.equals(this.loanRepaymentScheduleDetail.getPrincipal().getAmount());
```
vorburger: this looks quite suspicious, to me... so we're comparing an `approvedPrincipal` with a `getPrincipal().getAmount()`- is that really correct? Could someone with a better functional understanding of this code review this?

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanproduct/domain/LoanProductConfigurableAttributes.java
```diff
@@ -18,19 +18,18 @@
  */
 package org.apache.fineract.portfolio.loanproduct.domain;
 
-import javax.persistence.Column;
-import javax.persistence.Entity;
-import javax.persistence.JoinColumn;
-import javax.persistence.ManyToOne;
-import javax.persistence.Table;
-
+import org.apache.commons.lang.builder.EqualsBuilder;
```
vorburger: as above, why use Commons Lang here but java.lang.Objects helpers everywhere else? The Commons Lang helpers are less efficient, because they create intermediate very short live objects (to be GC'd), which can be easily avoided.

fineract-provider/src/main/java/org/apache/fineract/portfolio/paymenttype/data/PaymentTypeData.java
```diff
@@ -18,7 +18,12 @@
  */
 package org.apache.fineract.portfolio.paymenttype.data;
 
-public class PaymentTypeData {
+import org.apache.commons.lang.builder.EqualsBuilder;
```
vorburger: Again - consistency? Of course this does also work, so if you are short on time, we can leave all of these in, the other points raised are more important than these.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccountCharge.java
```diff
@@ -693,17 +679,35 @@ public boolean hasCurrencyCodeOf(final String matchingCurrencyCode) {
...
-                .isEquals();
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (!(o instanceof SavingsAccountCharge)) return false;
+        SavingsAccountCharge that = (SavingsAccountCharge) o;
+        return penaltyCharge == that.penaltyCharge &&
+                paid == that.paid &&
+                waived == that.waived &&
+                status == that.status &&
+                savingsAccount.equals(that.savingsAccount) &&
```
vorburger: again, this is not null safe (use Objects.equals helper)

fineract-provider/src/main/java/org/apache/fineract/portfolio/self/loanaccount/data/SelfLoansDataValidator.java
```diff
@@ -93,7 +93,7 @@ public void validateRetrieveTransaction(UriInfo uriInfo) {
 
         final String loanTypeParameterName = "loanType";
         final String loanTypeStr = this.fromApiJsonHelper.extractStringNamed(loanTypeParameterName, element);
-        baseDataValidator.reset().parameter(loanTypeParameterName).value(loanTypeStr).notNull().equals("individual");
```
vorburger: why remove this .equals("individual") validation? (Twice, both here, and below.)

fineract-provider/src/main/java/org/apache/fineract/template/service/TemplateMergeService.java
```diff
@@ -162,7 +150,7 @@ private static String getStringFromInputStream(final InputStream is) {
         String line;
         try {
 
-            br = new BufferedReader(new InputStreamReader(is));
+            br = new BufferedReader(new InputStreamReader(is, Charset.defaultCharset()));
```
vorburger: again IMHO just hard coding UTF-8 here is safer (as more stable and reproducible) than relying on platform default encoding

fineract-provider/src/test/java/org/apache/fineract/mix/report/service/XBRLBuilderTest.java
```diff
@@ -76,7 +76,8 @@ public void shouldCorrectlyBuildMap() {
         System.out.println(result);
         NodeList nodes = null;
         try {
-            nodes = DocumentBuilderFactory.newInstance().newDocumentBuilder().parse(new ByteArrayInputStream(result.getBytes()))
+            nodes =
+                    DocumentBuilderFactory.newInstance().newDocumentBuilder().parse(new ByteArrayInputStream(result.getBytes(Charset.defaultCharset())))
```
vorburger: again, just use UTF-8

vorburger (changes requested):
@awasum this is amazing work! I've done a detailed review, and have some feedback for you... 

Just add additional commits, do you not yet squash, that makes it easier to re-review.

PS: Re. the build problem, with #673 having also failed, I'm now confused whether this PR is causing it, or it's a more general problem; let's see if a trivial #674 passes.

vorburger:
> Looks..like the ShareAccount Integration tests are failing(locally) 

Oh! Silly me, I had not even noticed that (which is why I, confusingly, wrote about Docker above).

How about #675 to make this more obvious in the future? Both for stupid myself :smiling_imp: and others.. 

> Anyone reading this who could help pinpoint the problem before me..that will be great.

See if any of the review comments I just posted (before reading this) may actually help to fix this?

> Last resort, I will revert the changes in the share account module and add one change at a time while run the integration tests till we get it.

Yeah, I was thinking of suggesting doing some of the changes proposed here individually, but then didn't want to bother you with extra work. Especially those I picked up on, perhaps easier to get more eyes on them if smaller changes? Up to you.

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/CenterDomain.java
awasum: > this looks suspicious, and worth of a closer investigation..

Call to Builder() never passes accountNo, so setting the class var to itself makes no sense to me. Maybe, I am wrong. Advice? SpotBugs flags this as something to be removed.

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/accounting/GLAccountBuilder.java
awasum: In the moment..I did not care about the other fields..I just wanted SpotBugs test to pass as there were too many violations. Maybe I should clean everything up now.

fineract-provider/src/main/java/org/apache/fineract/accounting/common/AccountingConstants.java
awasum: Yes, you are right. I am correcting this by re-adding. Sorry. Thanks for catching this.

fineract-provider/src/main/java/org/apache/fineract/accounting/glaccount/domain/TrialBalance.java
awasum: Ohh...It does. the JPA entity looks like amount and closingBalance are never null (when saved to DB). That confused me and I did not check to see the corresponding equals for it. Fixing this now. Thanks.

fineract-provider/src/main/java/org/apache/fineract/accounting/provisioning/domain/LoanProductProvisioningEntry.java
awasum: Thanks. Fixing this and also adjusting equals() to use Objects.equals() too.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/AbandonedConnectionCleanupShutdownListener.java
awasum: I will add a logger. I forgot this. We will catch more of these easily with the findbugs-slf4j plugin you suggested in another issue.

fineract-provider/src/main/java/org/apache/fineract/portfolio/charge/service/ChargeReadPlatformServiceImpl.java
awasum: excludeClause is never used in this method but gets assigned a value here which is been used by another method. So to please spotbugs, I added the redundant line at the end.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
awasum: Type of loanCharge and transaction.getLoanChargePaid() are not compatible and so spotbugs throw an error. If they are incompatible, how is this going to ever be TRUE as the second expression in the if statement will be FALSE? I removed this so we can discuss with someone who has more domain knowledge and understands the code. I might be missing something here. @vishwasbabu  @nazeer1100126  @Nayan Help me understand this..please...

I may put this back and suppress the spotbug check for now.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
awasum: This came as a result of Incompatible Caparison rule violation Exception thrown by SpotBugs.
approvedPrincipal is of Type BigDecimal (Java Type) while loanRepaymentScheduleDetail.getPrincipal() returns type Money (Fineract type which does not extend BigDecimal, this is an object with various other properties). Will the JVM be able to compare BigDecimal and Money? I reasoned this wont work so I logically used the amount (of Type BigDecimal) to compare with approvedPrincipal. I might be missing something here. Advice. @vishwasbabu , @nazeer1100126 , @Nayan .

I could just also suppress this if its a false positive and my reasoning is wrong.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/data/LoanTransactionData.java
awasum: Thanks

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
awasum: I am putting this back and suppressing SpotBugs for this method. We can take a look at it later.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
awasum: I am putting this back and suppressing SpotBugs for this method. We can take a look at it later.

fineract-provider/src/main/java/org/apache/fineract/portfolio/self/loanaccount/data/SelfLoansDataValidator.java
awasum: SpotBug Said: ```This method calls equals(Object) on two references of different class types and analysis suggests they will be to objects of different classes at runtime. Further, examination of the equals methods that would be invoked suggest that either this call will always return false, or else the equals method is not be symmetric (which is a property required by the contract for equals in class Object).```

My think was since the types are different..it always returns false...which maybe wrong..so i removed it.

I will add this back and suppress this for now..We probably need a better way to handle this in the near future.

awasum:
@vorburger , If you have some time before xmas, will you like to review this again and merge if Ok? Travis build is now working.

awasum:
> @vorburger , If you have some time before xmas, will you like to review this again and merge if Ok? Travis build is now working.

Maybe, its time I squashed this?

vorburger:
@awasum FYI I haven't forgotten about this, I'm hoping to make time to re-review some time in the coming days!

.travis.yml
```diff
@@ -61,6 +61,6 @@ script:
 # using "&&" instead of several "-" means that integrationTest does not run if test fails,
 # and Docker test does not run if integration test fails, which makes PR failure easier to understand.
 # @see https://docs.travis-ci.com/user/job-lifecycle/#customizing-the-build-phase
-  - ./gradlew --console=plain licenseMain licenseTest licenseIntegrationTest test  &&  ./gradlew --console=plain integrationTest  &&  sudo service mysql stop  &&  docker-compose build  &&  docker-compose up -d  &&  sleep 30s  &&  http --verify=no --timeout 240 --check-status get https://localhost:8443/fineract-provider/actuator/health || docker logs fineract_fineract-server_1
+  - ./gradlew --console=plain licenseMain licenseTest licenseIntegrationTest check  &&  ./gradlew --console=plain integrationTest  &&  sudo service mysql stop  &&  docker-compose build  &&  docker-compose up -d  &&  sleep 30s  &&  http --verify=no --timeout 240 --check-status get https://localhost:8443/fineract-provider/actuator/health || docker logs fineract_fineract-server_1
```
vorburger, line 64: Changing `gradle test` to `check` is due to https://stackoverflow.com/questions/50104666/gradle-difference-between-test-and-check and looks great here! (Just recording this so that we remember it in the future.)

vorburger:
@awasum I'd like to get this in (sorry my delay-...), so I'm going to go about reviewing it differently - I'll pull your branch, squash all of your work into a single commit (with you as the commit's author!), quickly  resolve the (trivial) merge conflict caused by #678 myself, and then add one single new commit with changes I would have suggested you make in the final round of review, and then raise a new PR for you to re-review and self-merge - hope that's OK with you?

awasum:
> @awasum I'd like to get this in (sorry my delay-...), so I'm going to go about reviewing it differently - I'll pull your branch, squash all of your work into a single commit (with you as the commit's author!), quickly resolve the (trivial) merge conflict caused by #678 myself, and then add one single new commit with changes I would have suggested you make in the final round of review, and then raise a new PR for you to re-review and self-merge - hope that's OK with you?

Thanks. Go ahead and do it. I have been waiting for this so i can do the other related checkstyle, pmd and other quality checks.

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/LoanReschedulingWithinCenterTest.java
```diff
@@ -151,7 +151,7 @@ public void testCenterReschedulingLoansWithInterestRecalculationEnabled() {
 
         // VERIFY THE INTEREST
         Float interestDue = (Float) ((HashMap) loanRepaymnetSchedule.get(2)).get("interestDue");
-        assertEquals(String.valueOf(interestDue), "90.82");
+        assertEquals(String.valueOf(interestDue), "42.38");
```
vorburger, line 154: As per #682, this, and the other change in line 283 below, doesn't seem right...

vorburger (changes requested):

vorburger:
@awasum if you don't mind I would suggest that we (I'll let you) close this PR, and continue (together) based on #682?  (You could of course any time due the same I did and re-fork that branch from me to try to make further changes, it's probably just easier if we go forward from the "cleaner" single squashed rebased one.)

vorburger:
Definitely closing this in favour of #682 (which is now green).

PR #679, opened 2020-01-06
Fix flaky LoanReschedulingWithinCenterTest so that it also passes on Sunday (FINERACT-820)

vorburger:
see https://issues.apache.org/jira/browse/FINERACT-820


vorburger:
Hm, looks like fixing FINERACT-820 is not quite as trivial as this? I'll close this again, for now.

PR #546, opened 2019-03-11
Consolidating changes: Bulk import-fixing importhandler integration t…

wkk91193:
Opening a new PR  for #444 changes.

PR #680, opened 2020-01-08
Fixing bulk import integration test failures

wkk91193:
Opening a new PR for #546  changes.


wkk91193:
@vorburger @vishwasbabu @edcable 
Ready to be merged

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/bulkimport/importhandler/savings/SavingsImportHandlerTest.java
```diff
@@ -64,92 +64,99 @@ public void setup() {
...
+
+            FileInputStream fileInputStream = new FileInputStream(location);
+            Workbook Outputworkbook = new HSSFWorkbook(fileInputStream);
+            Sheet OutputSavingsSheet = Outputworkbook.getSheet(TemplatePopulateImportConstants.SAVINGS_ACCOUNTS_SHEET_NAME);
+            Row row = OutputSavingsSheet.getRow(1);
+            Assert.assertEquals("Imported", row.getCell(SavingsConstants.STATUS_COL).getStringCellValue());
+            
+        } catch (AssertionError error) {
+            // Output expected AssertionErrors.
+            System.out.println("AssertionError: "+error);
```
vorburger: This does not look right, to me... a test which `catch (AssertionError error)`, and then `System.out.println("AssertionError: "+error)` it, is unusual. If you are trying to express that you expect something int this test to fail, then that should be written differently. Perhaps an existing `Assert` can simply be changed/improved/fixed? it should also be more "fine grained", only around the section expecting to fail. For example, the way you currently have it, the test will PASS if any of those `Assert.assertNotNull()` FAIL - I doubt that's what you want here?

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/Utils.java
```diff
@@ -257,5 +257,19 @@ public static String performServerOutputTemplateLocationGet(final RequestSpecifi
...
         return templateLocation.substring(1,templateLocation.length()-1);
     }
+
+    public static void waitUntilFileCreation(String location){
+        Long start= System.currentTimeMillis();
+
+        while (!new File(location).exists()) {
+            try {
+                Thread.sleep(1000);
+            } catch (InterruptedException e) {
```
vorburger, line 271: don't do this, instead simply add `throws InterruptedException`  to progagate this up through until the test.

vorburger (changes requested):

vorburger:
> /rebase

please just ignore this comment (see https://issues.apache.org/jira/browse/FINERACT-829)

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/bulkimport/importhandler/savings/SavingsImportHandlerTest.java
wkk91193: @vorburger The AssertionError was thrown from LineNo 139  String importDocumentId = savingsAccountHelper.importSavingsTemplate(file); I enclosed it in a try catch block in the hope of seeing the stacktrace. However, I could not see it anywhere in the log. Numerous attempts to recreate the issue locally failed too. Would I able be to access the detail log from the build server in order to investigate the cause behind this?

vorburger:
@wkk91193 this fails the `SavingsImportHandlerTest` integration test:

```
org.apache.fineract.integrationtests.bulkimport.importhandler.savings.SavingsImportHandlerTest > testSavingsImport FAILED

    java.lang.AssertionError at SavingsImportHandlerTest.java:143
```

wkk91193:
> @wkk91193 this fails the `SavingsImportHandlerTest` integration test:
> 
> ```
> org.apache.fineract.integrationtests.bulkimport.importhandler.savings.SavingsImportHandlerTest > testSavingsImport FAILED
> 
>     java.lang.AssertionError at SavingsImportHandlerTest.java:143
> ```
@vorburger  In order to better understand the root of the problem I need to have the complete stacktrace of the AssertionError. There isn't an Assertion statement at line 143 nor I'm able recreate this locally on my end.

vorburger:
> isn't an Assertion statement at line 143 nor 

indeed the only assertion around there is the `Assert.assertNotNull(importDocumentId);` which is in line 145 not 143... I would just assume it's that one that's failing here (just because there is no other one), and not loose too much sleep about that (slight) line number mismatch. I can't quite explain it either, but if you made me guess, I would suspect it may be related to and may just have been solved by (my) recent #688 re. CR/LF... you could rebase to the latest develop, to check. If it's still a little bit off, just assume it's that one, and try to fix that.

> need to have the complete stacktrace of the AssertionError. 

Travis unfortunately doesn't let you see it, what we do is locally reproduce and debug.

> I'm able recreate this locally on my end.

But are you really running `./gradlew integrationTest` (not just `test` or `build`) ?  I'm pretty sure you should locally hit it as well... unless this is some mess due to time zone or something, but that seems quite unlikely.

wkk91193:
@vorburger SavingsImportHandlerTest is passing on my end.
Right before the end of the log I see this.
"There were failing tests. See the report at: file:///home/travis/build/apache/fineract/build/reports/tests/integrationTest/index.html"

Do you have access to this index.html file? There may be some useful pointers as to why it's failing there.

vorburger:
> Do you have access to this index.html file?

Travis unfortunately doesn't let us see it.

wkk91193:
@vorburger just wondering why the test is failing. It's passing on my end. I also tried pointing to a fresh tenant locally but couldn't reproduce this. Any contact who has access to build server logs?

vorburger:
@wkk91193 it's not possible to get access to more logs from Travis CI. The only thing you could do is make Fineract's build / tests log more itself. But before you try that, how about just rebase this on latest, resolve minor merge conflict, and see if perhaps it already just passes now?

wkk91193:
@vorburger Now I'm seeing some new errors after rebasing.
Build log line 1914

> 5291  jpa-pu  INFO   [Daemon worker] openjpa.Tool - No metadata was found for 271 classes; they have been enhanced as persistence-aware. If you intended for any of these classes to be persistence-capable, then this means that OpenJPA could not find any metadata for those classes. This can happen if the directory containing your metadata is not in your CLASSPATH, or if your metadata files are not named properly. See the documentation on metadata placement for more information. Persistence-aware classes...

vorburger:
@wkk91193 that line is just an INFO, like many other such lines, you can just safely ignore that. The real reason for the build failure on this PR are simply 14 Checkstyle violations in 4 of your files, you can see them by clicking on the Details link above (on the right side of "continuous-integration/travis-ci/pr — The Travis CI build failed") and then scrolling down. These are easy to see locally as well if you build locally before pushing to GitHub, which is always a good idea.

xurror (commented):
Still have a couple of checkstyle violations. Do you mind checking them out?

vorburger:
created FINERACT-913 so that we have a trace about this when it will get automatically closed in 30 days. @nnatarajan would you perhaps happen to have any interest to pick up and finish this, given that @wkk91193 seems to have abandoned it?

nnatarajan:
@vorburger , I apologize, I just saw this. Please let me attempt.

vorburger:
#980 fixed the test

PR #682, opened 2020-01-10
use SpotBugs and fix over 300 high priority Rule Violations (FINERACT-702)

vorburger:
This PR, so far, is just @awasum's #670 rebased and conflicts resolved, should otherwise be the same.

I'm hoping to shortly be adding a 2nd commit with some additional proposed related clean up here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,6 +61,6 @@ script:
 # using "&&" instead of several "-" means that integrationTest does not run if test fails,
 # and Docker test does not run if integration test fails, which makes PR failure easier to understand.
 # @see https://docs.travis-ci.com/user/job-lifecycle/#customizing-the-build-phase
-  - ./gradlew --console=plain licenseMain licenseTest licenseIntegrationTest test  &&  ./gradlew --console=plain integrationTest  &&  sudo service mysql stop  &&  docker-compose build  &&  docker-compose up -d  &&  sleep 30s  &&  http --verify=no --timeout 240 --check-status get https://localhost:8443/fineract-provider/actuator/health || docker logs fineract_fineract-server_1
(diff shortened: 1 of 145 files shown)


vorburger:
Wait a second @awasum FYI I screwed up the rebase & squash... :smiling_imp: this (as is) was NOT your latest #670 but an intermediate previous version; I'm redoing it now, hope I didn't confuse you if you already looked at this - I'll ping you again when this is fully ready.

vorburger:
@awasum doing a rebase and squash of your #670, so that we have a simple clean git history, was much less trivial than I thought... :smiley:  because of the merge commits created by GitHub when you accepted those change proposals I made, and then made further changes.

This was supposed to be it, but it's not 1000% identical to your #670, as I had to "manually" re-port some of the changes.. thought it was very close, but seems to be missing something, as 3 ITs are now failing here. :sob: I'll see if I can figure out the difference, shout if you do first. If we can't, we could always merge yours (but this would be nicer).

I'll be raising a new PR for some additional proposed related clean up from myself based off the 2nd commit on https://github.com/apache/fineract/compare/develop...vorburger:spotbugs-extra. (We can decide later if we cherry pick those on top of #670 if you go for merging that after all, or if we can figure this one out.)

vorburger:
@awasum the remaining integration test failures are in `LoanReschedulingWithinCenterTest`. I noticed only now that your original PR #670 changed the expected "interest due" in that same test. You seem to have added that in your last commit re. "Fix more spotbugs failures and integration tests" in the original PR.

That's... interesting, and probably merits more investigation? It looks like one of the changes we're making here for SpotBugs is causing a discrepancy. Either we found a pretty major real functional bug, or we broke the main functionality... which one is it? :question:

PS: Interestingly, my PR expects a different result than yours; the original was 90.82 & 41.05, you changed it to 42.38 & 19.16, whereas my results are 84.76 & 38.31 (so closer).

PPS: The "expected" and "was" in the `ComparisonFailure` is the wrong way around.

vorburger:
@awasum this is nwo green. Notice the 2nd commit on this PR, which reverts that test's change. Merge?

awasum:
This is looking good. What did I miss?

awasum (approved):

PR #683, opened 2020-01-10
FINERACT-824: NullPointerException is thrown when submitting a loan account application with an existing externalId

angelboxes:
Removed the second validation since realCause.getCause() is null if it is equal to realCause

## Description
When I tried to create a new loan account with an already used external Id, a NullPointerException was thrown with no description of why the loan account creation failed. I removed a validation that returned a null value, this kind of validation was only used with integrity issues on loan accounts and added for some OpenJPA Integration Changes. 

https://issues.apache.org/jira/browse/FINERACT-824

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanApplicationWritePlatformServiceJpaRepositoryImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanApplicationWritePlatformServiceJpaRepositoryImpl.java
@@ -1036,12 +1036,12 @@ public void checkForProductMixRestrictions(final Loan loan) {
      */
     private void handleDataIntegrityIssues(final JsonCommand command, final Throwable realCause, final Exception dve) {
     	
-        if (realCause.getMessage().contains("loan_account_no_UNIQUE") || realCause.getCause().getMessage().contains("loan_account_no_UNIQUE")) {
+        if (realCause.getMessage().contains("loan_account_no_UNIQUE") || (realCause.getCause() != null && realCause.getCause().getMessage().contains("loan_account_no_UNIQUE"))) {
 
             final String accountNo = command.stringValueOfParameterNamed("accountNo");
             throw new PlatformDataIntegrityException("error.msg.loan.duplicate.accountNo", "Loan with accountNo `" + accountNo
                     + "` already exists", "accountNo", accountNo);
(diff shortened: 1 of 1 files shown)


fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanApplicationWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -1036,12 +1036,11 @@ else if(existingLoanApplication.getGroup() != null){
      */
     private void handleDataIntegrityIssues(final JsonCommand command, final Throwable realCause, final Exception dve) {
     	
-        if (realCause.getMessage().contains("loan_account_no_UNIQUE") || realCause.getCause().getMessage().contains("loan_account_no_UNIQUE")) {
-
+        if (realCause.getMessage().contains("loan_account_no_UNIQUE")) {
             final String accountNo = command.stringValueOfParameterNamed("accountNo");
             throw new PlatformDataIntegrityException("error.msg.loan.duplicate.accountNo", "Loan with accountNo `" + accountNo
                     + "` already exists", "accountNo", accountNo);
-        } else if (realCause.getMessage().contains("loan_externalid_UNIQUE") || realCause.getCause().getMessage().contains("loan_externalid_UNIQUE")) {
```
vorburger, line 1044: @angelboxes I don't know much about this code, but it looks like someone thought that it would be useful to also check the cause... your [FINERACT-824](https://issues.apache.org/jira/browse/FINERACT-824) indicates that this can lead to NPEs, so instead of just completely removing the `getCause()` related check, do you think it may be a good idea to check it like this (for both):

    } else if (realCause.getMessage().contains("loan_externalid_UNIQUE") || (realCause.getCause() != null && realCause.getCause().getMessage().contains("loan_externalid_UNIQUE"))) {

Just a thought, as this would perhaps seem more prudent in case that extra check for the cause was there for a good reason. If you are totally confident that this cause is never set and always null, then this is also OK for me, but I thought it would be worth to clarify?

angelboxes: Yeah, actually that was my first approach to this issue however I saw that this is not the only place where unique values exceptions are thrown but this is the only place where such validation is done. and thought that it was one of those changes that worked when it was added but a different one made it useless and that maybe it wasn't needed,. Don't worry I will change it as you propose.

vorburger (approved):
LGTM.  I'll merge this one when the build passed.

PR #684, opened 2020-01-11
 more SpotBugs related / inspired code clean up (see FINERACT-702) 

vorburger:
based upon code review of changes made by @awasum in previous commit:

 - fix clear contains() GC_UNRELATED_TYPES bug in Loan

 - fix clear equals() EC_UNRELATED_TYPES bug in Loan

 - fix more equals() & hashCode() in...
    * Charge
    * TrialBalance
    * LoanProductBorrowerCycleVariations
    * LoanProductProvisioningEntry

 - fix AccountingConstants to avoid using @FindBugsSuppressWarnings

 - fix SelfLoansDataValidator instead of @FindBugsSuppressWarnings

 - fix up weird static in GLAccountBuilder

 - minor clean up in build.gradle (spotbugs has replaced findbugs)

PS: This change originally included also fixing the missing `hashCode()` and `equals()` in `ShareAccountCharge`, but that turned out to be non trivial, see [FINERACT-827](https://issues.apache.org/jira/browse/FINERACT-827), and is thus not yet fixed by this.

PR #685, opened 2020-01-11
 clean up GLAccountBuilder, following SpotBugs (FINERACT-702) 

vorburger:


awasum:
Which one are you merging first. This or #682 ?

@vorburger thanks very much for taking the time to review and augment the spotbug integration work. I am grateful.

vorburger:
> Which one are you merging first. This or #682 ?

#682 just merged, this one was actually just an extension with 1 more additional commit over that one. There's no real good way to do this better in GitHub (that I know of), so what I "imply" (but should say more clearly?) when I raise a PR like this one for review is to manually click on the last commit of such a PR (in the _Commits_ tab) in review that, instead of everything on the _Files changed_ tab (or use that Changes from drop down on _Files changed_ to review only last commit instead of _from all commits_).

> @vorburger thanks very much 

You are the one to thank, this (that we've finally properly adopted SpotBugs here) is amazing!

If this small PR is OK with you, just go ahead and merge it (use _Rebase and merge_ to avoid the merge commit, because #684 is "parallel"). -- PS: I hope this doesn't fail in Integration Test again due to [FINERACT-820](https://issues.apache.org/jira/browse/FINERACT-820) - as it's Sunday...  if it does, we really should figure out how to fix that problem.. or ask on list if anyone would like to take that on.

vorburger:
>  PS: I hope this doesn't fail in Integration Test again due to [FINERACT-820](https://issues.apache.org/jira/browse/FINERACT-820) - as it's Sunday... if it does, we really should figure out how to fix that problem.. or ask on list if anyone would like to take that on.

It did, this is really annoying.. are you willing to just review and merge this ignoring that `LoanReschedulingWithinCenterTest` failure? Or motivated to investigate how that could be fixed properly once and for all? I'll drop an email to the list as well.

awasum:
This looks good merging now.

PR #512, opened 2019-02-07
FINERACT-428 Parallelization of Jobs :Parallelizing and paging of update loan summary

wkk91193:
Opening a new PR based on feedback from PR #464

PR #653, opened 2019-11-23
FINERACT-724: upgrade Spring boot to 2.1.7 

awasum:
## Description
FINERACT-724: upgrade Spring boot to 2.1.7. Intial PR (contributed by  @ivan333m) and reviews (by @vorburger ) here: https://github.com/apache/fineract/pull/642

This just makes sure the commit history is consistent as specified here: https://github.com/apache/fineract/pull/642#pullrequestreview-321923274 and 
https://github.com/apache/fineract/pull/642#issuecomment-557822708

## Checklist
Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Commit message starts with the issue number from https://issues.apache.org/jira/projects/FINERACT/. Ex: FINERACT-646 Pockets API.

- [x] Coding conventions at https://cwiki.apache.org/confluence/display/FINERACT/Coding+Conventions have been followed.

- [x] API documentation at https://github.com/apache/fineract/blob/develop/api-docs/apiLive.htm has been updated with details of any API changes.

- [ ] Integration tests have been created/updated for verifying the changes made.

- [x] All Integrations tests are passing with the new commits.

- [x] Submission is not a "code dump".  (Large changes can be made "in repository" via a branch.  Ask on the list.)

Our guidelines for code reviews is at https://cwiki.apache.org/confluence/display/FINERACT/Code+Review+Guide

PR #687, opened 2020-01-11
FINERACT-428 Parallelization of Jobs :Parallelizing and paging of update loan summary

wkk91193:
Reopened a new PR after a rebase #512


vorburger:
@wkk91193 this now fails to build due to the recent Spring version upgrade ... That should be trivial to fix though, you will just need to replace `findOne()` with `findById(yourId).orElseThrow(() -> new SomeNotFoundException(YourId)`, see e.g. #653 and #642 for examples.

PR #691, opened 2020-01-13
add /rebase comment support for Pull Requests (FINERACT-829)

vorburger:
https://github.com/marketplace/actions/automatic-rebase

https://github.com/cirrus-actions/rebase

@awasum and anyone else reading this, how about adding this?

### DIFF
--- a/.github/workflows/rebase.yml
+++ b/.github/workflows/rebase.yml
@@ -1,0 +1,23 @@
+# https://github.com/marketplace/actions/automatic-rebase  (https://github.com/cirrus-actions/rebase)
+on:
+  issue_comment:
+    types: [created]
+name: Automatic Rebase
+jobs:
+  rebase:
+    name: Rebase
+    if: github.event.issue.pull_request != '' && contains(github.event.comment.body, '/rebase')
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


vorburger:
#693 will fix the build failure seen here

awasum:
@vorburger Do we need this documented somewhere on the readme. It can be done after this has been merged if you want. Looks like just writing /rebase in a PR comment should activate this action and rebase the PR. Is this correct?

awasum (approved):

vorburger:
self merging this one based on @awasum comment in #692 ("no experience with Github Actions. Merge when you feel comfortable with this.") - I don't expect any other committers could have a problem with this.

vorburger:
> @vorburger Do we need this documented somewhere on the readme. It can be done after this has been merged if you want. Looks like just writing /rebase in a PR comment should activate this action and rebase the PR. Is this correct?

Yeah, but doesn't seem to quite work yet.. watch https://issues.apache.org/jira/browse/FINERACT-829.  Once it does work, I'll announce it on the mailing list. We can also add it to the README, good idea. (Though in practice I think folks typically "learn" about such things less by reading about it in doc and more through seeing others use it in PRs and then by copy cat...)

PR #692, opened 2020-01-13
add GitHub Action to autoclose stale pull requests (FINERACT-829)

vorburger:
It will automatically add a comment after 30 days, and then another 30 days later (so after a total of 60 days of inactivity) automatically close Pull Requests.

https://github.com/marketplace/actions/close-stale-issues
https://github.com/actions/stale/blob/master/action.yml
https://github.com/actions/stale

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,20 @@
+# https://github.com/marketplace/actions/close-stale-issues
+# https://github.com/actions/stale/blob/master/action.yml
+# https://github.com/actions/stale
+name: Mark stale issues and pull requests
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


vorburger:
@awasum and anyone else reading this, how about adding this?

vorburger:
The test build failure seen here seems arbitrary, I've created [FINERACT-828](https://issues.apache.org/jira/browse/FINERACT-828), let's just retry.

awasum:
> @awasum and anyone else reading this, how about adding this?

This makes sense to have a 60 day auto close rule. I have no experience with Github Actions. Merge when you feel comfortable with this.

vorburger:
see https://issues.apache.org/jira/browse/FINERACT-829

PR #694, opened 2020-01-19
fix broken Dockerfile build (FINERACT-833)

vorburger:
see https://issues.apache.org/jira/browse/FINERACT-833


vorburger:
This build will most likely fail, because it's Sunday, due to https://issues.apache.org/jira/browse/FINERACT-820 (what a PITA, is anyone ever going to fix that?).  Having locally tested this fix, I'm therefore exceptionally merging this minor change WITHOUT Travis CI passing.

PR #689, opened 2020-01-12
introduce Checkstyle, start enforcing whitespace related rules (FINERACT-821)

vorburger:
Once #688 which bulk cleans up all existing whitespace issues,

after a rebase and verification (or more PRs) that there are no other new issues,

this PR contributes an initial Checkstyle configuration which will fails builds/PR for new whitespace issues.

The initial Checkstyle XML configuration contributed here has many more rules, but all non whitespace related ones are still commented out for now; the idea being that we (myself or, hopefully, other contributors!) then gradually enable more of those rules, after fixing existing violations.

See https://issues.apache.org/jira/browse/FINERACT-821 for full background.

PR #688, opened 2020-01-12
clean up all whitespace issues in Java files (FINERACT-821)

vorburger:
1. convert all Java files from DOS to UNIX line ending:
     find . -type f -name "*.java" -exec dos2unix '{}' ';'

  2. replace Tabs with 4 Spaces:
     rpl -eRxjava "\t" "    " .

  3. remove all trailing spaces:
     find . -type f -name "*.java" -exec sed -i 's/ *$//' '{}' ';'

PR #629, opened 2019-08-25
Swagger documentation gsoc2019

kangbreder:
## Description
For GSoC, I have improved the Swagger document with the following

- Reflected all API descriptions on swagger UI

- Formatted API descriptions

- Added swagger docs for new APIs

- Added swagger-gradle-plugin for automatic save of json during build.

- Leveraged swagger codegen for generation of language-specific libraries using gradle-swagger-generator plugin.

Link to JIRA TICKET: https://jira.apache.org/jira/browse/FINERACT-733

- [COMPLETED] Complete Swagger documentation on all API’s of Fineract

- [COMPLETED] Description inside of swagger docs

- [IN PROGRESS] Try it out button for API’s from swagger-UI

- [IN PROGRESS] Automatically generate the swagger-docs -> building with Fineract

- [NOT COMPLETED] Navigation Bar on swagger-UI(For easy access of documentation of API's)

- [NOT COMPLETED] Migrating the complete overview section in swagger-docs/swagger-docs

- [COMPLETED] Swagger guide for new Api’s or how to add a swagger doc to any Fineract API

- [IN PROGRESS] QA round to check if all the API’s are reflecting or not(Correct)

## Checklist
Please make sure these boxes are checked before submitting your pull request - thanks!

- [ ] Commit message starts with the issue number from https://issues.apache.org/jira/projects/FINERACT/. Ex: FINERACT-646 Pockets API.

- [x] Coding conventions at https://cwiki.apache.org/confluence/display/FINERACT/Coding+Conventions have been followed.

- [x] API documentation at https://github.com/apache/fineract/blob/develop/api-docs/apiLive.htm has been updated with details of any API changes.

- [ ] Integration tests have been created/updated for verifying the changes made.

- [x] All Integrations tests are passing with the new commits.

- [x] Submission is not a "code dump".  (Large changes can be made "in repository" via a branch.  Ask on the list.)

Our guidelines for code reviews is at https://cwiki.apache.org/confluence/display/FINERACT/Code+Review+Guide

PR #695, opened 2020-01-19
Initial Swagger related changes from GSOC 2019 contrib (FINERACT-733)

vorburger:


vorburger:
@sanyam96 @kangbreder @awasum FYI this is as per my last comment in #629 ...

I expect that this build will most likely fail, just because it's Sunday, due to https://issues.apache.org/jira/browse/FINERACT-820 (what a PITA, is anyone ever going to fix that?).  Having locally tested that this still builds, I'm therefore exceptionally merging this PR even WITHOUT it Travis CI passing (we should doing that in general, but I really want to wrap this up, and unblock #688 to unblock #689).

awasum (approved):
Merge when Travis Pass..or just merge this as Travis might fail due to https://issues.apache.org/jira/browse/FINERACT-820.

Thanks for fixing this long over due PR, @vorburger 

Also I have not been active this weekend due to abnormally slow internet across most parts of Africa, due to: https://www.bloomberg.com/news/articles/2020-01-17/african-internet-slowdown-continues-after-undersea-cables-break

kangbreder:
@vorburger Thank you for your help in fixing this. I just got my OS up few days ago but we have been experiencing very slow internet connection here. Thank you for the fixes.

PR #696, opened 2020-01-23
FINERACT-839: NPE when creating a Loan for a client fix

adamsaghy:
## Description

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanproduct/data/LoanProductData.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanproduct/data/LoanProductData.java
@@ -49,6 +49,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import net.sf.ehcache.util.FindBugsSuppressWarnings;
 
 /**
  * Immutable data object to represent loan products.
@@ -473,10 +474,10 @@ public class LoanProductData implements Serializable {
     }
 
     public static LoanProductData loanProductWithFloatingRates(final Long id, final String name,
-            final boolean isLinkedToFloatingInterestRates, final Integer floatingRateId, final String floatingRateName,
-            final BigDecimal interestRateDifferential, final BigDecimal minDifferentialLendingRate,
-            final BigDecimal defaultDifferentialLendingRate, final BigDecimal maxDifferentialLendingRate,
-            final boolean isFloatingInterestRateCalculationAllowed) {
+                                                               final boolean isLinkedToFloatingInterestRates, final Integer floatingRateId, final String floatingRateName,
+                                                               final BigDecimal interestRateDifferential, final BigDecimal minDifferentialLendingRate,
+                                                               final BigDecimal defaultDifferentialLendingRate, final BigDecimal maxDifferentialLendingRate,
+                                                               final boolean isFloatingInterestRateCalculationAllowed) {
         final String shortName = null;
         final String description = null;
         final CurrencyData currency = CurrencyData.blank();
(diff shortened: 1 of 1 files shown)


awasum:
Make sure this passes Spotbugs checks on your local machine.

Lets wait for Travis to build.

awasum:
On your local machine, run ./gradlew clean check build and make sure checkstyle and spotbugs rules are not voilated.

awasum:
Then squash your commits into one when done.

adamsaghy:
The problem is the SpotBug is a little bit too restrictive:
Bad practice Warnings
Code	Warning
NP	org.apache.fineract.portfolio.loanproduct.data.LoanProductData.isCompoundingToBePostedAsTransaction() has Boolean return type and returns explicit null
Details
NP_BOOLEAN_RETURN_NULL: Method with Boolean return type returns explicit null
A method that returns either Boolean.TRUE, Boolean.FALSE or null is an accident waiting to happen. This method can be invoked as though it returned a value of type boolean, and the compiler will insert automatic unboxing of the Boolean value. If a null value is returned, this will result in a NullPointerException.

Before the spotBug the business logic required to return null...as it is a valid option from business perspective.

Do you have any advice?

awasum:
> The problem is the SpotBug is a little bit too restrictive:
> Bad practice Warnings
> Code Warning
> NP org.apache.fineract.portfolio.loanproduct.data.LoanProductData.isCompoundingToBePostedAsTransaction() has Boolean return type and returns explicit null
> Details
> NP_BOOLEAN_RETURN_NULL: Method with Boolean return type returns explicit null
> A method that returns either Boolean.TRUE, Boolean.FALSE or null is an accident waiting to happen. This method can be invoked as though it returned a value of type boolean, and the compiler will insert automatic unboxing of the Boolean value. If a null value is returned, this will result in a NullPointerException.
> 
> Before the spotBug the business logic required to return null...as it is a valid option from business perspective.
> 
> Do you have any advice?

Maybe we add a spotbugs exception to omit checks of this type..Or just for this case in the code base.

@vorburger  some help please.

awasum:
What is wron with instead of returning Null we instead do ` return Boolean.FALSE;`, will this still cause issues?

adamsaghy:
From business perspective it is different to return FALSE or NULL. I reckon there was a consideration to choose Boolean as object to make it able to return NULL.

A warning would be better.

awasum:
> From business perspective it is different to return FALSE or NULL. I reckon there was a consideration to choose Boolean as object to make it able to return NULL.
> 
> A warning would be better.

Ok...In your PR..Suppress spotbugs for the methods having the error..Just above it, place the statement: `@FindBugsSuppressWarnings("NP_BOOLEAN_RETURN_NULL")` and spotbug check will ignore it.

awasum:
Have u functionally tested this and the problem has gone away? @adamsaghy . I dont have the right environment to test. If everything is fine on your end..let me know to merge this since every spotbugs and checkstyle test is also passing.

Thanks for your patience.

adamsaghy:
I did a quick test as now I was able to open a Loan to the client but havent checked all the iteration. 
Basically it is a revert back to the original business logic , which we had before the introduced spotbugs changes.

awasum:
> I did a quick test as now I was able to open a Loan to the client but havent checked all the iteration.
> Basically it is a revert back to the original business logic , which we had before the introduced spotbugs changes.

Please...check everything and let us know if this is Ok to be merged.

adamsaghy:
As it worked for 3 years before the spotbugs commit, and we just reverted back to the original behaviour, in my opinion it is safe to merge. ;)

awasum:
Send more PRs as you identify bugs. Thanks

vorburger:
> What is wron with instead of returning Null we instead do ` return Boolean.FALSE;`, will this still cause issues?

That's what I would have suggested as well... have commented in FINERACT-839. 

@adamsaghy are you motivated to raise a follow-up PR for that?

PR #698, opened 2020-01-24
FINERACT-806 - Allow customizing Tenants JDBC password, username and driverClassName (not just URL)

xurror:
In addition to being able to customize the JDBC URL of the Tenant DB (in the Docker Image, but even otherwise) as introduced in FINERACT-773, it would be good if that JDBC connection's username, password and driverClassName could be overriden via a simple environment variable as well.

This should be relatively straightforward (now) - just apply the "pattern" from https://github.com/apache/fineract/pull/648/files and introduce additional Java System properties (in lower-case) like the fineract_tenants_url from FINERACT-773.


docker/server.xml
```diff
@@ -43,8 +43,8 @@
         factory="org.apache.tomcat.jdbc.pool.DataSourceFactory"
         driverClassName="org.drizzle.jdbc.DrizzleDriver"
         url="${fineract_tenants_url}"
-        username="root"
-        password="mysql"
+        username="${username}"
+        password="${password}"
```
vorburger: Please use ’fineract_tenants_uid’ (and _pwd) instead of "username" and "password", to make more clear what these system properties are for.

vorburger (changes requested):

vorburger:
In addition to the variable name, this is also missing the equivalent customization for the demo tenant DB, as in https://github.com/apache/fineract/pull/648/files

awasum:
@xurror .. Are you looking forward to addressing feedback provided by @vorburger ?

xurror:
Definitely

xurror:
This should wrap up this issue

vorburger (changes requested):
@xurror sorry when I previously wrote *Please use ’fineract_tenants_uid’ (and _pwd) instead of "username" and "password", to make more clear what these system properties are for.* I meant `fineract_tenants_pwd` not `_pwd` .. would you be willing to change this? Again sorry for the confusion.

xurror:
Oh, sure.
My bad for misinterpreting

vorburger:
@xurror sorry I noticed only now that, more importantly than the naming, this doesn't actually work... explaining what needs to be done to fix it takes me the same amount of time as doing it, so let me raise a follow-up PR which takes this commit and adds another one doing the needful - hope OK for you.

xurror:
No  problem, do I need to close this?

vorburger:
closing this in favour of new #713

PR #713, opened 2020-02-09
Allow customizing Tenants DB JDBC uid/pwd/driver (FINERACT-806)

vorburger:


vorburger:
@xurror this is your initial #698 (for credit where credit is due) with a fix from my on top of it.

xurror (approved):
I see where I was failing, thanks for the update

awasum (approved):

PR #710, opened 2020-02-05
FINERACT-807  -  Use underscore instead of dash in tenants and default tenant database names

xurror:
## Description
The Fineract tenants and tenant database names by default are currently mifosplatform-tenants and mifostenant-default respectively.

The Google Cloud Platform console for managed MySQL (Cloud SQL) prevents creating MySQL databases containing dashes; the UI says Database name: Needs to follow the MySQL identifier rules.

While this limitation appears to only be enforced in the Google Cloud UI (it's still possible to create databases containing '-' dashes using the gcloud CLI), the linked doc does seem to suggest to use only underscores.

To make this simpler, and comply with mySQL doc recommendation, let's change mifosplatform-tenants to fineract_tenants and mifostenant-default to fineract_default. (The change from mifosplatform to fineract is just because this seems a good opportunity to make that change, and because it's probably less and not more confusing to change the entire name, and not just a dash to and underscore.)

## Checklist
Please make sure these boxes are checked before submitting your pull request - thanks!

- [ ] Commit message starts with the issue number from https://issues.apache.org/jira/projects/FINERACT/. Ex: FINERACT-646 Pockets API.

- [ ] Coding conventions at https://cwiki.apache.org/confluence/display/FINERACT/Coding+Conventions have been followed.

- [ ] API documentation at https://github.com/apache/fineract/blob/develop/api-docs/apiLive.htm has been updated with details of any API changes.

- [ ] Integration tests have been created/updated for verifying the changes made.

- [ ] All Integrations tests are passing with the new commits.

- [ ] Submission is not a "code dump".  (Large changes can be made "in repository" via a branch.  Ask on the list.)

Our guidelines for code reviews is at https://cwiki.apache.org/confluence/display/FINERACT/Code+Review+Guide

PR #722, opened 2020-03-05
Fineract-796 - Use Hikari instead of Tomcat Connection Pool

xurror:
## Description
Use Hikari instead of Tomcat Connection Pool

## Checklist


awasum:
Please @xurror and @percyashu . Fix one issue before moving to another. The issues will not run away and more beginner issues can always be created if they get finished. Fineract has a lot of bugs which needs fixing so dont go about opening alot of PRs without actually finishing them and wasting our time and your energy. Please, pick one issue and work on them for a few days or week before picking another one after u have TRIED EVERYTHING and asked for help...then you can move to pick something else...Please

awasum:
Please....run ./gradlew clean build and ./gradlew clean integrationTest on your local environment before sending PRs. if you do, you realise that there is a Checkstyle voilation on this PR.

docker/server.xml
```diff
@@ -40,23 +40,14 @@
...
-        testOnReturn="true"
-        testWhileIdle="true"
-        timeBetweenEvictionRunsMillis="30000"
-        minEvictableIdleTimeMillis="60000"
-        logAbandoned="true"
-        suspectTimeout="60"
+        minimumIdle="3"
+        maximumPoolSize="10"
+        idleTimeout="60000"
       />
```
vorburger: because for FINERACT-796 we (you) are correctly and as expected replacing the `<jee:jndi-lookup ...>` by a simple regular local Spring `<bean>`, this entire JNDI Resource in Tomcat's server.xml should not be required anymore - please remove this, it should still work.

fineract-provider/src/test/resources/META-INF/context.xml
```diff
@@ -23,7 +23,6 @@
 
 <Context>
 	<Resource type="javax.sql.DataSource" name="jdbc/mifosplatform-tenants"
-		driverClassName="org.drizzle.jdbc.DrizzleDriver" url="jdbc:mysql:thin://localhost:3306/mifosplatform-tenants"
-		username="root" password="mysql" validationQuery="select 1" maxActive="10"
-		maxIdle="4" />
+		driverClassName="org.drizzle.jdbc.DrizzleDriver" jdbcUrl="jdbc:mysql:thin://localhost:3306/mifosplatform-tenants"
+		username="root" password="mysql" />
```
vorburger: as above, it should be possible to remove this entirely

fineract-provider/src/main/resources/META-INF/spring/jndi.xml
```diff
@@ -29,6 +29,12 @@
...
 	<!-- name here must match TestDataSourceConfiguration -->
-	<jee:jndi-lookup jndi-name="java:comp/env/jdbc/mifosplatform-tenants"
-		id="tenantDataSourceJndi" />
+
+    <bean id="tenantDataSourceJndi" 
+        class="com.zaxxer.hikari.HikariDataSource">
+        <property name="driverClassName" value="org.drizzle.jdbc.DrizzleDriver" />
+        <property name="jdbcUrl" value="jdbc:mysql:thin://localhost:3306/mifosplatform-tenants" />
+        <property name="username" value="root" />
+        <property name="password" value="mysql" />
```
vorburger: we cannot hard-code these parameters here, you must use the same placeholders as in the original `docker/server.xml` - makes sense? (This is also why the `docker-compose` test fails on Travis; the `environment` from `docker-compose.yml` is no longer effective.)

fineract-provider/src/main/resources/META-INF/spring/jndi.xml
```diff
@@ -29,6 +29,12 @@
     	http://www.springframework.org/schema/jee http://www.springframework.org/schema/jee/spring-jee-3.0.xsd">
 
 	<!-- name here must match TestDataSourceConfiguration -->
-	<jee:jndi-lookup jndi-name="java:comp/env/jdbc/mifosplatform-tenants"
-		id="tenantDataSourceJndi" />
+
+    <bean id="tenantDataSourceJndi"
```
vorburger: It would be clearer to rename this bean name and drop the *Jndi suffix (because not it's not using JNDI anymore...), and according change the use of this bean in the code (search for it...).  Also, it then probably should not be in `jndi.xml` anymore now.

vorburger (changes requested):

vorburger:
Thanks for working on this! I'd love to get this ready to merge.

So the Checkstyle violation seems to have been fixed, and as far as I can tell from the latest log `integrationTest `passes here now - great! It's just failing in the final `docker-compose` test, see `README.md` and last line in `.travis.yml` for how to launch that locally.  The reason for that is related to the required changes review feedback I just posted - does it make sense?

vorburger:
Also note the just merged https://github.com/apache/fineract/pull/710 which will cause minor merge conflicts with this one which you should manually resolve when doing the required rebase when you work on this next.

xurror:
Thanks.
Makes a lot of sense.
Will send update soon

fineract-provider/src/test/resources/META-INF/context.xml
xurror: Seems to me that this resource file is used in build.gradle in operations such as "tomcatrunwar" etc..., if you see what I mean.

vorburger: well spotted.. but you should be able to either just remove that `configFile = file('src/test/resources/META-INF/context.xml')` in [build.gradle](https://github.com/apache/fineract/blob/develop/fineract-provider/build.gradle), because according to https://github.com/bmuschko/gradle-tomcat-plugin it should not be required; or if it's giving you trouble, you could just keep an empty [context.xml](https://github.com/apache/fineract/blob/develop/fineract-provider/src/test/resources/META-INF/context.xml) (but I don't think you need to; I expect you'll find that you can simply entirely remove that file, and any references to it that you'll find, as part of your work on this PR)

xurror:
@vorburger, I've made changes and this passes travis CI now. It was quite challenging to set a new name.

Please can you re-review?

fineract-provider/build.gradle
```diff
@@ -82,6 +82,7 @@ dependencyManagement {
      dependency 'joda-time:joda-time:+'
      dependency 'org.apache.commons:commons-email:1.3.3'
      dependency 'org.apache.commons:commons-lang3:+'
+     dependency 'com.zaxxer:HikariCP:3.4.2'
```
vorburger: Would it be possible to get the HikariCP version from the Spring BOM, like others? May be just using '+' as the version is enough, or perhaps you need to list it elsewhere, like other dependencies?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/boot/db/DataSourceConfiguration.java
```diff
@@ -38,19 +39,19 @@
     @Autowired JDBCDriverConfig config ;
 
     @Bean
-    public DataSourceProperties dataSourceProperties() {
-    return new DataSourceProperties(config.getDriverClassName(), config.getProtocol(), config.getSubProtocol(), config.getPort());
+    public DataSourceProperties DataSourceProperties() {
```
vorburger: why make this method name upper case? It works, by convention, class names are upper case, and method names are lower case, and it's is a little bit confusing to readers especially when there's an identically named class and method.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/boot/db/DataSourceProperties.java
```diff
@@ -78,31 +77,23 @@ public DataSourceProperties(String driverClassName, String protocol, String subP
...
-        setTestOnBorrow(true);
-        setTestOnReturn(true);
-        setTestWhileIdle(true);
-        setTimeBetweenEvictionRunsMillis(30000);
-        setTimeBetweenEvictionRunsMillis(60000);
-        setLogAbandoned(true);
-        setSuspectTimeout(60);
-
-        setJdbcInterceptors("org.apache.tomcat.jdbc.pool.interceptor.ConnectionState;"
-                + "org.apache.tomcat.jdbc.pool.interceptor.StatementFinalizer;org.apache.tomcat.jdbc.pool.interceptor.SlowQueryReport");
```
vorburger, line 95: I hope nobody is going to miss this... :smiling_imp:

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/TomcatJdbcDataSourcePerTenantService.java
```diff
@@ -85,26 +85,30 @@ private DataSource createNewDataSourceFor(final FineractPlatformTenantConnection
...
-         String jdbcUrl = this.driverConfig.constructProtocol(tenantConnectionObj.getSchemaServer(), tenantConnectionObj.getSchemaServerPort(), tenantConnectionObj.getSchemaName()) ;
-        //final String jdbcUrl = tenantConnectionObj.databaseURL();
-        final PoolConfiguration poolConfiguration = new PoolProperties();
-        poolConfiguration.setDriverClassName(this.driverConfig.getDriverClassName());
-        poolConfiguration.setName(tenantConnectionObj.getSchemaName() + "_pool");
-        poolConfiguration.setUrl(jdbcUrl);
-        poolConfiguration.setUsername(tenantConnectionObj.getSchemaUsername());
-        poolConfiguration.setPassword(tenantConnectionObj.getSchemaPassword());
-
-        poolConfiguration.setInitialSize(tenantConnectionObj.getInitialSize());
```
vorburger, line 97: Noteworthy: We're loosing the initial size; according to https://www.javadoc.io/doc/com.zaxxer/HikariCP/latest/com/zaxxer/hikari/HikariDataSource.html apparently Hikari doesn't have such a property. Oh well.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/TomcatJdbcDataSourcePerTenantService.java
```diff
@@ -85,26 +85,30 @@ private DataSource createNewDataSourceFor(final FineractPlatformTenantConnection
...
-
-        poolConfiguration.setTestOnBorrow(tenantConnectionObj.isTestOnBorrow());
-        poolConfiguration.setValidationQuery("SELECT 1");
```
vorburger, line 100: shouldn't we keep this `setValidationQuery` ? Especially because above in `DataSourceProperties.java` you did keep it..

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/TomcatJdbcDataSourcePerTenantService.java
```diff
@@ -85,26 +85,30 @@ private DataSource createNewDataSourceFor(final FineractPlatformTenantConnection
...
+
+        HikariConfig config = new HikariConfig();
+        config.setDriverClassName(this.driverConfig.getDriverClassName());
+        config.setPoolName(tenantConnectionObj.getSchemaName() + "_pool");
+        config.setJdbcUrl(jdbcUrl);
+        config.setUsername(tenantConnectionObj.getSchemaUsername());
+        config.setPassword(tenantConnectionObj.getSchemaPassword());
+
+        config.setValidationTimeout(tenantConnectionObj.getValidationInterval());
+
```
vorburger, line 99:
```suggestion
// https://github.com/brettwooldridge/HikariCP/wiki/MySQL-Configuration
```

reading https://github.com/brettwooldridge/HikariCP#statement-cache I at first wasn't sure what the following properties do, but then found https://github.com/brettwooldridge/HikariCP/wiki/MySQL-Configuration ... if these are inspired by that, I suggest putting that link inline into the code here, for future readers.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/TomcatJdbcDataSourcePerTenantService.java
```diff
@@ -85,26 +85,30 @@ private DataSource createNewDataSourceFor(final FineractPlatformTenantConnection
...
+        config.addDataSourceProperty("prepStmtCacheSize", "250");
+        config.addDataSourceProperty("prepStmtCacheSqlLimit", "2048");
+        config.addDataSourceProperty("useServerPrepStmts", "true");
+        config.addDataSourceProperty("useLocalSessionState", "true");
+
+        config.addDataSourceProperty("rewriteBatchedStatements", "true");
+        config.addDataSourceProperty("cacheResultSetMetadata", "true");
+        config.addDataSourceProperty("cacheServerConfiguration", "true");
+        config.addDataSourceProperty("elideSetAutoCommits", "true");
+        config.addDataSourceProperty("maintainTimeStats", "true");
```
vorburger: https://github.com/brettwooldridge/HikariCP/wiki/MySQL-Configuration suggests `false` instead of `true` for `maintainTimeStats` .. any reason you chose the opposite in (only) this case? Just curious.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/TomcatJdbcDataSourcePerTenantService.java
```diff
@@ -85,26 +85,30 @@ private DataSource createNewDataSourceFor(final FineractPlatformTenantConnection
...
+
+        config.setAutoCommit(true);
 
         /**
```
vorburger: I would delete this entire old comment at this opportunity.

fineract-provider/src/main/resources/META-INF/spring/jdbc.properties
```diff
@@ -20,4 +20,9 @@
...
 PROTOCOL:jdbc
 SUB_PROTOCOL:mysql:thin
-PORT:3306
\ No newline at end of file
+PORT:3306
+
+fineract_tenants_driver:org.drizzle.jdbc.DrizzleDriver
+fineract_tenants_url:jdbc:mysql:thin://localhost:3306/fineract_tenants
+fineract_tenants_uid:root
+fineract_tenants_pwd:mysql
```
vorburger, line 28: FTR: Seeing this surprised me a little at first (especially given that there's no equivalent before; given that the (now deleted) `fineract-provider/src/test/resources/META-INF/context.xml` was scoped test only, but then I understood how this works. Given that the docker compose test passes proves that overriding these properties actually really is working, so all good.

fineract-provider/src/main/resources/META-INF/spring/hikariDataSource.xml
```diff
@@ -29,6 +29,15 @@
...
+    <bean id="hikariTenantDataSource" class="com.zaxxer.hikari.HikariDataSource">
+        <property name="driverClassName" value="${fineract_tenants_driver}" />
+        <property name="jdbcUrl" value="${fineract_tenants_url}" />
+        <property name="username" value="${fineract_tenants_uid}" />
+        <property name="password" value="${fineract_tenants_pwd}" />
+        <property name="minimumIdle" value="3" />
+        <property name="maximumPoolSize" value="10" />
+        <property name="idleTimeout" value="60000" />
+        <property name="ConnectionTestQuery" value="SELECT 1" />
+    </bean>
```
vorburger, line 63: Let's also have this here:

```suggestion
         <!-- https://github.com/brettwooldridge/HikariCP/wiki/MySQL-Configuration -->
         <property name="dataSource.cachePrepStmts" value="true" /> 
         ... TODO etc.
    </bean>
```

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/boot/db/DataSourceConfiguration.java
```diff
@@ -38,19 +39,19 @@
...
+        return new DataSourceProperties(config.getDriverClassName(), config.getProtocol(), config.getSubProtocol(), config.getPort());
     }
 
     @Bean
-    public DataSource tenantDataSourceJndi() {
-    PoolConfiguration p = getProperties();
-        org.apache.tomcat.jdbc.pool.DataSource ds = new org.apache.tomcat.jdbc.pool.DataSource(p);
-        logger.info("Created new DataSource; url=" + p.getUrl());
+    public DataSource hikariTenantDataSource() {
```
vorburger: So this class is actually one I wrote myself, many many years ago, to get pure Spring Boot working (which subsequently got broken again somehow; fixing it is the point of https://issues.apache.org/jira/browse/FINERACT-764 and https://issues.apache.org/jira/browse/FINERACT-730), because as long as there was a DS in JNDI from Tomcat's server.xml, this was needed. But now that with this change there is an equivalent `hikariTenantDataSource` bean already, we should be able to just get rid of this class entirely... do you want to try to just delete it, and see if that just works? Anything that requires this bean should now get it from your new `hikariDataSource.xml` - makes sense?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/boot/db/DataSourceProperties.java
```diff
@@ -31,8 +31,7 @@
  * -Dfineract.datasource.port=3307.
  */
 // NOT a @Component - we do not want this to picked up by component scan, only explicitly declared in DataSourceConfiguration (if that's active)
-public class DataSourceProperties extends PoolProperties {
-
+public class DataSourceProperties extends HikariConfig {
```
vorburger: If you manage to delete `DataSourceConfiguration.java` (above), then I suspect we can delete this entire class as well.. Or is there anything else left that uses this?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/TomcatJdbcDataSourcePerTenantService.java
```diff
@@ -85,26 +85,30 @@ private DataSource createNewDataSourceFor(final FineractPlatformTenantConnection
...
+        config.addDataSourceProperty("cachePrepStmts", "true");
+        config.addDataSourceProperty("prepStmtCacheSize", "250");
+        config.addDataSourceProperty("prepStmtCacheSqlLimit", "2048");
+        config.addDataSourceProperty("useServerPrepStmts", "true");
+        config.addDataSourceProperty("useLocalSessionState", "true");
+
+        config.addDataSourceProperty("rewriteBatchedStatements", "true");
+        config.addDataSourceProperty("cacheResultSetMetadata", "true");
+        config.addDataSourceProperty("cacheServerConfiguration", "true");
+        config.addDataSourceProperty("elideSetAutoCommits", "true");
```
vorburger: Actually, I wonder if it would somehow be possible to read this from the bean for the tenants DB, instead of having to repeat and hard-cod it here... but we could look into that in a follow-up PR as well, your call.

vorburger (changes requested):
@xurror this is GREAT WORK - thanks!!  

I have some new and now more minor feedback, but I think we'll able to merge this very soon.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/boot/db/DataSourceConfiguration.java
xurror: Oh, little mistake on my behalf. It escaped my attention. Will patch that.
Thanks

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/boot/db/DataSourceProperties.java
xurror: Is it better I just comment it out?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/TomcatJdbcDataSourcePerTenantService.java
xurror: Yeah, I tried as much as possible to replicate the properties that are common to the 2 connection pools

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/boot/db/DataSourceConfiguration.java
xurror: I would check that out.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/boot/db/DataSourceProperties.java
vorburger: No, just delete it, it's fine.

xurror:
@vorburger I just applied the suggested changes and the build is successful. 
I have completely removed the highlighted classes and also the hikar imports in the gradle files, turns out the later versions of spring ship with it as default. 
I also added Hikari to logback, hooe that's fine too

xurror:
@vorburger please, any follow up on this?

vorburger:
I'll re-review this hopefully hopefully this weekend, tomorrow or day after.

>

fineract-provider/src/main/resources/logback.xml
```diff
@@ -44,6 +44,7 @@
 	<logger name="org.springframework.beans" level="info" />
 	<logger name="net.sf.ehcache" level="error" />
 	<logger name="org.hibernate.cache" level="error" />
+	<logger name="com.zaxxer.hikari" level="debug" additivity="false" />
```
vorburger, line 47: I'm wondering, won't this (at level debug instead of maybe warn) create a lot of extra logging? 

I don't mind merging it as is (I'm just re-reading everything else), we can always change this log level later.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/TomcatJdbcDataSourcePerTenantService.java
vorburger: @xurror it looks like we lost this again? I'll merge this, and then raise a follow-up PR for you to review where I'll propose to add something like this back again...

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/TomcatJdbcDataSourcePerTenantService.java
vorburger: @xurror this also didn't seem to have been done? I'll propose to change this in my follow-up PR.

vorburger (approved):

PR #725, opened 2020-03-08
Fix audit trails filter (FINERACT-808)

vorburger:
As per the great analysis by Manthan Surkar (@thesmallstar) in the original PR #723, the cause of
the [FINERACT-808](https://issues.apache.org/jira/browse/FINERACT-808) bug was that "the backend would treat "UPDATE" and similar strings as SQL injection".

The root cause of that was that (IMHO..) how Fineract has implemented SQL injection prevention is more of a workaround (it's just blacklisting some keywords and doing some heuristic checks in `SQLInjectionValidator.validateSQLInput`) then how this really should be done (by always and everywhere only ever using JDBC Prepared statements with arguments for all external data, instead inlining them into SQL String concatenation).

This also lays the foundation for more like this in [FINERACT-854](https://issues.apache.org/jira/browse/FINERACT-854).

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/commands/api/AuditsApiResource.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/commands/api/AuditsApiResource.java
@@ -38,17 +38,16 @@ import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.UriInfo;
-import org.apache.commons.lang.StringUtils;
 import org.apache.fineract.commands.data.AuditData;
 import org.apache.fineract.commands.data.AuditSearchData;
 import org.apache.fineract.commands.service.AuditReadPlatformService;
-import org.apache.fineract.infrastructure.core.api.ApiParameterHelper;
 import org.apache.fineract.infrastructure.core.api.ApiRequestParameterHelper;
 import org.apache.fineract.infrastructure.core.data.PaginationParameters;
 import org.apache.fineract.infrastructure.core.serialization.ApiRequestJsonSerializationSettings;
 import org.apache.fineract.infrastructure.core.serialization.DefaultToApiJsonSerializer;
 import org.apache.fineract.infrastructure.core.service.Page;
 import org.apache.fineract.infrastructure.security.service.PlatformSecurityContext;
+import org.apache.fineract.infrastructure.security.utils.SQLBuilder;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Scope;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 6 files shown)


vorburger:
@thesmallstar I'm curious if (and how) you like this idea / approach... it's obviously very much inspired by your work in #723!  If you do do like this, and are in favour of this approach, then you can "LGTM" this PR, and perhaps would be willing to close your #723 ?  

If you are further interested in the general topic of security and SQL injections, you would be very very welcome to continue working in this area! Some ideas: 1. Implement the `TODO` about the `toString()` in `SQLBuilder`[`Test`] (which I didn't have time for); 2. similarly change `CenterReadPlatformServiceImpl`'s and `GroupReadPlatformServiceImpl`'s remaining use of `ApiParameterHelper.sqlEncodeString` to this approach (and then delete `sqlEncodeString`), and then more work as described in [FINERACT-854](https://issues.apache.org/jira/browse/FINERACT-854)...

thesmallstar:
I am very satisfied with this approach(I would have tested it and update).
I am willing to work on the files and issue you mentioned 😄
Thanks - Manthan

awasum:
@thesmallstar  Have u been able to test this with the front end? So you pull this PR to a branch and run this with the front end to see if the issue is solved functionally.

thesmallstar:
@awasum  sorry for the late response, I am bedridden due to high fever.
Will test and tell in an hour.
thanks- Manthan

thesmallstar:
@vorburger  @awasum 
The API is not working  with any action name filter, Status code: 500 
message: "criteria column name must match [a-z]: aud.action_name ="

image

Note: it is working fine without any filter:
image

I am trying to find the bug.
Edit: The problem is with regex, trying to fix it.

thanks- Manthan

vorburger:
Oh I see the problem, that's really easy to fix: That ATOZ RegExp in my new class SqlBuilder is much too strict.. it should allow (match) not only a-z but also . and - and _ and.. maybe a few more special characters that typically appear in column names. Do you want to try adding that to fix? Once you have it working, you could push a new PR, with my original commit, and then a 2nd Fix commit attributing your work. PS: Get well first, don't work with fever, nobody is dying ;-) because of this bug, this can wait until you're feeling better.

thesmallstar:
@vorburger, yes I will fix it :) 
thanks - Manthan

thesmallstar:
@vorburger 

It turns out everything is a valid column name as long as we use it between quote :P
check this
image


For our convenience though, we can limit it to this form of regex [validtablename.]validcolumnname
[ ] -> means optional, and valid table name and valid column name comes from: [here](https://stackoverflow.com/questions/4977898/check-for-valid-sql-column-name)
Regex: [a-zA-Z_][a-zA-Z0-9_]*

any suggestions?

vorburger:
Closing as this was included in #729 (this plus a required fix).

PR #719, opened 2020-03-03
FINERACT-809 - Upgrade rest-assured

xurror:
## Description
Upgrade rest-assured to latest spring boot compatible version https://issues.apache.org/jira/browse/FINERACT-809

## Checklist
Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Commit message starts with the issue number from https://issues.apache.org/jira/projects/FINERACT/. Ex: FINERACT-646 Pockets API.

- [x] Coding conventions at https://cwiki.apache.org/confluence/display/FINERACT/Coding+Conventions have been followed.

- [ ] API documentation at https://github.com/apache/fineract/blob/develop/api-docs/apiLive.htm has been updated with details of any API changes.

- [ ] Integration tests have been created/updated for verifying the changes made.

- [x] All Integrations tests are passing with the new commits.

- [ ] Submission is not a "code dump".  (Large changes can be made "in repository" via a branch.  Ask on the list.)

Our guidelines for code reviews is at https://cwiki.apache.org/confluence/display/FINERACT/Code+Review+Guide

PR #726, opened 2020-03-08
upgrade RestAssured version (FINERACT-809)

vorburger:
Co-authored-by: percyashu <percyayukashu@gmail.com>


vorburger:
This is just a rebase of  #719 from @xurror and @percyashu and I'm just curious if this passes... I ran this locally x3 times, and it failed with https://issues.apache.org/jira/browse/FINERACT-855 every time - may be that is not just a flaky test, but really something that this PR breaks, for some (strange, yes) reason.

NB that other PRs passed today (e.g. #723 and #725) so if this still fails, perhaps the new RestAssured version is somehow changing some timing or something which causes IT test failures?!

vorburger:
Now this failed due to yet another new IT failure, see [FINERACT-857](https://issues.apache.org/jira/browse/FINERACT-857)..

awasum:
@percyashu or @xurror ..Please can we look into this?

vorburger:
#719

PR #728, opened 2020-03-08
throw exception with details in InteropServiceImpl (FINERACT-860)

vorburger:
instead of the NoSuchElementException from Optional.get(), we're now
throwing an UnsupportedOperationException which includes the details of
the missing object.  (This kind of detailed exception was already thrown
in two existing places in the same code after a null check, but it never
reached there, because the Optional.get() failed earlier.  Another
option would have been to return null for an absent Optional, but this
is better.)

The other unrelated changes are automatic clean up actions (in Eclipse).

see https://issues.apache.org/jira/projects/FINERACT/issues/FINERACT-860


vorburger:
@awasum want to review this one? See the JIRA for some background.

awasum (approved):
LGTM. Merging now. Thanks @vorburger

PR #730, opened 2020-03-10
Added: SQL builder toString function

thesmallstar:
## Description
To continue to work done in #725 by @vorburger this PR intends to complete one of the remaining tasks.
Please use the following for functional testing: http://tpcg.io/QfG1joot [outdated] (directly run for sample output)
After this PR, I intend to write proper test statements as required.
Note: Another approach could have been directly using sb, I decided to do it this way any suggestions are welcome :) 

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/SQLBuilder.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/SQLBuilder.java
@@ -43,6 +43,8 @@ public class SQLBuilder {
     // This holds the arguments, in the order of the '?' placeholders in sb
     private final List<Object> args = new ArrayList<>();
 
+    // This holds the criterias, where nth element corresponds to nth element in args
+    private final ArrayList<String> crts = new ArrayList<String>();
     /**
      * Adds a criteria for a SQL WHERE clause.
      * All criteria are appended by AND (support for OR, or nesting, can be added when needed).
@@ -84,6 +86,7 @@ public class SQLBuilder {
         }
         sb.append(trimmedCriteria);
         sb.append(" ?");
+        crts.add(trimmedCriteria);
         args.add(argument);
     }
 
@@ -118,9 +121,38 @@ public class SQLBuilder {
     /*
      * Returns a String representation suitable for debugging and log output.
      * This is ONLY intended for debugging in logs, and NEVER for passing to a JDBC database.
+     */
     @Override
     public String toString() {
-        return "SQLBuilder{..."; // TODO implement this...
+       StringBuilder whereClause  = new StringBuilder("SQLBuilder{");
+       for (int i=0;i<args.size();i++)
+        {
+            if (i!=0)
+            {
+                whereClause.append("  AND  ");
+            }
+            else
+            {
+                whereClause.append("WHERE  ");
+            }
+            Object currentArg = args.get(i);
+            whereClause.append(crts.get(i));
+            whereClause.append(" ");
+            whereClause.append("[");
+            if(currentArg instanceof String)
+            {
+                whereClause.append("'");
+                whereClause.append(currentArg);
+                whereClause.append("'");
+            }else if(currentArg == null)
+            {
+                whereClause.append("null");
+            }else{
+                whereClause.append(String.valueOf(currentArg));
+            }
+            whereClause.append("]");
+        }
+         whereClause.append("}");
+         return whereClause.toString();
     }
-     */
 }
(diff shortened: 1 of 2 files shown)


fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/SQLBuilder.java
```diff
@@ -118,9 +121,34 @@ public String getSQLTemplate() {
     /*
      * Returns a String representation suitable for debugging and log output.
      * This is ONLY intended for debugging in logs, and NEVER for passing to a JDBC database.
+     */
     @Override
     public String toString() {
-        return "SQLBuilder{..."; // TODO implement this...
+        StringBuilder whereClause  = new StringBuilder();
```
vorburger:
```suggestion
        StringBuilder whereClause  = new StringBuilder("SQLBuilder{ ");
```

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/SQLBuilder.java
```diff
@@ -118,9 +121,34 @@ public String getSQLTemplate() {
...
+            {
+                whereClause.append("  AND  ");
+            }
+            if(whereClause.length()==0)
+            {
+                whereClause.append("WHERE  ");
+            }
+            whereClause.append(crts.get(i) + " " + modifiedArgs.get(i));
+        }
+        return "SQLBuilder{"+ whereClause.toString() +"}";
```
vorburger:
```suggestion
        whereClause.append(" }");
        return whereClause.toString();
```

The point of this is to avoid unnecessary String concatenation object creation.

You should also avoid any other String `+` in this method - just `append()` everything.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/SQLBuilder.java
```diff
@@ -118,9 +121,34 @@ public String getSQLTemplate() {
     /*
      * Returns a String representation suitable for debugging and log output.
      * This is ONLY intended for debugging in logs, and NEVER for passing to a JDBC database.
+     */
     @Override
     public String toString() {
-        return "SQLBuilder{..."; // TODO implement this...
+        StringBuilder whereClause  = new StringBuilder();
+        ArrayList<String> modifiedArgs = new ArrayList<String>();
```
vorburger: you can probably avoid creating this intermediate list? Just iterate and append to the `StringBuilder`, no? Or perhaps I'm missing something here (quick review only; didn't read the code in full details).

vorburger (changes requested):
>  After this PR, I intend to write proper test statements as required.

If I may be a pain :smiling_imp: it is typically considered much better practice to do something like this and have the test in the same PR.. in this particular case, just commenting out the x3 related `TODO` in [SQLBuilderTest.java](https://github.com/apache/fineract/blob/develop/fineract-provider/src/test/java/org/apache/fineract/infrastructure/security/utils/SQLBuilderTest.java) as part of this PR would be great.

thesmallstar:
Thank you for the review, working on the required changes.
Thanks- Manthan

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/SQLBuilder.java
```diff
@@ -118,9 +121,40 @@ public String getSQLTemplate() {
...
+            {
+                whereClause.append("  AND  ");
+            }
+            else
+            {
+                whereClause.append("WHERE  ");
+            }
+            Object currentArg = args.get(i);
+            String currentCri = crts.get(i);
+            StringBuilder modifiedArg  = new StringBuilder("[");
```
vorburger: I think you could still save this `StringBuilder` and just `whereClause.append(...)` as you go? Just a thought.

vorburger (changes requested):
Great! Minor feedback, if you want to further optimize? Just for the fun of it... ;-) 

PS: Also formatting is non-standard, but I can't hold that against you so won't be a pain, let's ignore; it will all get bulk re-formatted at some future point (what we really need is more Checkstyle.. want to help with that?)

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/SQLBuilder.java
thesmallstar: Yes, on it.

thesmallstar:
@vorburger No pain, I am willing to learn the correct way( I plan to stay with fineract for long :P)
1. Can you give a quick link to refer the correct formatting?
2. Please tell how can we optimize this? ( Maybe something I suggested in the Note above, replacing '?')
3. Want to help with the Checkstyle :D

vorburger (approved):
LGTM! I'll merge when build passed.

vorburger:
PS re Checkstyle, see https://issues.apache.org/jira/browse/FINERACT-821, as per https://lists.apache.org/thread.html/r40d39423999b55a4ac2fa96aba0a16b91f80ac9f8e4ecb4c372c61c8%40%3Cdev.fineract.apache.org%3E (also on https://markmail.org/message/hk42ip7feruoaooo), gradually fixing formatting problems allowing to enable new rules in https://github.com/apache/fineract/blob/develop/fineract-provider/config/checkstyle/checkstyle.xml is the way to go.

PR #732, opened 2020-03-13
fix broken build due to small logging error detected by SpotBugs

vorburger:
significant reformatting due to Eclipse auto-save action (someone really
needs to finish the Checkstyle uniform formatting)

see https://github.com/apache/fineract/pull/592

PR #716, opened 2020-03-02
FINERACT-696 - add spotbug bug-pattern plugin

xurror:
## Description
While code reviewing https://github.com/apache/fineract/pull/464, I came accross a catch & printStackTrace, which is of course wrong (it should instead be, correctly, logged, instead).

In an ideal world, it would not take human code review to catch this, but just an automated build failure. This is possible using SpotBugs (not FindBugs anymore; SpotBugs is the new FindBugs, it's a fully compatible successor)  and https://github.com/KengoTODA/findbugs-slf4j/ (for this particular case https://github.com/KengoTODA/findbugs-slf4j/issues/70, but using findbugs-slf4j for Fineract would have value even before that's implemented).

## Checklist
Please make sure these boxes are checked before submitting your pull request - thanks!

- [ ] Commit message starts with the issue number from https://issues.apache.org/jira/projects/FINERACT/. Ex: FINERACT-646 Pockets API.

- [ ] Coding conventions at https://cwiki.apache.org/confluence/display/FINERACT/Coding+Conventions have been followed.

- [ ] API documentation at https://github.com/apache/fineract/blob/develop/api-docs/apiLive.htm has been updated with details of any API changes.

- [ ] Integration tests have been created/updated for verifying the changes made.

- [ ] All Integrations tests are passing with the new commits.

- [ ] Submission is not a "code dump".  (Large changes can be made "in repository" via a branch.  Ask on the list.)

Our guidelines for code reviews is at https://cwiki.apache.org/confluence/display/FINERACT/Code+Review+Guide

PR #733, opened 2020-03-14
Removed Console output and added checkstyle module

thesmallstar:
## Description
There were many TODOs, to get started and learn the correct method I thought this would be the easiest one.
The build is passing but I would require following confirmations:
1. Is this the correct approach, should I continue the same way for other TODOs.
2. Since we are removing the console outputs, was it expected to log it?
3. In Listener.java the build failed if the method msg.getText() is not invoked, current implementation seems like a temporary fix, what is the correct way?
4. Removed Console outputs and replaced them with a new line(blank line) any suggestions on this? 

## Checklist

### DIFF
--- a/fineract-provider/config/checkstyle/checkstyle.xml
+++ b/fineract-provider/config/checkstyle/checkstyle.xml
@@ -47,6 +47,7 @@
         <property name="fileExtensions" value="java"/>
     </module>
   -->
+  
     <module name="TreeWalker">
         <module name="EqualsHashCode"/>
 
@@ -59,6 +60,11 @@
             <property name="customImportOrderRules" value="STATIC###THIRD_PARTY_PACKAGE"/>
         </module>
 
+        <module name="RegexpSinglelineJava">
+            <property name="format" value="System\.(out)|(err)\.print(ln)?\("/>
+            <property name="message" value="Line contains console output."/>
+            <property name="ignoreComments" value="false" />
+        </module>
 <!-- TODO Enable many more checks (go about this one by one, step by step, raise separate PRs fixing and then enforcing):
 
         <module name="HiddenField">
--- a/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/AccountNumberPreferencesTest.java
+++ b/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/AccountNumberPreferencesTest.java
@@ -43,9 +43,11 @@ import org.apache.fineract.integrationtests.common.system.CodeHelper;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class AccountNumberPreferencesTest {
-
+    private final static Logger LOG = LoggerFactory.getLogger(AccountNumberPreferencesTest.class);
     private RequestSpecification requestSpec;
     private ResponseSpecification responseSpec;
     private ResponseSpecification responseValidationError;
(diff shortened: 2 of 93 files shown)


thesmallstar:
The build is failing probably #732 will fix it.

awasum:
Try to fix root cause of build generally failing... Will you like to help? Please check flyway mismatch numbering problem with latest merge...

vorburger:
@thesmallstar rebase this, and we'll review it when build passed!

thesmallstar:
@vorburger the build looks good now :)

vorburger (changes requested):
> Is this the correct approach, should I continue the same way for other TODOs.

YES! Please. That's exactly the right way to do this!

> Since we are removing the console outputs, was it expected to log it?

It's only in integration tests.. where it probably can be useful to debug failing tests. If you have the time to change them to LOG.info statements, I think that would be better.

I've only really reviewed it now, sorry for not noticing before.

>  In Listener.java the build failed if the method msg.getText() is not invoked, current implementation seems like a temporary fix, what is the correct way?

LOG.debug()

> Removed Console outputs and replaced them with a new line(blank line) any suggestions on this?

I didn't understand what you meant here.

thesmallstar:
> I didn't understand what you meant here.

@vorburger , I just realized that was a repeated question, I meant if the method implemented in this PR is correct( turns out it is ).

Replacing with LOG statements, thank you for the review.

thesmallstar:
@vorburger I have replaced System.out with LOG statements, but Spotbug is bothered with log performance and suggesting some changes, Is there a way to solve this rather than going individually through all 200-250 lines?

vorburger:
> @vorburger I have replaced System.out with LOG statements, but Spotbug is bothered with log performance and suggesting some changes, 

Yeah, that's due to the JUST introduced #716, see [FINERACT-696](https://issues.apache.org/jira/browse/FINERACT-696) for background.

 > Is there a way to solve this rather than going individually through all 200-250 lines?

I pulled this locally and noticed there are 270 violations... too many to manually fix up? :smile: Do a bunch every day, instead of all together?

Alternatives I can think of could be to configure Checkstyle to not run this particular check on src/test and only src/main, but I think that will be a PITA to configure (you would need separate configuration files, and somehow tell the Gradle plugin to run different configs for different dirs; I'm not sure that's possible), and also confusing for future developers, so I wouldn't do that and keep it simple.

In terms of best use of time, what I would probably do if I were you is maybe pause this one, and first move on to gradually activating other Checkstyle checks, and then come back to this one later / at the end.

thesmallstar:
@vorburger agreed, actually, I manually added those logger variables to 93 files(each one being unique with class name) and hoped for the reward of merged PR  😸 
I would go for editing this daily then, should be done in a week without being bored :P 

-thanks

thesmallstar:
@vorburger I am planning to add This into my GSOC proposal, Adding all check-style modules and working on removing concatenated SQL strings(using SQL builder) and solving other security issues that are confirmed by the plugin, along with solving some pending bugs on Jira. 
Considering the amount of work, I thought this is a would-be a worthy candidate I have also discussed the same with @awasum  who liked the idea. 

What do think? Do you think this is a good and worthy Idea?

vorburger:
> What do think? Do you think this is a good and worthy Idea?

absolutely, please go for it. (I'm not sure it's a full GSoC project, but I would say start with this; I expect you'll probably be able to complete it sooner than you think - and then you can look for more work!)

awasum:
Status on this?

thesmallstar:
@awasum fixed a part of it, will be done soon :)

thesmallstar:
This will be covered in other PR(closing this one).

PR #755, opened 2020-04-12
FINERACT-826 Remove unused getTimeZone

percyashu:
## Description
Describe the changes made and why they were made. Ignore if these details are present on the associated Jira ticket: https://issues.apache.org/jira/browse/FINERACT-826

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/DateUtils.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/DateUtils.java
@@ -41,7 +41,6 @@ public class DateUtils {
         DateTimeZone zone = null;
         if (tenant != null) {
             zone = DateTimeZone.forID(tenant.getTimezoneId());
-            TimeZone.getTimeZone(tenant.getTimezoneId());
         }
         return zone;
     }


awasum:
Why did you send this as a draft? @percyashu ?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/DateUtils.java
```diff
@@ -41,7 +41,7 @@ public static DateTimeZone getDateTimeZoneOfTenant() {
         DateTimeZone zone = null;
         if (tenant != null) {
             zone = DateTimeZone.forID(tenant.getTimezoneId());
-            TimeZone.getTimeZone(tenant.getTimezoneId());
+//            TimeZone.getTimeZone(tenant.getTimezoneId());
```
awasum: Just completely remove the line...Dont comment it out.

awasum (changes requested):

percyashu:
> Why did you send this as a draft? @percyashu ?

I also want to check if very tenant is guaranteed to always have a TZ.   

@vorburger seems it falls dependent on the OS / JVM TZ if tenant TZ is null.

vorburger:
@percyashu I guess the IT failures here are unrelated to this change, see [FINERACT-885](https://jira.apache.org/jira/browse/FINERACT-885). Just keep trying once a day or so, until we have a green build? I know it's stupid.. if you can dig more into FINERACT-885, perhaps try fixing [FINERACT-723](https://jira.apache.org/jira/browse/FINERACT-723), that would be cool! 

> I also want to check if very tenant is guaranteed to always have a TZ.
> @vorburger seems it falls dependent on the OS / JVM TZ if tenant TZ is null.

if it's not mandatory (`NON NULL`) in the DB, then you can't be sure, right? It would break backwards compatibility for any existing installations out there, agreed? The only thing we (you) could do if you want to fix this is to make a Flyway migration script that sets it to required? Then anyone upgrading would fail, and would be forced to set a TZ for each of their Tenants, if they don't already have one. Then you could make this code assume there always is one, and throw a (clear... include reference to a JIRA!) exception.

percyashu:
@vorburger will make changes.

vorburger:
@percyashu re. build failure, please subscribe to and watch (or help fix??) #764 and #761 for [FINERACT-852](https://issues.apache.org/jira/browse/FINERACT-852)... that's a mess, and help would be most welcome, see recent thread on the mailing list.

vorburger:
@percyashu following #764 and #761 this will pass the build (or not at least not fail because of integration tests) if you rebase this now.

vorburger:
see https://issues.apache.org/jira/browse/FINERACT-899 re. build failure...

vorburger (approved):

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/DateUtils.java
vorburger: @awasum this was done, so I'm going to merge this it - one less

PR #759, opened 2020-04-13
set fineract_tenants_* more nicely in docker-compose.yml

vorburger:
We can use environment variables instead of system properties.

We originally had to use system properties and could not use environment variables, because these used to read by Tomcat, but since the switch to Hikari in FINERACT-796 these are read by Spring Boot instead of Tomcat, so we can now use system properties.

This both "looks nicer" and is useful e.g. for FINERACT-881.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -43,7 +43,11 @@ services:
     depends_on:
       - fineractmysql
     environment:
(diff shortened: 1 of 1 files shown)


vorburger:
This failed due to what #762 will fix (work around, deactivating test); let's rebase this once that is in.

@xurror you're probably interested in this change, see my comment in FINERACT-881.

awasum:
#764 has been merged. lets run this or rebase with develop and see.

awasum:
This is now passing. Do you want to merge this @vorburger ?

PR #764, opened 2020-04-13
comment out flaky SchedulerJobsTest (FINERACT-852)

vorburger:
Also adds a new assertion for a clearer failure message than the
confusing low-level technical ArrayIndexOutOfBoundsException for
whenever this test will be re-activated later.

PR #761, opened 2020-04-13
comment out flaky ClientSavingsIntegrationTest testSavingsAccount_DormancyTracking (FINERACT-852)

vorburger:

PR #765, opened 2020-04-13
improve ClientSavingsIntegrationTest error reporting (FINERACT-852)

vorburger:
Adds a new assertion for a hopefully clearer failure message than the
confusing low-level technical ArrayIndexOutOfBoundsException we
regularly hit on Travis CI.

Similar to https://github.com/apache/fineract/pull/764


vorburger:
@awasum how would you feel about already merging this one? Despite the failing build, which is what #764 would avoid... this change, contrary to #761 does NOT actually comment (deactivate) out either of those two test failing tests, but just attempts to make the error reporting from one of those test failures more clear, for future investigations. 

If LGTY would you be wiling to merge it even with the IT build failure? Risk seems minimal, to me.

@thesmallstar @fynmanoj FYI

awasum (approved):
And I will also merge the other PR which ignores the failing test so as to move development forward.I hope we can come back to fixing them soon.

PR #774, opened 2020-04-18
Ignore :( flaky :( AccountingScenarioIntegrationTest (FINERACT-899)

vorburger:
only checkPeriodicAccrualAccountingFlow_OVER_PAYMENT(), not the others

PR #778, opened 2020-04-21
Display git build info (FINERACT-883)

xurror:
Display fineracts git build info actuator/info endpoint

## Description

## Checklist

### DIFF
--- a/fineract-provider/build.gradle
+++ b/fineract-provider/build.gradle
@@ -49,6 +49,7 @@ buildscript {
       // See: https://github.com/spotbugs/spotbugs-gradle-plugin/issues/128#issuecomment-535864882
      classpath 'com.google.guava:guava:28.1-jre'
      classpath "gradle.plugin.com.github.andygoossens:gradle-modernizer-plugin:1.3.0"
+     classpath "gradle.plugin.com.gorylenko.gradle-git-properties:gradle-git-properties:2.2.2"
   }
 }
 
@@ -68,6 +69,7 @@ apply plugin: "com.github.spotbugs"
 apply plugin: 'checkstyle'
 apply plugin: 'com.github.andygoossens.gradle-modernizer-plugin'
 apply plugin: 'jacoco'
+apply plugin: "com.gorylenko.gradle-git-properties"
 // apply plugin: 'pmd'
 
 dependencyManagement {
@@ -250,8 +252,8 @@ rat {
 
 	//BSD License
 	'**/assets/uglify.js',
-         //Ignore out folder
-         '**/out/**'
+    //Ignore out folder
+    '**/out/**',
 	]
 }
 
@@ -403,7 +405,15 @@ war {
 
 license {
     header rootProject.file('../APACHE_LICENSETEXT.md')
-    excludes(["**/*.html", "**/*.mustache", "**/package-info.java", "**/keystore.jks", "**/swagger-ui/**"])
+    excludes([
+        "**/rat-report.txt",
+        "**/git.properties",
+        "**/*.html",
+        "**/*.mustache",
+        "**/package-info.java",
+        "**/keystore.jks",
+        "**/swagger-ui/**",
+    ])
     strictCheck true
 }
 
@@ -557,6 +567,7 @@ def qualifyVersionIfNecessary(version) {
 springBoot {
     mainClassName = 'org.apache.fineract.ServerWithMariaDB4jApplication'
 }
+
 bootWar {
     mainClassName = 'org.apache.fineract.ServerWithMariaDB4jApplication'
 }
@@ -567,3 +578,16 @@ tasks.withType(SpotBugsTask) {
     reports.html.enabled = true
     reportLevel = "high"
 }
+
+gitProperties {
+    gitPropertiesResourceDir = "${project.rootDir}/src/main/resources"
+    dateFormat = "yyyy-MM-dd'T'HH:mmZ"
+    dateFormatTimeZone = "GMT"
+}
+
+generateGitProperties.outputs.upToDateWhen { false }
+
+clean {
+    delete "${project.rootDir}/src/main/resources/git.properties"
+}
+
--- a/fineract-provider/build.gradle
+++ b/fineract-provider/build.gradle
@@ -43,7 +43,7 @@ buildscript {
      // below
      classpath "org.apache.openjpa:openjpa:$openJPAVersion"
      classpath 'com.radcortez.gradle:openjpa-gradle-plugin:3.1.0'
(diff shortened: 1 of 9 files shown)


vorburger:
@xurror can you locally run and fix `./gradlew rat` ? Some license header is missing.

BTW this has nothing to do with [FINERACT-905](https://jira.apache.org/jira/browse/FINERACT-905), confusion.

PS please ignore the failed Build - Summary, and only look at Travis (see [FINERACT-907](https://jira.apache.org/jira/browse/FINERACT-907)).

xurror:
> PS please ignore the failed Build - Summary, and only look at Travis (see FINERACT-907).

Was wondering where it came from. Thanks for the notice

> @xurror can you locally run and fix ./gradlew rat ? Some license header is missing.

Looking into it. It's kind of weird rat-report.txt is getting checked for license.

xurror:
./gradlew rat  passes locally.

xurror:
Oh now this interesting. Seems .git directory cannot be found here since it is a file which is automatically ignored

This also fails with docker build.

xurror:
It's hard for me to figure out how to pull this off.

I'm thinking to set the option failOnNoDotGitFound to false. @vorburger, @awasum do you think we can go with that.

nikpawar89:
@xurror  i saw the build, in your case following test case is failing: 
 org.apache.fineract.integrationtests.AccountingScenarioIntegrationTest > checkPeriodicAccrualAccountingFlow FAILED
    java.lang.AssertionError at AccountingScenarioIntegrationTest.java:588

For now may be you want to ignore the the testcase . But definitely it is not a long time solution.
It appears to me a concurrency issue on Loan entity object and is not limited to integration tests.

fineract-provider/build.gradle
```diff
@@ -24,32 +24,33 @@ gradle clean tomcatrunwar
...
+    dependencies {
+        classpath 'io.spring.gradle:dependency-management-plugin:1.0.9.RELEASE'
+        classpath 'com.bmuschko:gradle-tomcat-plugin:2.5'
+        classpath 'nl.javadude.gradle.plugins:license-gradle-plugin:0.11.0'
+        classpath 'org.zeroturnaround:gradle-jrebel-plugin:1.1.2'
+        classpath 'org.springframework.boot:spring-boot-gradle-plugin:2.2.4.RELEASE'
+        // below
+        classpath "org.apache.openjpa:openjpa:$openJPAVersion"
+        classpath 'com.radcortez.gradle:openjpa-gradle-plugin:3.1.0'
+        classpath "org.nosphere.apache:creadur-rat-gradle:0.6.0"
```
vorburger: This is interesting - why? Just newer version? And what does this have to do :smiling_imp: with FINERACT-883? Would you consider undoing this unrelated change here, and raising a separate (JIRA and?) PR for this?  Or does radle-git-properties somehow require this other RAT?

fineract-provider/build.gradle
```diff
@@ -173,86 +175,85 @@ jacocoTestReport{
 }
 
 rat {
-  xmlOutput = false
```
vorburger, line 176: as above, this to me seems totally unrelated to FINERACT-883 ?

fineract-provider/build.gradle
```diff
@@ -403,7 +404,14 @@ war {
 
 license {
     header rootProject.file('../APACHE_LICENSETEXT.md')
-    excludes(["**/*.html", "**/*.mustache", "**/package-info.java", "**/keystore.jks", "**/swagger-ui/**"])
+    excludes([
```
vorburger, line 411: Nice, but in a perfect world better raised in a very small separate PR. (I won't be a PITA and still merge it if you leave this one here though.  But that RAT related change really would be better to separate out.)

fineract-provider/build.gradle
```diff
@@ -567,3 +576,11 @@ tasks.withType(SpotBugsTask) {
...
     reportLevel = "high"
 }
+
+gitProperties {
+    gitPropertiesResourceDir = "$buildDir/classes/java/main/resources"
+    dateFormat = "yyyy-MM-dd'T'HH:mmZ"
+    dateFormatTimeZone = "GMT"
+}
+
+generateGitProperties.outputs.upToDateWhen { false }
```
vorburger, line 592: May be it would be useful to document what this does inline? According to https://github.com/n0mer/gradle-git-properties/blob/master/README.md#other-usages:

```suggestion
// make sure the generateGitProperties task always executes (even when git.properties is not changed)
generateGitProperties.outputs.upToDateWhen { false }
```

vorburger:
@xurror re. test failure, follow #782 to find FINERACT-899 and note #774, I think if you rebase this, that will disappear.

> Oh now this interesting. Seems .git directory cannot be found here since it is a file which is automatically ignored

I don't think the `.git/` _is a file which is automatically ignored_, but I suspect there are several situations where it's not present... this project certainly should still build even if there is no `.git/` directory present.  This could happen if someone builds from the Release ZIP, and a few other situations I can imagine...

> I'm thinking to set the option failOnNoDotGitFound to false. @vorburger, @awasum do you think we can go with that.

... therefore something like that would probably be acceptable / good idea; see FINERACT-911.

However I do not find an option `failOnNoDotGitFound` so I suspect this was a typo?

vorburger (changes requested):

xurror:
> @xurror re. test failure, follow #782 to find [FINERACT-899](https://issues.apache.org/jira/browse/FINERACT-899) and note #774, I think if you rebase this, that will disappear.
> 
> > Oh now this interesting. Seems .git directory cannot be found here since it is a file which is automatically ignored
> 
> I don't think the `.git/` _is a file which is automatically ignored_, but I suspect there are several situations where it's not present... this project certainly should still build even if there is no `.git/` directory present. This could happen if someone builds from the Release ZIP, and a few other situations I can imagine...
> 
> > I'm thinking to set the option failOnNoDotGitFound to false. @vorburger, @awasum do you think we can go with that.
> 
> ... therefore something like that would probably be acceptable / good idea; see [FINERACT-911](https://issues.apache.org/jira/browse/FINERACT-911).
> 
> However I do not find an option `failOnNoDotGitFound` so I suspect this was a typo?

Oops, typo. It's failOnNoGitDirectory.

fineract-provider/build.gradle
xurror: Ok. Felt the line was getting long with what I added, so decided to break it.

xurror:
@vorburger, this passes now. Hope the you OK with the actual changes

fineract-provider/build.gradle
vorburger: Sure, OK.

vorburger (approved):
LGTM!

PR #757, opened 2020-04-13
escape all `grouping` in SQL files for MySQL 8 compatibility

vorburger:
Indirectly related to https://issues.apache.org/jira/browse/FINERACT-762 - not sure if strictly required, but seems like a good idea anyway at this opportunity, so why not.

As originally on https://github.com/apache/fineract/pull/638/files

PR #780, opened 2020-04-22
FINERACT-897 -  Added missing `grouping` keywords from https://github.com/apache/fi…

nemux:
Some keywords escaped from `grouping` are missing at https://github.com/apache/fineract/pull/757.

### DIFF
--- a/fineract-provider/src/main/resources/sql/migrations/core_db/V134_2__permissions_spelling_correction.sql
+++ b/fineract-provider/src/main/resources/sql/migrations/core_db/V134_2__permissions_spelling_correction.sql
@@ -17,4 +17,4 @@
 -- under the License.
 --
 
-update m_permission set `grouping`="organisation" where grouping = "organistion";
+update m_permission set `grouping`="organisation" where `grouping` = "organistion";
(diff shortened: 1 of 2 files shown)


vorburger (approved):

vorburger:
Thanks! So this is strictly required for MySQL v8? These things have a tendency to creep back in, over time... ;-) Any thoughts about how to prevent that? Would you like to try switching the Travis CI configuration to use MySQL v8?

nemux:
> Thanks! So this is strictly required for MySQL v8? These things have a tendency to creep back in, over time... ;-) Any thoughts about how to prevent that? Would you like to try switching the Travis CI configuration to use MySQL v8?

Yes @vorburger , according to the MySQL reference manual GROUPING became a reserved word in 8.0.1 https://dev.mysql.com/doc/refman/8.0/en/keywords.html

imagen

Let me make a test in a previous MySQL version using GROUPING quoted, but according to the documentation should be allowed quoted or not because in previous versions is not a reserved word.

imagen

nemux:
And I think we are going to have problems in Travis CI with that switch because the current version of Drizzle. The code itself do not support the new version because little changes:
imagen
The @@tx_isolation was renamed in 8.0+ to @@transaction_isolation and this need to be changed in Drizzle to work https://github.com/krummas/DrizzleJDBC/pull/48/commits/ccd8a2c5cd1b34c822744f0fbf276671d413e925 unfortunately as you saw, the project looks dead

PR #782, opened 2020-04-26
add Policy re. unstable tests to README

vorburger:
@awasum OK for you? 

@nikpawar89 makes sense?

@xurror @maektwain @fynmanoj @thesmallstar @DBryzz @wkk91193 @nemux @nnatarajan FYI!

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,14 +239,33 @@ Governance and Policies
 [Becoming a Committer](https://cwiki.apache.org/confluence/display/FINERACT/Becoming+a+Committer)
 documents the process through which you can become a committer in this project.
 
+
+Pull Requests
+-------------
+
+If your PR is failing to pass our CI build due to a test failure, then:
+
+1. Understand if the failure is due to your PR or an unrelated unstable test.
+1. If you suspect it is because of a "flaky" test, and not due to a change in your PR, then please do not simply wait for an active maintainer to come and help you, but instead be a proactive contributor to the project - see next steps.
+1. Search for the name of the failed test on https://issues.apache.org/jira/, e.g. for `AccountingScenarioIntegrationTest` you would find [FINERACT-899](https://issues.apache.org/jira/browse/FINERACT-899).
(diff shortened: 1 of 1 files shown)


vorburger:
The latest "force push" just now just added `@Ignore // TODO FINERACT-123` and _Of course, we very much appreciate you then jumping onto any such bugs and helping us figure out how to fix all ignored tests!_.

awasum:
Why is the other Travis PR test failing?

I see this: 
```org.apache.fineract.integrationtests.AccountingScenarioIntegrationTest > checkPeriodicAccrualAccountingFlow FAILED
    java.lang.AssertionError at AccountingScenarioIntegrationTest.java:588```

Is this a flaky test? Something must be broken somewhere.

xurror (commented):
I suggest that instead of having all this details stated out in one big paragraph, wouldn't it be better if the instructions were listed out in points form?. Reading a huge paragraph is kind of tedious and difficult to track. I have no problem with it though. Just think it would look better that way

vorburger:
> I suggest that instead of having all this details stated out in one big paragraph, wouldn't it be better if the instructions were listed out in points form?. Reading a huge paragraph is kind of tedious and difficult to track. I have no problem with it though. Just think it would look better that way

That is an excellent point, you are absolutely right, it was a bit "dense"... how about how it's now - better?

xurror:
> > I suggest that instead of having all this details stated out in one big paragraph, wouldn't it be better if the instructions were listed out in points form?. Reading a huge paragraph is kind of tedious and difficult to track. I have no problem with it though. Just think it would look better that way
> 
> That is an excellent point, you are absolutely right, it was a bit "dense"... how about how it's now - better?

Yeah, definitely. That looks more like it.

nikpawar89:
+1 from me.

vorburger:
I've exceptionally self merged this README only change PR based on the 2 votes of support expressed by @xurror and @nikpawar, because I have too much gray hear to re-re-base it following the successful build and merge of #785 and then wait for it to build and turn green...

PR #786, opened 2020-04-26
add @Ignore to another ClientLoanIntegrationTest method (FINERACT-885)

vorburger:
loanWithFlatCahargesAndPeriodicAccrualAccountingEnabled()


vorburger:
@awasum @xurror @nikpawar89 no objections to also ignoring this one, I assume? See FINERACT-885.

Based on agreement reached today in #782.

PR #791, opened 2020-05-02
change community-app port in docker compose from 80 to 9090 (FINERACT-845)

vorburger:
This makes it much easier and more secure to use it "rootless",
e.g. with https://podman.io's podman-compose instead (Moby) Docker.

Also includes minor README improvements useful for beginners.

re. FINERACT-845


vorburger:
@xurror is this OK with you?

xurror (approved):
This follow up is awesome. I can only like it.
LGTM

PR #795, opened 2020-05-02
remove './gradlew tomcatRun' - it's both un-used and appears totally broken anyway

vorburger:


maektwain:
This test also failed, are we going to use @Ignore again, seems like this is more consistent now.

vorburger:
I'm converting this to Draft, because I created it re. https://markmail.org/thread/jnlerankgfbq6zhe, but it looks like Petri Tuomola may actually use tomcatRun for FINERACT-730, so let's hold this.

PS: @maektwain Yes, as per https://github.com/apache/fineract#pull-requests. Unrelated to this PR.

vorburger:
Closing this one, because both `tomcatRun` and `tomcatRunWAR` are (about to be) removed in #807.

@ptuomola

PR #793, opened 2020-05-02
add @Ignore to more SchedulerJobsTestResults methods (FINERACT-857)

vorburger:
testUpdateAccountingRunningBalancesJobOutcome()
testTransferFeeForLoansFromSavingsJobOutcome()

PR #797, opened 2020-05-02
add new /echo debug API endpoint (FINERACT-915)

vorburger:


maektwain:
This seems quite great stuff, I was reading about the http tracing,(contains useful data) are you suggesting by introducing this (functionality) that we should trace HTTP request. (store)?

Have you thought of using an external tool or just internal spring file based logging.?

It seems like you are out of luck with test's failing today :)

vorburger:
@maektwain thanks for the feedback!  No, this is a simple debug API endpoint, not a full/real HTTP tracing.  I'm interested in possibly exploring tracing separately later, for performance measurement and tuning, but that's totally separate from this small PR.

Re. build failure, this is awaiting #800, then rebase (will also get just merged #793).

vorburger:
Ready for review - do you want to merge this @awasum or any other active committers?

PR #798, opened 2020-05-02
add server.forward-headers-strategy=framework (FINERACT-914)

vorburger:
FINERACT-914

### DIFF
--- a/fineract-provider/properties/basicauth/application.properties
+++ b/fineract-provider/properties/basicauth/application.properties
@@ -19,4 +19,9 @@
 
 spring.profiles.default=basicauth
 management.health.jms.enabled=false
+
+# FINERACT-883
 management.info.git.mode=FULL
+
+# FINERACT-914
+server.forward-headers-strategy=framework
--- a/fineract-provider/properties/basicauth/twofactor/application.properties
+++ b/fineract-provider/properties/basicauth/twofactor/application.properties
@@ -20,4 +20,9 @@
 spring.profiles.default=basicauth
 spring.profiles.active=basicauth,twofactor
 management.health.jms.enabled=false
+
+# FINERACT-883
 management.info.git.mode=FULL
+
+# FINERACT-914
+server.forward-headers-strategy=framework
--- a/fineract-provider/properties/oauth/application.properties
+++ b/fineract-provider/properties/oauth/application.properties
@@ -20,4 +20,9 @@
 spring.profiles.default=basicauth
 spring.profiles.active=oauth
 management.health.jms.enabled=false
+
+# FINERACT-883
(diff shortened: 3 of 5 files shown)


vorburger:
awaiting #800, then rebase (will also get just merged #793)

vorburger:
Ready for review - do you want to merge this @awasum or any other active committers?

awasum:
I am not very familiar with this feature. How is `server.forward-headers-strategy=framework` different from `server.forward-headers-strategy=native`?

Is the `server.forward-headers-strategy=framework` going to affect good old fashion tomcat deployments?

If these are irrelevant questions, you can ignore them and self merge as I am not familiar with the problem. @vorburger

vorburger:
> How is server.forward-headers-strategy=framework different from server.forward-headers-strategy=native

@awasum E.g. https://github.com/spring-projects/spring-boot/issues/18667 explains this a bit, linking to https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.2-Release-Notes#deprecations-in-spring-boot-22, note, quote:

> (...) previously server.use-forward-headers=true would use the web server native support. You can now achieve the same with server.forward-headers-strategy=native. Because each server has a specific behavior, we now offer an alternative which relies on Spring’s ForwardedHeaderFilter: server.forward-headers-strategy=framework. Developers should use the option that fits best their use case.

It seems to me that for us to use a Spring Framework Filter is better than to rely on Tomcat's for this, because this means it will work the same if we ever switched e.g. from Tomcat to Jetty, or if someone were to be crazy enough to deploy the Fineract WAR in something like WAS or WLS... :smiling_imp: Also, `frameworks` actually does work and solves the problem... :smile: see details in FINERACT-914.

> Is the server.forward-headers-strategy=framework going to affect good old fashion tomcat deployments?

No, it should not. This thing basically just checks for certain HTTP Headers related to SSL proxies and relevant for "serious" deployments, such as https://www.fineract.dev. In our "traditional" deployment environments, such as Tomcat running locally, or inside a Docker Container for Docker Compose, those HTTP Headers will not be present, and this PR should be a NOOP for those.

Thanks for the questions and doing a real review!

vorburger:
@awasum OK for you to merge, or LGTM for me to self merge?

vorburger:
@awasum was the explanation above helpful for you to merge this one?

I'll self merge it let's on Sunday (day after tomorrow, 1 week) if I don't hear back, hope fair.

awasum:
I am sorry...I forget about this because I saw your reply on mobile but never returned to merge this. Your explanation was good enough and I dont have any more questions. We can merge this now. Sorry for delay.

PR #799, opened 2020-05-02
use throw instead of confusing return null inside catch

vorburger:
in JournalEntryWritePlatformServiceJpaRepositoryImpl


vorburger:
@awasum how about this? It's apparently done wrong (pattern) in many classes in the code.

@maektwain FYI

maektwain:
Gotcha , should I wait for this to be merged?

maektwain:
We should gtmit

vorburger:
@maektwain I've just self merged it, assuming no objections, to unblock you so that you can rebase.

PR #801, opened 2020-05-02
remove hard-coded versions of Spring Security OAuth2 & JMS (use BOM)

vorburger:
**TODO** `./gradlew tomcatRunWAR` fails due to `Caused by: java.lang.ClassNotFoundException: javax.jms.JMSContext` ...

### DIFF
--- a/fineract-provider/build.gradle
+++ b/fineract-provider/build.gradle
@@ -77,7 +77,7 @@ dependencyManagement {
     }
 
     dependencies {
-        dependency 'org.springframework.security.oauth:spring-security-oauth2:2.3.6.RELEASE'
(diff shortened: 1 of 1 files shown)


vorburger:
I wonder if FINERACT-919 will help to fix this? Probably not, but I'll retry this later, with that.

vorburger:
see also #863 and #832

vorburger:
What I'm suggesting here is actually contradictory to what we discussed in FINERACT-805 and what @xurror wants to do... so I'll probably close this PR when we have a proposal for FINERACT-805 from @xurror.

vorburger:
#1035

PR #803, opened 2020-05-03
add classpath duplicates detection test anf fix problems found (FINERACT-919)

vorburger:
FINERACT-919

PR #807, opened 2020-05-03
FINERACT-730: Fixing bootRun and removing embedded Tomcat

ptuomola:
## Description
Fixed bootRun / bootJar by: 
- Rewriting ServerApplication class based on SpringBoot documentation
- Fixing SSL configuration for Tomcat by removing trust store config
- Reworking dependencies to remove conflicts (conflicting HikariCP, missing commons-logging and commons-io, 

To avoid conflict between embedded Tomcat and Spring Boot Tomcat:
- Added Cargo plugin for Gradle to start a separate Tomcat instance
- Changed integrationTest to use separate Tomcat instance through Cargo plugin
- Removed embedded Tomcat and Gradle Tomcat plugin

TODO:
- Get the dev profile (using MariaDB4j) to work in the same way with integrationTest etc
- Update Confluence pages to remove references to tomcatRunWar etc

## Checklist

- [X] Commit message starts with the issue number from https://issues.apache.org/jira/projects/FINERACT/. Ex: FINERACT-646 Pockets API.

- [X] Coding conventions at https://cwiki.apache.org/confluence/display/FINERACT/Coding+Conventions have been followed.

- [X] API documentation at https://github.com/apache/fineract/blob/develop/api-docs/apiLive.htm has been updated with details of any API changes.

- [X] Integration tests have been created/updated for verifying the changes made.

- [X] All Integrations tests are passing with the new commits.

- [X] Submission is not a "code dump".  (Large changes can be made "in repository" via a branch.  Ask on the list.)


awasum:
Seems there are checkstyle exceptions with the incoming changes. See: https://travis-ci.org/github/apache/fineract/builds/682516455

Thanks for this very important PR @ptuomola

vorburger:
@ptuomola would you consider "squashing" the commits into a single one? Or object if we did, before merging?

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/SchedulerJobHelper.java
```diff
@@ -115,6 +115,9 @@ public void executeJob(String jobName) throws InterruptedException {
 
                 // Executing Scheduler Job
                 runSchedulerJob(this.requestSpec, jobId.toString());
+                
+                // Need to give the job an opportunity to start - otherwise below might return false as job not yet started
+                Thread.sleep(15000);
```
vorburger: @ptuomola would you consider raising this change as a separate PR instead of as part of this one? Or we could try to make a real better fix, based on what you described in FINERACT-855...

ptuomola: Of course - will squash the commits and remove this. Based on the emails, I believe Awasum was going to send a fix to FINERACT-855.

awasum: @ptuomola, I am looking into it but I cannot seem to replicate the issue locally as when running ```./gradlew integrationTest --tests ClientSavingsIntegrationTest ``` 
Everything passes locally. Could it be that our Travis Vm is not powerful enough?
If you have time to get this done earlier than me. Do it.

My fix for FINERACT-855 involved doing something like this where we are accessing the jobHistoryData...like so:

` Assert.assertEquals("Verifying Last Scheduler Job Status", "success", jobHistoryData.get(jobHistoryData.size() == 0 ? jobHistoryData.size() : jobHistoryData.size() - 1).get("status")); ` 

So that we stop getting the ArrayOutOfBounceException.

That may not be a good solution.

vorburger: My suggestion would be that we wrap up this PR ASAP by strictly and naively :imp:  following the process outlined on https://github.com/apache/fineract#pull-requests ... @ptuomola just try to get this PR to fully pass the build by raising separate PRs, and then rebasing, for any tests that fail on you here. The "full" (proper) solution, whether `Thread.sleep(15000)` or something else, should IMHO be worked on completely separate from this PR.

awasum: You are right @vorburger . lets just get his green

vorburger:
@ptuomola this now fails in the (brand new...) `ClasspathHellDuplicatesCheckRuleTest` introduced in https://github.com/apache/fineract/pull/803/files, see FINERACT-919 for background. Please shout here or there if that is not clear enough and you require assistance - without hearing from you, I'll assume that you'll figure it out, whenever you have time for this PR. - Again, we appreciate this contribution very much!

ptuomola:
> @ptuomola this now fails in the (brand new...) `ClasspathHellDuplicatesCheckRuleTest` introduced in https://github.com/apache/fineract/pull/803/files, see [FINERACT-919](https://issues.apache.org/jira/browse/FINERACT-919) for background. Please shout here or there if that is not clear enough and you require assistance - without hearing from you, I'll assume that you'll figure it out, whenever you have time for this PR. - Again, we appreciate this contribution very much!

Thanks - I've fixed the duplicates, and also had to upgrade the Docker Tomcat image to Tomcat 9 in order to get all of checks to pass. So that should all be OK now.

My only concern about this pull request is: after this, Fineract will need Tomcat 9 or above to run. The WAR does not work on Tomcat 7/8 anymore as it has been compiled against Tomcat 9 JARs, and there are some classes that are not present in 7/8. Is that going to be a problem? I.e. are there people who are using Tomcat 7 and expecting Fineract to work? Do we need to communicate this to everyone somehow so that they are aware?

If this is an issue, it might be possibility to build two versions of the WAR - one for Tomcat (including 7/8) and another one for Spring Boot. But if no one is using Tomcat 7/8 (or if we can expect people to upgrade to 9), then of course it would be good to keep things simple...

vorburger:
Re. Tomcat versions, that's totally fine IMHO. But we should document it - do you want to add just 1 bullet point about it on https://github.com/apache/fineract#requirements as part of this PR, like this:

* Java >= 1.8 (Oracle JVMs have been tested)
* Tomcat 9 (since FINERACT-730; Tomcat 7/8, now 9 *is* required)
* MySQL 5.5

Dockerfile
```diff
@@ -37,18 +37,20 @@ RUN ./gradlew clean -x rat -x test war
 
 # =========================================
 
-FROM bitnami/tomcat:7.0.94 as fineract
+FROM bitnami/tomcat:latest as fineract
```
vorburger:
```suggestion
FROM bitnami/tomcat:9.0 as fineract
```

because `:latest` is always dangerous (it could silently break us when Tomcat 10 is released)

docker/server.xml
```diff
@@ -108,7 +108,7 @@
     -->
 
     <!-- Define an AJP 1.3 Connector on port 8009 -->
-    <Connector port="8009" protocol="AJP/1.3" redirectPort="8443" />
+    <Connector port="8009" protocol="AJP/1.3" redirectPort="8443" secret="xyz123" />
```
vorburger, line 112: Hm, a "magical" :imp: `secret`... I doubt anyone uses the AJP Connector with Tomcat in a Container, and for better security than shipping with a default password (bad best practice) I suggest that we just remove (or at least comment out) this `Connector` entirely here, what do you think?

vorburger (changes requested):

docker/server.xml
ptuomola: Thanks - that makes sense. I only added it to suppress an error message from Tomcat 9. But you're right that a better approach would be to disable the connector altogether. Will do that next. 

And the magical secret was reuse of the keystore password in the same config file :-)

vorburger (approved):

Dockerfile
vorburger: @ptuomola did you mean to Commit Suggestion to accept this, or manually work this into your Commit? This Conversation was resolved, and as I was looking this over about to merge it, I've noticed that it hasn't actually been changed... or perhaps I'll just merge this, and raise a quick follow-up PR myself proposing to change it.

Dockerfile
```diff
@@ -37,18 +37,20 @@ RUN ./gradlew clean -x rat -x test war
...
 USER root
 RUN apt-get update -qq && apt-get install -y wget
 
 COPY --from=builder /fineract/build/libs/fineract-provider.war /opt/bitnami/tomcat/webapps
+COPY --from=builder /usr/local/openjdk-8/jre /usr/local/openjdk-8/jre
 
 RUN keytool -genkey -keyalg RSA -alias tomcat -keystore /opt/bitnami/tomcat/tomcat.keystore -keypass xyz123 -storepass xyz123 -noprompt -dname "CN=Fineract, OU=Fineract, O=Fineract, L=Unknown, ST=Unknown, C=Unknown"
 COPY ./docker/server.xml /opt/bitnami/tomcat/conf
 RUN chmod 664 /opt/bitnami/tomcat/conf/server.xml
+#RUN chown -R tomcat /opt/bitnami/tomcat
```
vorburger: @ptuomola partially out of curiosity, what's this for, being commented out? If it's not required, is cool to remove this? We can do this in a follow-up PR, just to get this one in... :smiley:

ptuomola: There's a problem with the Docker image for Tomcat: the temp directory is owned by root rather than tomcat - which means Tomcat spits out an error message when starting. 

I was first going to fix it by changing owner of all of the directories to Tomcat (i.e. this line), but then realised there are actually two temp directories: /opt/bitnami/tomcat/tmp and /opt/bitnami/tomcat/temp. 

/opt/bitnami/tomcat/tmp has the right permissions & owner, but for some reason the environment variable for temp directory in the Bitnami image points to /opt/bitnami/tomcat/temp. 

I decided that this was probably worthy of a separate issue & PR, but looks like I forgot to remove the line. This could still be an acceptable solution but it might be better just to update the environment variable to point to the right directory - i.e. needs to be investigated. I will remove this now.

ptuomola: Pushed a new commit with this removed

Dockerfile
ptuomola: Sorry. I did Commit Suggestion to accept it, but then looks like when I pushed a new commit with the other changes you suggested (i.e. README.md etc) that got overwritten. I've now pushed a new commit that hopefully addresses all of the comments you raised. Thanks again for all the feedback!

PR #808, opened 2020-05-03
 un-static-ify SchedulerJobHelper

vorburger:
TODO Un-Draft and rebase once other issues in SchedulerJobHelper are solved, and tests un-ignored

re. FINERACT-922

### DIFF
--- a/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/SchedulerJobHelper.java
+++ b/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/SchedulerJobHelper.java
@@ -113,7 +113,7 @@ public class SchedulerJobHelper {
         return response;
     }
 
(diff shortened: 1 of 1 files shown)


vorburger:
Most of this is now in #809 #811 #814 - I'll hold this, and rebase it once those are in.

awasum:
Ok...It all looks good. Why is the other one #814  failing.... its just weird

vorburger:
I've actually done most of what was in this PR was actually done in other PRs meanwwhile, but there were 2 very small changes left here after we re-base which we could still merge (very minor).

vorburger:
Actually, what little is left here now (just removing x2 `static`) will contradict https://errorprone.info/bugpattern/MethodCanBeStatic which I'm assuming (hoping) @percyashu will shortly be adding, so let me abandon this PR instead of merging this.

PR #812, opened 2020-05-03
improve logging in SchedulerTriggerListener (re. FINERACT-922)

vorburger:

PR #813, opened 2020-05-03
add (Quartz's) LoggingTriggerHistoryPlugin (re. FINERACT-922)

vorburger:
To have more visibility into scheduler jobs.

FINERACT-922

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/jobs/service/JobRegisterServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/jobs/service/JobRegisterServiceImpl.java
@@ -47,6 +47,7 @@ import org.quartz.Scheduler;
 import org.quartz.SchedulerException;
 import org.quartz.Trigger;
 import org.quartz.TriggerListener;
+import org.quartz.plugins.history.LoggingTriggerHistoryPlugin;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -342,7 +343,10 @@ public class JobRegisterServiceImpl implements JobRegisterService, ApplicationLi
         schedulerFactoryBean.setQuartzProperties(quartzProperties);
         schedulerFactoryBean.afterPropertiesSet();
         schedulerFactoryBean.start();
-        return schedulerFactoryBean.getScheduler();
+        Scheduler scheduler = schedulerFactoryBean.getScheduler();
+        LoggingTriggerHistoryPlugin loggingTriggerListener = new LoggingTriggerHistoryPlugin();
+        loggingTriggerListener.initialize("Fineract's LoggingTriggerHistoryPlugin", scheduler, null);
+        return scheduler;
     }
 
     private JobDetail createJobDetail(final ScheduledJobDetail scheduledJobDetail) throws Exception {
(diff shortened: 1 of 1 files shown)


vorburger:
Please don't merge this one just yet; I want to do some more local tests to compare this VS #812.

vorburger:
> Please don't merge this one just yet; I want to do some more local tests to compare this VS #812.

Let's actually merge this... together with, not instead of #812 - having both shouldn't do any harm - more logging for eyes is probably a very Good Thing in the case of the scheduler, with all the problems we're having with it.

vorburger:
@awasum @ptuomola review and merge this?

ptuomola:
Looking at the source code, this plugin logs at loglevel INFO. Given the logging guidelines, do we want to therefore enable "permanently" i.e. including non-test configurations?

An option could be to add this to a property file so you can then comment the property in / out when testing - but without needing to recompile:

org.quartz.plugin.triggHistory.class = \
  org.quartz.plugins.history.LoggingTriggerHistoryPlugin

awasum:
> Looking at the source code, this plugin logs at loglevel INFO. Given the logging guidelines, do we want to therefore enable "permanently" i.e. including non-test configurations?
> 
> An option could be to add this to a property file so you can then comment the property in / out when testing - but without needing to recompile:
> 
> org.quartz.plugin.triggHistory.class =
> org.quartz.plugins.history.LoggingTriggerHistoryPlugin

Is this something you can handle in a follow up PR? @ptuomola  Or Is this critical to the usefulness of this PR?

awasum:
I made a mistake and rebased this yesterday. It was not necessary.

ptuomola:
> Is this something you can handle in a follow up PR? @ptuomola Or Is this critical to the usefulness of this PR?

Well... this is a one liner PR to introduce logging for triggers. I'm suggesting another oneliner change that should achieve exactly the same thing but make it easy to turn on/off, so allowing this change to comply with our logging policy. 

So of course we can merge this PR and then someone else can raise another PR to remove the line added here and add another line instead. Or we can just change this PR to use the other line straightaway :-). Either is naturally OK with me, so will let @vorburger decide how he'd like to handle this...

vorburger:
I was going to add it (and test), no worries, just didn't get to it yet (not urgent).

vorburger:
Wait a sec, so iText v5 is AGPL licensed?! Then we can't use it in Fineract, see https://www.apache.org/legal/resolved.html#category-x ...

Reading https://en.m.wikipedia.org/wiki/IText and seeing on https://search.maven.org/search?q=iText that v4.2.2 is ancient from 2015.

I suggest that we use this as the opportunity to switch to https://github.com/LibrePDF/OpenPDF instead. That's dual LGPL (NOK) and BSD (fine).

vorburger:
> add this to a property file so you can then comment the property in / out when testing - but without needing to recompile: `org.quartz.plugin.triggHistory.class = org.quartz.plugins.history.LoggingTriggerHistoryPlugin`

@ptuomola yeah, makes sense, as per http://www.quartz-scheduler.org/documentation/quartz-2.3.0/configuration/ConfigPlugins.html. But actually, looking at this against after a bit of time, and re-comparing it with my other #812, I see that this probably won't really add any more additional useful information anyway, and the logging levels in (my) `SchedulerTriggerListener` are better.

>> Please don't merge this one just yet; I want to do some more local tests to compare this VS #812.
> Let's actually merge this... together with, not instead of #812 - having both shouldn't do any harm - 

I'll therefore "abandon" (close without merge) this proposal, after all.

PR #814, opened 2020-05-03
more clean up in SchedulerJobsTest (re. FINERACT-922)

vorburger:


vorburger:
@awasum @ptuomola does this LGTY? It's ongoing clean-up, while I get familiar with this code...

awasum:
This fails again due to the Scheduler problem.  Other than that...Coding to Interface as you done and other clean ups are relevant. @vorburger . Maybe we should assemble a couple of these scenarios for new contributors and our upcoming GSoC students so they use as starter issues.

This LTGM except we want to add a line to ignore the failing test. if not related to these changes.

vorburger:
#815 adds an ignore to the IT that failed here.

vorburger:
@awasum OK, thanks for super quick review, I've merged this now one.   @ptuomola FYI.

PR #816, opened 2020-05-03
add java.util.Date ISO-8601 GSON type adapter (FINERACT-926)

vorburger:
FINERACT-926

PR #817, opened 2020-05-03
Step 1 of improving SchedulerJobHelper to be more reliable and prevent flaky tests (FINERACT-922)

vorburger:
see FINERACT-922

This PR includes #816 (FINERACT-926). Re. FINERACT-922, please only review the last commit, as the other commits will disappear when I rebase this once we've reached agreement about the best solution re. FINERACT-926 on the dev list, merged that, and rebased this.

### DIFF
--- a/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/SchedulerJobHelper.java
+++ b/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/SchedulerJobHelper.java
@@ -18,23 +18,32 @@
  */
 package org.apache.fineract.integrationtests.common;
 
+import static java.time.Instant.now;
+import static org.awaitility.Awaitility.await;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.Is.is;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.fail;
 
 import com.google.gson.Gson;
 import io.restassured.builder.ResponseSpecBuilder;
 import io.restassured.specification.RequestSpecification;
 import io.restassured.specification.ResponseSpecification;
-import java.util.ArrayList;
+import java.time.Duration;
+import java.time.Instant;
+import java.time.format.DateTimeFormatter;
+import java.time.temporal.ChronoUnit;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.function.ToIntFunction;
+import java.util.concurrent.Callable;
+import java.util.function.Function;
 import java.util.stream.Collectors;
-import org.junit.Assert;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-@SuppressWarnings({ "rawtypes", "unchecked" })
 public class SchedulerJobHelper {
     private final static Logger LOG = LoggerFactory.getLogger(SchedulerJobHelper.class);
     private final RequestSpecification requestSpec;
@@ -53,16 +62,24 @@ public class SchedulerJobHelper {
         this.response202Spec = responseSpec;
     }
 
-    private List getAllSchedulerJobs() {
+    private List<Map<String, Object>> getAllSchedulerJobs() {
         final String GET_ALL_SCHEDULER_JOBS_URL = "/fineract-provider/api/v1/jobs?" + Utils.TENANT_IDENTIFIER;
         LOG.info("------------------------ RETRIEVING ALL SCHEDULER JOBS -------------------------");
-        final ArrayList response = Utils.performServerGet(requestSpec, response200Spec, GET_ALL_SCHEDULER_JOBS_URL, "");
+        List<Map<String, Object>> response = Utils.performServerGet(requestSpec, response200Spec, GET_ALL_SCHEDULER_JOBS_URL, "");
+        assertNotNull(response);
         return response;
     }
 
+    private <T> List<T> getAllSchedulerJobDetails(Function<Map<String, Object>, T> mapper) {
+        return getAllSchedulerJobs().stream().map(mapper).collect(Collectors.toList());
+    }
+
     public List<Integer> getAllSchedulerJobIds() {
-        ToIntFunction<Map> mapper = map -> (Integer) map.get("jobId");
-        return getAllSchedulerJobs().stream().mapToInt(mapper).boxed().collect(Collectors.toList());
+        return getAllSchedulerJobDetails(map -> (Integer) map.get("jobId"));
+    }
+
+    public List<String> getAllSchedulerJobNames() {
+        return getAllSchedulerJobDetails(map -> (String) map.get("displayName"));
     }
 
     public Map<String, Object> getSchedulerJobById(int jobId) {
@@ -103,14 +120,15 @@ public class SchedulerJobHelper {
         return new Gson().toJson(map);
     }
 
-    public List getSchedulerJobHistory(int jobId) {
+    @SuppressWarnings("unchecked")
+    private List<Map<String, Object>> getSchedulerJobHistory(int jobId) {
         final String GET_SCHEDULER_STATUS_URL = "/fineract-provider/api/v1/jobs/" + jobId + "/runhistory?" + Utils.TENANT_IDENTIFIER;
         LOG.info("------------------------ RETRIEVING SCHEDULER JOB HISTORY -------------------------");
-        final Map response = Utils.performServerGet(requestSpec, response200Spec, GET_SCHEDULER_STATUS_URL, "");
-        return (ArrayList) response.get("pageItems");
+        final Map<String, Object> response = Utils.performServerGet(requestSpec, response200Spec, GET_SCHEDULER_STATUS_URL, "");
+        return (List<Map<String, Object>>) response.get("pageItems");
     }
 
-    public static void runSchedulerJob(final RequestSpecification requestSpec, final String jobId) {
+    private void runSchedulerJob(int jobId) {
         final ResponseSpecification responseSpec = new ResponseSpecBuilder().expectStatusCode(202).build();
         final String RUN_SCHEDULER_JOB_URL = "/fineract-provider/api/v1/jobs/" + jobId + "?command=executeJob&" + Utils.TENANT_IDENTIFIER;
         LOG.info("------------------------ RUN SCHEDULER JOB -------------------------");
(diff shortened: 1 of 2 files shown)


maektwain:
aha ! moving to functional style, strongly advise that we move other test cases also to await ability 
though did you check this out [Await](https://github.com/awaitility/awaitility/wiki/Usage)  

some question what about thread configuration ?

Though I would like to see this happening, 

```java 
await().atMost(5, SECONDS).until(scheduler());
```

What I see from your example is that we can do this 

```java 
@Test(timeout = 2000)
    public void awaitOperationSupportsSpecifyingZeroAsPollInterval() {
        new Asynch(fakeRepository).perform();
        with().pollDelay(TWO_HUNDRED_MILLISECONDS).pollInterval(Duration.ZERO).then().await().until(fakeRepositoryValueEqualsOne());
        assertEquals(1, fakeRepository.getValue());
    }
```

Sources are awaitility package

vorburger:
@maektwain thanks for reminding me of Awaitility! I had actually used it in a past life, but kind of forgotten about it, and I guess re-created something like it here... :smile: Let me propose to introduce it in a simpler case in #851, and then use it here as well.

vorburger:
I expect this PR will fail to fully pass integration tests, because it will highlight the problem described in FINERACT-958. Not having the time to deal with FINERACT-958 myself, I'm slightly torn how to proceed here... temporarily slapping an `@Ignore` on the (new) entire `testTriggeringManualExecutionOfAllSchedulerJobs()` seems like such a shame. Should I just hard-code skipping that one Job (`JobName`) which currently fails due to FINERACT-958, so that we can merge this and at least get non-regression coverage for the other jobs, for now? Inputs welcome.

vorburger:
Huh - this, to my surprise, actually passed the build! In that case, let's review and possibly merge...

vorburger:
@awasum @ptuomola would you review and LGTM this for merge if I resolved the conflicts here?

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/SchedulerJobHelper.java
```diff
@@ -122,45 +139,112 @@ private static String runSchedulerJobAsJSON() {
...
+        runSchedulerJob(jobId);
+
+        // Await JobDetailData.lastRunHistory [JobDetailHistoryData] jobRunStartTime >= beforeExecuteTime (or timeout)
+        await().atMost(TIMEOUT).pollInterval(PAUSE).until(jobLastRunHistorySupplier(jobId), lastRunHistory -> {
+            String jobRunStartText = lastRunHistory.get("jobRunStartTime");
+            if (jobRunStartText == null) {
+                return false;
             }
+            Instant jobRunStartTime = df.parse(jobRunStartText, Instant::from);
+            return jobRunStartTime.equals(jobRunStartTime) || jobRunStartTime.isAfter(beforeExecuteTime);
```
ptuomola: Shouldn't the first condition be comparing beforeExecuteTime and jobRunStartTime?

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/SchedulerJobHelper.java
```diff
@@ -51,21 +60,28 @@ public SchedulerJobHelper(final RequestSpecification requestSpec, final Response
...
+        return getAllSchedulerJobDetails(map -> (Integer) map.get("jobId"));
+    }
+
+    public List<String> getAllSchedulerJobNames() {
+        return getAllSchedulerJobDetails(map -> (String) map.get("displayName"));
     }
 
     public Map<String, Object> getSchedulerJobById(int jobId) {
         final String GET_SCHEDULER_JOB_BY_ID_URL = "/fineract-provider/api/v1/jobs/" + jobId + "?" + Utils.TENANT_IDENTIFIER;
-        System.out.println("------------------------ RETRIEVING SCHEDULER JOB BY ID -------------------------");
```
ptuomola: Just a cosmetic comment: with this, the logging from the different requests is not uniform. All the other calls log the "--- <call type> ---" text, this one logs the response. Is there a reason for this?

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/SchedulerJobHelper.java
```diff
@@ -122,45 +139,112 @@ private static String runSchedulerJobAsJSON() {
...
-                Integer jobId = (Integer) allSchedulerJobsData.get(jobIndex).get("jobId");
+                return (Integer) allSchedulerJobsData.get(jobIndex).get("jobId");
+            }
+        }
+        throw new IllegalArgumentException("No such named Job (see org.apache.fineract.infrastructure.jobs.service.JobName enum):" + jobName);
+    }
 
-                // Executing Scheduler Job
-                runSchedulerJob(this.requestSpec, jobId.toString());
+    @Deprecated // FINERACT-922 TODO Gradually replace use of this method with new executeAndAwaitJob() below, if it proves to be more stable than this one
```
ptuomola: Should we not just replace the old implementation with the new, and skip with the gradual transition? Or are there concerns that the new one may not work in some cases?

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/SchedulerJobHelper.java
```diff
@@ -122,45 +139,112 @@ private static String runSchedulerJobAsJSON() {
...
+        // Verify status & propagate jobRunErrorMessage and/or jobRunErrorLog (if any)
+        String status = finalLastRunHistory.get("status");
+        if (!status.equals("success")) {
+            fail("Job status is not success: " + finalLastRunHistory.toString());
         }
+
+        // PS: Checking getSchedulerJobHistory() [/runhistory] is pointless, because the lastRunHistory JobDetailHistoryData is already part of JobDetailData anyway.
+    }
+
+    @SuppressWarnings("unchecked")
```
ptuomola, line 243: Would be great to get rid of the SuppressWarnings... I wonder if there's any clever solution to achieve that?

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/SchedulerJobHelper.java
vorburger: @ptuomola thank you for having taken the time for a "real" (full, detailed - as opposed to just "waving through") code review - I appreciate that very much! So yeah testing that jobRunStartTime is equal to itself was definitely wrong here. But I don't think we want to beforeExecuteTime and jobRunStartTime either - that should always be the case (job should always have started just on or later than the test start). What I really wanted to do here is `jobRunEndTime.equals(jobRunStartTime) || jobRunEndTime.isAfter(jobRunStartTime)` - makes sense? Now so done. Please do ask if unclear!

ptuomola: @vorburger I think the second check (row 228) is correct as you state (i.e. jobRunEndTime.equals(jobRunStartTime) || jobRunEndTime.isAfter(jobRunStartTime))

But the first check I was commenting on (row 217) would still make most sense to me as:

jobRunStartTime.equals(beforeExecuteTime) || jobRunStartTime.isAfter(beforeExecuteTime)

in order to make sure that the job has had a chance to start, before we start looking at its history. 

If we don't do this and go straight to comparing jobRunStartTime and jobRunEndTime, I think there's a risk of hitting the same issue as I observed earlier: i.e. that we started looking for the history before the job even had had a chance to start for this test. 

In such case we would get the previous execution history of the job, jobRunEndTime would still be >= jobRunStartTime - but this would be the previous run of the job, not the current one we had triggered.

Or am I missing something here?

vorburger: No you're not missing anything, I was just looking at the wrong line! Yes, absolutely agree - better now? However this actually now reveals a new real bug! See FINERACT-992... do you happen to have any interest in taking that? Just asking! I'm marking this PR Draft again after all, as we probably have to fix that issue before this can be merged.

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/SchedulerJobHelper.java
vorburger: It's just something that was useful while debugging.. as it's a test, let's keep it, if you don't mind. OK to resolve this point?

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/SchedulerJobHelper.java
vorburger: As far as I can see, no, not easily unfortunately. Why? Vaguely related to what I blabbered on about in FINERACT-949 re. JSON mapping. I do want to think further about that some fine day, but not as part of this PR. OK to resolve this point?

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/SchedulerJobHelper.java
vorburger: When I tried that while originally creating this, it didn't "just work" yet, so I wanted to get this in, and then more gradually attempt to "switch over", one by one, and further investigate related problems. Wanna help? :smiley_cat: OK to resolve this point?

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/SchedulerJobHelper.java
ptuomola: Of course - makes sense. One solution that I thought could work for FINERACT-949 in general would be to use the responseSpec for checking response correctness (instead of looking at return values). But let's continue the discussion under that issue. So OK to resolve.

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/SchedulerJobHelper.java
ptuomola: Sounds good - OK to resolve.

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/SchedulerJobHelper.java
ptuomola: Yep - OK to resolve.

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/SchedulerJobHelper.java
vorburger: @ptuomola I'm resolving this because I have just closed  FINERACT-992. 

Please re-open if NOK for you.

vorburger:
Un-drafting this now that I have clarified FINERACT-992. 

This is now ready for final review and can hopefully be merged very soon.

@ptuomola no objections? @awasum merge this soon-ish?

ptuomola:
Looks great to me!

PR #820, opened 2020-05-04
FINERACT-846: Migrating to Java 11

ptuomola:
## Description
Made changes to Gradle configuration and dependencies to migrate Fineract to Java 11

## Checklist

- [X] Commit message starts with the issue number from https://issues.apache.org/jira/projects/FINERACT/. Ex: FINERACT-646 Pockets API.

- [X] Coding conventions at https://cwiki.apache.org/confluence/display/FINERACT/Coding+Conventions have been followed.

- [X] Integration tests have been created/updated for verifying the changes made.

- [X] All Integrations tests are passing with the new commits.

- [X] Submission is not a "code dump".  (Large changes can be made "in repository" via a branch.  Ask on the list.)

### DIFF
--- a/fineract-provider/build.gradle
+++ b/fineract-provider/build.gradle
@@ -21,6 +21,8 @@ Run as:
 gradle clean bootRun'''
 
 project.ext.jerseyVersion = '1.19.4'
+ext["groovy.version"] = '3.0.3' 
+
 
 buildscript {
     ext {
@@ -103,12 +105,17 @@ dependencyManagement {
         dependency 'io.swagger:swagger-jersey-jaxrs:1.5.15'
         dependency 'jakarta.management.j2ee:jakarta.management.j2ee-api:1.1.4'
         dependency 'jakarta.jms:jakarta.jms-api:2.0.3'
+	dependency 'jakarta.xml.bind:jakarta.xml.bind-api:2.3.3'
         dependency 'org.apache.activemq:activemq-broker:+'
         dependency 'org.apache.bval:org.apache.bval.bundle:2.0.2'
         dependency 'junit:junit:+'
         dependency 'org.mockito:mockito-core:+'
-        dependency 'org.slf4j:slf4j-simple:1.7.30'
+        dependency 'org.slf4j:slf4j-simple:1.8.0-beta4'
+        dependency 'org.slf4j:slf4j-api:1.8.0-beta4'
         dependency 'com.mockrunner:mockrunner-jms:2.0.1'
+	dependency 'dom4j:dom4j:1.6.1'
+        dependency 'xerces:xercesImpl:2.12.0'
+        dependency 'io.github.classgraph:classgraph:4.8.43'
 
         dependencySet(group: 'com.sun.jersey', version: jerseyVersion) {
             entry 'jersey-core'
@@ -120,18 +127,24 @@ dependencyManagement {
             entry 'jersey-spring'
             entry 'jersey-multipart'
         }
-        dependencySet(group: 'org.apache.poi', version: '3.9') {
+        dependencySet(group: 'org.apache.poi', version: '4.1.2') {
             entry 'poi'
             entry 'poi-ooxml'
             entry 'poi-ooxml-schemas'
-        }             
+        }
+        dependencySet(group: 'io.rest-assured', version: '4.3.0')
+        {
+            entry 'rest-assured'
+            entry 'json-path'
+            entry 'xml-path'
+        }
     }
 }
 
 /* define the valid syntax level for source files */
-sourceCompatibility = JavaVersion.VERSION_1_8
+sourceCompatibility = JavaVersion.VERSION_11
 /* define binary compatibility version */
-targetCompatibility = JavaVersion.VERSION_1_8
+targetCompatibility = JavaVersion.VERSION_11
 
 project.ext.mysqlUser='root'
 project.ext.mysqlPassword='mysql'
@@ -270,6 +283,12 @@ configurations {
 	runtime
 	all*.exclude module: 'jul-to-slf4j' // see FINERACT-700 re. ClassCircularityError: java/util/logging/LogRecord
 }
+
+tasks.withType(JavaCompile) {
+    options.compilerArgs << '-Xlint:unchecked'
+    options.deprecation = true
+}
+
 /* Pick up dependencies based on the environemnt, defaults to production */
 if (project.hasProperty('env') && project.getProperty('env') == 'dev') {
     apply from:  'dev-dependencies.gradle'
--- a/fineract-provider/dependencies.gradle
+++ b/fineract-provider/dependencies.gradle
@@ -83,6 +83,7 @@ dependencies {
             'org.apache.poi:poi',
             'org.apache.poi:poi-ooxml',
             'org.apache.poi:poi-ooxml-schemas',
+            'dom4j:dom4j',
 
             'com.lowagie:itext',
 
@@ -96,7 +97,10 @@ dependencies {
             // it's useful to have this for the Spring Boot TestRestTemplate http://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/#boot-features-rest-templates-test-utility
             'org.apache.httpcomponents:httpclient',
 
-            'org.apache.bval:org.apache.bval.bundle'
+            'org.apache.bval:org.apache.bval.bundle',
+
+             // JAX-B dependencies for JDK 9+
+             "jakarta.xml.bind:jakarta.xml.bind-api"
     )
     implementation ('io.swagger:swagger-jersey-jaxrs') {
 		exclude group: 'javax.validation'
@@ -106,8 +110,9 @@ dependencies {
     }
 
     testCompile 'junit:junit',
+            'xerces:xercesImpl',
             'org.mockito:mockito-core',
-            'io.github.classgraph:classgraph:4.8.43',
+            'io.github.classgraph:classgraph',
             'com.google.code.gson:gson',
             'org.springframework:spring-jms',
             'joda-time:joda-time',
@@ -115,12 +120,13 @@ dependencies {
             'org.apache.poi:poi-ooxml',
             'org.springframework:spring-context-support',
             'com.sun.jersey:jersey-core'
-    testCompile ('com.mockrunner:mockrunner-jms') {
-        exclude group: 'regexp'
-    }
+
     testCompile ('io.rest-assured:rest-assured') {
         exclude group: 'commons-logging'
         exclude group: 'org.apache.sling'
+    }
+    testCompile ('com.mockrunner:mockrunner-jms') {
+        exclude group: 'regexp'
     }
     testCompile ('org.springframework.boot:spring-boot-starter-test') {
         exclude group: 'com.jayway.jsonpath', module: 'json-path'
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ dist: trusty
 language: java
 
 jdk:
-  - openjdk8
+  - openjdk12
 
 services:
   - mysql
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 #
-FROM openjdk:8 AS builder
+FROM openjdk:11 AS builder
 
 # COPY . fineract is slow e.g. when using Podman instead of Docker (because it doesn't honor .dockerignore and copies all of .git/** into the container..), so let's explicitly list only what we need:
 COPY fineract-provider/src/main fineract/fineract-provider/src/main/
@@ -39,13 +39,10 @@ RUN ./gradlew clean -x rat -x test war
 
 FROM bitnami/tomcat:9.0 as fineract
 
-ENV JAVA_HOME /usr/local/openjdk-8/jre
-
 USER root
 RUN apt-get update -qq && apt-get install -y wget
 
 COPY --from=builder /fineract/build/libs/fineract-provider.war /opt/bitnami/tomcat/webapps
-COPY --from=builder /usr/local/openjdk-8/jre /usr/local/openjdk-8/jre
 
 RUN keytool -genkey -keyalg RSA -alias tomcat -keystore /opt/bitnami/tomcat/tomcat.keystore -keypass xyz123 -storepass xyz123 -noprompt -dname "CN=Fineract, OU=Fineract, O=Fineract, L=Unknown, ST=Unknown, C=Unknown"
 COPY ./docker/server.xml /opt/bitnami/tomcat/conf
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ If you are interested in contributing to this project, but perhaps don't quite k
 
 Requirements
 ============
-* Java >= 1.8 (Oracle JVMs have been tested)
+* Java >= 11 (Oracle JVMs have been tested)
 * MySQL 5.5
 * Tomcat 9, if deploying to a separate servlet container. Until FINERACT-730, Tomcat 7/8 were also supported, but now Tomcat 9 is required. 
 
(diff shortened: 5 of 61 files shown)


xurror (commented):
The build is failing. Please make sure the changes passes all the tests locally.
run
`./gradlew clean build`
`./gradlew clean check`
`./gradlew clean IntegrationTest`
and be sure they all pass. Also, look at https://github.com/apache/fineract#pull-requests on how to handle failed builds

awasum:
Hello @ptuomola . Thanks for taking on this work. You forgot to change the JDK version to 11 in Travis.

awasum:
Integration Tests are failing. Maybe run them locally and take time to see if you can fix them. This is valuable work. Thanks.

awasum:
Seems to be very similar problems for all the test. Looks like when one is fixed..you can easily copy to the other failing tests and fix them too.

` java.lang.RuntimeException at RecurringDepositTest.java:109
        Caused by: javax.net.ssl.SSLPeerUnverifiedException at RecurringDepositTest.java:109
    java.lang.NullPointerException at RecurringDepositTest.java:3202
org.apache.fineract.integrationtests.RecurringDepositTest > testRecurringDepositAccountWithAmountInterestRateChart_PERIOD_VARIATION FAILED
    java.lang.RuntimeException at RecurringDepositTest.java:109
        Caused by: javax.net.ssl.SSLPeerUnverifiedException at RecurringDepositTest.java:109
    java.lang.NullPointerException at RecurringDepositTest.java:3202
org.apache.fineract.integrationtests.RecurringDepositTest > testRecurringDepositAccountWithClosureTypeTransferToSavings_WITH_HOLD_TAX FAILED
    java.lang.RuntimeException at RecurringDepositTest.java:109
        Caused by: javax.net.ssl.SSLPeerUnverifiedException at RecurringDepositTest.java:109
    java.lang.NullPointerException at RecurringDepositTest.java:3202
org.apache.fineract.integrationtests.RecurringDepositTest > testRecurringDepositAccountUndoApproval FAILED
    java.lang.RuntimeException at RecurringDepositTest.java:109
        Caused by: javax.net.ssl.SSLPeerUnverifiedException at RecurringDepositTest.java:109
    java.lang.NullPointerException at RecurringDepositTest.java:3202
org.apache.fineract.integrationtests.RecurringDepositTest > testMaturityAmountForDailyCompoundingAndMonthlyPosting_With_360_Days FAILED
    java.lang.RuntimeException at RecurringDepositTest.java:109
        Caused by: javax.net.ssl.SSLPeerUnverifiedException at RecurringDepositTest.java:109
    java.lang.NullPointerException at RecurringDepositTest.java:3202
org.apache.fineract.integrationtests.RecurringDepositTest > testRecurringDepositAccountWithAmountAndPeriodInterestRateChart_AMOUNT_VARIATION FAILED
    java.lang.RuntimeException at RecurringDepositTest.java:109
        Caused by: javax.net.ssl.SSLPeerUnverifiedException at RecurringDepositTest.java:109
    java.lang.NullPointerException at RecurringDepositTest.java:3202
org.apache.fineract.integrationtests.RecurringDepositTest > testRecurringDepositAccountWithPeriodInterestRateChart_PERIOD_VARIATION FAILED
    java.lang.RuntimeException at RecurringDepositTest.java:109
        Caused by: javax.net.ssl.SSLPeerUnverifiedException at RecurringDepositTest.java:109
    java.lang.NullPointerException at RecurringDepositTest.java:3202
org.apache.fineract.integrationtests.RecurringDepositTest > testMaturityAmountForDailyCompoundingAndMonthlyPosting_With_365_Days FAILED
    java.lang.RuntimeException at RecurringDepositTest.java:109
        Caused by: javax.net.ssl.SSLPeerUnverifiedException at RecurringDepositTest.java:109
    java.lang.NullPointerException at RecurringDepositTest.java:3202
org.apache.fineract.integrationtests.RecurringDepositTest > testPrematureClosureAmountWithPenalInterestForWholeTerm_With_360_Days FAILED
    java.lang.RuntimeException at RecurringDepositTest.java:109
        Caused by: javax.net.ssl.SSLPeerUnverifiedException at RecurringDepositTest.java:109
    java.lang.NullPointerException at RecurringDepositTest.java:3202`

ptuomola:
Thanks everyone for your comments. This is still a bit WIP - I've just fixed the depreciated and unchecked warnings, as well as some other build errors. Still a few integration tests failing - will fix them next.

ptuomola:
@xurror @awasum Interestingly I don't get the same failures with integration tests when I run them on my machine locally. 

The error seems to caused by the test client not trusting the HTTPS certificate provided by the server. But that should never happen as we are using relaxedHTTPSValidation() from RestAssured, which should skip all of the certificate validations. And on my local machine that seems to work as expected. 

Anyway - will investigate further tomorrow, maybe I can find a way to reproduce the error locally.

vorburger:
@ptuomola I've run into this kind of thing before - it could be caused e.g. by Oracle Java vs. OpenJDK, which include different root certificates. Haven't investigated this particular problem, but thought I'd mention it here in case that is it.

ptuomola:
This is still work in progress - but looked like if I kept it as draft then Travis would not build it. Will let you know once I've fixed the integration tests and then it's ready for review.

xurror:
It's OK, just ping any of us when you think it's ready for review.

ptuomola:
> @ptuomola I've run into this kind of thing before - it could be caused e.g. by Oracle Java vs. OpenJDK, which include different root certificates. Haven't investigated this particular problem, but thought I'd mention it here in case that is it.

@vorburger Thanks! Found the issue: there is a bug in OpenJDK 11.0.2. They've implemented SSL connection reuse, but the reuse discards all the certificates shared during the earlier use of the connection, which results in this error. relaxedHTTPSValidation doesn't help as it still expects some certificate, and in this case there is none. 

The issue is fixed in later OpenJDK 11 builds (e.g. 11.0.7) which is what I have on my own machine, and hence I did not see this. However Travis pulls in 11.0.2 and I was not able to find an easy way to get it to use a later minor version. 

So I fixed the Travis build by getting it to use OpenJDK 12 which has this fix. I'm assuming that will still be OK, because in the Gradle file we specify JDK 11. Alternatively need to find some way to get Travis to download and install 11.0.7 instead...

ptuomola:
@xurror @awasum I found the SSL issue (bug in OpenJDK) and all the tests now seem to pass, so this should be ready for your review.

xurror (commented):
This is some really great work you have done here. I just have some few changes to suggest

fineract-provider/dependencies.gradle
```diff
@@ -80,9 +80,10 @@ dependencies {
 
             'org.drizzle.jdbc:drizzle-jdbc',
 
-            'org.apache.poi:poi',
-            'org.apache.poi:poi-ooxml',
-            'org.apache.poi:poi-ooxml-schemas',
+            'org.apache.poi:poi:4.1.2',
```
xurror: Since https://github.com/apache/fineract/pull/662, we don't want to set version numbers for prject dependencies in dependencies.gradle but in the build.gradle.
Version numbers don't really fit here anymore. Look at https://github.com/apache/fineract/pull/662 for more about this.

fineract-provider/build.gradle
```diff
@@ -21,6 +21,9 @@ Run as:
 gradle clean bootRun'''
 
 project.ext.jerseyVersion = '1.19.4'
+ext["rest-assured.version"] = '4.3.0'
```
xurror: Also, I don't think this will be necessary since you can just define the dependency as is done https://github.com/apache/fineract/pull/662

fineract-provider/dependencies.gradle
```diff
@@ -80,9 +80,10 @@ dependencies {
...
+            'org.apache.poi:poi-ooxml:4.1.2',
```
xurror: I also noticed some of these dependencies were already defined in build.gradle though just older, so you could just go and update the version numbers there. Is that OK with you??

xurror (commented):
I also ran this locally, though it passes, I saw some warnings which caught my attention.
`
> Task :compileIntegrationTestJava
/C/Users/Kaze/Documents/src/fineract/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/StaffTest.java:135: warning: [unchecked] unchecked cast
        final List<HashMap> responseActive = (List<HashMap>)StaffHelper.getStaffListWithState(requestSpec, responseSpec, "active");
                                                                                             ^
  required: List<HashMap>
  found:    Object
/C/Users/Kaze/Documents/src/fineract/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/StaffTest.java:144: warning: [unchecked] unchecked cast
        final List<HashMap> responseInactive = (List<HashMap>)StaffHelper.getStaffListWithState(requestSpec, responseSpec, "inactive");
                                                                                               ^
  required: List<HashMap>
  found:    Object
2 warnings`

Think you might want to check that. Shouldn't be a problem to fix that.

xurror:
Aside from that, this is really a good job. We might have to do something about the java version in travis.yml.
@vorburger please can you have a look at this?

fineract-provider/build.gradle
ptuomola: The reason why I did it this was that with a single line you can ensure that all the rest-assured libraries are brought in with the right version. I was first doing it through dependencies, but that seemed to require separately setting the version for every single rest-assured jar. That is of course much more error-prone: if you miss one, suddenly you are in classpath hell again. But if that's preferred I can of course set it jar-by-jar...

fineract-provider/dependencies.gradle
ptuomola: OK - will have a look to see if there are duplicates, and fix that. Thanks!

fineract-provider/build.gradle
xurror: Yeah, about that. spring dependency has an option to define a set of dependencies(dependencySet) for libraries which have thesame version number for different artifacts. There are samples of that in https://github.com/apache/fineract/pull/662. So you should be able to easily get the gist once you look that. With that you don't have to define dependencies jar-by-jar. Do you get what I mean here?

ptuomola: Sorry - I thought the dependencySet only set the version for the JARs you've liste in the set, but not for the dependencies that those dependencies bring in (i.e. the transitive dependencies). 

But if it works automatically for the non-listed transitive dependencies as well, then of course that will have the same result as this. I'll try it out and change if it works.

ptuomola: Looks like it does not work for transitive dependencies - you need to explicitly list all the jars for which you want the version to apply. I've done this and listed the JARs we are currently using. But of course if we add some in the future then this list needs to be updated to avoid incompatible classes...

awasum:
Good work Petri. Thank you.

ptuomola:
> I also ran this locally, though it passes, I saw some warnings which caught my attention.
> `
> 
> > Task :compileIntegrationTestJava
> > /C/Users/Kaze/Documents/src/fineract/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/StaffTest.java:135: warning: [unchecked] unchecked cast
> > final List responseActive = (List)StaffHelper.getStaffListWithState(requestSpec, responseSpec, "active");
> > ^
> > required: List
> > found:    Object
> > /C/Users/Kaze/Documents/src/fineract/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/StaffTest.java:144: warning: [unchecked] unchecked cast
> > final List responseInactive = (List)StaffHelper.getStaffListWithState(requestSpec, responseSpec, "inactive");
> > ^
> > required: List
> > found:    Object
> > 2 warnings`
> 
> Think you might want to check that. Shouldn't be a problem to fix that.

Yes - I fixed a few hundred depreciated / unchecked warnings, but looks like I missed this one. 

It's actually a bit of a funny case: in some tests, getStaffListWithState returns ArrayList<<HashMap<String, Object>> whereas in some other tests it returns LinkedHashMap. The former is in cases where it is called with expected parameters - the latter when called with invalid parameters. Given this, the obvious solution of just changing the return type from Object to something more specific doesn't really work. 

I suppose I could just get the compiler to shut up with either SuppressWarnings or with an explicit Class.cast(). Any other ideas?

xurror:
> I suppose I could just get the compiler to shut up with either SuppressWarnings or with an explicit Class.cast(). Any other ideas?

I think suppressing the warning is a good idea.

xurror (approved):
Well, this LGTM. Maybe @vorburger will have more insights as far as this is concerned. Thanks again for the great work.

vorburger:
@ptuomola wow, great job, nice find re. JDK!!! I'm just curious, did you notice this by comparison of Travis's VS local JDK version numbers, or do you have a link to a OpenJDK bug about this?

Using JDK 12 instead of 11 can't stay long, because even though Gradle specifies 11, that will only cause javac to allow only 11 SYNTAX, but it will still allow to build with use of 12 LIBRARIES - I've dealt with related problems in other projects, over the years. But what I think we should do due to OpenJDK bug is still do this, for now, merge, but then change it "soon". I volunteer to look into it for Travis later. Perhaps create a new JIRA dedicated to "Switch Travis from Java 12 to Java 11" (and assign it to me)?

As for javac compiler warnings, let's fix in follow up PRs. Personally I may even have left the warning instead of suppressing, as a reminder, but it really doesn't matter, ignore. 

I was also going to suggest that this could have been broken up into separate PRs first for all the code changes, while still on Java 8, and then figure out make the Java 11 build changes, but seeing your breakthrough, I'll get this in ASAP; so more of a suggestion for future smaller PRs.

Let me fully review this by staring at the diff a little longer after work today, and likely merge ASAP.

awasum:
This is another major change for the project. So a mailing list message
about this will be great. Thank you all.

ptuomola:
Many thanks @vorburger for your feedback and comments

> @ptuomola wow, great job, nice find re. JDK!!! I'm just curious, did you notice this by comparison of Travis's VS local JDK version numbers, or do you have a link to a OpenJDK bug about this?

I found it by turning on the SSL debug logging and stepping through the SSL handshakes / negotiation. But once I worked out what's going on, it was easy to find a JDK bug that someone else had raised on the same issue and that has been fixed in a later JDK minor version (https://bugs.openjdk.java.net/browse/JDK-8212885 - TLS 1.3 resumed session does not retain peer certificate chain)
 
> Using JDK 12 instead of 11 can't stay long, because even though Gradle specifies 11, that will only cause javac to allow only 11 SYNTAX, but it will still allow to build with use of 12 LIBRARIES - I've dealt with related problems in other projects, over the years. But what I think we should do due to OpenJDK bug is still do this, for now, merge, but then change it "soon". I volunteer to look into it for Travis later. Perhaps create a new JIRA dedicated to "Switch Travis from Java 12 to Java 11" (and assign it to me)?

That makes sense. There are some possible ways suggested on how to get a specific version of OpenJDK on different sites - e.g. https://stackoverflow.com/questions/29636754/can-you-specify-minor-jdk-version-for-travis-ci. However none of them seemed particularly "clean". 

I also considered upgrading the whole Linux distro being used by Travis to a newer version (probably a good idea anyway). But unfortunately the newer distro version still uses the same script to pull OpenJDK, and hence seemed to end up with the same JDK version. 

> As for javac compiler warnings, let's fix in follow up PRs. Personally I may even have left the warning instead of suppressing, as a reminder, but it really doesn't matter, ignore.

I've left this for now. 

> I was also going to suggest that this could have been broken up into separate PRs first for all the code changes, while still on Java 8, and then figure out make the Java 11 build changes, but seeing your breakthrough, I'll get this in ASAP; so more of a suggestion for future smaller PRs.

Thanks for the suggestion - will in the future submit smaller PRs. There are still a couple of cleanups I think would make sense - e.g. there's still a depreciation warning from Gradle, and some of the Gradle plugins are not at their latest versions - but I can submit those as separate PRs. 

Thanks again

vorburger:
FINERACT-937 created re Java 12 -> 11 on Travis.

PR #815, opened 2020-05-03
ignore flaky SchedulerJobsTestResults.testInterestTransferForSavings()

vorburger:
see FINERACT-857

PR #821, opened 2020-05-05
add new testNumberOfJobs() to SchedulerJobsTest (re. FINERACT-922)

vorburger:
This will catch problems such as the integration tests running "too early" (before Flyway has fully initialized the database), which I have a suspicion could be one of (many) causes of the unstable scheduler tests, notably because of the IllegalArgumentException at SchedulerJobsTest.java:90 in #817.

see FINERACT-922 and FINERACT-952 re. removed JobName


vorburger:
Failed due to `testInterestTransferForSavings()` and thus requires #815 ...

vorburger:
@awasum @ptuomola @xurror wanna review this one from 2 weeks ago - OK to merge?

ptuomola:
Looks good to me!

xurror:
Couldn't help but notice references to 2 different issues in your commits. Changes are very little though. I don't know if they are co-dependent but would have been much better to raise 2 different PRs for each of them. But as I said, very little changes so, I don't mind if you merge it like this

PR #831, opened 2020-05-08
add "Instructions to build the JAR file" to README

vorburger:
it occurred to me that this may not be so obvious to everyone, and worth pointing out more explicitly

@ptuomola wording OK for you?


ptuomola:
Looks great to me

PR #829, opened 2020-05-08
FINERACT-938: Upgrading nekohtml and removing direct reference to Xerces

ptuomola:
## Description
Version of nekohtml pulled in by Mockrunner uses an old version of Xerces, leading to problems with JDK11. This was first fixed by explicitly pulling in a new version of Xerces. However, this pull request fixes it by upgrading nekohtml to a newer version instead. Also upgrading dom4j to the newest version. 

## Checklist
Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Commit message starts with the issue number from https://issues.apache.org/jira/projects/FINERACT/. Ex: FINERACT-646 Pockets API.

- [x] Coding conventions at https://cwiki.apache.org/confluence/display/FINERACT/Coding+Conventions have been followed.

- [x] API documentation at https://github.com/apache/fineract/blob/develop/api-docs/apiLive.htm has been updated with details of any API changes.

- [x] Integration tests have been created/updated for verifying the changes made.

- [x] All Integrations tests are passing with the new commits.

- [x] Submission is not a "code dump".  (Large changes can be made "in repository" via a branch.  Ask on the list.)

Our guidelines for code reviews is at https://cwiki.apache.org/confluence/display/FINERACT/Code+Review+Guide

PR #832, opened 2020-05-08
remove (useless) SenderTest + mockrunner-jms/nekohtml (FINERACT-938)

vorburger:
follow-up to #829

see FINERACT-938

@ptuomola does this LGTY?


awasum:
@Anh3h Just adding you to review this PR as you also worked on the Notification module. You may have a better view than me.

ptuomola:
@vorburger Definitely great if we can get rid of Mockrunner with its awful dependency tree. What I can't comment on is the value of this test - was it testing anything useful - and whether we had plans to build more tests like it?

vorburger:
> value of this test - was it testing anything useful 

Just from looking at it, I don't see this `SenderTest` doing anything particularly useful. It doesn't `assert` on anything. I doubt this adds any real value in testing Fineract's JMS support; e.g. I don't think this test would reliably catch any possible regressions in Fineract's JMS support. (Which isn't exactly it's top most used feature anyway, as far as I know.)

> and whether we had plans to build more tests like it?

according to [git log history](https://github.com/apache/fineract/commits/develop/fineract-provider/src/test/java/org/apache/fineract/notification/SenderTest.java), it was written 3 years ago, by @adhyans ... Adhyan, if you are still alive and have any interest in Fineract, would be interested to chime in here and let us know if you can still sleep well at night :smile: if we delete this `SenderTest`? It's dependency on `mockrunner-jms` was giving us a hard time (see FINERACT-846 and FINERACT-938), and we don't' see much value in that particular test anymore, compared to the dependency management headache its use of `mockrunner-jms` (which is, at least now, only used in that test), and we'd like to remove it. Do you have any strong objections?

vorburger:
Anyone reading, chime in on FINERACT-944 for a more general discussion re. the Notifications feature.

vorburger:
I've just rebased this and resolved conflicts. @awasum given that we haven't heard back from @Anh3h within 1 week, and that @ptuomola expresses support for _"great if we can get rid of Mockrunner with its awful dependency tree"_ and my explanations about the value of this test, would you have any strong objections to merging this? It's a simplification of our dependencies which we will only be grateful for in the future, at the loss of a IMHO useless test.

awasum:
Please, merge it now. I just wanted someone familiar with the notification
feature to look at it. But we can't keep waiting for that forever. So merge
this ASAP. Thanks and sorry for holding this  that's precious time but it
was for a good reason.

PR #834, opened 2020-05-08
upgrade Spring Boot from 2.2.4 to 2.2.7 (FINERACT-941)

vorburger:
FINERACT-941


vorburger:
@ptuomola @awasum @percyashu makes sense?

percyashu:
@vorburger LGTM

ptuomola:
@vorburger definitely makes sense... and not sure why I had left so much whitespace at the end of those lines! Probably need to configure vim to show whitespace.

PR #842, opened 2020-05-08
fix generic warnings in StaffTest (re. FINERACT-949)

vorburger:
FINERACT-949


vorburger:
@ptuomola @awasum @xurror OK for you?

But more importantly, please chime in to the more general question raised in FINERACT-949 ...

awasum:
LGTM.

PR #844, opened 2020-05-08
rm application.properties from Git, and .gitignore it (because we copy it during build)

vorburger:


vorburger:
@awasum @ptuomola @xurror how about this? 

I noticed a diff in my local application.properties (CR/LF, after #798) and to me this seems the right solution.

PR #845, opened 2020-05-08
clean up CommandHandlerProviderTest

vorburger:


vorburger:
exceptionally self merged totally trivial code clean up change which passed the build

PR #847, opened 2020-05-08
add integration test for (new) ISO-8601 Date Format (FINERACT-926)

vorburger:
This is the missing integration test for #816

I've put it into SchedulerJobsTest, just because that's convenient.

FINERACT-926


vorburger:
@awasum merge if this LGTY? And then close FINERACT-926 - with this, it's fully done.

PR #849, opened 2020-05-08
bulk replace org.apache.commons.lang with lang3 imports (FINERACT-954)

vorburger:
We apparently currently have both on the classpath, but really nowadays
should have, and can, only use lang3 with this minor clean up.

FINERACT-954


vorburger:
@xurror makes sense, right? LGTY and OK to merge?

xurror (approved):
It's good. I think you can merge this

PR #850, opened 2020-05-08
clean up dependencies.gradle and clarify scopes

vorburger:
All of those are ALSO api and/or implementation, so there's no need to
duplicate them in testCompile.

Also adds a new comment inline hopefully clarifying this for future
users.

### DIFF
--- a/fineract-provider/build.gradle
+++ b/fineract-provider/build.gradle
@@ -495,7 +495,7 @@ sourceSets {
 }
 
 configurations {
-    integrationTestCompile.extendsFrom testCompile
+    integrationTestCompile.extendsFrom testImplementation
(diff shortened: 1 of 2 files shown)


vorburger:
@xurror @awasum @ptuomola agreed?

vorburger:
Oh wait, integrationTest needs those.. but shouldn't there be a separate scope for that then?

vorburger:
@lingocoder given [your past interest in this subject](https://github.com/vorburger/fineract/pull/1), does this make sense to you?

xurror (approved):
LGTM

ptuomola:
Looks great to me

PR #851, opened 2020-05-09
improve SchedulerJobsTest.testDateFormat() by introducing Awaitility

vorburger:
This recently introduced new test (FINERACT-926) would locally pass, but
then fail on the next run, because another test in the same class would
modify job state to be inactive.  It now sets the first jobs state as it
requires it to be, and properly awaits the expected outcome.

Related to FINERACT-922.


vorburger:
@maektwain does this LGTY? Anyone else interested in reviewing this?

maektwain:
Yeah I just reviewed this.  Looks good to me.  I reviewed it seems solid to me.

vorburger:
> Yeah I just reviewed this. Looks good to me. I reviewed it seems solid to me.

Thanks! I've self merged based on this. Feel free to help with more code reviews... :smile: appreciate it.

PR #853, opened 2020-05-09
remove fixed Slfj4j versions to fix broken logging (FINERACT-956)

vorburger:
FINERACT-956


vorburger:
@ptuomola @xurror @awasum OK for you?

ptuomola:
Hi

The reason why I added this was because some of the Gradle plugins (spotbugs etc?) were compiled against 1.8, whereas the default version being pulled in by gradle was an older one (1.7). This resulted in errors in the gradle log. 

However you've now removed the explicit version, but I don't see the errors reappearing. So I'm assuming some other dependency change has fixed this. In which case of course no issue in changing back to default versions...

PR #854, opened 2020-05-09
fix bug causing NullPointerException in SavingsAccount (FINERACT-957)

vorburger:
FINERACT-957


nemux:
I think is a good and simple fix.

vorburger:
> I think is a good and simple fix.

OK, thanks. I've "self-merged" this based on this review comment (and thumbs up from @trackcc).

If anyone else reading this understands this better and thinks this is the wrong fix (is this just addressing a "symptom" rather the real "root cause"?), then please (comment on to) re-open FINERACT-957 and let's discuss further there.

PR #846, opened 2020-05-08
use Java >11.0.2 instead of Java 12 on Travis (FINERACT-937)

vorburger:
FINERACT-937

PR #857, opened 2020-05-09
SQL scripts made MySQL v8 compatible, but still 5.x as well (FINERACT-897)

xurror:
Upgrade to mysql v8


.travis.yml
```diff
@@ -34,6 +34,15 @@ addons:
     # https://httpie.org is used by the docker-compose test below
     - httpie
 
+before_script:
```
vorburger: FYI (minor) this will conflict with #846, maybe we can get that one merged first?

Dockerfile
```diff
@@ -49,13 +49,5 @@ COPY ./docker/server.xml /opt/bitnami/tomcat/conf
 RUN chmod 664 /opt/bitnami/tomcat/conf/server.xml
 
 WORKDIR /opt/bitnami/tomcat/lib
-# org.drizzle.jdbc.DrizzleDriver is used by default for both the all tenants and demo tenant DB DataSource
-RUN wget https://repo1.maven.org/maven2/org/drizzle/jdbc/drizzle-jdbc/1.4/drizzle-jdbc-1.4.jar
```
vorburger: this is required, both because it's the current default, and because it is part of Fineract - we can't have it as a dependency in Gradle, but then not have it in the Dockerfile... because the container image that is built by this Dockerfile is distributed by the ASF, I'm assuming we're still bound by LEGAL-462.

vorburger:
I've very briefly glanced over this, and put some new thoughts into FINERACT-897...

Dockerfile
xurror: I've reverted to drizzle.
Mysql 8 won't work with drizzle though, but I guess that's fine for now.
Migrating to mysql v8 in the future should be alot easier with few db tweaks I've done.

xurror:
@awasum, @vorburger, @ptuomola is it for you?

fineract-provider/build.gradle
```diff
@@ -94,7 +94,7 @@ dependencyManagement {
         dependency 'com.lowagie:itext:2.1.7'
         dependency 'com.lowagie:itext-rtf:2.1.7'
         dependency 'org.mnode.ical4j:ical4j:1.0.7'
-        dependency 'com.googlecode.flyway:flyway-core:2.1.1'
+        dependency 'org.flywaydb:flyway-core:6.4.1'
```
vorburger: @xurror wait, waaaat - this work now?! Could I invite you to first raise a small separate PR just re. this, referencing (and closing?) FINERACT-810 ?

vorburger:
> @awasum, @vorburger, @ptuomola is it for you?

I would prefer to see the Flyway upgrade raised a separate PR with ref. to FINERACT-810. Make FINERACT-810 a blocker for FINERACT-897 if that is the case (I'm assuming that you have found that it is).

Let us then clarify by updating both the title of FINERACT-897 and the commit comment in this PR to more precisely express what exactly this is, and isn't... I think what we have concluded in FINERACT-897 is that this now actually is NOT (anymore) a "Upgrade to mysql v8", but... "make SQL and code work on both MySQL 5.7 as well as more recent versions such as v8" - kind of wording?

As for reviewing the (massive, wow, how did you do that, hats off!) SQL diff in this PR, I'm not the most qualified - I'm hoping other parties interested in MySQL would chime in to review this better than I could.

xurror:
> I would prefer to see the Flyway upgrade raised a separate PR with ref. to FINERACT-810. Make FINERACT-810 a blocker for FINERACT-897 if that is the case (I'm assuming that you have found that it is).

Not a problem. I can chop that change off this PR and raise a different PR for it.

> As for reviewing the (massive, wow, how did you do that, hats off!) SQL diff in this PR, I'm not the most qualified - I'm hoping other parties interested in MySQL would chime in to review this better than I could.

It was not much of a struggle. All the scripts are basically thesame as before. I just updated some deprecated features such int(10).. etc. Most respect to those who wrote the scripts. They are quite robust.

vorburger:
Separate PR was raised in #882, and meanwhile merged. 

This now only changes lots of SQL files, and make minor change in 2 Java classes.

I was hoping other parties interested in MySQL would chime in to review this better than I could... no?

I think I'll just merge this day after tomorrow (Sunday), if there is no other feedback.

vorburger:
@xurror perhaps if you could just address this feedback which I posted earlier:

> Let us then clarify by updating both the title of FINERACT-897 and the commit comment in this PR to more precisely express what exactly this is, and isn't... I think what we have concluded in FINERACT-897 is that this now actually is NOT (anymore) a "Upgrade to mysql v8", but... "make SQL and code work on both MySQL 5.7 as well as more recent versions such as v8" - kind of wording?

xurror:
Sure. I'll update the refs.

ptuomola:
As this change updates a lot of the Flyway SQL migration scripts, I was just wondering: what happens with existing databases where these scripts have already been ran?  I'm assuming Flyway will not try to reapply the scripts that it thinks it has already ran. Will their column types be converted to supported somehow magically by MySQL as part of their MySQL upgrade? 

Or will we end up with two types of schemas: ones which have been created with the Flyway scripts after these changes have been committed and therefore work with MySQL 8 - and other ones which have been created before, and continue to use the deprecated features?

xurror:
Just wish to clarify that the changes here are literally very minor. The schemas and architectures are basically thesame as they were before: same names, same rules etc. The only changes that are done here are purely syntactic(Ex: changing deprecated declaration int(2) to the int which works fine with both v8 and v5.7 also changing charsets). Same logic, different syntax and again even the syntatic changes are minor. Unless I'm mistaken, but this will still look thesame in the DB once migrated so I can hardly see an issue here.
I don't know if you follow??

vorburger:
@ptuomola I'm totally with you on the principle that in theory we should never touch existing Flyway SQL migration scripts, only write new ones (that's kind of the point of Flyway). I suspect that you are correct that this will not give us automagic migration, but at least the script is cleaner for new installs? In this particular instance, it's probably "not worth it" to do the same via migration scripts - what do you think? Personally I don't really have a problem with this getting merged.

ptuomola:
@vorburger Fully agreed that if none of the changes have any impact on functionality, then it doesn't matter and doing this via a migration script is "not worth it". 

Unfortunately I don't have enough knowledge about MySQL's innards to assess whether that's the case for these changes, so happy to defer to the judgement of others...

PR #879, opened 2020-05-10
clean up .github/workflows/rebase.yml (FINERACT-829)

vorburger:
re. FINERACT-829 (still doesn't seem to work, but if I do get it to work later, this is needed anyway)


vorburger:
@xurror I've also self merged this, to try if and see if `/rebase` works e.g. on #881 now..

PR #887, opened 2020-05-14
FINERACT-830: Configuring Dockerfile to use Google Distroless base image

ptuomola:
Using Google Distroless with OpenJDK 11 as base image for Fineract Docker build

FINERACT-830

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,31 +17,30 @@
 #
 FROM openjdk:11 AS builder
 
-COPY . fineract
-WORKDIR fineract
-RUN ./gradlew --no-daemon -x rat -x test war
-
-# =========================================
-
-FROM bitnami/tomcat:9.0 as fineract
-
-USER root
 RUN apt-get update -qq && apt-get install -y wget
 
-COPY --from=builder /fineract/build/libs/fineract-provider.war /opt/bitnami/tomcat/webapps
+COPY . fineract
+WORKDIR /fineract
 
-RUN keytool -genkey -keyalg RSA -alias tomcat -keystore /opt/bitnami/tomcat/tomcat.keystore -keypass xyz123 -storepass xyz123 -noprompt -dname "CN=Fineract, OU=Fineract, O=Fineract, L=Unknown, ST=Unknown, C=Unknown"
-COPY ./docker/server.xml /opt/bitnami/tomcat/conf
-RUN chmod 664 /opt/bitnami/tomcat/conf/server.xml
+RUN ./gradlew --no-daemon -x rat -x test bootJar
 
-WORKDIR /opt/bitnami/tomcat/lib
-# org.drizzle.jdbc.DrizzleDriver is used by default for both the all tenants and demo tenant DB DataSource
-RUN wget https://repo1.maven.org/maven2/org/drizzle/jdbc/drizzle-jdbc/1.4/drizzle-jdbc-1.4.jar
+WORKDIR /fineract/target
+RUN jar -xf /fineract/build/libs/fineract-provider.jar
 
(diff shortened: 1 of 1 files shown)


xurror:
Seems this works fine. @vorburger, @awasum is this OK for you?

vorburger:
@ptuomola thanks for this!! I'll try to review this (and test it on https://www.fineract.dev...) ASAP.

Dockerfile
```diff
@@ -32,30 +32,11 @@ COPY *LICENSE* fineract/
...
-# org.drizzle.jdbc.DrizzleDriver is used by default for both the all tenants and demo tenant DB DataSource
-RUN wget https://repo1.maven.org/maven2/org/drizzle/jdbc/drizzle-jdbc/1.4/drizzle-jdbc-1.4.jar
-
-# https://issues.apache.org/jira/browse/LEGAL-462
-# https://issues.apache.org/jira/browse/FINERACT-762
-# We include an alternative JDBC driver (which is faster, but not allowed to be default in Apache distribution)
-# allowing implementations to switch the driver used by changing start-up parameters (for both tenants and each tenant DB)
-# The commented out lines in the docker-compose.yml illustrate how to do this.
-# To be sure that this instead of Drizlle is used, comment out wget above.
-RUN wget https://repo1.maven.org/maven2/mysql/mysql-connector-java/8.0.19/mysql-connector-java-8.0.19.jar
```
vorburger, line 47: This loss of the `mysql-connector-java-8.0.19.jar` is a problem for those who want to use that instead of Drizzle (e.g. I actually do, on https://www.fineract.dev...). It means e.g. that if you (quote) _"comment out lines in the docker-compose.yml illustrate how to do this"_ it won't find it anymore after this PR.

@ptuomola do you think we could keep somehow that? Unfortunately, you can't just add it to `dependencies.gradle` - because we cannot distribute an LGPL library with Fineract release ZIPs that can be downloaded from https://fineract.apache.org. I know and totally agree that this is a bit of a mess; ask for references if you want more details. I guess you could still keep the `wget` here, and add it to the classpath when launching the JAR? BTW I'm sure you've also come across posts suggesting that a bit of startup time can be gained by "exploding" (uncompressing) "FAT JARs", and directly specifying the main class. I don't know if that's something you would be willing to already do here? Or we could of course look into that later. It just occurred to me that it may go naturally together with adding an additional JAR.

vorburger: @ptuomola due to FINERACT-980 I have just created FINERACT-982 for future discussion on this front, if you want to chime in there. But in the very short term, to get this PR in, IMHO we should still keep this as is? Unless you have any better ideas...

ptuomola: To get MySQL driver option to work again, I've added in downloading the JAR for the driver back in. At the same time, I also exploded the JAR and changed Dockerfile to run the main class directly as suggested. Changes seem to work OK at least locally - tested with both Drizzle and MySQL connectors. Let me know if this looks ok...

vorburger: @ptuomola thanks! I'd love to manually test deploy this on https://www.fineract.dev before I merge it - I'll try to do that in the coming days, and assuming that it "just works" (which I expect that it will, just to be extra careful...), I'll merge it; hope that's OK.

vorburger: @xurror in the mean time, do you want to review this already, based on your comments I just saw in https://issues.apache.org/jira/browse/FINERACT-830 ? Please don't merge it yet, just review and LGTM - I'll merge it after I've tested it as described above - if it's LGTY.

vorburger:
@ptuomola I haven't forgotten about this (on the contrary), and am testing this right now...

PR #888, opened 2020-05-14
FINERACT-972: Ensure Tomcat is started after compiling integration tests

ptuomola:
## Description
Tomcat was being started before compiling integration tests, meaning that if the integration test compilation failed then it was left running. After adding the dependencies, it now is started only after successful compilation of integration tests. 

FINERACT-972


xurror:
I don't know much about this. I'll just pull @vorburger to have a look

vorburger:
This is great! I had been wondering why, as your FINERACT-972 says, Tomcat sometimes didn't shut down.... now I know! Thanks for having spotted and fixed this.

PR #763, opened 2020-04-13
Added Empty Statement Check and enforced checkstyle

thesmallstar:
## Description
No functional changes, Code formatting , indent fix and empty statement checktyle enforce :) 

## Checklist
Please make sure these boxes are checked before submitting your pull request - thanks!

- [ ] Commit message starts with the issue number from https://issues.apache.org/jira/projects/FINERACT/. Ex: FINERACT-646 Pockets API.

- [ ] Coding conventions at https://cwiki.apache.org/confluence/display/FINERACT/Coding+Conventions have been followed.

- [ ] API documentation at https://github.com/apache/fineract/blob/develop/api-docs/apiLive.htm has been updated with details of any API changes.

- [ ] Integration tests have been created/updated for verifying the changes made.

- [ ] All Integrations tests are passing with the new commits.

- [ ] Submission is not a "code dump".  (Large changes can be made "in repository" via a branch.  Ask on the list.)

Our guidelines for code reviews is at https://cwiki.apache.org/confluence/display/FINERACT/Code+Review+Guide

PR #889, opened 2020-05-14
Ignore Flaky testNotificationStorage test FINERACT-828

thesmallstar:
This caused failed test in #763 
Background: https://issues.apache.org/jira/browse/FINERACT-828
For reference: 
image


vorburger:
@thesmallstar how about #892 instead?

thesmallstar:
@vorburger yes, definitely fixing it is a better idea.
Closing this PR.

PR #891, opened 2020-05-15
FINERACT-942 Remove bad logging pattern(Use of system.out) and added checkstyle 

thesmallstar:
Background: https://issues.apache.org/jira/browse/FINERACT-942

With a similar approach, I will be dealing with the rest of the files(about 84 of them).
Made an early PR, if in case someone has any suggestions?  Meanwhile, I will be updating other files.

### DIFF
--- a/fineract-provider/src/test/java/org/apache/fineract/notification/Listener.java
+++ b/fineract-provider/src/test/java/org/apache/fineract/notification/Listener.java
@@ -22,13 +22,15 @@ import javax.jms.JMSException;
 import javax.jms.Message;
 import javax.jms.Session;
 import javax.jms.TextMessage;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.jms.listener.SessionAwareMessageListener;
 
 public class Listener implements SessionAwareMessageListener {
-
+    private final static Logger LOG = LoggerFactory.getLogger(Listener.class);
     @Override
     public void onMessage(Message message, Session session) throws JMSException {
         TextMessage msg = (TextMessage) message;
-        System.out.println("Received: " + msg.getText());
+        LOG.info("Received: {}" ,msg.getText());
     }
 }
--- a/fineract-provider/config/checkstyle/checkstyle.xml
+++ b/fineract-provider/config/checkstyle/checkstyle.xml
@@ -47,7 +47,13 @@
         <property name="fileExtensions" value="java"/>
     </module>
   -->
+   
     <module name="TreeWalker">
+    <module name="RegexpSinglelineJava">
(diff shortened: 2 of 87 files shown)


fineract-provider/src/test/java/org/apache/fineract/notification/Listener.java
```diff
@@ -29,6 +29,6 @@
     @Override
     public void onMessage(Message message, Session session) throws JMSException {
         TextMessage msg = (TextMessage) message;
-        System.out.println("Received: " + msg.getText());
+        LOG.debug("Received: " + msg.getText());
```
vorburger: it's probably OK, and likely better/clearer, to simply always use `LOG.info()` instead of `debug()` to replace `System.out.println()` in tests. In "real" (non-test) code, it's... not always super obvious which level is right. But, hopefully, you won't find any `System.out.println()` in non-test code?

vorburger:
The direction looks all right to me. See inline comment above re. log level.

fineract-provider/src/test/java/org/apache/fineract/notification/Listener.java
thesmallstar: Yes, no System.out  in a non-test folder.
Replacing all of them with LOG.info :)


-thanks 
Manthan

thesmallstar:
The build passed locally, I see my IDE messed some files up.
Fixing them in the next commit.

thesmallstar:
@vorburger If the build passes, we can see the "messed files".

thesmallstar:
Few points:
1. Some of the (30-40 out of 900) of the system.out were commented and not converted to log.info, since the build failed(with a weird reason), one of it was deleted 
2.  There are few files 4-5, to which my ide decided to beautify without asking me :p, if it is not big issue we can merge this, or maybe I can copy the old version of that file and make those changes again?

vorburger:
@thesmallstar this mostly LGTM now. Feedback:

0. can you amend the commit message, and PR title/description, with the related JIRA, please?

1. not sure what that's about, probably doesn't matter; let's ignore and move forward?

2. can you chagne your IDE to avoid that in the future? But let's not redo this, but keep going. (The beautificating your IDE did behind your back appears to be related to max. line length? That is actually one of the Checkstyle modules... note `LineLength` - but let's get to that later, only; it has less priority (value) than e.g. the logging and exception handling related checks. But when we do get to that, we can discuss just exactly how long we want lines to be maximum. I think your IDE has it set to be too short; there's no need to break everything at 80 - the (right now still commented out) config currently has it at 120, but personally I'd probably even go for something bigger, like 160 or perhaps 140. We all have big screens nowadays! :smile:

thesmallstar:
Updated the commit message, will keep all of your suggestion(commit-msg,ide) in mind :D

PR #892, opened 2020-05-15
fix flaky StorageTest.testNotificationStorage() (FINERACT-828)

vorburger:
FINERACT-828


vorburger:
@thesmallstar do you want to review and LGTM if it looks right to you, so that I can exceptionally "self merge" this? Unless @xurror or @awasum you would like to do the honours (after it passes the build, of course).

thesmallstar:
LGTM @vorburger 

-thanks
Manthan

PR #894, opened 2020-05-16
bump json-path from 0.9.1 to 2.4.0 (FINERACT-964)

xurror:
bump json-path from 0.9.1 to 2.4.0

FINERACT-964

### DIFF
--- a/fineract-provider/build.gradle
+++ b/fineract-provider/build.gradle
@@ -99,7 +99,7 @@ dependencyManagement {
         dependency 'com.amazonaws:aws-java-sdk-s3:1.11.779'
         dependency 'net.sf.ehcache:ehcache:+'
         dependency 'com.github.spullara.mustache.java:compiler:0.8.12'
-        dependency 'com.jayway.jsonpath:json-path:0.9.1'
+        dependency 'com.jayway.jsonpath:json-path:2.4.0'
         dependency 'org.apache.tika:tika-core:1.24.1'
         dependency 'org.apache.httpcomponents:httpclient:+'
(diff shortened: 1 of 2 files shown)


xurror:
@vorburger, @ptuomola @awasum, is this OK for you?

vorburger:
> @vorburger, @ptuomola @awasum, is this OK for you?

Looks like it builds! But I don't know if that Batch ResolutionHelper class has any test coverage... perhaps you could have a quick look using https://github.com/apache/fineract#code-coverage-reports if it those? If the lines you're changing here is covered by tests, then I'm OK to merge this. If not, perhaps someone should review that `JsonModel` => `JsonPath` + `ReadContext` business a little more closely... anyone reading this interested in having a closer look and doing a real code review?

PS: It would be neat if somehow automatically code coverage would show in PRs... anyone got an idea?

vorburger:
@xurror have you had a chance to check if that Batch ResolutionHelper class has any test coverage?

xurror:
Yes. One test.

vorburger:
> Yes. One test.

which one? I just ran https://github.com/apache/fineract#code-coverage-reports, and found that `ResolutionHelper` basically has 0% unit test coverage (except for its constructor). We therefore can't really know if these change break that class, or everything still works as before. There may be 2 separate issues - FINERACT-985 ITs are not covered by Jacoco + `ResolutionHelper` may or may not be covered by any IT (I don't know).

I'll just merge his anyway, but it's theoretically possible that it break some batch related feature.

PR #895, opened 2020-05-16
FINERACT-976 : Client date of birth can not be greater than the client submitted on date

kiran-vsm:
## Description
Describe the changes made and why they were made. Ignore if these details are present on the associated Jira ticket

## Checklist
Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Commit message starts with the issue number from https://issues.apache.org/jira/projects/FINERACT/. Ex: FINERACT-646 Pockets API.

- [x] Coding conventions at https://cwiki.apache.org/confluence/display/FINERACT/Coding+Conventions have been followed.

- [ ] API documentation at fineract-provider/src/main/resources/static/api-docs/apiLive.htm has been updated with details of any API changes.

- [ ] Integration tests have been created/updated for verifying the changes made.

- [ ] All Integrations tests are passing with the new commits.

- [ ] Submission is not a "code dump".  (Large changes can be made "in repository" via a branch.  Ask on the list.)

Our guidelines for code reviews is at https://cwiki.apache.org/confluence/display/FINERACT/Code+Review+Guide

PR #896, opened 2020-05-16
rm .github/workflows/first-interaction.yml

vorburger:
because it's broken, see
https://github.com/actions/first-interaction/issues/10

as seen on #895


vorburger:
@xurror do you want to review and merge this? 

@awasum FYI, this was a nice idea, but doesn't seem to work... :sob: 

I just set up a Save Reply on GitHub to do the same manually some times! :smiley:

PR #898, opened 2020-05-16
use --no-daemon in Dockerfile (and don't clean; really no need) [FINERACT-970]

vorburger:
FINERACT-970


vorburger:
@ptuomola makes sense?

ptuomola:
Makes perfect sense to me!

PR #899, opened 2020-05-16
automatically deploy to https://Fineract.dev [FINERACT-970]

vorburger:
see FINERACT-970

### DIFF
--- a/.github/workflows/fineract.dev.yaml
+++ b/.github/workflows/fineract.dev.yaml
@@ -1,0 +1,31 @@
+# This is the GitHub Action which
+# automatically updates https://www.fineract.dev
+#
+# see https://jira.apache.org/jira/browse/FINERACT-970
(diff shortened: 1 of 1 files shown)


vorburger:
I'm going to self-merge this in order to be able to test FINERACT-970. We could always revert it if NOK.

awasum:
Thanks. Looks good

One thing which may or may not be important is, why don't you use a stable
version of Ubuntu say 18.04 or 20.04 LTS than the Ubuntu-latest image?

Ignore if the above comment is stupid. I don't know enough about GitHub
actions, just curious.

vorburger:
> One thing which may or may not be important is, why don't you use a stable version of Ubuntu say 18.04 or 20.04 LTS than the Ubuntu-latest image?

To be honest I just copy/pasted that line... I don't think it matters much - this is just for a very short lived container, to run an action. BTW GitHub Actions seems pretty cool!

PR #901, opened 2020-05-16
gcloud components install beta to deploy to https://Fineract.dev [FINERACT-970]

vorburger:
FINERACT-970


vorburger:
self merged like #900 and #899 previously, in order to be able to test this (easier like this than on branch)

PR #904, opened 2020-05-16
copy .git/** into build container so git.properties can be generated [FINERACT-983]

vorburger:
read details in FINERACT-983

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -18,12 +18,9 @@
 build/*
 .gradle/*
 
+*.jar
+*.war
+
 # Incremental "docker build" is faster if we can modify the Dockerfile
 # without "COPY ." re-transferring everything into the builder container.
 Dockerfile
-
-# https://github.com/containers/buildah/issues/1582
-.git/*
-.git/*/*
-.git/*/*/*
-.git/*/*/*/*
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ catalina.base_IS_UNDEFINED/
 keystore.jks
 bin/
 fineract-provider/src/main/resources/application.properties
+.lock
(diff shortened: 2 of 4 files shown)


vorburger:
@xurror @ptuomola @awasum check this out.. do you have any better ideas? Merge?

.dockerignore
```diff
@@ -18,12 +18,9 @@
 build/*
 .gradle/*
 
+*.jar
```
xurror, line 21: Why are we ignoring this?

xurror (approved):
This looks good to me

.dockerignore
vorburger: any *.jar that have been built locally should not get copied into the build container

xurror: Ok, makes sense.

vorburger:
@awasum Is this OK for you too?

vorburger:
@ptuomola no objections to this? See linked JIRA. If @awasum or @xurror merge this ASAP, then I'll send an email to the dev list to announce the auto-updating of fineract.dev, which finally fully works end2end, since yesterday! I'd love this to be merged before I announce, because this change helps to "see" what's actually running on Fineract.dev.

ptuomola:
Looks good - except the previous comment that copying .git slows things down, and that is why it was excluded. Is that no longer the case - ie do we not end up with the same issue now that we are again including .git? It's basically copying the entire repository history, right?

awasum:
Can we get the CI system to copy just the latest revision of the history? Will that help?
Like:
`git clone -–depth [depth] [remote-url]`

awasum:
Maybe we do this in a follow up PR? Or even rethink the issue of adding latest git revision in /actuator/info... IS there a better way?

awasum:
I will merge this and someone can cycle back to this in the near future.

vorburger:
https://demo.fineract.dev/fineract-provider/actuator/info - yay!! (This was fully automated.)

@ptuomola @awasum see my comments in FINERACT-983 and FINERACT-911 re. your points.

PR #858, opened 2020-05-09
add toString() to Notification + fix setActor()

vorburger:
to help debug FINERACT-828 flaky test

PR #905, opened 2020-05-17
fix moar flaky StorageTest.testNotificationStorage() [FINERACT-828]

vorburger:
similar fix to earlier Pull Request for the same

also added some more toString()

see FINERACT-828


vorburger:
@thesmallstar @xurror review this?

fineract-provider/src/main/java/org/apache/fineract/useradministration/domain/AppUser.java
```diff
@@ -663,4 +663,8 @@ public boolean isSelfServiceUser() {
         return newAppUserClientMappings;
     }
 
+    @Override
+    public String toString() {
```
xurror, line 667: Just curious, I'm not seeing this method being used anywhere, I wish to know why you wrote it

thesmallstar: Hello @xurror, if I am not wrong, it is explicitly used for debugging(if the test fails). 
Refer to this how it helped: #858  and https://issues.apache.org/jira/browse/FINERACT-828

thesmallstar (approved):
LGTM

PR #833, opened 2020-05-08
clean up XBRLBuilderTest (re. FINERACT-940)

vorburger:
This doe NOT actually fix FINERACT-940, it's a clean up, while I looked at that problem.

PR #906, opened 2020-05-18
FINERACT-942 Remove use of printStackTrace and added checkstyle

thesmallstar:
Background: https://issues.apache.org/jira/browse/FINERACT-942

Requesting an early review, since there are a lot of files to change(Around 40).
Just FYI, I used "throws" keyword on some functions so as to deal with chained exceptions.

REF: #833

### DIFF
--- a/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/AccountingScenarioIntegrationTest.java
+++ b/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/AccountingScenarioIntegrationTest.java
@@ -556,7 +556,7 @@ public class AccountingScenarioIntegrationTest {
 
     @Test
     @Ignore // TODO FINERACT-899
-    public void checkPeriodicAccrualAccountingFlow() {
+    public void checkPeriodicAccrualAccountingFlow() throws InterruptedException,ParseException {
         final Account assetAccount = this.accountHelper.createAssetAccount();
         final Account incomeAccount = this.accountHelper.createIncomeAccount();
         final Account expenseAccount = this.accountHelper.createExpenseAccount();
@@ -587,11 +587,8 @@ public class AccountingScenarioIntegrationTest {
         this.journalEntryHelper.checkJournalEntryForAssetAccount(assetAccount, this.EXPECTED_DISBURSAL_DATE, assetAccountInitialEntry);
 
         final String jobName = "Add Accrual Transactions";
-        try {
-            this.schedulerJobHelper.executeJob(jobName);
-        } catch (InterruptedException e) {
-            e.printStackTrace();
-        }
+
+        this.schedulerJobHelper.executeJob(jobName);
 
         // MAKE 1
         LOG.info("Repayment 1 ......");
@@ -673,7 +670,7 @@ public class AccountingScenarioIntegrationTest {
 
     @Test
     @Ignore // TODO https://issues.apache.org/jira/browse/FINERACT-899
-    public void checkPeriodicAccrualAccountingFlow_OVER_PAYMENT() {
+    public void checkPeriodicAccrualAccountingFlow_OVER_PAYMENT() throws InterruptedException,ParseException {
         final Account assetAccount = this.accountHelper.createAssetAccount();
         final Account incomeAccount = this.accountHelper.createIncomeAccount();
         final Account expenseAccount = this.accountHelper.createExpenseAccount();
@@ -704,11 +701,8 @@ public class AccountingScenarioIntegrationTest {
         this.journalEntryHelper.checkJournalEntryForAssetAccount(assetAccount, this.EXPECTED_DISBURSAL_DATE, assetAccountInitialEntry);
 
         final String jobName = "Add Accrual Transactions";
-        try {
-            this.schedulerJobHelper.executeJob(jobName);
-        } catch (InterruptedException e) {
-            e.printStackTrace();
-        }
+
+        this.schedulerJobHelper.executeJob(jobName);
 
         // MAKE 1
         LOG.info("Repayment 1 ......");
@@ -732,7 +726,7 @@ public class AccountingScenarioIntegrationTest {
 
     @Test
     @Ignore // TODO https://issues.apache.org/jira/browse/FINERACT-899
-    public void checkPeriodicAccrualAccountingTillCurrentDateFlow() {
+    public void checkPeriodicAccrualAccountingTillCurrentDateFlow() throws InterruptedException,ParseException {
         final Account assetAccount = this.accountHelper.createAssetAccount();
         final Account incomeAccount = this.accountHelper.createIncomeAccount();
         final Account expenseAccount = this.accountHelper.createExpenseAccount();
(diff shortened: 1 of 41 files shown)


fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/AccountingScenarioIntegrationTest.java
```diff
@@ -587,11 +587,8 @@ public void checkPeriodicAccrualAccountingFlow() {
         this.journalEntryHelper.checkJournalEntryForAssetAccount(assetAccount, this.EXPECTED_DISBURSAL_DATE, assetAccountInitialEntry);
 
         final String jobName = "Add Accrual Transactions";
-        try {
-            this.schedulerJobHelper.executeJob(jobName);
-        } catch (InterruptedException e) {
-            e.printStackTrace();
```
xurror, line 593: I suggest you using a logger but maybe @vorburger  or @awasum, @ptuomola or anyone else would have a different opinion as far as this is concerned.
It's more a convention to use here, whether checked or unchecked exceptions.

vorburger: No, in a test, exceptions typically should not be logged but just "fall through" (propagate, with 'testXYZ() throws SomeException, Another # Exception...'), so that the test fails, if the exception happens. PS: We need to write this and more like this up into Logging Guidelines, e.g. on the README. Do you want to make a start with that in this PR?

vorburger:
The general direction LGTM here... please tag me again when it builds green for a full review.

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/AccountingScenarioIntegrationTest.java
thesmallstar: Yes, I will update the readme.

thesmallstar:
@vorburger there are many instances where printStacktrace is used in a non-test file, for that case should I comment or remove the line only? Or Am I expected to follow the same approach(which turns out to be weird due to inheritance).

vorburger:
@thesmallstar no in non-test it depends.. definitely never printStackTrace() but either catch(SomeException e) { LOG.error("explanation", e) -OR- (re, "chain") throw NewEception("Details", e) - but typically never BOTH log and re-throw but only either-or. HTH?

thesmallstar:
@vorburger  This build should pass, in the file MultiException the printStacktrace method was overridden and a custom print-stream/writer was used for specific logging purpose, I was unsure how to tackle that? For now, I have deleted the lines(definitely needs a better approach), what do you think?

thesmallstar:
The build has passed on Travis I am not sure why is this still yellow? 
https://travis-ci.org/github/apache/fineract/builds/688892634

vorburger (changes requested):
LOG.error("{}",ex.getMessage())

must all be changed to:

LOG.error("some sort of explanation here please", ex)

thesmallstar:
@vorburger whoops, I missed that from the last message, Updating it.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exception/MultiException.java
```diff
@@ -87,7 +86,6 @@ public void printStackTrace(PrintStream s) {
         int i = 0;
         for (Throwable e : throwables) {
             s.print(++i + ".");
-            e.printStackTrace(s);
```
vorburger: @thesmallstar this needs to be kept. It's "special", and SupressWarning is OK here (only)

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exception/MultiException.java
```diff
@@ -97,7 +95,6 @@ public void printStackTrace(PrintWriter s) {
...
```
vorburger: as above

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exception/MultiException.java
```diff
@@ -77,7 +76,7 @@ public void printStackTrace() {
         int i = 0;
         for (Throwable e : throwables) {
             LOG.info("{}.",++i);
-            e.printStackTrace();
```
vorburger: as below

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exception/MultiException.java
```diff
@@ -65,7 +65,6 @@ public String getMessage() {
             sb.append(++i);
             sb.append(". ");
             Writer w = CharStreams.asWriter(sb);
-            e.printStackTrace(new PrintWriter(w, true));
```
vorburger: as below, must keep this as is

awasum:
Build shows that Travis passed: https://travis-ci.org/github/apache/fineract/builds/688892634

But this is still showing pending on the UI? Open and Close? or Rebase?

awasum:
Also seems there are some issues to be resolved as noted by @vorburger .

thesmallstar:
@vorburger 
Just a to make sure, 
 LOG.error("Runtime exception occurred in importEntity function",ex);
Is this the expected explanation? If not do I need to dive deeper on what might cause the exception for each function try-catch and write an explanation?

vorburger:
Instead of "Runtime exception occurred in...", I would probably just use"problem occurred in..." because the , e) you're adding at the end automatically logs the details, incl. whether or not it was a RuntimeException or what not, so you don't need to repeat that the message. Does that help?

thesmallstar:
Thank for the quick response :D, on it :)

thesmallstar:
@xurror can you review why the build is failing? Mostly not related to my PR(The build passed locally)
Edit: Hope that the build passes after a rebase.

thesmallstar:
@vorburger  this is ready for review :)

fineract-provider/src/main/java/org/apache/fineract/portfolio/calendar/service/CalendarUtils.java
```diff
@@ -114,7 +116,7 @@ private static Date convertToiCal4JCompatibleDate(final LocalDate inputDate) {
         try {
             formattedDate = new Date(seedDateStr, "yyyy-MM-dd");
         } catch (final ParseException e) {
-            e.printStackTrace();
+            LOG.error("{}",e.getMessage());
```
vorburger: Two thoughts on this this: a) your LOG.error() is wrong, as everywhere else, please:

```suggestion
            LOG.error("Invalid date: " + seedDateStr, e);
```

but, b) more importantly, isn't this catch plain wrong, not required, and can just be removed?! `new Date()` does not throw ParseException - or am I missing something obvious?

vorburger (changes requested):
This is looking great now! Just 1 thing left - hoping to merge this... today? :smile:

fineract-provider/src/main/java/org/apache/fineract/portfolio/calendar/service/CalendarUtils.java
thesmallstar: It does throw a parse exception, a possible explanation is the string received might be complete gibberish and could not be parsed as a date?
image

thesmallstar: Replacing it with:
LOG.error("Invalid date: {} ", seedDateStr, e);
or spot bugs won't like it :)

vorburger (approved):
LGTM now, and ready for merge as soon as build passes! 

Thanks for this one @thesmallstar - another great contribution from you.

fineract-provider/src/main/java/org/apache/fineract/portfolio/calendar/service/CalendarUtils.java
vorburger: LGTM!

PR #912, opened 2020-05-19
 add new Logging Guidelines section to README (re. FINERACT-942)

vorburger:
@thesmallstar would this LGTY as a start? You can subsequently extend it, if there's more to say...

@ptuomola @xurror @awasum FYI and feel free to amend in follow-up PRs, if you have additional points.

related to FINERACT-942 and motivated by #906

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,6 +270,25 @@ Governance and Policies
 documents the process through which you can become a committer in this project.
 
 
+Error Handling Guidelines
+------------------
+* When catching exceptions, either rethrow them, or log them.  Either way, include the root cause by using `catch (SomeException e)` and then either `throw AnotherException("..details..", e)` or `LOG.error("...context...", e)`.
+* Completely empty catch blocks are VERY suspicous!  Are you sure that you want to just "swallow" an exception?  Really, 100% totally absolutely sure?? ;-) Such "normal exceptions which just happen sometimes but are actually not really errors" are almost always a bad idea, can be a performance issue, and typically are an indication of another problem - e.g. the use of a wrong API which throws an Exception for an expected condition, when really you would want to use another API that instead returns something empty or optional.
+* In tests, you'll typically never catch exceptions, but just propagate them, with `@Test void testXYZ() throws SomeException, AnotherException`..., so that the test fails if the exception happens.  Unless you actually really want to test for the occurence of a problem - in that case, use [JUnit's Assert.assertThrows()](https://github.com/junit-team/junit4/wiki/Exception-testing) (but not `@Test(expected = SomeException.class)`).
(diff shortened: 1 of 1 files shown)


ptuomola:
In line with your comments in the JIRA ticket, do we also want to say something about when to use different log levels - i.e. what is the meaning of ERROR/WARN vs INFO/DEBUG/TRACE?

awasum:
LGTM. @ptuomola comment can be addressed in a follow up PR. @ptuomola Want to handle that?

README.md
```diff
@@ -268,6 +268,13 @@ Governance and Policies
 documents the process through which you can become a committer in this project.
 
 
+Logging Guidelines
+------------------
+* Never, ever, use `System.out` and `System.err` or `printStackTrace()` anywhere, but `LOG.info()` or `LOG.error()` instead.
+* When catching exceptions, either rethrow them, or log them.  Either way, include the root cause by using `catch (SomeException e)` and then either `throw AnotherException("..details..", e)` or `LOG.error("...context...", e)`.
+* In tests, you'll typically never catch exceptions, but just propagate them, with `@Test void testXYZ() throws SomeException, AnotherException`...`, so that the test fails if the exception happens.  Unless you actually really want to test for the occurence of a problem - in that case, use [JUnit's Assert.assertThrows()](https://github.com/junit-team/junit4/wiki/Exception-testing) (but not `@Test(expected = SomeException.class)`).
```
xurror:
```suggestion
* In tests, you'll typically never catch exceptions, but just propagate them, with `@Test void testXYZ() throws SomeException, AnotherException`..., so that the test fails if the exception happens.  Unless you actually really want to test for the occurence of a problem - in that case, use [JUnit's Assert.assertThrows()](https://github.com/junit-team/junit4/wiki/Exception-testing) (but not `@Test(expected = SomeException.class)`).
```

I noticed an extra "`" in the guide. It didn't look good in the preview

xurror (changes requested):

thesmallstar:
LGTM, we can extend as we add more checkstyles related to logging :D.

README.md
vorburger: done - thanks for noticing

vorburger:
> In line with your comments in the JIRA ticket, do we also want to say something about when to use different log levels - i.e. what is the meaning of ERROR/WARN vs INFO/DEBUG/TRACE?

@ptuomola thanks for pointing that out - I've now amended this. How do you all like it like it now?

xurror (approved):
LGTM

vorburger:
@ptuomola OK for you as is?

ptuomola:
@vorburger Looks great to me!

PR #913, opened 2020-05-19
remove direct explicit Guava dependency (FINERACT-963)

vorburger:
We *ARE* using Guava, but, unless we have a real need for a more recent
specific version, it seems "safer" (and less hassle to maintain bumping)
just using whatever version we transitively inherit from other libraries
we use that themselves depend on Guava (currently Swagger, Mustache and
activemq-broker).

Currently that "whatever version" is a 28.2 (we used 28.1 before).

This removes two fairly different kinds of Guava dependencies we have:
One for run-time (above), and one for build time which seems to have
been added due to some SpotBugs related issue (but it now works without
that, so that must have been some issue in SpotBugs that meanwhile go
fixed).

### DIFF
--- a/fineract-provider/build.gradle
+++ b/fineract-provider/build.gradle
@@ -45,9 +45,6 @@ buildscript {
         classpath 'com.radcortez.gradle:openjpa-gradle-plugin:3.1.0'
         classpath 'gradle.plugin.org.nosphere.apache:creadur-rat-gradle:0.3.1'
         classpath "gradle.plugin.com.github.spotbugs.snom:spotbugs-gradle-plugin:4.0.8"
-        // Use Guava version 23+ as a workaround to spotbug intergration.
-        // See: https://github.com/spotbugs/spotbugs-gradle-plugin/issues/128#issuecomment-535864882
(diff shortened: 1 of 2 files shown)


vorburger:
@ptuomola are you in favour of this?

ptuomola:
Yes - probably the simplest solution. Just out of interest: which version do we end up with this? Hopefully the newest of the dependencies, rather than just what has been cached or is referenced first?

vorburger:
> which version do we end up with this?

_Currently that "whatever version" is a 28.2 (we used 28.1 before)._

@awasum @xurror given @ptuomola's vote of support here, would one of you like to merge this? (Or I can self merge it in 24 hours.)

awasum:
Given that Travis and Spotbugs have not complaint, the problem is fixed. Also given the spotbugs problem required using a Guava version above 23, this has now, surely(i have not looked closely) been fixed given the Guava version on the classpath via transitive dep will be higher than 23. We can merge this ASAP. We can always reverse in case of problems. Thanks @vorburger and @ptuomola for this.

PR #914, opened 2020-05-20
FINERACT-723: Fixing date comparisons

ptuomola:
## Description
Three integration tests were using the local server timezone rather than using the correct utility function to retrieve the tenant timezone in test

FINERACT-723

### DIFF
--- a/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/shares/ShareAccountIntegrationTests.java
+++ b/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/shares/ShareAccountIntegrationTests.java
@@ -201,8 +201,7 @@ public class ShareAccountIntegrationTests {
             } else if (transactionType.equals("charge.payment")) {
                 Assert.assertEquals("2.0", String.valueOf(transaction.get("amount")));
                 Assert.assertEquals("0", String.valueOf(transaction.get("amountPaid")));
-                Date transactionDate = DateUtils.getDateOfTenant() ;
-                Assert.assertEquals(simple.format(transactionDate), simple.format(date));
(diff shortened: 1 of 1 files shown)


awasum (approved):
This looks good to me.
In https://issues.apache.org/jira/browse/FINERACT-723 , Seems there are other cases like this in other parts of the code base? Do we want to look at that later? in a separate issue?

awasum:
See https://github.com/apache/fineract/pull/627 do we need to remove the hard coded Timezone we are setting in Travis now with this PR? @vorburger ..Maybe?

ptuomola:
> This looks good to me.
> In https://issues.apache.org/jira/browse/FINERACT-723 , Seems there are other cases like this in other parts of the code base? Do we want to look at that later? in a separate issue?

You're right: based on the JIRA, there are probably other scenarios that also lead to failures. However these are the only test that fail on my machine every day if I run the integration tests in the evening - so I had a personal interest in fixing them. My suggestion would be to incorporate this PR if it looks OK to you, but we can still keep the JIRA open for other fixes of the same nature.

ptuomola:
> See #627 do we need to remove the hard coded Timezone we are setting in Travis now with this PR? @vorburger ..Maybe?

The Utils class used by RestAssured also has the tenant timezone hardcoded:

in org.apache.fineract.integrationtests.common.Utils:
    public static final String TENANT_TIME_ZONE = "Asia/Kolkata";

A better solution would of course be to change the Utils class to look this up from the database.

After my PR, the current IT scope seems to work as everything is hardcoded to point to the same timezone (Asia/Kolkata). 

To make everything work with any timezone for tenant/database/application server without any hardcoding, then we will need to review/update all code and queries using either application server date and database date.

PR #915, opened 2020-05-20
FINERACT-986: Update spotbugs to v4.2.0

xurror:
Update spotbugs to v4.2.0

## Description

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccountCharge.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccountCharge.java
@@ -25,6 +25,7 @@ import static org.apache.fineract.portfolio.savings.SavingsApiConstants.feeInter
 import static org.apache.fineract.portfolio.savings.SavingsApiConstants.feeOnMonthDayParamName;
 import static org.apache.fineract.portfolio.savings.SavingsApiConstants.localeParamName;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.math.BigDecimal;
 import java.math.MathContext;
 import java.util.Date;
@@ -423,6 +424,8 @@ public class SavingsAccountCharge extends AbstractPersistableCustom {
         }
     }
 
+
+    @SuppressFBWarnings(value="NP_NULL_PARAM_DEREF_NONVIRTUAL") // https://issues.apache.org/jira/browse/FINERACT-987
     public Map<String, Object> update(final JsonCommand command) {
 
         final Map<String, Object> actualChanges = new LinkedHashMap<>(7);
@@ -465,6 +468,7 @@ public class SavingsAccountCharge extends AbstractPersistableCustom {
             final BigDecimal newValue = command.bigDecimalValueOfParameterNamed(amountParamName);
             actualChanges.put(amountParamName, newValue);
             actualChanges.put(localeParamName, localeAsInput);
+
             switch (ChargeCalculationType.fromInt(this.chargeCalculation)) {
                 case INVALID:
                 break;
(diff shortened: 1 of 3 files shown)


xurror:
@vorburger @awasum LGTY?

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccountCharge.java
```diff
@@ -474,7 +475,7 @@ public void update(final BigDecimal amount, final LocalDate dueDate, final Month
                 break;
                 case PERCENT_OF_AMOUNT:
                     this.percentage = newValue;
-                    this.amountPercentageAppliedTo = null;
+                    this.amountPercentageAppliedTo = BigDecimal.ZERO;
```
awasum: Hoping this is not checked for null somewhere in the code base?

awasum (commented):
+1. Except for the one small comment above.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccountCharge.java
xurror: Can't tell for sure, Just know it's a nullable field and as null, spotbugs throws an error when the value is passed to `update.percentageOf(BigDecimal, BigDecimal)`. So I just replaced that particular instance where I needed it to be something other than null

maektwain: I think we cannot have this zero, since it would affect the calculation, there is a reason why they were using null, Zero means there is some value provided ie zero, did you test it though?

xurror: Hmm, This is troublesome. I thought this was the most subtle way of going around this. I don't think multiplication by null is possible. This is actually some faulty method but I think I'll see how to make spotsbug ignore this just to stay on the safe side

maektwain: Yes true, might be just initializing it with null and then populating with real values... Could be like that

vorburger:
Of you have to use Suppress (it would be better to avoid it), then please
create a JIRA and add a comment like this: @SuppressWarning("...") // TODO
FINERACT-123

>

xurror:
> Of you have to use Suppress (it would be better to avoid it), then please create a JIRA and add a comment like this: @SuppressWarning("...") // TODO FINERACT-123
> […](#)

Suppress is not helping here. Unfortunately.

xurror:
I tried forcing the function to be able to accept a null parameter, but that just won't do:

```
org.apache.fineract.portfolio.savings.domain.SavingsAccountCharge since first historized release unspecified
In class org.apache.fineract.portfolio.savings.domain.SavingsAccountCharge
In method org.apache.fineract.portfolio.savings.domain.SavingsAccountCharge.percentageOf(BigDecimal, BigDecimal)
Parameter value
At SavingsAccountCharge.java:[lines 540-548]
value must be non-null but is marked as nullable
This parameter is always used in a way that requires it to be non-null, but the parameter is explicitly annotated as being Nullable. Either the use of the parameter or the annotation is wrong.
```



This got me thinking, why would the default value for a decimal be `null` instead of zero. It's a glitch that has worked well so far, probably because we never stumble on the crash. The method is vouched to fail if a null value is passed through.

maektwain:
The method is updating the values, in the savings account charge now the null value was there to initialise it , might be the case that it was checking for null values somewhere else while applying the charges and then based on that computing from zero. 

This could be interesting to note that we can start to remove all the null values where it should be a bigDecimal as zero functionally .

What are your thoughts ?

xurror:
Had some other concern, passing a ZERO to the function and performing a ZERO multiplication means the method returns a ZERO. I fear what a bunch of ZEROs everywhere may do to the core system functions.

xurror:
I just created https://issues.apache.org/jira/browse/FINERACT-987 for this.

xurror:
@awasum, @vorburger, @maektwain can we conclude this thread on these changes?

vorburger (approved):
LGTM

PR #916, opened 2020-05-20
FINERACT-942 Added and Enforced EmptyCatchBlock checkstyle

thesmallstar:
@vorburger I hope this is the correct approach to deal with empty catch blocks :)

 FINERACT-942

### DIFF
--- a/fineract-provider/config/checkstyle/checkstyle.xml
+++ b/fineract-provider/config/checkstyle/checkstyle.xml
@@ -77,6 +77,9 @@
         </module>
            <module name="OneStatementPerLine"/>
            <module name="EmptyStatement"/>
+         <module name="EmptyCatchBlock">
+            <property name="exceptionVariableName" value="expected|ignore"/>
+        </module>
              <module name="RegexpSinglelineJava">
             <property name="format" value="\.printStackTrace?\("/>
             <property name="message" value="Line contains printStacktrace()."/>
--- a/fineract-provider/src/main/java/org/apache/fineract/accounting/glaccount/service/GLAccountWritePlatformServiceJpaRepositoryImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/accounting/glaccount/service/GLAccountWritePlatformServiceJpaRepositoryImpl.java
@@ -184,6 +184,7 @@ public class GLAccountWritePlatformServiceJpaRepositoryImpl implements GLAccount
                 throw new GLAccountDisableException();
             }
         } catch (EmptyResultDataAccessException e) {
+            LOG.error("Problem encountered in validateForAttachedProduct()",e);
         }
     }
 
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/campaigns/sms/service/SmsCampaignDropdownReadPlatformServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/campaigns/sms/service/SmsCampaignDropdownReadPlatformServiceImpl.java
@@ -34,6 +34,8 @@ import org.apache.fineract.infrastructure.core.exception.PlatformDataIntegrityEx
 import org.apache.fineract.portfolio.calendar.domain.CalendarWeekDaysType;
 import org.apache.fineract.portfolio.calendar.service.CalendarEnumerations;
 import org.apache.fineract.portfolio.common.domain.PeriodFrequencyType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.core.ParameterizedTypeReference;
 import org.springframework.http.HttpEntity;
@@ -45,7 +47,7 @@ import org.springframework.web.client.RestTemplate;
 
 @Service
 public class SmsCampaignDropdownReadPlatformServiceImpl implements SmsCampaignDropdownReadPlatformService {
-
+    private final static Logger LOG = LoggerFactory.getLogger(SmsCampaignDropdownReadPlatformServiceImpl.class);
     private final RestTemplate restTemplate;
 
     private final SmsConfigUtils smsConfigUtils ;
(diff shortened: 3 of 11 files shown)


vorburger:
> @vorburger I hope this is the correct approach to deal with empty catch blocks :)

Your job here is going to be a little bit more challenging, thus more interesting! :smiling_imp: 

What you are getting into here is kind of related to what I alluded to in FINERACT-932... there are 2 cases: You're finding a number of obvious plain wrong **empty** `catch` statements. For example, catching a `javax.mail.MessagingException` and then just ignoring and doing absolutely nothing about it is most certainly a clear bug! If your code is trying to send out an email, and that fails, probably your SMTP server is fried, or whatever - and you certainly do want an "operator" who supervises error logs (e.g. like I do on https://www.fineract.dev, using https://cloud.google.com/error-reporting) do know about that. E.g. that `GeneralSecurityException` you spotted most likely also fall into this category.

However, for e.g. `EmptyResultDataAccessException` I'm personally not entirely sure yet. What does that exception signal? Presumably Spring is indicating that some JDBC ResultSet was empty, when it should not have been? Is that OK, or a clear indication of a database corruption? That depends on the specific code... But if it is OK and can happen during "normal" execution, then... there is another problem - then some other Spring API which returns an empty list or an Optional or something like that should have been used instead of expecting an exception! Do you know what I mean?

BTW can you write up what I'm explaining here into a follow-up PR to #912 ? :smile:

vorburger:
> BTW can you write up what I'm explaining here into a follow-up PR to #912 ?

FYI I kind of just did that myself - hope you don't mind. Let's get #912 merged, and you can then improve upon it in this or future PRs, if anything is new or missing.

thesmallstar:
@vorburger I am looking into what you said, will update shortly on this :)

thesmallstar:
@vorburger I will continue of this now :D
Thank you for the detailed view on how to work on this.
From what I understand, I will go with two approaches firstly if it is "really an exception" and needs to inform to the operator I will log sufficient description of the error. 
Otherwise, If this might indicate a "bad approach to solve a problem" and might need a complete restructure of the code, I will try to do it :)

In both cases, I think looking what exactly an exception means is important in this case?  and is the solution :)

vorburger:
> @vorburger I will continue of this now :D

@thesmallstar don't forget about this one, it would be great to see this wrapped up... this week?

thesmallstar:
Yes, already working on this would be done in a day or two max :)

thesmallstar:
@vorburger I wrote how I approached most of the exceptions in comment.
I did not see a major problem(That like what you suggested) please suggest if you have anything on the comments :) 
If you don't have anything to add, we can just remove the comment and merge this.

EDIT: please ignore that I removed the check of line endings, will be fixed before me merge this after review of comments.

thesmallstar:
Another mistake in comments I have explained that:
EmptyResultDataAccessException is thrown when more than one row is returned by the query. It is actually thrown when atleast one row is expected and zero received
(Sorry for that)
The explanation for the count (*) still remains the same.
It is bound to return a single row

fineract-provider/config/checkstyle/checkstyle.xml
```diff
@@ -35,10 +35,10 @@
         <property name="format" value="\s+$"/>
         <property name="message" value="Line has trailing spaces."/>
     </module>
-    <module name="RegexpMultiline">
+    <!-- <module name="RegexpMultiline">
```
vorburger: why comment this out here for making this change?

fineract-provider/src/main/java/org/apache/fineract/accounting/glaccount/service/GLAccountWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -184,6 +184,9 @@ private void validateForAttachedProduct(Long glAccountId) {
                 throw new GLAccountDisableException();
             }
         } catch (EmptyResultDataAccessException e) {
+            // EmptyResultDataAccessException is thrown when more than one row is returned by queryForObject()
+            // count(*) Enforces the return of a single row and hence we should never arrive in this catch block
+            logger.error("Problem encountered in validateForAttachedProduct()",e);
```
vorburger: :+1:

fineract-provider/src/main/java/org/apache/fineract/infrastructure/campaigns/sms/service/SmsCampaignDropdownReadPlatformServiceImpl.java
```diff
@@ -81,6 +83,9 @@ public SmsCampaignDropdownReadPlatformServiceImpl(final SmsConfigUtils smsConfig
             if (!responseOne.getStatusCode().equals(HttpStatus.OK)) {
             }
         } catch (Exception e) {
+            //Developer was too lazy to figure out whwat can go wrong?
```
vorburger: you could already change this now? (or later, up to you; later is perfectly fine with me too)

vorburger (changes requested):

fineract-provider/config/checkstyle/checkstyle.xml
thesmallstar: Oh I think you missed it, I wrote a note above already that this is unintentional and will be fixed when I remove the comments

thesmallstar:
Will update this now :) 

Update: Faced some minor issues, still working on this should be done by tomorrow.

thesmallstar:
Update:
1. I have gone through all the empty catch in detail, I found no catch block that was suspicious, in most of the cases they were left empty because we were in a special case where we won't enter the catch block.
2. My past observations are on this branch if anyone wants to refer: https://github.com/thesmallstar/fineract/tree/refer
Note that, EmptyResultDataAccessException is thrown when more than one row is returned by the query. It is actually thrown when atleast one row is expected and zero received
(I wrote it incorrectly in comments, Sorry for that)
The explanation for the count (*) still remains the same.
It is bound to return a single row
3. Use of Exception class is not yet done, it will be the very next checkstyle I work on. I have not done it here to collectively address them all in a single PR.

vorburger (approved):
LGTM

PR #918, opened 2020-05-21
ignore StorageTest [FINERACT-828]

vorburger:
FINERACT-828


vorburger:
@awasum @xurror merge? Unless you're motivated to fix `StorageTest`... I'm giving up on it! :smile:

xurror:
I just don't get the thing with the storageTest. Seems to be some inconsistences only funny part is these inconsistences does not happen all the time.

xurror:
I'm merging this

PR #574, opened 2019-06-04
FINERACT-609 Group Topup Loan Bug Fix.

DylanRobson:
Fix FINERACT-609 issue where topup loan wasn't working for group loan accounts. This was because the server wasn't sending the JSON array (named activeClientLoanOptions) which contained the IDs of active group loan accounts that could be closed in order to create the topup group loan account.
Now when creating a group topup loan, the server correctly replies with active group loans (named activeClientLoanOptions) so that an account to close can be selected in order to create the new group topup loan account.

## Description
Describe the changes made and why they were made. Ignore if these details are present on the associated Jira ticket
Added query: retrieveGroupActiveLoanAccountSummary(Long groupId) to get the active group loans (that can be closed) in order to complete the group topup loan application.

## Checklist
Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Commit message starts with the issue number from https://issues.apache.org/jira/projects/FINERACT/. Ex: FINERACT-646 Pockets API.

- [x] Coding conventions at https://cwiki.apache.org/confluence/display/FINERACT/Coding+Conventions have been followed.

- [x] API documentation at https://github.com/apache/fineract/blob/develop/api-docs/apiLive.htm has been updated with details of any API changes.
I don't think any update is required.

- [ ] Integration tests have been created/updated for verifying the changes made.

- [ ] All Integrations tests are passing with the new commits.
I'm creating this PR as advised by my GSoC mentors to see whether the integration tests on Travis-CI succeed. Even my local upstream/develop branch has failing integration tests - possibly due to https://issues.apache.org/jira/browse/FINERACT-707

- [x] Submission is not a "code dump".  (Large changes can be made "in repository" via a branch.  Ask on the list.)

Our guidelines for code reviews is at https://cwiki.apache.org/confluence/display/FINERACT/Code+Review+Guide

PR #924, opened 2020-05-22
FINERACT-609 TopUp

percyashu:
Build from https://github.com/apache/fineract/pull/574 
## Description
Describe the changes made and why they were made. Ignore if these details are present on the associated Jira ticket  https://issues.apache.org/jira/browse/FINERACT-609

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/accountdetails/service/AccountDetailsReadPlatformServiceJpaRepositoryImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/accountdetails/service/AccountDetailsReadPlatformServiceJpaRepositoryImpl.java
@@ -157,7 +157,8 @@ public class AccountDetailsReadPlatformServiceJpaRepositoryImpl implements Accou
         return retrieveLoanAccountDetails(loanWhereClause, new Object[] { groupId, loanOfficerId });
     }
 
-    @Override public Collection<LoanAccountSummaryData> retrieveClientActiveLoanAccountSummary(final Long clientId) {
+    @Override
+    public Collection<LoanAccountSummaryData> retrieveClientActiveLoanAccountSummary(final Long clientId) {
         final String loanWhereClause = " where l.client_id = ? and l.loan_status_id = 300 ";
         return retrieveLoanAccountDetails(loanWhereClause, new Object[] { clientId });
     }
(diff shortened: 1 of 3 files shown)


fineract-provider/src/main/java/org/apache/fineract/portfolio/accountdetails/service/AccountDetailsReadPlatformServiceJpaRepositoryImpl.java
```diff
@@ -171,6 +172,12 @@ public AccountSummaryCollectionData retrieveGroupAccountDetails(final Long group
         return this.jdbcTemplate.query(sql, rm, new Object[]{groupId , glimAccount});
     }
 
+    @Override
+    public Collection<LoanAccountSummaryData> retrieveGroupActiveLoanAccountSummary(final Long groupId) {
+        final String loanWhereClause = " where l.group_id = ? and l.loan_status_id = 300 ";
```
vorburger: This still doesn't seem to take the code review feedback that @vishwasbabu had provided in #574 into account?

vorburger (changes requested):

fineract-provider/src/main/java/org/apache/fineract/portfolio/accountdetails/service/AccountDetailsReadPlatformServiceJpaRepositoryImpl.java
percyashu: From 

> (since you check if client ID is not null in the service method), it might introduce bugs in the future when someone tries to reuse this method. I would suggest returning only loans of type 3.

I removed the conditional statement to check if client ID is null from LoanReadPlatformServiceImpl

vorburger: @vishwasbabu does this LGTY?

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanReadPlatformServiceImpl.java
```diff
@@ -1428,15 +1428,15 @@ public LoanAccountData retrieveLoanProductDetailsTemplate(final Long productId,
             }
         }
 
-        Collection<LoanAccountSummaryData> clientActiveLoanOptions = null;
-        if(loanProduct.canUseForTopup() && clientId != null){
-            clientActiveLoanOptions = this.accountDetailsReadPlatformService.retrieveClientActiveLoanAccountSummary(clientId);
```
vishwasbabu, line 1433: We shouldn't be removing the existing condition which checks for client loans, rather we should be adding an else if for additionally checking for group loans

fineract-provider/src/main/java/org/apache/fineract/portfolio/accountdetails/service/AccountDetailsReadPlatformServiceJpaRepositoryImpl.java
vishwasbabu: I would append a clause to the effect of "l.client_id is NULL", this ensures that only group loans are fetched by the query all JLG loans are ignored (loans which are associated with clients and groups and fetched as a part of the retrieveClientActiveLoanAccountSummary method)

vishwasbabu (changes requested):

percyashu:
@vorburger and @vishwasbabu LGTY?

vorburger:
@vishwasbabu I'll let you review (and merge, if/when OK..) this one, and ignore it; I've got (more than enough..) other PRs to tend to! ;)

PR #925, opened 2020-05-22
FINERACT-997

maektwain:
While debugging, the previous code was returning null even with data. Tested on the current fineract.dev (does not work)
findProductIdAndProductTypeAndFinancialAccountTypeAndChargeId

## Description
When you make a savings product with charges and associate them with accounting using advanced accounting rules you can map them to INCOME and FEES based operating headers.

Once any transaction associated to this charges will be diverted in those account headers properly.

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/domain/ProductToGLAccountMappingRepository.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/domain/ProductToGLAccountMappingRepository.java
@@ -30,8 +30,9 @@ public interface ProductToGLAccountMappingRepository extends JpaRepository<Produ
     ProductToGLAccountMapping findByProductIdAndProductTypeAndFinancialAccountTypeAndPaymentTypeId(Long productId, int productType,
             int financialAccountType, Long paymentType);
 
-    ProductToGLAccountMapping findByProductIdAndProductTypeAndFinancialAccountTypeAndChargeId(Long productId, int productType,
-            int financialAccountType, Long chargeId);
+    @Query("select mapping from ProductToGLAccountMapping mapping where mapping.productId= :productId and mapping.productType= :productType and mapping.financialAccountType= :financialAccountType and mapping.charge.id= :chargeId")
+    ProductToGLAccountMapping findProductIdAndProductTypeAndFinancialAccountTypeAndChargeId(@Param("productId") Long productId, @Param("productType") int productType,
+            @Param("financialAccountType") int financialAccountType, @Param("chargeId") Long ChargeId);
 
     @Query("select mapping from ProductToGLAccountMapping mapping where mapping.productId =:productId and mapping.productType =:productType and mapping.financialAccountType=:financialAccountType and mapping.paymentType is NULL and mapping.charge is NULL")
     ProductToGLAccountMapping findCoreProductToFinAccountMapping(@Param("productId") Long productId, @Param("productType") int productType,
--- a/fineract-provider/src/main/java/org/apache/fineract/accounting/journalentry/service/AccountingProcessorHelper.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/accounting/journalentry/service/AccountingProcessorHelper.java
@@ -1124,7 +1124,7 @@ public class AccountingProcessorHelper {
         if (accountMappingTypeId == CashAccountsForLoan.INCOME_FROM_FEES.getValue()
                 || accountMappingTypeId == CashAccountsForLoan.INCOME_FROM_PENALTIES.getValue()) {
             final ProductToGLAccountMapping chargeSpecificIncomeAccountMapping = this.accountMappingRepository
-                    .findByProductIdAndProductTypeAndFinancialAccountTypeAndChargeId(loanProductId, PortfolioProductType.LOAN.getValue(),
+                    .findProductIdAndProductTypeAndFinancialAccountTypeAndChargeId(loanProductId, PortfolioProductType.LOAN.getValue(),
                             accountMappingTypeId, chargeId);
             if (chargeSpecificIncomeAccountMapping != null) {
                 accountMapping = chargeSpecificIncomeAccountMapping;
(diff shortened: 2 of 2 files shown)


vorburger:
> While debugging, the previous code was returning null even with data. Tested on the current fineract.dev (does not work)

@maektwain is this related to anything linked from FINERACT-932? If not, this sounds like it deserves a dedicated JIRA - will you create one?

maektwain:
Not that I know, I was going through some backlogs, I realised that there are some potentials areas where I can help? Could not grasp what are we trying to fix avoid null values or fix the Logging style or catching exception ?

I tested the issue, on Fineract.dev so did not work.

fineract-provider/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/domain/ProductToGLAccountMappingRepository.java
```diff
@@ -30,8 +30,9 @@
     ProductToGLAccountMapping findByProductIdAndProductTypeAndFinancialAccountTypeAndPaymentTypeId(Long productId, int productType,
             int financialAccountType, Long paymentType);
 
-    ProductToGLAccountMapping findByProductIdAndProductTypeAndFinancialAccountTypeAndChargeId(Long productId, int productType,
-            int financialAccountType, Long chargeId);
+    @Query("select mapping from ProductToGLAccountMapping mapping where mapping.productId= :productId and mapping.productType= :productType and mapping.financialAccountType= :financialAccountType and mapping.charge.id= :chargeId")
+    ProductToGLAccountMapping findProductIdAndProductTypeAndFinancialAccountTypeAndChargeId(@Param("productId") Long productId, @Param("productType") int productType,
+            @Param("financialAccountType") int financialAccountType, @Param("chargeId") Long ChargeId);
```
vorburger, line 35: @maektwain I'd like to better understand what problem this fixes how... At first glance, to me this looks like we are just writing an explicit `@Query` for what Spring Data otherwise automagically internally generates with from its `findBy` syntax... can you help me understand what I'm missing? What Query does the current version generate, and what is wrong with it which this explicit version fixes? How is the auto-generated query wrong?

maektwain: So, with findBy  query it does not return data , it return null value, even data is available, making it more dumb and simple.

maektwain: So, I tested it across steps to reproduce the above  now.

Create a savings product, define accounting and associate charges with advanced accounting mapping to Income and Fees.

Make a transaction, Charges associate to advanced accounting should go to those headers defined in previous step. 

Debugged and realised its coming null and with this query its appropriately returning and posting in journal entries.

vorburger: > So, with findBy query it does not return data , it return null value, even data is available, making it more dumb and simple.
 
I found this puzzling, and had a closer look... and noticed that `AndChargeId` is curious - there is no `chargeId` getter and field on `ProductToGLAccountMapping` - only a `charge`. Do you want to try to see if just renaming the method instead of having to manually specify the query also fixes this? If that is so, then I'm interested how to find a way to validate those Spring Data Repository interfaces...

>  So, I tested it across steps to reproduce the above now.

do you think you would be able to write an integration test for this problem in this PR? Just asking!

maektwain: Didn't realise that JPA binds them with getters ( :() ) I think no need , the test is already there I guess , if it's  not there then I can write.

maektwain: So, if we rename the method then we eventually have to change the existing method sigs, since it uses the ChargeId , and then in that case we need to pass the Charge. This fixes the problem.

maektwain: Fields Types are different and usage is also different if you look here 

File is AccountingProcessorHelper.java

``` java
if (accountMappingTypeId == CashAccountsForLoan.INCOME_FROM_FEES.getValue()
                || accountMappingTypeId == CashAccountsForLoan.INCOME_FROM_PENALTIES.getValue()) {
            final ProductToGLAccountMapping chargeSpecificIncomeAccountMapping = this.accountMappingRepository
                    .findByProductIdAndProductTypeAndFinancialAccountTypeAndChargeId(loanProductId, PortfolioProductType.LOAN.getValue(),
                            accountMappingTypeId, chargeId);
            if (chargeSpecificIncomeAccountMapping != null) {
                accountMapping = chargeSpecificIncomeAccountMapping;
            }
        }
```
It uses the chargeId 

and perviously it gets passed here 

```
final GLAccount chargeSpecificAccount = getLinkedGLAccountForSavingsCharges(savingsProductId, accountTypeToBeCredited.getValue(),
                chargePaymentDTO.getChargeId());
```

vorburger: @maektwain thanks for your explanations, I now (kind of) see what you mean - I had previously missed the `Charge` vs. `Long chargeId` ... then I guess this fix makes sense. I'm still puzzled as to why Spring doesn't flag this, but let's pick that up separately in FINERACT-998 later. I'm going to LGTM this PR, but not merge it, to let others review it as well, and give you a chance to still add an IT for it, that would be nice, but I don't want to blocking merging this even if no IT.

vorburger (approved):

fineract-provider/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/domain/ProductToGLAccountMappingRepository.java
vorburger: @maektwain small practical thing, normally the commit message itself should include the JIRA, which is missing here, as you just added it to the PR title afterwards.. perhaps you want to amend the commit message?

maektwain: Thanks for the understanding . ACK.  

It has been that way, I will try to add IT too soon.

maektwain: Sure

vorburger (approved):
LGTM.

PR #926, opened 2020-05-22
FINERACT-977

maektwain:
## Description
Probable fix, please test before or replay , though these scenario are typically because of null values stored in the data structure .

fix possible NPE in AbstractLoanScheduleGenerator

Need to come up with better way to have default values in such case and calculation criteria.

## Checklist


vorburger:
@nnatarajan following along the code review conversation in this Pull Request could interest you.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/AbstractLoanScheduleGenerator.java
```diff
@@ -517,9 +517,11 @@ private LoanScheduleModelPeriod handleRecalculationForTransactions(final MathCon
                                 scheduleParams.getTotalCumulativePrincipal().plus(
                                         currentPeriodParams.getPrincipalForThisPeriod().minus(principalProcessed)),
                                 scheduleParams.getPeriodNumber() + 1, mc));
-                        if (loanApplicationTerms.getAmortizationMethod().isEqualInstallment()
-                                && fixedEmiAmount.compareTo(loanApplicationTerms.getFixedEmiAmount()) != 0) {
-                            currentPeriodParams.setEmiAmountChanged(true);
+                        if (loanApplicationTerms.getAmortizationMethod().isEqualInstallment())
+                          if(fixedEmiAmount != null) {
+                            if (fixedEmiAmount.compareTo(loanApplicationTerms.getFixedEmiAmount()) != 0) {
```
vorburger: The 3 nested `if` are not necessary, and this should instead be written like so (correctly formatted).

```suggestion
                        if (loanApplicationTerms.getAmortizationMethod().isEqualInstallment()
                          && fixedEmiAmount != null
                          && fixedEmiAmount.compareTo(loanApplicationTerms.getFixedEmiAmount()) != 0) {
```

More importantly, while I don't understand much about this functionality, it seems to me that as-is we may be missing a subtle corner case... what if one of `fixedEmiAmount` and `loanApplicationTerms.getFixedEmiAmount()` are `null` but (respectively) the other is not (anymore?) - should that also trigger `setEmiAmountChanged(true)` ? I don't know details about this, but that would kind of seem "logical", to me. If you agree, then the check should take that case into account.

BTW we are assuming here that it is valid for an "Emi Amount" (whatever that is???) to be `null`. Again, I don't know anything about this functionality, so unless you know more, I guess it's a fair assumption. If someone know actually understands the details of this can chime in here with knowledge if it's expected to (sometimes) be null, or if that is really an indication of another problem, that could be interesting, but if we don't learn more, let's just go for making it null safe.

In an ideal world, an integration test for this would be nice, but I'd rather merge it without one than not having this fixed... ;-)

The same feedback as above should be applied below as well.

Lastly, the commit message here needs to be improved, instead of "Fix for Null Values" e.g. "fix possible NPE in AbstractLoanScheduleGenerator [FINERACT-977]" would be better.

@pramodn02 according to Git Blame log, these lines of code were introduced by you, long ago, for [MIFOSX-579 ](https://mifosforge.jira.com/browse/MIFOSX-579)... if you still have any interest in this project, perhaps you'd like to chime in here!

vorburger (changes requested):

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/AbstractLoanScheduleGenerator.java
maektwain: I thought about this, @vorburger , in case where fixedAmount is null , then the third condition will simply throw null exception. Don't you think?

Let me simplify the case for you.

Given a loan scenario where loan product type Amortisation is set as Equal Instalment so, you could set the fixedEmi Amount which is your repayment as when you create a loan product you can set this boolean value here 

Allow fixing of the installment amount []

if this and other condition is true of equal instalment then while making a loan you can set the EMI of your own choice like in case of 100 USD loan you can choose to set every month 10 USD instalment then system is forced to have this as every month due.

Now in the above case there are two parts, when we decide to set the FixedEMI or not and when we don't then there is null value and then you can imagine what might happen.

Now in code level, we should actually take care of null values since we know the cases, but it was not taken care here.

It won't trigger `setEmiAmountChanged (true)`if fixedEMI is null,  and in the edge case its null, so better we check if fixedEmi is null and then apply the method .

maektwain: This part is based on the fact that `fixedEmiAmount.compareTo(loanApplicationTerms.getFixedEmiAmount()) != 0)`
fixedEmiAmount is not going to be null ? 
It will fail even if loanApplicationTerms.getFixedEmiAmount() this is going to be null, since the first hand requires not to be null. 

We need to broadly think of the null checks in the application, this way we can make code more clean and easier to such edge cases, 

In my opinion there are many places like this where null values are sometimes checked and if the field dependent to some fields is not logically set it fails at null exception

vorburger:
*@maektwain* Java expressions evaluation respects parentheses (and && ||
Precedenc), so it first checks "fixedEmiAmount != null" and if that is
false, it does not evaluate "fixedEmiAmount.compareTo()" anymore, so this
is safe.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/AbstractLoanScheduleGenerator.java
vorburger: why did you use the && style below as discussed, but kept if/if/if style here? This is inconsistent, please change.

vorburger (changes requested):
NOK, see reopened prior discussion.

Also, 3 commits must be squashed into 1 please.

vorburger:
@maektwain respectfully pinging you re. this, after 1 week. Are you planning to wrap this up?

maektwain:
Does it makes sense to you ? oh didn't realise I had this merge on top?

vorburger:
@maektwain LGTM but as previously requested, all commits must be squashed into 1 please.

maektwain:
Let me reopen a new PR

vorburger:
#960 @maektwain FYI you don't actually have to open a new PR, you can just "force push" (next time)

PR #931, opened 2020-05-23
FINERACT-821 Added and enforced NeedBraces Checkstyle

thesmallstar:

### DIFF
--- a/fineract-provider/config/checkstyle/checkstyle.xml
+++ b/fineract-provider/config/checkstyle/checkstyle.xml
@@ -87,6 +87,7 @@
             <property name="allowNonPrintableEscapes" value="true"/>
         </module>
           <module name="UpperEll"/>
+          <module name="NeedBraces"/>
 <!-- TODO Enable many more checks (go about this one by one, step by step, raise separate PRs fixing and then enforcing):
 
         <module name="HiddenField">
@@ -126,7 +127,7 @@
             <property name="option" value="TEXT"/>
             <property name="tokens" value="LITERAL_TRY, LITERAL_CATCH, LITERAL_FINALLY, LITERAL_IF, LITERAL_ELSE, LITERAL_SWITCH"/>
         </module>
-        <module name="NeedBraces"/>
+      
         <module name="LeftCurly" />
         <module name="RightCurly">
             <property name="option" value="alone"/>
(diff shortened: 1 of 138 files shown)


thesmallstar:
@vorburger this is the most not-so-obedient checkstyle :(

vorburger:
WOW! Did you do this manually, or with a tool? IDE like Eclipse or IDEA could maybe help with something like this.. Also I'm trying to decide if I should review this carefully line by line, or just merge... @awasum @Anh3h thoughts? Should be safe?

thesmallstar:
@vorburger I tried to automate this with every plugin/ide that I knew. The best help I could get was a quick fix bulb icon in VSC, which on visiting such if/for loop would help me identify/add braces, but still I had to visit each one of them individually(not a very exciting job tbh). A detailed review will be better? although unlikely if there is a  problem, we might not even notice it after this gets merged. Also, It would be great if we could merge this is 3/4 days to avoid multiple merge conflicts :)

awasum:
> WOW! Did you do this manually, or with a tool? IDE like Eclipse or IDEA could maybe help with something like this.. Also I'm trying to decide if I should review this carefully line by line, or just merge... @awasum @Anh3h thoughts? Should be safe?

He is just adding needed braces where necessary, which makes code clearer. Would have been good to look through line by line. I just schemed through and nothing seems to be wrong with the PR. Even though I have not looked closely, it should not be a problem if we merge this.

awasum:
Will a rebase fix the conflicts? or it needs manual resolution?

vorburger (approved):
> Even though I have not looked closely, it should not be a problem if we merge this.

@awasum yeah OK, then @thesmallstar could you rebase this, and I'm OK that we merge this ASAP.

thesmallstar:
@vorburger this is ready for merge.

vorburger:
Why comment out the RegExpMultiline?

thesmallstar:
Argh! Sorry for another mistake.
UNINTENTIONAL AGAIN :( 
Actually I use windows for this specific checkstyle (VSCcode plugin I told earlier worked smoother). The build fails in windows due to that specific checkstyle(We can probably fix that?). So I removed that for testing the code. 
Adding it back in :)

PR #940, opened 2020-05-26
FINERACT-965 Upgrade, Replace itext 2.1.7 with Openpdf 1.1.0

nnatarajan:
FINERACT-965

### DIFF
--- a/fineract-provider/build.gradle
+++ b/fineract-provider/build.gradle
@@ -87,8 +87,7 @@ dependencyManagement {
         dependency 'org.apache.commons:commons-email:1.5'
         dependency 'org.apache.commons:commons-io:+'
         dependency 'org.drizzle.jdbc:drizzle-jdbc:1.4'
-        dependency 'com.lowagie:itext:2.1.7'
-        dependency 'com.lowagie:itext-rtf:2.1.7'
+        dependency 'com.github.librepdf:openpdf:1.1.0'
         dependency 'org.mnode.ical4j:ical4j:3.0.18'
         dependency 'org.flywaydb:flyway-core:6.4.1'
         dependency 'org.quartz-scheduler:quartz:+'
@@ -160,6 +159,7 @@ openjpa {
 jacoco {
     toolVersion = jacocoVersion
     reportsDir = file("$buildDir/reports/jacoco")
+
 }
 
 jacocoTestReport{
--- a/fineract-provider/dependencies.gradle
+++ b/fineract-provider/dependencies.gradle
@@ -65,7 +65,7 @@ dependencies {
             'org.flywaydb:flyway-core',
 
             'net.sf.ehcache:ehcache',
-            'com.lowagie:itext',
+            'com.github.librepdf:openpdf',
             'com.github.spullara.mustache.java:compiler',
             'com.jayway.jsonpath:json-path',
 


fineract-provider/build.gradle
```diff
@@ -87,8 +87,9 @@ dependencyManagement {
         dependency 'org.apache.commons:commons-email:1.5'
         dependency 'org.apache.commons:commons-io:+'
         dependency 'org.drizzle.jdbc:drizzle-jdbc:1.4'
-        dependency 'com.lowagie:itext:2.1.7'
-        dependency 'com.lowagie:itext-rtf:2.1.7'
+        dependency 'com.lowagie:itext:4.2.2'
+        dependency 'com.lowagie:itext-rtf:4.2.2'
+        dependency 'com.itextpdf:itextpdf:5.5.13'
```
vorburger: The mix of v4 and v5 here seems curious... is this really right?

vorburger (changes requested):

fineract-provider/build.gradle
nnatarajan: When I updated itext, I got error, requiring com.itextpdf:itextpdf. It seems that they are independent vendors. Version 5 is the most current version and dates back to 2011.

nnatarajan: https://mvnrepository.com/artifact/com.itextpdf/itextpdf

vorburger:
Wait a sec, so iText v5 is AGPL licensed?! Then we can't use it in Fineract, see https://www.apache.org/legal/resolved.html#category-x ...

Reading https://en.m.wikipedia.org/wiki/IText and seeing on https://search.maven.org/search?q=iText that v4.2.2 is ancient from 2015.

I suggest that we use this as the opportunity to switch to https://github.com/LibrePDF/OpenPDF instead. That's dual LGPL (NOK) and MPL (fine).

nnatarajan:
> Wait a sec, so iText v5 is AGPL licensed?! Then we can't use it in Fineract, see https://www.apache.org/legal/resolved.html#category-x ...
> 
> Reading https://en.m.wikipedia.org/wiki/IText and seeing on https://search.maven.org/search?q=iText that v4.2.2 is ancient from 2015.
> 
> I suggest that we use this as the opportunity to switch to https://github.com/LibrePDF/OpenPDF instead. That's dual LGPL (NOK) and MPL (fine).

Is the latest version ok -- OpenPDF 1.3.17 ?

nnatarajan:
> Wait a sec, so iText v5 is AGPL licensed?! Then we can't use it in Fineract, see https://www.apache.org/legal/resolved.html#category-x ...
> 
> Reading https://en.m.wikipedia.org/wiki/IText and seeing on https://search.maven.org/search?q=iText that v4.2.2 is ancient from 2015.
> 
> I suggest that we use this as the opportunity to switch to https://github.com/LibrePDF/OpenPDF instead. That's dual LGPL (NOK) and MPL (fine).

Dependency Openpdf: 1.1.0 is the most recent version to fit all class requirements.

vorburger (changes requested):
> Dependency Openpdf: 1.1.0 is the most recent version to fit all class requirements.

https://github.com/LibrePDF/OpenPDF says latest is 1.3.17 ... does that not "fit all class requirements"? Can you elaborate on what you mean by that? It should, hopefully, be possible to upgrade to the latest, even if some import statements or some code has to be changed...

nnatarajan:
> > Dependency Openpdf: 1.1.0 is the most recent version to fit all class requirements.
> 
> https://github.com/LibrePDF/OpenPDF says latest is 1.3.17 ... does that not "fit all class requirements"? Can you elaborate on what you mean by that? It should, hopefully, be possible to upgrade to the latest, even if some import statements or some code has to be changed...

@vorburger 
OpenPDF: 1.1.0 from 2018 is the latest version, which will recognize this import in a few classes -- com.lowagie.text.pdf.codec.Base64.

More recent versions of OpenPDF do not recognize Base64

vorburger:
> More recent versions of OpenPDF do not recognize Base64

thanks for having identified that. FYI Base64 is a (very common) "encoding scheme", read https://en.wikipedia.org/wiki/Base64, and as per http://www.docjar.com/docs/api/com/lowagie/text/pdf/codec/Base64.html => http://iharder.sourceforge.net/current/java/base64/ they just ditched that because now there is https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html

I've done the "conversion" in new #948 (that PR includes your commit, for correct "attribution").

OK to close this?

nnatarajan:
> > More recent versions of OpenPDF do not recognize Base64
> 
> thanks for having identified that. FYI Base64 is a (very common) "encoding scheme", read https://en.wikipedia.org/wiki/Base64, and as per http://www.docjar.com/docs/api/com/lowagie/text/pdf/codec/Base64.html => http://iharder.sourceforge.net/current/java/base64/ they just ditched that because now there is https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html
> 
> I've done the "conversion" in new #948 (that PR includes your commit, for correct "attribution").
> 
> OK to close this?

@vorburger  Ok.

PR #933, opened 2020-05-24
FINERACT-821 Added multiple checkstyles

thesmallstar:
I feel sorry for someone who have to look and consider this PR :P , but anyways. 

So I came across this extremely cool repo: https://github.com/google/google-java-format
It formats java code according to google checkstyles(it is by google). 

It made out code look nicer, and helped implement some of the hardest to implement checkstyles otherwise(Even with plugins). 

Definitely not necessary to merge this, but what do you feel about it?
@vorburger  @awasum  @xurror  @ptuomola

PR #704, opened 2020-01-27
Activate and Enforce new Checkstyle check related to import statement order (FINERACT-821)

vorburger:
Please see the individual commit message for details about the steps.

Note how the reformatting and other changes are in separate commit, to ease review of this PR.

PR #943, opened 2020-05-27
FINERACT-1006 Added spotless to auto format source code

thesmallstar:
FINERACT-1006

@vorburger  fineractdev-eclipse-preferences.epf, the file that you provided, was not suitable to be used as config. I imported and exported the same as XML(there is a guide on spotless on how to do it https://github.com/diffplug/spotless/blob/master/ECLIPSE_SCREENSHOTS.md ).

The following works great:
For testing, you can use ./gradlew spotlessJavaCheck to run the checks and ./gradlew spotlessApply to fix violations. 

The problem now? 
I don't think fineractdev-eclipse-preferences.xml that we currently use is in the correct format as required by us, it adds trailing spaces after we run ./gradlew spotlessApply at some specific points.

Rest this is a really good improvement :) 
#933


thesmallstar:
The test fails due to spotless and that is expected currently.

thesmallstar:
@vorburger Do you think it is a good idea to let spotless handle our indentation?
Every one of them including google java format seems to have some variation. 

Take this case for example->
I tried adding the same indentation as suggested by google checks and ran google java format, but the checks still won't pass, there are some ambiguities present.

So to let spotless handle out indentation and let checkstyles not interfere with it, is a good idea? 
In fact every such checkstyle which is better if automated like spaces, can we let spotless take care?
Since with one command we would then fix the error, and we cannot have them at both places since they won't completely agree on "all" the places(I wonder why).

thesmallstar:
Update:
1. I have decided to go with google-checks instead of our old XML, since it is more complete. As suggested by @vorburger I have updated the indentation to 4.
2. I have commented all the checks that formats "comments/javadocs"
3. Spotless loves to add trailing spaces, so I have added action(see spotless section) in build.gradle and asked it to fix itself ;) 
4. If this is merged many hard-to-implement checkstyles like, the one related to spaces and line-wraps can be directly added (I have verified spaces checkstyles)
5. Indentation checkstyle still do not directly match with what spotless offers, but spotless can take good care of indentation (I believe).

What do you think @vorburger  @awasum  @xurror

xurror (approved):
Good work on this. There so many changed files It's hard to follow but the few I checked looked good to me. Hopefully it's that consistent throughout. Given that this follows a certain standard and there is no apparent conflict with checkstyle so far, I can only approve this. I'm not very knowledgeable about all this though. Maybe @vorburger will have some things to suggests but LGTM

thesmallstar:
I will pause the work on checkstyles until we can final this PR since it would be extremely hard to deal with merge conflicts otherwise. Looking into other planned issue currently. I hope we can make a decision on this in 2-3 days with required changes(If any).

thesmallstar:
@vorburger  @awasum  your views on this?

vorburger:
@thesmallstar So this idea look very interesting, to me; thanks for having investigated this! But I'd like to fully understand exactly what we are doing here a little more myself, before merging it... please know that, even though I've originally suggested you look into Spotless in FINERACT-1006, I've actually not read much about it myself... :smile: so I'm looking for your guidance to fully learn about it myself.

In order to make a full review and merging more easy and avoid conflicts, could I make a suggestion, just from a "practical" point of view? How about in this PR, you revert the .java code changes and ONLY propose whatever we need to add to Gradle (and README, hopefully?) and config files, but do not actually include the result of having run it to change the Java files? That makes it easier for me to completely review it properly, and hopefully eventually merge. We could then, in a separate next step, actually run and merge the result (or I could just do that myself, and merge it "quickly", to avoid delays).

One thing I haven't completely understood yet here is if this "just" adds a new way to Auto Format by launching a certain Gradle task (include doc in README...) or if this, like Checkstyle and Error Prone & Co. will actually verify formatting, and fail the build? Or is the idea that we leave that up to Checkstyle only, but that folks can then use this to make changes comply? Or both? It's something worth clearly documenting in the README, IMHO.

I guess what would be "ideal" is if this would both verify formatting, fail the build if NOK (like Checkstyle; I wouldn't mind if this PR fails to build, because of that), but if such failures would show a clear message à la "Hey buddy, you can easily fix your code's ugly formatting by running `./gradlew something`....

thesmallstar:
Yes, I will update this.
From what I have understood until now, Spotless and Checkstyles have different approaches. Spotless(Based on the configuration) is more about code formatting and checkstyle is a little bit more intelligent than just code formatting(Not a very good explanation). 

Keeping both seems a good idea, there are a few checks like naming conventions which cannot be handled by spotless whereas some like indentation/whitespace it would be a really good idea to let spotless take "more" care of it since adding those spaces myself after writing some code would be really troublesome, having said that for a few checks like "space" where spotless and checkstyle agree on everything(either both pass or both fail) we can add a note in check style error message to use spotless to fix this(makes sense?) and keep them both place. Only for a few checks, where they won't agree but it would be a really nice idea to have it automated(like indentation), we can let only spotless handle it. 

This is what I think would be a good approach if it sounds good to you, should I proceed and add it to readme?

vorburger:
Yes, please proceed in this direction! (And do remove the changes to the
.java, to make this PR much easier to review.)

>

thesmallstar:
@vorburger LGTY?

README.md
```diff
@@ -195,10 +195,19 @@ Instructions to run Apache RAT (Release Audit Tool)
 2. Run `./gradlew rat`. A report will be generated under build/reports/rat/rat-report.txt
 
 
-Checkstyle
+Checkstyle and Spotless
 ============
 
-This project enforces its code conventions using [checkstyle.xml](fineract-provider/config/checkstyle/checkstyle.xml).  It is configured to run automatically during the normal Gradle build, and fail if there are any style violations detected.
+This project enforces its code conventions using [checkstyle.xml](fineract-provider/config/checkstyle/checkstyle.xml) through Checkstyle and [google-checks](config/google-checks-updated.xml) through Spotless. They are configured to run automatically during the normal Gradle build, and fail if there are any violations detected. You can run the following command to automatically fix spotless violations
```
vorburger: super minor "nit pick" (which you can also address in a follow-up PR), the doc is missing a semi-colon here:

```suggestion
This project enforces its code conventions using [checkstyle.xml](fineract-provider/config/checkstyle/checkstyle.xml) through Checkstyle and [google-checks](config/google-checks-updated.xml) through Spotless. They are configured to run automatically during the normal Gradle build, and fail if there are any violations detected. You can run the following command to automatically fix spotless violations:
```

fineract-provider/build.gradle
```diff
@@ -48,6 +48,7 @@ buildscript {
         classpath "gradle.plugin.com.github.andygoossens:gradle-modernizer-plugin:1.3.0"
         classpath "gradle.plugin.com.gorylenko.gradle-git-properties:gradle-git-properties:2.2.2"
         classpath "net.ltgt.gradle:gradle-errorprone-plugin:1.1.1"
+        classpath "com.diffplug.spotless:spotless-plugin-gradle:4.0.1"
```
vorburger: do you want to switch to using the latest already in this PR? (Renovate would probably propose to automatically upgrade it soon anyway, but why not start with the very latest here?)

```suggestion
        classpath "com.diffplug.spotless:spotless-plugin-gradle:4.1.0"
```

README.md
```diff
@@ -195,10 +195,19 @@ Instructions to run Apache RAT (Release Audit Tool)
...
-Checkstyle
+Checkstyle and Spotless
 ============
 
-This project enforces its code conventions using [checkstyle.xml](fineract-provider/config/checkstyle/checkstyle.xml).  It is configured to run automatically during the normal Gradle build, and fail if there are any style violations detected.
+This project enforces its code conventions using [checkstyle.xml](fineract-provider/config/checkstyle/checkstyle.xml) through Checkstyle and [google-checks](config/google-checks-updated.xml) through Spotless. They are configured to run automatically during the normal Gradle build, and fail if there are any violations detected. You can run the following command to automatically fix spotless violations 
+
+    `./gradlew spotlessApply`
+
+Since some checks are present in both Checkstyle and Spoltess, the same command can help you fix some of the checkstyle violations, but other checkstyle violations need to fixed manually.
```
vorburger: typo (Spoltess/Spotless) + minor style improvement suggestion

```suggestion
Since some checks are present in both Checkstyle and Spotless, the same command can help you fix some of the Checkstyle violations (but not all, other Checkstyle violations need to fixed manually).
```

README.md
```diff
@@ -195,10 +195,19 @@ Instructions to run Apache RAT (Release Audit Tool)
...
+
+You can run the following command to check for Spotless violations
```
vorburger: Perhaps useful to re-clarify & repeat here what I know you basically already wrote above, but just to make it extra clear:

```suggestion
You can also check for Spotless violations (only; but normally don't have to, because the regular build full already includes this anyway):
```

vorburger (changes requested):
@thesmallstar OK only _now_ I actually **can** properly review this... :smile: So, couple of points:

First of all, again, if I haven't made it clear before - this is really cool! Great job.

1. I have some smaller feedback, which I've provide "inline".

2. I don't understand your `config/fineractdev-eclipse-preferences.xml` -VS- `config/google-checks-updated.xml` ... so they're both (XML representations of) Eclipse Preferences (*.epf), and both are `<profile kind="CodeFormatterProfile"`, and seem to contain many similar entries (although one is 365 lines, and the other one is only 337 lines). Seeing how in `fineract-provider/build.gradle` you're using `eclipse().configFile '../config/google-checks-updated.xml'` I gathered that's the one used? Or both? With duplicate entries? Did you perhaps just forget to delete the `config/fineractdev-eclipse-preferences.xml` and we should have only one of them, or am I missing something about how this works?

3. I think I understand what you did with https://github.com/diffplug/spotless/blob/master/ECLIPSE_SCREENSHOTS.md, but I see a problem I'd like us to address before merging this PR... aren't "my" (old) config/fineractdev-eclipse-preferences.epf and "your" (new) at least partially "redundant", now? What did "mine" have additionally that "yours" does not include? It seems to be that if you have to split the old one up into a new XML format, into 1 or N (?) new files, then fine - but in that case, you need to "switch" (and update the respective documentation...), and not "duplicate" it. Otherwise this will be confusing to maintain in the future. Hope you agree? As far as I can tell, "mine" includes not just the Formatter, but also a few other things I had spent some time on fine-tuning when I had created #704 in January, notably some `jdt.core.compiler` and `org.eclipse.jdt.ui/cleanup` and a few other such settings. Note that there is a huge XML line hidden embedded inside the `/instance/org.eclipse.jdt.ui/org.eclipse.jdt.ui.cleanupprofiles` and `/instance/org.eclipse.jdt.ui/org.eclipse.jdt.ui.formatterprofiles`. Could you not also similarly export those as XML, and ditch the *.epf?

4. Combining 2. & 3. I'm now trying to understand how this is "Google Code Format" specific? Couldn't we just keep "my" (old) config/fineractdev-eclipse-preferences.epf, saved as XML, and have Spotless apply that? Or did that not work? What I'm really asking I guess is where did you take `config/google-checks-updated.xml` from, and why do we even need a new one?

5. https://github.com/diffplug/spotless/blob/master/ECLIPSE_SCREENSHOTS.md mentions you should be able to Export the Organize Imports. That would seem like Spotless could also fully automate what I had done in #704 in January, which is part of "my" (old) config/fineractdev-eclipse-preferences.epf, using the `importOrderFile` (best with `removeUnusedImports`) mentioned on https://github.com/diffplug/spotless/tree/master/plugin-gradle#applying-to-java-source. Is that something you would be willing to add here already? I feel like this would be really useful for the project notably for anyone not using Eclipse.

6. use `lineEndings 'UNIX'` as per https://github.com/diffplug/spotless/tree/master/plugin-gradle#line-endings-and-encodings-invisible-stuff to match what we're already enforcing in `checkstyle.xml` anyway. Having Spotless do this conversion will help people contribute in the future, I feel.

7. While we're at it, let's not just use this for Java, but for other files as well? Something like this seems super useful, to me (if it works, not tested; you may have to tweak it, a little bit):

```
exclude '**/build/**', '**/bin/**' // as in .gitignore

format 'misc', {
    target '**/*.md', '**/*.properties', '**/*.gradle', '**/.gitignore', '**/*.yml', '**/*.xml', '**/**.json', '**/*.sql'

    trimTrailingWhitespace()
    indentWithSpaces(4)
    endWithNewline()
  }```

**Comment by thesmallstar** [comment]
For 2,3,4:
I tried to see how many checkstyles can be automatically added through the generated XML file from your ".epf", it turned out space-related checkstyles were not automatically included. Then I found google-checks, this is just what google-java-formatter would do, and it happened to help automatically add the space-related check styles, I probably came to an incorrect conclusion of this being a bigger set and happen to use google-checks.

If I am not wrong "epf" includes all preferences(also the formatting ), but XML only has formatting settings (which is required by formator). They are not interchangeable, we need to have them both. 
And spotless won't support the epf, it only takes XML as an argument hence I made it.
Edit: preferences cannot be exported as XML, and spotless only wants .XML (hence this extra step was needed

google-checks vs our-old-xml
->  Plan update, I will go around and stick to the generated XML from epf, and then add more tags as required to automate the checks from checkstyle. I was taking a very lazy approach earlier :P 

I need to see more for point 3 and removing .epf, would update on that.

For 5,6,7
Good idea, working on it :)

**Comment by thesmallstar** [comment]
if anyone already knows about this (@xurror @awasum  @vorburger )
I am trying to do a git reset (soft), and the indentation changes in build.gradle won't appear, but, they appear on Github. 
Is this some issue with space and tab? Or I am missing something oblivious here?

File: fineract-provider/build.gradle (hunk 2 above)
Comment (thesmallstar) [reply fineract-provider/build.gradle]: yes :)

**Comment by xurror** [comment]
Maybe the reset is actually not happening.

**Comment by vorburger** [comment]
@thesmallstar I have full faith in you that you'll be able to figure out your `git reset` issues! :smiley_cat: 

> If I am not wrong "epf" includes all preferences(also the formatting ), but XML only has formatting settings (which is required by formator). They are not interchangeable, we need to have them both.
And spotless won't support the epf, it only takes XML as an argument hence I made it. Edit: preferences cannot be exported as XML, and spotless only wants .XML (hence this extra step was needed

Understood, but https://github.com/diffplug/spotless/blob/master/ECLIPSE_SCREENSHOTS.md gives me the impression that x1 EPF can be exported as xN XML - one for formatting, another one for Organize Imports, right? Let's do that. One that the Spotless doc doesn't show is for _Preferences > Java > Code Style > Clean Up_ (which in the EPF I've extensively customized for Fineract) - that's your 3rd XML.  You can ignore Code Templates (because I have not customized those; although it could be nice to have them with the License header... but that's separate).

Once you have done that, for any other Preferences see what you can spot in the EPF that is still not in the XML. What we should do IMHO, to avoid "redundancy", is just to manually edit and strip out anything from the EPF that you have put into XML, and only keep a very small EPF! At least 2 things which I suspect will be left that I see (in Eclipse) are the _Preferences > Java > Compiler > ..._ settings, as well as the _Preferences > Java > Editor > Save Actions_. Those are great, and in January I have extensively customized those for Fineract. [It looks like there is no easy way to export those](https://stackoverflow.com/questions/22891425/eclipse-importexport-actions-in-save-actions-option) (link), so just keep anything related to that in a manually edited small EPF.

PS: Here's an idea for a future follow-up, not for this PR: I think the right way to eventually replace what will be left in an EPF is to try to put it all into a Gradle configuration. But reading https://github.com/gradle/gradle/issues/856, it looks like this is not (easily) possible, yet... :sob: An alternative in the mean time may be to simply switch from workspace global properties in the EPF to actually just have `fineract-provider/.settings/org.eclipse.jdt.core.prefs` in Git (instead of `.gitignore`) instead of using EPF, assuming that it contain what will be left in the EPF. I can make that switch later, once we are done with this PR (unless you want to already do it as part of this here).

Hope this makes sense and helps and is not too confusing.

**Comment by thesmallstar** [comment]
Any particular reason why you removed the import order from .epf? (just curious)

**Comment by thesmallstar** [comment]
@vorburger please review two things:
1. What I removed from .epf is correct(doubtful about end part)
2. Did I use the wrong import order? I found /instance/org.eclipse.jdt.ui/org.eclipse.jdt.ui.importorder=java;javax;org;com;

It seems something is wrong, should we be doing it alphabetically(Also as in checkstyle) , 
and also consider ->
import io.* type too?

**Comment by vorburger** [comment]
> What I removed from .epf is correct(doubtful about end part)

I'm struggling to (easily) see the diff in the .epf, because you've included all changed .java files again (by mistake?), can you remove those, and keep that local, to make both my review and the eventual merge easier? That way, whenever we'll be good here, I'll just run it myself and immediately merge it - thus avoid conflicts. Makes sense?

BTW are you also going to remove `config/google-checks-updated.xml` ? We agreed that we don't care about that, but want to just migrate the (my) Fineract specific EPF to XML - right?

> Did I use the wrong import order? I found /instance/org.eclipse.jdt.ui/org.eclipse.jdt.ui.importorder=java;javax;org;com; It seems something is wrong, should we be doing it alphabetically(Also as in checkstyle) , and also consider -> import io.* type too?

Yes you are right we want pure alphabetical import order, no grouping or java first... but I'm confused, where did you find `org.eclipse.jdt.ui.importorder=java;javax;org;com` ? https://github.com/apache/fineract/blob/develop/config/fineractdev-eclipse-preferences.epf#L463 has `/instance/org.eclipse.jdt.ui/org.eclipse.jdt.ui.importorder=;` and https://github.com/apache/fineract/blob/develop/fineract-provider/config/checkstyle/checkstyle.xml#L64 has `<module name="CustomImportOrder"> <property name="thirdPartyPackageRegExp" value=".*"/> <property name="sortImportsInGroupAlphabetically" value="true"/>` which to me seems right, that's what we want - agreed?

**Comment by thesmallstar** [comment]
@vorburger sorry I don't understand what do you mean by java files? 
I see 6 file changes and none of them are .java?

**Comment by thesmallstar** [comment]
Also, I don't see Google-checks? You probably didn't pull the new changes?
Currently looking into import order.

File: config/importConfig.importorder [hunk 5]
```diff
@@ -0,0 +1,6 @@
+#Organize Import Order
+#Thu Jun 04 03:11:49 IST 2020
+0=java
+1=javax
+2=org
+3=com
```
vorburger: Nope, we do NOT want this, agreed? See longer comment in main discussion thread.

fineract-provider/build.gradle
```diff
@@ -192,6 +194,27 @@ jacocoTestReport{
     }
 }
 
+spotless {
+  format 'misc', {
+   target '**/*.md', '**/*.properties', '**/*.gradle', '**/.gitignore', '**/*.yml', '**/*.xml', '**/**.json', '**/*.sql'
+   targetExclude '**/build/**', '**/bin/**'
+   indentWithSpaces(4)
+   trimTrailingWhitespace()
```
vorburger: could `trimTrailingWhitespace()` + `endWithNewline()` be moved "out" of 'misc' and 'java' at the start (or end) of the spotless configuration, together with `lineEndings`, so that it applies to everything? Or is that not how this configuration works?

vorburger (changes requested):
@thesmallstar huh, yeah, somehow my browser cached something?! Ignore me, I can see it better now.

> What I removed from .epf is correct(doubtful about end part)

What's your doubt? It looks great to me!

Just some other more minor feedback, now. Looks like we're getting there!

fineract-provider/build.gradle
thesmallstar: I already tried doing that, it won't let us do it :(

config/importConfig.importorder
thesmallstar: yes!

thesmallstar:
@vorburger LGTY?

thesmallstar:
After we merge this we can basically automate:
Indentation, GenericWhitespace,MethodParamPad, OperatorWrap,WhitespaceAround, LeftCurly,RightCurly, LineLength,SeparatorWrap. 9 Checkstyles, and only 10-12 checkstyles will remain mostly related to naming conventions. A great win isn't it?

xurror:
Definitely. This thing rules. Great work
Travis is still failing though, some violations.

fineract-provider/build.gradle
vorburger: OK, resolving.

config/importConfig.importorder
vorburger: done (replaced by `importOrder() //sort imports alphabetically`)

vorburger:
@thesmallstar While exploring actually using this myself a little bit, to verify that I can actually run `./gradlew spotlessApply` myself, that after it ran `./gradlew spotlessCheck` is happy, AND that both a full `./gradlew build integrationTest` still worked (which it currently actually does not, but a rebase to grab #990 did the trick), I was blown away about how fast Spotless was... just 2/20s to reformat almost 2900 files - holy crap! :airplane: 

One thing that I ran into and which surprised me and that could be a minor productivity PITA is that, apparently, spotless, out of the box, runs at the end, after Java build? I found that a little surprising, as I would have expected a source code quality (formatting) tool to run BEFORE Gradle compiles Java. It would make the "turn-around" while developing slightly more efficient. If that's easy enough to change, perhaps it would be worth to still do that as part of this PR? Otherwise forget about it.

I've also just realized that we forgot something (minor) here.. now that you've split the EPF, the respective paragraph in the README still needs to be updated to say that a developer must now manually import `config/fineractdev-cleanup.xml` and `config/fineractdev-formatter.xml` as well, agreed? You absolutely and totally should also explain in this doc update that the very same formatter XML one can use in the (Eclipse) IDE is also used by Spotless to BOTH validate the formatting AND to autoformat on the CLI - because that's actually really really cool, and very much worth pointing out explicitly.

>  Travis is still failing though, some violations.

@xurror that's totally expected, here... :smile_cat: I'll handle this.

The question now becomes when exactly we want to merge this (with a second commit, which I'll make, which will be the massive reformat). I think this kind of thing is "big enough" that an announcement, with a bit of time given, is appropriate. I'll post something to the dev list ASAP.

thesmallstar:
@vorburger  should I add more formatting tags in the XML? So that we don't have to make many PRs with the lot of file changes?
By adding more tags I mean the one that will be useful for enforcing checkstyles.

vorburger:
> The question now becomes when exactly we want to merge this
>  I'll post something to the dev list ASAP.

As per email thread "Massive Fineract automated code reformatting (FINERACT-1006)", I'm going to do this in the coming hours. I'll attempt to merge any open passing PRs BEFORE doing the bulk reformat. FYI @thesmallstar @ptuomola @awasum @xurror @percyashu 

> should I add more formatting tags in the XML? 

Please don't touch it anymore now, and put follow-up in future new PRs after merge.

fineract-provider/build.gradle
```diff
@@ -450,6 +472,7 @@ modernizer {
 
 compileJava{
     dependsOn rat
+    dependsOn spotlessCheck
 }
```
thesmallstar, line 476: @vorburger  FYI, I added this to the commit(on Friday) for making spotless run before compile java as you suggested. The other change is the readme(explaning about .XML). Everything else is the same.

vorburger: Great, thanks. FYI I'm going to be merging this shortly (started working on it right now).

vorburger:
==> #1055.

PR #946, opened 2020-05-28
fix broken build due to @Deprecated error [FINERACT-1012]

vorburger:
FINERACT-1012


vorburger:
@awasum @xurror I've taken the liberty to "self merge" this to fix broken build; details are recorded in FINERACT-1012.

PR #949, opened 2020-05-28
FINERACT-933 Fixed ArrayIndexOutOfBoundsException at ClientPersonImportHandler

nnatarajan:
## Description
It appears that the exception may have been caused by the formatting of dashes in which the code expects a dash, but the input doesn't contain dashes or the possibility that the spreadsheet has an empty cell. If so, then there would be a similar or the same exception for the logic directly below it.

FYI -- Due to recent changes in the file, the exception moved down a few lines from 149 to 154.

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/bulkimport/importhandler/client/ClientPersonImportHandler.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/bulkimport/importhandler/client/ClientPersonImportHandler.java
@@ -105,7 +105,7 @@ public class ClientPersonImportHandler implements ImportHandler {
         Long clientTypeId = null;
         if (clientType!=null) {
             String[] clientTypeAr = clientType.split("-");
-            if (clientTypeAr[1] != null) {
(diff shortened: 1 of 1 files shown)


vorburger (changes requested):
as we discussed, I think there's an even better way to fix this problem

nnatarajan:
@vorburger Should just the logic be changed or should error handling be
included as well?

vorburger:
> @vorburger Should just the logic be changed or should error handling be included as well?

not entirely sure what you mean, but remember how you spotted that there is several places in this class which could easily have the same ArrayIndexOutOfBoundsException and how I said that instead of `isEmpty()` you can make the code more robust by checking for the condition which actually causes an `ArrayIndexOutOfBoundsException` ? So just "hardening" the code in this class, in all places where this can happen, would fix FINERACT-933, which is the goal here. I think think there's any expectation to add more sophisticated additional error handling.

vorburger:
@nnatarajan you have to rebase this to get rid of the `RecurringDepositTest` it ran into.

vorburger (approved):
LGTM

PR #951, opened 2020-05-29
FINERACT-822 Enable Warnings CompareToZero

percyashu:
FINERACT-822

### DIFF
--- a/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/CenterDomain.java
+++ b/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/CenterDomain.java
@@ -231,7 +231,7 @@ public class CenterDomain implements Comparable<CenterDomain> {
         }
         if (this.status != null) {
             if ((Double) this.status.get("id") >= 0) {
-                hash += (Double) this.status.get("id");
+                hash = (int) (hash + (Double) this.status.get("id"));
             }
             if ((String) this.status.get("code") != null) {
                 hash += this.status.get("code").hashCode();
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/interestratechart/data/InterestRateChartSlabDataValidator.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/interestratechart/data/InterestRateChartSlabDataValidator.java
@@ -154,7 +154,7 @@ public class InterestRateChartSlabDataValidator {
         }
 
         if (amountRangeFrom != null && amountRangeTo != null) {
-            if (amountRangeFrom.compareTo(amountRangeTo) > 1) {
+            if (amountRangeFrom.compareTo(amountRangeTo) > 0) {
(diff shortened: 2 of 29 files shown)


percyashu:
@vorburger, @awasum, @xurror LGTY?

xurror (commented):
I'm seeing you are enabling many checks at once. I would propose you instead raise small PRs for each of them. It will become a real PITA to review this if to files and changes done here. Is that OK for you?

percyashu:
OK will revert to the first commit

fineract-provider/src/main/java/org/apache/fineract/portfolio/interestratechart/data/InterestRateChartSlabDataValidator.java
```diff
@@ -239,7 +239,7 @@ public void validateChartSlabsUpdate(final JsonElement element, final DataValida
         }
 
         if (amountRangeFrom != null && amountRangeTo != null) {
-            if (amountRangeFrom.compareTo(amountRangeTo) > 1) {
+            if (amountRangeFrom.compareTo(amountRangeTo) > 0) {
```
thesmallstar, line 242: Just curious of this change, @percyashu can you please tell why they mean the same thing?

percyashu: Always comparing to 0 is the safest use of the return value when using compareTo since the magnitude of the value return is not certain but the sign is. See https://errorprone.info/bugpattern/CompareToZero

vorburger: This is actually "fun"... I also had to read https://errorprone.info/bugpattern/CompareToZero - and now I understand it. What's really interesting and not just fun is that in N places of this PR you're just doing something to make EP happy, but in exactly 4 occurrences, you've actually just fixed real bugs - see FINERACT-1014; thank You, Error Prone!

vorburger:
The build failure here was "just" due to the (very interesting!) finding of @ptuomola in FINERACT-885.

FINERACT-1014 explains how this PR isn't only "nice", but actually fixes real bugs in API validation .

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/CenterDomain.java
```diff
@@ -231,7 +231,7 @@ public int hashCode() {
         }
         if (this.status != null) {
             if ((Double) this.status.get("id") >= 0) {
-                hash += (Double) this.status.get("id");
+                hash = (int) (hash + (Double) this.status.get("id"));
```
vorburger, line 234: @percyashu ==> https://github.com/apache/fineract/pull/954

fineract-provider/src/main/java/org/apache/fineract/portfolio/interestratechart/data/InterestRateChartSlabDataValidator.java
thesmallstar: @vorburger @percyashu this is interesting!

PR #953, opened 2020-06-01
minor clean-up to eliminate 'dead code' in CodeValue

vorburger:
This fixes one of too many ;) warnings Eclipse shows.

It is correct that the condition this checked is useless,
because that is already checked for in the method it calls.
(BTW it's actually pretty impressive that JDT, the Eclipse
Java compiler, can figure this out; it's "smarter" than javac).


vorburger:
@xurror would you like to review and merge this one?

xurror (approved):
LGTM

PR #956, opened 2020-06-01
executeAndAwaitJob() !executeJob() ClientLoanIntegrationTest un-ignore

vorburger:
re. FINERACT-885 and FINERACT-922


vorburger:
@ptuomola how about this - LGTY? I'm not 100% sure this isn't still flaky, but it's worth to try...

ptuomola:
Looks good to me!

vorburger:
@awasum @xurror does this one LGTY, merge?

PR #958, opened 2020-06-01
executeAndAwaitJob() !executeJob() ClientSavingsIntegrationTest

vorburger:
re. FINERACT-852 and FINERACT-922


vorburger:
@ptuomola how about this - LGTY? I'm not 100% sure this isn't still flaky, but it's worth to try...

ptuomola:
Looks great to me. Even if there are still other reasons why this test fails, I'm sure this is a step in the right direction...

vorburger:
@awasum @xurror does this one LGTY, merge?

xurror (approved):

PR #961, opened 2020-06-01
FINERACT-822 enable ArgumentSelectionDefectChecker warning

percyashu:
FINERACT-822

### DIFF
--- a/fineract-provider/build.gradle
+++ b/fineract-provider/build.gradle
@@ -324,7 +324,6 @@ tasks.withType(JavaCompile) {
                 "BadImport",
                 "UnusedMethod",
                 "ModifiedButNotUsed",
-                "ArgumentSelectionDefectChecker",
                 "JodaPlusMinusLong",
                 "TypeParameterUnusedInFormals",
                 "DoubleBraceInitialization",
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/creditbureau/service/CreditBureauReadPlatformServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/creditbureau/service/CreditBureauReadPlatformServiceImpl.java
@@ -58,7 +58,7 @@ public class CreditBureauReadPlatformServiceImpl implements CreditBureauReadPlat
             final String cbSummary = rs.getString("cbSummary");
             final long implementationKey = rs.getLong("implementationKey");
 
-            return CreditBureauData.instance(id, name, product, country, cbSummary, implementationKey);
+            return CreditBureauData.instance(id, name, country, product, cbSummary, implementationKey);
 
         }
     }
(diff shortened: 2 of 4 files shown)


vorburger:
@awasum will you review this one?

fineract-provider/build.gradle
```diff
@@ -324,7 +324,6 @@ tasks.withType(JavaCompile) {
                 "BadImport",
                 "UnusedMethod",
                 "ModifiedButNotUsed",
-                "ArgumentSelectionDefectChecker",
```
vorburger, line 327: @percyashu I see on https://errorprone.info/bugpattern/ArgumentSelectionDefectChecker that by default this is just a Warning. Given that this seems to fix real bugs (which you are seem to be proposing to fix in this PR? I have not looked closely at each...), should we not configure this, and every other currently disabled check as you go along, to actually be an Error instead of just a Warning?

percyashu:
@vorburger I thought so too but since compiler warning are set to errors with "-Werror", it will still fail if picked up. Nevertheless I can still move it to the error check list.

awasum (approved):
Woow..thanks to Error Prone. We are actually catching these types of bugs. Humans might not have caught this one. Wrong parameters usually with same data type were passed to methods. See more here: https://errorprone.info/bugpattern/ArgumentSelectionDefectChecker
Why has an integrationTest not caught this one for us and failed? Even the IDE missed this?

I will merge this now but @vorburger, I will allow you respond to @percyashu  after this has been merge and maybe He can send another PR.

awasum:
> @vorburger I thought so too but since compiler warning are set to errors with "-Werror", it will still fail if picked up. Nevertheless I can still move it to the error check list.

Please..when responding to a comment...make sure you respond in the right section. @vorburger gave a review feedback and its just natural for you @percyashu to respond to that feedback just below it on same Conversation not for you to reply on a separte general comment section. I dont know if am making sense @percyashu ? See
image

vorburger:
> @vorburger I thought so too but since compiler warning are set to errors with "-Werror", it will still fail if picked up. Nevertheless I can still move it to the error check list.

That's a fair point... :smile: I didn't even think of that. Here are 2 reasons why I suggest you still do:

1. it's probably "clearer" to future new developers to the project if they see Errors from EP, instead of just Warnings, which then cause the build to anyway fail - just because, by "convention", Warnings _normally_ don't cause build failure, in Java. (I'm assuming that Error Prone's reporting issues errors instead of warnings, once you explicitly list everything.)

2. how much do you want to bet that at some point, in future years, when people who are currently active on his project have moved on, and others maintain it, someone disagrees with making all warnings errors, and removes that again? (Or someone may do that in a private in-house fork.) They would then also loose all Error Prone validation which by default is just a warning, but that you are working hard on fixing and enforcing. If you make all EP checks errors, then it's more "durable" - independent of the `-Werror` thing.

Makes sense? OK for you to raise a PR adding what is in so far, and doing so going forward?

@awasum FYI

PR #964, opened 2020-06-02
FINERACT-1004 Resolved NPE in AddressWritePlatformServiceImpl.updateC…

nnatarajan:
FINERACT-1004


vorburger:
Build failure here is FINERACT-855 - please help us fix it.

vorburger:
This will have to be re-based after #968 for FINERACT-1016 for the build to pass.

nnatarajan:
rebased

vorburger (changes requested):
while this would probably fix FINERACT-1004, it would "hide" a real problem.. have a closer look at the related code - when would `clientAddressObj` be null? That should probably be handled and cause an error (throw an exception, but the right one...), with a clear message.

vorburger (approved):
LGTM

PR #960, opened 2020-06-01
FINERACT-977-fix possible NPE in AbstractLoanScheduleGenerator

maektwain:
FINERACT-977

PR #967, opened 2020-06-02
remove line in AbstractLoanScheduleGenerator [FINERACT-977]

vorburger:
This line was wrongly added in #960 re. fix possible NPE in AbstractLoanScheduleGenerator for FINERACT-977.

This has absolutely nothing to do with FINERACT-855/FINERACT-1016 (I initially confused 2 unrelated things).


vorburger:
@maektwain see new inline review comment on (already merged) #960 re. this - do you agree this was wrong?

I'm not sure if it's causing FINERACT-855 or if that's an unrelated problem.

vorburger:
@maektwain this actually has absolutely nothing to do with FINERACT-855 (now FINERACT-1016), but it still seems to me that line in your PR #960 was clearly wrong and should still be removed here (totally unrelated to current test failure).

I'll change the commit message and PR title & description to make it clear that it has nothing to do with the current test failure (it was a coincidence, and I made a connection in my mind when this was totally unrelated).

This will have to be re-based after #968 for it to pass.

maektwain:
Sure let me fix this

vorburger:
> Sure let me fix this

You don't have to fix it, I did that with this PR for you - just confirm that this LGTY, if you agree the line was wrong and this is right?

vorburger:
I've "self merged" this, assuming consent from @maektwain ... @awasum @xurror FYI.

PR #972, opened 2020-06-02
FINERACT-822 enable variableNameSameAsType warning

percyashu:
FINERACT-822


vorburger (approved):
This LGTM except for the fact that I would not remove from disable what you enable, but move anything that is just a Warning to the error section, as per our discussion in #961.

percyashu:
OK will also do that for the warnings already removed

PR #975, opened 2020-06-02
FINERACT-822  Update and add more checks to be enabled

percyashu:
FINERACT-822

### DIFF
--- a/fineract-provider/build.gradle
+++ b/fineract-provider/build.gradle
@@ -345,6 +345,24 @@ tasks.withType(JavaCompile) {
                 "TypeParameterUnusedInFormals",
                 "UnusedNestedClass",
                 "UndefinedEquals",
+                "JdkObsolete",
+                "OperatorPrecedence",
+                "EmptyBlockTag",
+                "MissingSummary",
+                "MutablePublicArray",
+                "InvalidParam",
+                "ReturnFromVoid",
+                "AlmostJavadoc",
+                "PublicConstructorForAbstractClass",
+                "ProtectedMembersInFinalClass",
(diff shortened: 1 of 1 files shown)


percyashu:
From https://github.com/apache/fineract/pull/971

vorburger:
I expect this will fail to build, as per #971. 

Just do only the adding of the checks here, not don't upgrade, in this PR (or a new one). 

Adjust title and commit message.

percyashu:
@vorburger Will do only for the checks

vorburger (approved):
LGTM (if it passes the build, obviously).

percyashu:
@vorburger am sure this will fail since the checks I added aren't available for ver 2.3.4

percyashu:
@vorburger should I update it manually

fineract-provider/build.gradle
```diff
@@ -330,13 +330,32 @@ tasks.withType(JavaCompile) {
...
+                "MissingSummary",
+                "MutablePublicArray",
+                "InvalidParam",
+                "ReturnFromVoid",
+                "AlmostJavadoc",
+                "PublicConstructorForAbstractClass",
+                "ProtectedMembersInFinalClass",
+                "StaticAssignmentInConstructor",
+                "UnescapedEntity",
+                "EmptyCatch",
```
xurror, line 359: Seems this is not a valid check, maybe a typo or maybe you meant something else and wrote something else. I can merge this once this is resolved.

percyashu:
Since this passes, #971 can be closed

xurror (commented):
I'll just merge this now. I don't see why we need to hold this on longer.

PR #977, opened 2020-06-03
distch plus and use fixed versions instead(FINERACT-1002)

xurror:
https://issues.apache.org/jira/browse/FINERACT-1002


vorburger (commented):
I thought that at least some of these dependencies would be part of the Spring BOM, and thus not need version numbers, to inherit Spring's... none of these are? @ptuomola any interest in reviewing this?

ptuomola:
@vorburger Most of those dependencies (gson, quartz etc) are indeed part of spring-boot-dependencies BOM. 

But I've just noticed we are importing into dependency management spring-framework-bom - which contains just the Spring framework and doesn't contain any of these dependencies.  

So if we switch dependency management to import spring-boot-dependencies instead, we should be able to get rid of many of these versions. In fact, shouldn't we be able to avoid mentioning them altogether in build.gradle given that the imported POM already references them?

xurror:
There is the [spring dependency BOM](https://mvnrepository.com/artifact/org.springframework/spring-framework-bom/5.2.6.RELEASE) but only conatains spring package  which we are already using.

On the other hand, there is this massive [platform bom](https://mvnrepository.com/artifact/io.spring.platform/platform-bom/Cairo-SR8) which contains alot of our dependencies if not all including spring dependencies except it's way behind us on updated packages. Using this will save us the pain of having to define dependencies in build.gradle but will also mean we would have to revert most if not all of the packages upgraded so far to the version specified in the BOM unless otherwise overwriting them which is just thesame as what we are doing now.

Or we could create our own BOM but I hardly see the point of doing that. Let me know what you think.

ptuomola:
And there is also the [Spring Boot BOM](https://mvnrepository.com/artifact/org.springframework.boot/spring-boot-dependencies/2.3.0.RELEASE) which is even bigger - but also contains a lot of stuff we don't need / want (e.g. Hibernate).

Not sure what is the best way to go here. Naming our own versions of course allows us to use the latest versions (with any fixes / security patches) and control when we upgrade them, as well as being strict about what dependencies are introduced. 

But at the same time it of course results in a lot of "avoidable" work to keep upgrading specific dependencies. Depending on your perspective, this additional effort to manage the dependencies could be seen as a good thing (providing easy issues for people to start working on) or bad thing (take focus away from  improving Fineract itself)...

vorburger:
@xurror @ptuomola I gave this some more thought... I did a "verification" by running `./gradlew dependencies`, and convinced myself that this does not "regress" any of the now fixed versions but gives us exactly those we currently have (except activemq-broker which + now resolves to 5.15.13 instead 5.15.12), and therefore decided to merge this. The reason is that since the initial discussion in FINERACT-1002, I have made progress with FINERACT-962, and that "offsets" what you discussed above - makes sense?

ptuomola:
@vorburger sounds great - and Renovate definitely seems like a very good way to solve the manual work issue!

PR #980, opened 2020-06-04
FINERACT-1019: Fix import integration tests

ptuomola:
## Description

FINERACT-1019 Import integration tests were failing and ignored for various reasons (timeout, incorrect test data set-up etc)

## Checklist

### DIFF
--- a/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/bulkimport/importhandler/client/ClientEntityImportHandlerTest.java
+++ b/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/bulkimport/importhandler/client/ClientEntityImportHandlerTest.java
@@ -47,10 +47,13 @@ import org.apache.poi.ss.usermodel.Sheet;
 import org.apache.poi.ss.usermodel.Workbook;
 import org.junit.Assert;
 import org.junit.Before;
-import org.junit.Ignore;
 import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class ClientEntityImportHandlerTest {
+    private final static Logger LOG = LoggerFactory.getLogger(ClientEntityImportHandlerTest.class);
+
     private ResponseSpecification responseSpec;
     private RequestSpecification requestSpec;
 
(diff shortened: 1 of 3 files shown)


awasum (approved):
This looks good to me. Thanks very much @ptuomola . Has this been tested with the UI? Data import has a UI component. It will be great to test this against the UI.

ptuomola:
@awasum Thanks. I don't think I've changed anything that would change the behaviour of the API from the perspective of the UI - but I can try to see if I can test this with the UI as well.

vorburger (changes requested):
@ptuomola in #680, which FYI is abandoned (original author disappeared, Intern is struggling to work through Git rebase and merge conflicts) there was an attempt to avoid the `Thread.sleep()`, which are always bad as they slow down running tests, by introducing a new `Utils.waitUntilFileCreation()` (which was badly written in that PR, we could now just use Availability for that, very similarly to how I have in the new Scheduler test util). Is this something you would consider doing here? Or would be interested in doing a follow-up PR, if we already merged this as is? If you have a chance to look over that #680 and let us know if this PR makes that one redundant (and dupe FINERACT-913 to your new FINERACT-1019), or perhaps if there are any fixes in there that you think still make sense to integrate here (so that we can abandon that one), that would be great!

ptuomola:
@vorburger ah ok - awaiting for the file to appear definitely makes sense, I'll see if I can get that to work without too much hassle and will add that to this PR. 

In terms of the rest of the PR, as far as I can see it's either already covered by this - though for a number of changes in that PR, I'm not quite sure why they are needed...

ptuomola:
@vorburger Hmm... looks like it's not that easy after all. If I've understood this correctly, the current logic works like this: 

If we call getOutputTemplateLocation() before the import processing has completed, then we get the location where Fineract has stored the import document - not the output document! Only after the processing has completed, the same API call returns the output document location - but we have no way of knowing when the processing is ready. 

Personally I think that's not a great design of an API - to put it nicely :-). I hope I've missed something, but the testing seems to confirm this behaviour. 

So I can see a couple of options: 

1. Accept the Thread.sleep of 10 seconds in the integration tests
2. Implement some logic that repeatedly calls the API and tries to guess from the filename whether the processing has completed. But that might introduce flakiness...
3. Rework the entire API logic e.g. to return 204 No content if called before the processing has completed

Personally I think #3 would be the right way to solve this, but not sure what clients will be broken by this. Maybe I should install the community app and see how that deals with this...

vorburger:
Or let's just merge this as-is already, as a slow test is better than an
ignored test, and you can subsequently improve it? - I don't mind, at all
(seeing now how it's not quite so trivial).

>

ptuomola:
@vorburger I've just checked how the frontend works - smartly it doesn't rely on getOutputTemplateLocation immediately providing the right value straightaway. Given all we are doing here is trying to save a few seconds on an integration test, I don't think changing the API behaviour is worth it. So your suggestion to merge this as-is sounds good.

There is still something wrong with the import functionality when using the community app - it doesn't always seem to work. It won't have anything to do with these changes (especially as they have not yet been merged), but I'll investigate what's wrong and raise a JIRA as required...

ptuomola:
Confirmed that import is fixed now also for the community app - so nothing further required

PR #985, opened 2020-06-05
FINERACT-821 Enforced Membername checkstyle(1)

thesmallstar:
This is more for discussion currently:

So checkstyle believe a constant is a variable that is both "static" and "final" definitely true, but it a little big arguable do we really also need a static keyword to call it constant?

MemberName check style would want something like:
final int TOTAL_SUM to be changed to final int totalSum
That is name final variables not like constantNames.

Fineract extensively uses TOTAL_SUM way of naming things for final variables.
What should we follow?
@vorburger @awasum  @xurror

### DIFF
--- a/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/TemplateIntegrationTest.java
+++ b/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/TemplateIntegrationTest.java
@@ -34,9 +34,9 @@ import org.junit.jupiter.api.Test;
 
 public class TemplateIntegrationTest {
 
-    private final String GET_TEMPLATES_URL = "/fineract-provider/api/v1/templates?tenantIdentifier=default";
-    private final String GET_TEMPLATE_ID_URL = "/fineract-provider/api/v1/templates/%s?tenantIdentifier=default";
-    private final String RESPONSE_ATTRIBUTE_NAME = "name";
+    private static final String GET_TEMPLATES_URL = "/fineract-provider/api/v1/templates?tenantIdentifier=default";
+
+    private static final String RESPONSE_ATTRIBUTE_NAME = "name";
 
     private ResponseSpecification responseSpec;
     private RequestSpecification requestSpec;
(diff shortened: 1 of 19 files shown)


vorburger:
>  final int TOTAL_SUM to be changed to final int totalSum

or you can make it `final static int TOTAL_SUM`, right? Then it's happy - and that's better.

FYI in Java, as far as I know, always making all constants `final static` is much better than only `final`, because this saves memory... a `final static` is defined only ONCE in the Class, but a `final` is stored in every INSTANCE of said Class (I think, at least that's probably how it was originally; perhaps nowadays that gets optimized?). For a Test or Service class this doesn't matter (as there is typically only 1 instance of `YourService` or `SomethingTest`), but for a "domain" class of which there are N in production at run-time, this could make a real difference. For consistency, the rule of always making all constants `static final` seems sensible, to me - and is a very widely used pseudo standard convention in Java anyway.

vorburger:
@thesmallstar will you update this before it gets auto closed because of stale?

thesmallstar:
@vorburger on it.

awasum:
Is this still important? and also merge conflicts. @thesmallstar , If you have too many issues on your plate, please take one at a time and finish them except they are too complicated?

thesmallstar:
Yes, I am trying to focus on newer issues. However this is important, I will restart the checkstyle work next week and wrap this.

thesmallstar:
failed due to connection error, closing and opening again.

thesmallstar:
@xurror  @vorburger  ready for review.

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/TemplateIntegrationTest.java
```diff
@@ -34,9 +34,9 @@
 
 public class TemplateIntegrationTest {
 
-    private final String GET_TEMPLATES_URL = "/fineract-provider/api/v1/templates?tenantIdentifier=default";
-    private final String GET_TEMPLATE_ID_URL = "/fineract-provider/api/v1/templates/%s?tenantIdentifier=default";
-    private final String RESPONSE_ATTRIBUTE_NAME = "name";
+    private static final String GET_TEMPLATES_URL = "/fineract-provider/api/v1/templates?tenantIdentifier=default";
+
```
xurror, line 38: Why remove this variable?
It was only used once though so I don't mind leaving it as is.

xurror (commented):
Looks great. Won't mind merging this.

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/TemplateIntegrationTest.java
vorburger: Yeah, only used once, and with the parameter, I would agree that the `String.format()` is EASIER to read, now.

vorburger (approved):
LGTM. The only thing that's confusing is that this actually does NOT "enforce" any new Checkstyle rule.. but I'm assuming that this is the first step, in order to?

thesmallstar:
Yes, this PR does not add any new checkstyle but enforces a part of member name checkstyle.
In this PR I have updated to support for integrationTests.

PR #987, opened 2020-06-06
FINERACT-822 enable OperatorPrecedence error

percyashu:
FINERACT-822

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/configuration/domain/GlobalConfigurationProperty.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/configuration/domain/GlobalConfigurationProperty.java
@@ -114,7 +114,7 @@ public class GlobalConfigurationProperty extends AbstractPersistableCustom {
 
         final String passwordPropertyName = "force-password-reset-days";
         if (this.name.equalsIgnoreCase(passwordPropertyName)) {
-            if (((this.enabled == true) && command.hasParameter(valueParamName) && (this.value == 0)) || ((this.enabled == true)
+            if ((this.enabled == true && command.hasParameter(valueParamName) && (this.value == 0)) || (this.enabled == true
                     && !command.hasParameter(valueParamName) && (previousValue == 0))) { throw new ForcePasswordResetException(); }
         }
 
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/configuration/domain/GlobalConfigurationProperty.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/configuration/domain/GlobalConfigurationProperty.java
@@ -114,8 +114,8 @@ public class GlobalConfigurationProperty extends AbstractPersistableCustom {
 
         final String passwordPropertyName = "force-password-reset-days";
         if (this.name.equalsIgnoreCase(passwordPropertyName)) {
-            if (this.enabled == true && command.hasParameter(valueParamName) && this.value == 0 || this.enabled == true
-                    && !command.hasParameter(valueParamName) && previousValue == 0) { throw new ForcePasswordResetException(); }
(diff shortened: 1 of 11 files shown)


awasum (approved):
This LGTM. @xurror @ptuomola , @vorburger will you like to also take a closer look at this? and merge?

xurror (approved):
I don't know about this feature. Uh.., the compiler most certainly understands this stuff without the brackets, but on our end, readability can be tough with the bracket delimiters. But I'm OK merging this, it can't be a very big deal. Just resolve the merge conflicts and let me know when build passes.

percyashu:
@xurror done

fineract-provider/src/main/java/org/apache/fineract/infrastructure/configuration/domain/GlobalConfigurationProperty.java
```diff
@@ -114,8 +114,8 @@ public Date getDateValue(){
 
         final String passwordPropertyName = "force-password-reset-days";
         if (this.name.equalsIgnoreCase(passwordPropertyName)) {
-            if (this.enabled == true && command.hasParameter(valueParamName) && this.value == 0 || this.enabled == true
-                    && !command.hasParameter(valueParamName) && previousValue == 0) { throw new ForcePasswordResetException(); }
+            if (((this.enabled == true) && command.hasParameter(valueParamName) && (this.value == 0)) || ((this.enabled == true)
+                    && !command.hasParameter(valueParamName) && (previousValue == 0))) { throw new ForcePasswordResetException(); }
```
vorburger: Wait, this looks weird to me, shouldn't this just be:

```suggestion
            if ((this.enabled == true && command.hasParameter(valueParamName) && this.value == 0) || (this.enabled == true
                    && !command.hasParameter(valueParamName) && previousValue == 0)) { throw new ForcePasswordResetException(); }
```

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
```diff
@@ -2384,7 +2384,7 @@ public ChangedTransactionDetail disburse(final AppUser currentUser, final JsonCo
          **/
 
         if (isNoneOrCashOrUpfrontAccrualAccountingEnabledOnLoanProduct()
-                        && (isMultiDisburmentLoan() && getDisbursedLoanDisbursementDetails().size() == 1 || !isMultiDisburmentLoan())) {
+                        && ((isMultiDisburmentLoan() && getDisbursedLoanDisbursementDetails().size() == 1) || !isMultiDisburmentLoan())) {
```
vorburger: Wouldn't this be better & clearer like this, instead:

    if ((isNoneOrCashOrUpfrontAccrualAccountingEnabledOnLoanProduct()
        && isMultiDisburmentLoan() && getDisbursedLoanDisbursementDetails().size() == 1) || !isMultiDisburmentLoan()) {

vorburger (changes requested):

xurror:
Still some merge conflicts and maybe also apply suggested changes.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
percyashu: OK will try this but the parentheses grouping might change the logic of the check

fineract-provider/src/main/java/org/apache/fineract/infrastructure/configuration/domain/GlobalConfigurationProperty.java
percyashu: yeah will modify this

xurror:
Good work.

PR #988, opened 2020-06-06
FINERACT-822 enable ReferenceEquality error

percyashu:
FINERACT-822


xurror:
resolve conflicts.

vorburger:
> resolve conflicts.

@percyashu a small tip re. this, similar to what I told @thesmallstar in another PR: If you "cheat" just a little bit, and put new checks not always as last ones, but a little further up or down, then you can easily avoid git merge conflicts in this particular case - do you know what I mean?

vorburger (approved):
Jeeeeeeeeeeeeeeeeeezus - these are REAL BUGS - how did Fineract ever work?! LOL.

Thank You, Error Prone. @awasum you can tweet again! :smiley_cat:

fineract-provider/src/main/java/org/apache/fineract/infrastructure/bulkimport/importhandler/loan/LoanImportHandler.java
```diff
@@ -431,7 +431,7 @@ private Integer importDisbursalData(CommandProcessingResult result, int rowIndex
             String linkAccountId = disbusalData.getLinkAccountId();
             GsonBuilder gsonBuilder = new GsonBuilder();
             gsonBuilder.registerTypeAdapter(LocalDate.class, new DateSerializer(dateFormat));
-            if (linkAccountId != null && linkAccountId != "") {
+            if (linkAccountId != null && !"".equals(linkAccountId)) {
```
thesmallstar, line 434: I am late again :( 
@percyashu  @vorburger I haven't seen "".equals being used before, should we not use linkAccountId.isEmpty() instead? Although it seems they mean the same thing.

vorburger: that would indeed be a little nicer, yes (I merged it anyway just to move forward, and because it was a "correct" fix in the "mechanical" sense)

PR #990, opened 2020-06-06
FINERACT-821 Added and Enforced EmptyBlock Checkstyle

thesmallstar:
I had doubt on one of the change If you can review that?
I thought maybe we can log that the response code received is not 200?
Or will you recommend to just delete it?

I have removed LITERAL_CATCH since it is already covered in emptyCatchBlock checkstyle.

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/campaigns/sms/service/SmsCampaignDropdownReadPlatformServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/campaigns/sms/service/SmsCampaignDropdownReadPlatformServiceImpl.java
@@ -30,6 +30,7 @@ import org.apache.fineract.infrastructure.campaigns.sms.constants.SmsCampaignEnu
 import org.apache.fineract.infrastructure.campaigns.sms.constants.SmsCampaignTriggerType;
 import org.apache.fineract.infrastructure.campaigns.sms.data.SmsProviderData;
 import org.apache.fineract.infrastructure.core.data.EnumOptionData;
+import org.apache.fineract.infrastructure.core.exception.PlatformDataIntegrityException;
 import org.apache.fineract.portfolio.calendar.domain.CalendarWeekDaysType;
 import org.apache.fineract.portfolio.calendar.service.CalendarEnumerations;
 import org.apache.fineract.portfolio.common.domain.PeriodFrequencyType;
@@ -70,19 +71,18 @@ public class SmsCampaignDropdownReadPlatformServiceImpl implements SmsCampaignDr
     public Collection<SmsProviderData> retrieveSmsProviders() {
         Collection<SmsProviderData> smsProviderOptions = new ArrayList<>();
         String hostName = "" ;
-        try {
             Map<String, Object> hostConfig = this.smsConfigUtils.getMessageGateWayRequestURI("smsbridges", null);
             URI uri = (URI) hostConfig.get("uri");
             hostName = uri.getHost() ;
             HttpEntity<?> entity = (HttpEntity<?>) hostConfig.get("entity");
             ResponseEntity<Collection<SmsProviderData>> responseOne = restTemplate.exchange(uri, HttpMethod.GET, entity,
                     new ParameterizedTypeReference<Collection<SmsProviderData>>() {});
+             if (!responseOne.getStatusCode().equals(HttpStatus.OK)) {
+                throw new PlatformDataIntegrityException("error.msg.mobile.service.provider.not.available",
+                "Mobile service provider not available.");
+             }
             smsProviderOptions = responseOne.getBody();
-            if (!responseOne.getStatusCode().equals(HttpStatus.OK)) {
-            }
-        } catch (Exception e) {
-        }
-        return smsProviderOptions;
+            return smsProviderOptions;
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


fineract-provider/src/main/java/org/apache/fineract/infrastructure/campaigns/sms/service/SmsCampaignDropdownReadPlatformServiceImpl.java
```diff
@@ -78,8 +78,9 @@ public SmsCampaignDropdownReadPlatformServiceImpl(final SmsConfigUtils smsConfig
             ResponseEntity<Collection<SmsProviderData>> responseOne = restTemplate.exchange(uri, HttpMethod.GET, entity,
                     new ParameterizedTypeReference<Collection<SmsProviderData>>() {});
             smsProviderOptions = responseOne.getBody();
-            if (!responseOne.getStatusCode().equals(HttpStatus.OK)) {
-            }
+            //Better approch? Maybe log/throw exception that this happen?
```
vorburger: let's throw some suitable exception in this case, don't just log it (and certainly not just ignore it, like the current bad code)

fineract-provider/src/main/java/org/apache/fineract/infrastructure/campaigns/sms/service/SmsCampaignDropdownReadPlatformServiceImpl.java
```diff
@@ -78,8 +78,9 @@ public SmsCampaignDropdownReadPlatformServiceImpl(final SmsConfigUtils smsConfig
...
+            // if (!responseOne.getStatusCode().equals(HttpStatus.OK)) {
+            // }
         } catch (Exception e) {
         }
```
vorburger: while you are already here anyway, let's get rid of this overly broad ` } catch (Exception e) {` as well? see if you can just remove it, or if not, make it catch a more specific exception, and then re-throw that. No need to log it if you rethrow. As per https://github.com/apache/fineract#error-handling-guidelines

vorburger (changes requested):

thesmallstar:
Will update this now :)

thesmallstar:
I have picked the exception to be thrown from another function which calls the current function.
It throws the same exception if the request fails or (we receive an empty response).

thesmallstar:
From the guidelines -> 

> Either way, including the root cause by using catch (SomeException e) and then either ..

The root cause, in this case, is not an exception (If I am not wrong), we are checking the response code and then throwing an exception. It won't make much sense, in this case, to rethrow the same exception twice? or any other exception. 

I hope this is a correct way

fineract-provider/src/main/java/org/apache/fineract/organisation/teller/service/TellerManagementReadPlatformServiceImpl.java
```diff
@@ -261,8 +261,10 @@ private String getTellerCriteria(final String sqlSearch, final Long officeId, fi
             extraCriteria.append(" and status = 300 ");
         } else if (status.equalsIgnoreCase("inActive")) {
             extraCriteria.append(" and status = 0 ");
-        } else if (status.equalsIgnoreCase("all")) {} else {
+        } else {
+            if (!status.equalsIgnoreCase("all")){
```
vorburger, line 265: Why changing this? Just curious..

fineract-provider/src/main/java/org/apache/fineract/infrastructure/campaigns/sms/service/SmsCampaignDropdownReadPlatformServiceImpl.java
```diff
@@ -70,19 +71,18 @@ public SmsCampaignDropdownReadPlatformServiceImpl(final SmsConfigUtils smsConfig
...
         Collection<SmsProviderData> smsProviderOptions = new ArrayList<>();
         String hostName = "" ;
-        try {
             Map<String, Object> hostConfig = this.smsConfigUtils.getMessageGateWayRequestURI("smsbridges", null);
             URI uri = (URI) hostConfig.get("uri");
             hostName = uri.getHost() ;
             HttpEntity<?> entity = (HttpEntity<?>) hostConfig.get("entity");
             ResponseEntity<Collection<SmsProviderData>> responseOne = restTemplate.exchange(uri, HttpMethod.GET, entity,
                     new ParameterizedTypeReference<Collection<SmsProviderData>>() {});
+             if (!responseOne.getStatusCode().equals(HttpStatus.OK)) {
```
vorburger, line 80: I'm not 100% sure if PlatformDataIntegrityException is "correct" here, but don't really have a better idea without more closely than I have time for, and it probably doesn't really matter all that much, anyway. One fine day we should look into and then write up Guidance in the README about which exact Fineract Exception type to use in which "layer" (e.g. *Resource VS *Service). @ptuomola I doubt that's something you would fancy digging into?

fineract-provider/src/main/java/org/apache/fineract/organisation/teller/service/TellerManagementReadPlatformServiceImpl.java
thesmallstar: Else if block is empty for condition (status.equalsIgnoreCase("all")), this is not supported.
But as else if block works, we cannot simply remove it, or even if the (status.equalsIgnoreCase("all")) value is true it will go in else.

We don't want that, so I have added a extra check in else block to make sure (status.equalsIgnoreCase("all")) is not true, and only then perform else action that what we received is incorrect and act on it.

vorburger: got it, thanks for clarifying!

fineract-provider/src/main/java/org/apache/fineract/infrastructure/campaigns/sms/service/SmsCampaignDropdownReadPlatformServiceImpl.java
vorburger: this was done

vorburger (approved):
LGTM! PS: There was post on the dev mailing list today, subject "Re: Message gateway configuration", that I have a... "hunch" may actually be related to what we've been able to fix here! Good stuff.

PR #983, opened 2020-06-05
FINERACT-822 Enabling ErrorProne error checks

percyashu:
FINERACT-822

PR #997, opened 2020-06-07
remove bad 'default:' in SavingsEnumerations switch

vorburger:
This default: is "wrong", because it's not required - the switch, in this particular case, *IS* actually already handling all cases.

But if in the future someone were to add an additional new SavingsAccountTransactionType, then we would want Error Prone to fail to signal that it needs to be handled here - but the "default:" would "hide" that.

This is a follow-up to https://github.com/apache/fineract/pull/983/ for FINERACT-822, and also related to the discussion started on https://github.com/apache/fineract/pull/989.

There originally used to be a "default:" in the middle of the switch, that obviously WAS completely wrong and an old copy-paste bug.


vorburger:
@percyashu @awasum @thesmallstar @ptuomola

ptuomola:
@vorburger looks good to me!

awasum (approved):
LGTM.

PR #999, opened 2020-06-07
un-@Ignore all tests in SchedulerJobsTestResults using FINERACT-922 [FINERACT-857]

vorburger:
using FINERACT-922 for FINERACT-857

### DIFF
--- a/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/SchedulerJobsTestResults.java
+++ b/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/SchedulerJobsTestResults.java
@@ -63,7 +63,6 @@ import org.apache.fineract.portfolio.account.domain.AccountTransferType;
 import org.junit.After;
 import org.junit.Assert;
 import org.junit.Before;
-import org.junit.Ignore;
 import org.junit.Test;
 
 @SuppressWarnings({ "unused", "unchecked", "rawtypes", "static-access", "cast" })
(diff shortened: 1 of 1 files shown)


vorburger:
@ptuomola if this actually passes on Travis (I'm curious...) I'm sure you're in favour of merging this? :smile_cat: 

@xurror @awasum feel free to do the honours (if it passes the build; it worked locally, but let's see).

ptuomola:
@vorburger unfortunately failed on Travis. My guess (based on no further analysis than looking at the Travis log) is that some other tests insert crap data (missing office_id?) that then breaks the scheduled jobs. If that's the case, shouldn't be too hard to find which tests insert the broken data...

awasum:
This may not be related to this PR but travis keeps showing pending on the UI when the CI already ran. Also our language on the Travis config is showing Ruby: https://travis-ci.org/github/apache/fineract/jobs/695680352/config . This may not be important but I just saw that and was curious.

ptuomola:
@awasum I think the "ruby" in the config is due to our .travis.yml no longer having any language tag in it. We install Java manually as the default Java version that you get by specifying java as language is broken for HTTPS certificate handling. 

But instead of having no language defined (defaulting to ruby), we could try "language: minimal" or "language: generic" - those may be more appropriate?
 
And yes, I also noticed the "pending" status for Travis in Github - but no idea why that happens...

vorburger:
> My guess (based on no further analysis than looking at the Travis log) is that some other tests insert crap data (missing office_id?) that then breaks the scheduled jobs. If that's the case, shouldn't be too hard to find which tests insert the broken data...

@ptuomola I don't know when I'll have the time to properly debug this one.. if you are looking for ways to contribute, and didn't have any other plans already, and would enjoy doing this, and... (your constraints here), please by all means feel free to jump into this! I would, very gladly, let you do this.

PS: The `.travis.yml` was addressed in #1008.

ptuomola:
@vorburger I looked into this, and interestingly it wasn't a test or test data issue at all, but actually two real bugs in Fineract. 

I've created a new PR that includes both moving to executeAndAwaitJob() as well as fixes to the bugs: https://github.com/apache/fineract/pull/1049. Let's see if that passes the Travis build.

vorburger:
==> #1049

PR #1010, opened 2020-06-08
FINERACT-821 Made changes to enforce ConstantName Checkstyle (3)

thesmallstar:
If anyone is wondering if am I converting those 40char names manually in correct format-
No, I wrote a code which does that for me https://github.com/thesmallstar/CorrectMyconstant

I saw there are about 1000 variables which need to change and this PR changes about 30-40 of them :P Long way to go, but sure it will be a great change.


vorburger (approved):
LGTM

PR #1016, opened 2020-06-09
Fix ResultSet type(FINERACT-995)

xurror:
[FINERACT-995](https://issues.apache.org/jira/browse/FINERACT-995)

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/accounting/closure/service/GLClosureReadPlatformServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/accounting/closure/service/GLClosureReadPlatformServiceImpl.java
@@ -25,22 +25,22 @@ import java.util.List;
 import org.apache.fineract.accounting.closure.data.GLClosureData;
 import org.apache.fineract.accounting.closure.exception.GLClosureNotFoundException;
 import org.apache.fineract.infrastructure.core.domain.JdbcSupport;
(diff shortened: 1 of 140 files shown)


xurror:
@vorburger, are you OK with this?
Can't tell for sure if it will fix our problem.
cc: @awasum @ptuomola

ptuomola:
@xurror Unfortunately I doubt this will work: MySQL Connector does not support scrollable cursors, so setting ResultSet.TYPE_SCROLL_INSENSITIVE should have no impact. 

From MySQLConnector documentation: 
MySQL does not support SQL cursors, and the JDBC driver does not emulate them, so setCursorName() has no effect.

I think we need to look at the job that fails and see what could lead to this... and why does Spring throw a TransientDataAccessException - some concurrency issue? Underlying data was modified by another job at the same time?

ptuomola:
@xurror @vorburger - I think I found the line that causes it - added the relevant code in JIRA as a comment

vorburger:
@xurror would you be willing to retract this PR in favour of new @ptuomola #1024 ? (And can / would you like to help to co-review that new PR? Please don't merge before we have input from several people.)

vorburger (changes requested):
@xurror would you be willing to retract this PR in favour of new @ptuomola #1024 ? (And can / would you like to help to co-review that new PR? Please don't merge before we have input from several people.)

xurror:
No problem

PR #1018, opened 2020-06-09
FINERACT-822 enable ImmutableEnumChecker error

percyashu:
FINERACT-822

### DIFF
--- a/fineract-provider/build.gradle
+++ b/fineract-provider/build.gradle
@@ -367,6 +367,7 @@ tasks.withType(JavaCompile) {
                 "VarTypeName",
                 "ArgumentSelectionDefectChecker",
                 "CompareToZero",
+                "ImmutableEnumChecker",
                 "NarrowingCompoundAssignment",
                 "MissingCasesInEnumSwitch",
                 "ReferenceEquality",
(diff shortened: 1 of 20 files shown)


fineract-provider/build.gradle
```diff
@@ -436,7 +436,8 @@ modernizer {
 
     ignorePackages = []
     ignoreClassNamePatterns = [
-        '.*AbstractPersistableCustom'
+        '.*AbstractPersistableCustom',
+        '.*EntityTables'
```
percyashu, line 448: Errorprone wants all fields in Enumeration to Immutable, hence Integer[] is require to be ImmutableList<Integer> and modernizer wants all ImmutableList.of to be List.of which Error prone doesn't consider as Immutable, hence had to add EntityTables to this list.

vorburger: understood. Thanks for the explanation.

vorburger (approved):

percyashu:
@awasum @xurror @vorburger merge this?

awasum:
Squash this? Otherwise looks good.

awasum (approved):

PR #1019, opened 2020-06-09
FINERACT-821 Added and Enforced HideUtilityClassConstructor checkstyle

thesmallstar:
FINERACT-821

HideUtilityClassConstructor  ->
Utility Classes like Helperand Constants are forced to have all methods static, and constructor non-public and non-default
If we keep constructor protected we cannot create an object of that class but support inheritance if required.  The other option was to use a private constructor(which won't support inheritance). I have decided to go with protected constructor any contradicting views?

The problems-> 
1. Some Utility classes don't have all the methods static and also use "this" keyword inside methods.
Solution-> remove this keywords and make the method static.
2. The constructor is used to set values in a utility class
Solution -> Make another method to set values and make constructor protected. 
3. Some methods return "this" 
Solution-> Very suspicious on this, why is a utility class returning its instance? Still looking into how to fix this.

There are about 150-200 changes so can anyone please verify this approach? Do you see any unintended side effects? 
@vorburger @awasum @ptuomola @xurror @percyashu

### DIFF
--- a/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/ClientStatusChecker.java
+++ b/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/ClientStatusChecker.java
@@ -24,9 +24,13 @@ import java.util.HashMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class ClientStatusChecker {
+public final class ClientStatusChecker {
 
     private static final Logger LOG = LoggerFactory.getLogger(ClientStatusChecker.class);
+
+    private ClientStatusChecker() {
+
+    }
 
     public static void verifyClientIsActive(final HashMap<String, Object> clientStatusHashMap) {
         assertEquals(300, (int) clientStatusHashMap.get("id"));
--- a/fineract-provider/config/checkstyle/checkstyle.xml
+++ b/fineract-provider/config/checkstyle/checkstyle.xml
@@ -161,6 +161,7 @@
              <message key="ws.notPreceded"
              value="GenericWhitespace ''{0}'' is not preceded with whitespace."/>
         </module>
+        <module name="HideUtilityClassConstructor"/>
           <module name="OneTopLevelClass"/>
         <module name="MethodName">
             <property name="format" value="^[a-z][a-zA-Z0-9_]*$"/>
@@ -203,7 +204,6 @@
             <property name="setterCanReturnItsClass" value="true" />
         </module>
 
-        <module name="HideUtilityClassConstructor"/>
 
         < ! - - TODO Checks for Exception Handling Anti-Patterns - - >
         <module name="IllegalCatch"/>
--- a/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/CalendarHelper.java
+++ b/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/CalendarHelper.java
@@ -28,7 +28,7 @@ import java.util.HashMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class CalendarHelper {
+public final class CalendarHelper {
 
     private static final Logger LOG = LoggerFactory.getLogger(CalendarHelper.class);
     private static final String BASE_URL = "/fineract-provider/api/v1/";
@@ -36,6 +36,10 @@ public class CalendarHelper {
     private static final String ENITY_NAME = "/calendars";
     private static final String CENTER_ENTITY = "centers/";
     private static final String EDIT_CALENDAR = "editcalendarbasedonmeetingdates/";
+
+    private CalendarHelper() {
+
+    }
 
     public static Integer createMeetingCalendarForGroup(final RequestSpecification requestSpec, final ResponseSpecification responseSpec,
             final Integer groupId, final String startDate, final String frequency, final String interval, final String repeatsOnDay) {
--- a/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/CenterHelper.java
+++ b/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/CenterHelper.java
@@ -31,13 +31,17 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 @SuppressWarnings({ "rawtypes", "unchecked" })
-public class CenterHelper {
+public final class CenterHelper {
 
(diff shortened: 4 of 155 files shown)


fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/ClientStatusChecker.java
```diff
@@ -25,6 +25,12 @@
 import org.slf4j.LoggerFactory;
 
 public class ClientStatusChecker {
+
+    protected ClientStatusChecker() {
+        // prevents calls from subclass(if any)
+        throw new UnsupportedOperationException();
+    }
```
vorburger: This is weird / un-usual... I think the typical solution for this is to make the constructor private, and/or (?) declare the class to be `final` so that there can be no subclasses.

```suggestion
    private ClientStatusChecker() {
    }
```

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/bulkimport/populator/client/ClientEntityWorkbookPopulatorTest.java
```diff
@@ -60,9 +60,8 @@ public void setup(){
     @Test
     public void testClientEntityWorkbookPopulate() throws IOException {
         //in order to populate helper sheets
-        StaffHelper staffHelper=new StaffHelper();
         requestSpec.header(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON);
-        Integer outcome_staff_creation =staffHelper.createStaff(requestSpec,responseSpec);
+        Integer outcome_staff_creation =StaffHelper.createStaff(requestSpec,responseSpec);
```
vorburger: Question: Is this (static invocation of static methods) something you have found a way to make your IDE do automatically? If yes, then this would even allow us to enable Error Prone's `MethodCanBeStatic` in FINERACT-822... but doing it manually for 301 cases (according to @percyashu) is probably not "worth it" (https://en.wikipedia.org/wiki/Diminishing_returns).

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/bulkimport/populator/loan/LoanWorkbookPopulatorTest.java
```diff
@@ -64,18 +64,15 @@ public void testLoanWorkbookPopulate() throws IOException {
         Assert.assertNotNull("Could not create office" ,outcome_office_creation);
 
         //in order to populate helper sheets
-        ClientHelper clientHelper=new ClientHelper(requestSpec,responseSpec);
-        Integer outcome_client_creation=clientHelper.createClient(requestSpec,responseSpec);
```
vorburger: this is actually a very nice clean up - as it makes absolutely no sense to pass requestSpec, responseSpec twice

vorburger (changes requested):
> we keep constructor protected we cannot create an object of that class but support inheritance if required. 

The point is that a utility class with only static methods does not require to ever support inheritance.. if it does, then it's a "real object", not a "utility". Did you know that the programming language Scala even has dedicated syntax support for this? They distinguish a `class`from an `object` (if I remember correctly, I haven't looked at it in years).

> The other option was to use a private constructor(which won't support inheritance). I have decided to go with protected constructor any contradicting views?

private constructor is better because you don't WANT to support inheritance (for this kind of classes)

> Some Utility classes don't have all the methods static and also use "this" keyword inside methods.

Then they are not real static utility classes, but a mess....

> Solution-> remove this keywords and make the method static.

... if you can, that makes sense in some cases.

> The constructor is used to set values in a utility class

That depends on the case, agreed? All of those IT Helper which take a `requestSpec, responseSpec` would actually be "nicer" if they took that as a constructor argument, and then you wouldn't have to keep repeating it when calling their (now non-static) methods... check out how I recently changed the https://github.com/apache/fineract/blob/develop/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/SchedulerJobHelper.java - does that pattern make sense? But now whether to change everything existing like that is another question - it may not be worth it... I want to leave that decision up to you.

> Solution -> Make another method to set values and make constructor protected.

Wait, that may be a worse OOP design. If the class is not a static utility, then having a constructor may be the right design.

> Some methods return "this". Solution-> Very suspicious on this, why is a utility class returning its instance? Still looking into how to fix this.

Wait, that probably depends a bit... do you have some concrete examples that you want me to look at more closely?

> There are about 150-200 changes so can anyone please verify this approach?

Hope above and the inline comments help.

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/ClientStatusChecker.java
thesmallstar: Agreed! Making the changes

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/bulkimport/populator/client/ClientEntityWorkbookPopulatorTest.java
thesmallstar: Nope, it is all manual clean-up. Just that my IDE(vscode) does give a suggestion for this while writing code which makes it easier to spot. 
In this case, since I am removing the object I have "no other option" that making it static it throws out an error.

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/bulkimport/populator/loan/LoanWorkbookPopulatorTest.java
thesmallstar: Yes!

awasum:
Is this still important?

vorburger:
@thesmallstar do you want to finish this one up?

xurror (commented):
@thesmallstar what's left to do here?

thesmallstar:
@vorburger probably some work was done in this space recently? The checkstyle was much simpler (although boring :( ) to implement now.
Just FYI, the only suppressed checks used are for the classes which were extended, and hence we could not make them follow the utility class constructor rules.

thesmallstar:
@ptuomola  merge this?

ptuomola:
@thesmallstar LGTM, and as far as I can see, you've addressed the feedback from @vorburger. So unless @vorburger has further comments,  I'm happy to merge this...

ptuomola (approved):

PR #1022, opened 2020-06-10
FINERACT-822 enable UnnecessaryDefaultInEnumSwitch error

percyashu:
FINERACT-822

### DIFF
--- a/fineract-provider/build.gradle
+++ b/fineract-provider/build.gradle
@@ -441,6 +441,7 @@ tasks.withType(JavaCompile) {
                 "EqualsGetClass",
                 "EqualsUnsafeCast",
                 "DoubleBraceInitialization",
+                "UnnecessaryDefaultInEnumSwitch",
                 "InconsistentCapitalization",
                 "MissingOverride",
 //                "InvalidBlockTag",
--- a/fineract-provider/src/main/java/org/apache/fineract/adhocquery/service/AdHocScheduledJobRunnerServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/adhocquery/service/AdHocScheduledJobRunnerServiceImpl.java
@@ -55,6 +55,7 @@ public class AdHocScheduledJobRunnerServiceImpl implements AdHocScheduledJobRunn
     @Transactional
     @Override
     @CronTarget(jobName = JobName.GENERATE_ADHOCCLIENT_SCEHDULE)
+    @SuppressWarnings("UnnecessaryDefaultInEnumSwitch")
     public void generateClientSchedule() {
         final Collection<AdHocData> adhocs = this.adHocReadPlatformService.retrieveAllActiveAdHocQuery();
         if (adhocs.size() > 0) {
@@ -89,7 +90,6 @@ public class AdHocScheduledJobRunnerServiceImpl implements AdHocScheduledJobRunn
                             default:
                                 throw new IllegalStateException();
                         }
-
                     }
                 }
 
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/documentmanagement/contentrepository/ContentRepositoryUtils.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/documentmanagement/contentrepository/ContentRepositoryUtils.java
@@ -47,6 +47,7 @@ public class ContentRepositoryUtils {
             return this.value;
         }
 
+        @SuppressWarnings("UnnecessaryDefaultInEnumSwitch")
         public static ImageMIMEtype fromFileExtension(ImageFileExtension fileExtension) {
             switch (fileExtension) {
                 case GIF:
--- a/fineract-provider/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/service/SavingsProductToGLAccountMappingHelper.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/service/SavingsProductToGLAccountMappingHelper.java
@@ -172,8 +172,6 @@ public class SavingsProductToGLAccountMappingHelper extends ProductToGLAccountMa
             break;
             case ACCRUAL_UPFRONT:
             break;
-            default:
-            break;
         }
         return changes;
     }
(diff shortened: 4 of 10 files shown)


fineract-provider/src/main/java/org/apache/fineract/adhocquery/service/AdHocScheduledJobRunnerServiceImpl.java
```diff
@@ -88,10 +88,8 @@ public void generateClientSchedule() {
                                 next = start.plusDays((int) (long) adhoc.getReportRunEvery());
                                 run = Days.daysBetween(start, end).getDays() >= adhoc.getReportRunEvery();
                                 break;
-                            default:
-                                throw new IllegalStateException();
                         }
-
+                        throw new IllegalStateException();
```
vorburger: wait... wat ??  By doing this, it will ALWAYS throw, no? This will completely break. @awasum please help..

vorburger (changes requested):

fineract-provider/src/main/java/org/apache/fineract/adhocquery/service/AdHocScheduledJobRunnerServiceImpl.java
awasum: @percyashu ..Maybe remove this exception? and put it back in the default case inside the } on line 91 above?

Also if you want to add a default step, you need to check the `ReportRunFrequency.fromId(adhoc.getReportRunFrequency()` method from line 70 above and see how many different cases needs to be treated, then try to handle them all. Do we absolutely need a default case in all implementations of Switch statement?
As seen here: https://errorprone.info/bugpattern/UnnecessaryDefaultInEnumSwitch, dont put a default statement if all the cases have been handled. Maybe skip it? Will that throw an error?

Or given that ReportRunFrequency is an enum and we/Fineract is already handling all the 5 cases here, we can just throw a `new AdHocScheduledJobFrequencyException()` error here as default case with a message like "Correct Frequency not specified, Check the API docs"... @vorburger  @ptuomola your take?

awasum (changes requested):
Address feedback...Please.

awasum:
Fix conflicts too.

percyashu:
@awasum I will do

vorburger:
@ptuomola was wondering if you would like to help code-review this PR as well? Just because it's this is very similar & seems related to #1006 on which you have engaged. If interested, then I or @awasum would merge this only after your LGTM.

fineract-provider/src/main/java/org/apache/fineract/adhocquery/service/AdHocScheduledJobRunnerServiceImpl.java
ptuomola: Isn't this ErrorProne check specifically highlighting switch statements where the default case is never used, because we are already handling all the cases specifically? Given that, wouldn't the right solution just be to remove the default including the throw, and not add anything?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/documentmanagement/contentrepository/ContentRepositoryUtils.java
```diff
@@ -55,9 +55,8 @@ public static ImageMIMEtype fromFileExtension(ImageFileExtension fileExtension)
                     return ImageMIMEtype.JPEG;
                 case PNG:
                     return ImageMIMEtype.PNG;
-                default:
-                    throw new IllegalArgumentException();
             }
+            throw new IllegalArgumentException();
```
ptuomola: Same here - shouldn't we just remove the default case altogether, not add anything back?

fineract-provider/build.gradle
```diff
@@ -338,7 +338,6 @@ tasks.withType(JavaCompile) {
                 "ClassCanBeStatic",
                 "SameNameButDifferent",
                 "StringSplitter",
-                "ImmutableEnumChecker",
```
vorburger: Is this being removed (not moved) intentional?

percyashu: It was moved to the error list in a previous PR.  Don't know how it is here (maybe during merge) so I just removed it

fineract-provider/src/main/java/org/apache/fineract/adhocquery/service/AdHocScheduledJobRunnerServiceImpl.java
percyashu: @ptuomola I thought so to but it requires a return. 
@vorburger I thought if the case is not found, switch default is used with exception thrown and if case is not found and no default then switch is not executed then exception is thrown. That's why I did it like this. Just undid what I did and added a suppressedwarning

percyashu:
@ptuomola @vorburger @awasum  LGTY?

fineract-provider/src/main/java/org/apache/fineract/adhocquery/service/AdHocScheduledJobRunnerServiceImpl.java
```diff
@@ -87,9 +88,8 @@ public void generateClientSchedule() {
                                 run = Days.daysBetween(start, end).getDays() >= adhoc.getReportRunEvery();
                             break;
                             default:
-                                throw new IllegalStateException();
+                                throw new IllegalArgumentException();
```
awasum: If you have suppressed this check..Do we still need this default?

awasum:
@percyashu  Address last set of feedback here and maybe @ptuomola will give it a LGTH

awasum (approved):

fineract-provider/src/main/java/org/apache/fineract/adhocquery/service/AdHocScheduledJobRunnerServiceImpl.java
percyashu: Was my mistake wasn't not suppose to change the exception will revert it. Yeah we need it because Error prone thinks it is not necessary since all cases are specified.

vorburger:
@ptuomola does this one have LGTM from you now?

ptuomola (approved):
Looks good to me!

PR #1024, opened 2020-06-10
FINERACT-995: Rewriting logic to remove call to rs.previous()

ptuomola:
FINERACT-995

## Description
Rewrote the logic to remove call to rs.previous() that was throwing an SQLException. 

## Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,6 +73,7 @@ script:
 # using "&&" instead of several "-" means that integrationTest does not run if test fails,
 # and Docker test does not run if integration test fails, which makes PR failure easier to understand.
 # @see https://docs.travis-ci.com/user/job-lifecycle/#customizing-the-build-phase
+# NOTE: Sleep after docker-compose increased to 60 seconds as often Travis would fail to get Docker up in 30 seconds
   - ./gradlew --console=plain licenseMain licenseTest licenseIntegrationTest check  &&  ./gradlew --console=plain integrationTest --fail-fast  &&  sudo service mysql stop  &&  docker-compose build  &&  docker-compose up -d  &&  sleep 30s  &&  http --verify=no --timeout 240 --check-status get https://localhost:8443/fineract-provider/actuator/health  &&  (( $(http --verify=no --timeout 30 --check-status --body get https://localhost:8443/fineract-provider/actuator/info | wc --chars) > 100 ))
 # We stop the mysql system service when running the Docker test to avoid port 3306 conflicts (unless we run the mysql in docker-compose on another port; req. FINERACT-773)
 # The fany /actuator/info test makes sure that has more than 100 characters of JSON to test that the git.properties worked (see FINERACT-983)
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanArrearsAgingServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanArrearsAgingServiceImpl.java
@@ -438,19 +438,14 @@ public class LoanArrearsAgingServiceImpl implements LoanArrearsAgingService, Bus
 
             while (rs.next()) {
                 Long loanId = rs.getLong("loanId");
-                List<LoanSchedulePeriodData> periodDatas = new ArrayList<>();
-                LoanSchedulePeriodData loanSchedulePeriodData = fetchLoanSchedulePeriodData(rs);
-                periodDatas.add(loanSchedulePeriodData);
-                while (rs.next()) {
-                    Long tempLoanId = rs.getLong("loanId");
-                    if (loanId.equals(tempLoanId)) {
-                        periodDatas.add(fetchLoanSchedulePeriodData(rs));
-                    } else {
-                        rs.previous();
-                        break;
-                    }
+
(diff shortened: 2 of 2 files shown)


ptuomola:
@vorburger @xurror - here's a quick rewrite to remove the call to rs.previous(). Unfortunately I have no sensible way of testing this - so no idea if it works! Does this look sensible and if yes, any ideas how we could test this?

vorburger:
@awasum @fynmanoj @avikganguly01 @vishwasbabu are you able to help review this PR for FINERACT-995?

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanArrearsAgingServiceImpl.java
```diff
@@ -423,18 +423,14 @@ public OriginalScheduleExtractor(final String loanIdsAsString) {
 
             while (rs.next()) {
                 Long loanId = rs.getLong("loanId");
-                List<LoanSchedulePeriodData> periodDatas = new ArrayList<>();
-                LoanSchedulePeriodData loanSchedulePeriodData = fetchLoanSchedulePeriodData(rs);
-                periodDatas.add(loanSchedulePeriodData);
-                while (rs.next()) {
-                    Long tempLoanId = rs.getLong("loanId");
-                    if (loanId.equals(tempLoanId)) {
```
xurror, line 446: I'm trying to understand this to the best I can but I just have the feeling that this piece of logic is missing. Can you maybe make it a little clearer for me?
Thanks.

ptuomola: As far as I can see, the logic is trying to build a map of (loanId, periodData). There may be multiple rows in the ResultSet and these must therefore be merged to be part of the same entry in the map.

The previous code does it by:

1. Read one LoanId from resultset and create a map entry
2. In a loop, keep fetching next row from result set. 
    - If the next row has the same loanId, then add it to the same map entry and fetch next
    - If the next row has a different loanId, go back one row in result set and start from step 1

The problem is the "go back one row" which is not supported by some JDBC drivers.

But an equivalent way of building the same map (at least in my view would be):

1. Iterate through the result set. For each row's loanId: 
- If we already have a map entry for that loanId, add the dates to the existing entry
- If we don't yet have a map entry for that loanId, create a new one and add the dates to that

No rs.previous() required, and either way you should end up with the same Map - as far as I can see..

But I don't have a way to test this code, so a bit concerned I may have missed something...

xurror: Totally makes sense now. Thanks for helping me understand this better.
I actually read a comment that suggested doing this or using caching or hashmap or any other collection to hold all the resultset and instead iterate through it as you did but it just wasn't clear to me.

Unfortunately we don't have a way testing this but I agree with your approach. Since it passes all IT tests, we can just blindly merge this 2 or 3 days if there are no objections and hope for the best 😁.

awasum:
I maybe able to look at this tomorrow..

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanArrearsAgingServiceImpl.java
vorburger: @ptuomola sorry for delay in reviewing this (I wanted to do it with a clear mind instead of clicking through it half asleep a night after day job work...) and after staring at this a little bit, I get it now - this makes good sense to me as well, and I have no objection to merging this even without having any way to functionally test this myself either (no idea if existing ITs, recently re-enable, would catch this, but doesn't matter, it "makes sense"). 

One very small feedback I would have is that, now that I understand this logic, it seems to me that moving that `scheduleDate.put(loanId, periodDatas);` now in line 449 (originally 434) up into the `if(periodDatas == null)` right after the `periodDatas = new ArrayList<>();` would make it a tiny bit more clear and explicit what we're doing here? It's too minor to hold back merging this important fix, so I'll just go ahead and merge it ASAP, and let you raise a very small minor follow-up PR, if you like (or not, fine). _I originally wrote this before seeing the Spotless style failure, if you're going over this again, might as well do that too here - if you agree that makes sense?_

vorburger (approved):
LGTM, but the rebase failed due to Spotless (probably the `{` needs to be on the same line as `if`..) - will you fix @ptuomola ? 

Thank You for this contribution - that's one more release Blocker issue (FINERACT-995) done!

.travis.yml
```diff
@@ -73,6 +73,7 @@ script:
 # using "&&" instead of several "-" means that integrationTest does not run if test fails,
 # and Docker test does not run if integration test fails, which makes PR failure easier to understand.
 # @see https://docs.travis-ci.com/user/job-lifecycle/#customizing-the-build-phase
+# NOTE: Sleep after docker-compose increased to 60 seconds as often Travis would fail to get Docker up in 30 seconds
```
vorburger, line 76: @ptuomola just noticed while merging this that you haven't actually changed the `sleep 30s` so this comment is perhaps a left over? I'll merge this PR anyway, you can always clean it up later, if you like.

PR #1027, opened 2020-06-10
FINERACT-799 Fixes Sorting for certain columns is not supported in Audit

thesmallstar:
Please refer: https://issues.apache.org/jira/browse/FINERACT-799
I have done the functional testing from the data I had and this looks good. 
Note that processingResult refers to status mentioned in the issue.

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/commands/service/AuditReadPlatformServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/commands/service/AuditReadPlatformServiceImpl.java
@@ -75,9 +75,9 @@ import org.springframework.stereotype.Service;
 public class AuditReadPlatformServiceImpl implements AuditReadPlatformService {
 
     private static final Logger LOG = LoggerFactory.getLogger(AuditReadPlatformServiceImpl.class);
-    private static final Set<String> supportedOrderByValues = new HashSet<>(
-            Arrays.asList("id", "actionName", "entityName", "resourceId", "subresourceId", "madeOnDate", "checkedOnDate", "officeName",
-                    "groupName", "clientName", "loanAccountNo", "savingsAccountNo", "clientId", "loanId"));
+    private static final Set<String> supportedOrderByValues = new HashSet<>(Arrays.asList("id", "actionName", "entityName", "resourceId",
+            "subresourceId", "madeOnDate", "checkedOnDate", "officeName", "groupName", "clientName", "loanAccountNo", "savingsAccountNo",
+            "clientId", "loanId", "maker", "checker", "processingResult"));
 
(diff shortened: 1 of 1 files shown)


vorburger (changes requested):
>  I have done the functional testing from the data I had and this looks good.

I can only trust that you actually did that manually, but if you could prove this to us with an integration test, that would be even better? :smile:

thesmallstar:
Yes! 😀

vorburger:
@thesmallstar just wanted to check if you'd still like to write an IT for this, or if we should just merge it as is? Please don't hesitate to ask if you need help with the IT.

thesmallstar:
@vorburger yes, I did not find any integrations tests for the audit section. Makes this harder and I am trying to figure out on this. 
I am not sure about the work-flow in this case, and I couldn't figure out much by looking at other integration tests. 
If you can address the following doubts(if they are too naive, I would research more)
1. Should I also be worried about tests for creating new audits?(since there are none now)
2. If not, what will be the basic logic of the test: I thought it would be just to visit the route and check if we get a success message. 
I am confused about what purpose does the integration test will play in this case, is it just point 2? 
Also if I were to create integration tests for creation will the workflow be:
  -  post request to create an audit 
  -  Get request to check if it stored in the server. 

I hope I am making some sense :P

vorburger:
@thesmallstar so these audits are not something you can create new audits of, you cannot really post a request to create an audit. They are basically just traces of changes made - read https://demo.fineract.dev/fineract-provider/api-docs/apiLive.htm#audits and then check out https://demo.fineract.dev/fineract-provider/api/v1/audits?tenantIdentifier=default (login with _mifos/password_) ... does that help? So an IT for this I guess would make 2-3 POSTs on a few other entities (perhaps create like a User, or whatever), and then do a GET with a sort and make sure that works.

But let me already merge this as-is anyway so we can progress and close FINERACT-799, and you can contribute an integration test in a separate follow-up PR. Fixed bugs without ITs are better than not fixed bugs!

vorburger:
@thesmallstar but this actually needs a rebase and a `./gradlew :spotlessApply` now...

thesmallstar:
> But let me already merge this as-is anyway so we can progress and close FINERACT-799, and you can contribute an integration test in a separate follow-up PR. Fixed bugs without ITs are better than not fixed bugs!

@vorburger  Yes please merge this, I will create a JIRA issue so that we can keep track of this integration test.

vorburger:
This failed to build with `JVM crash log found: file:///fineract/hs_err_pid39.log FAILURE: Build failed with an exception.` ... let's rebase and see if it disappears. Otherwise, upgrading JDK sometimes helps for crashes like this.

PR #1032, opened 2020-06-10
FINERACT-1034-encryption-infra

fynmanoj:
FINERACT-1034

## Description

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/api/CryptographyApiResource.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/api/CryptographyApiResource.java
@@ -1,0 +1,74 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.infrastructure.crypt.api;
+
+import io.swagger.annotations.Api;
+import io.swagger.annotations.ApiOperation;
+import io.swagger.annotations.ApiParam;
+import io.swagger.annotations.ApiResponse;
+import io.swagger.annotations.ApiResponses;
+import io.swagger.annotations.SwaggerDefinition;
+import io.swagger.annotations.Tag;
+import javax.ws.rs.Consumes;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+import org.apache.fineract.infrastructure.core.serialization.ToApiJsonSerializer;
+import org.apache.fineract.infrastructure.crypt.data.PublicKeyData;
+import org.apache.fineract.infrastructure.crypt.service.CryptographyReadPlatformService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Scope;
+import org.springframework.stereotype.Component;
+
+
+/**
+ * @author manoj
+ */
+@Path("crypt")
+@Component
+@Scope("singleton")
+@Api(tags = {"Cryptography"})
+@SwaggerDefinition(tags = {
+        @Tag(name = "Cryptography", description = "Provides RSA public keys for various modules(types)")
+})
+public class CryptographyApiResource {
+
+    private final ToApiJsonSerializer<PublicKeyData> toApiJsonSerializer;
+    private final CryptographyReadPlatformService cryptographyReadPlatformService;
+
+    @Autowired
+    public CryptographyApiResource(ToApiJsonSerializer<PublicKeyData> toApiJsonSerializer,
+                                   CryptographyReadPlatformService cryptographyReadPlatformService) {
+        this.toApiJsonSerializer = toApiJsonSerializer;
+        this.cryptographyReadPlatformService = cryptographyReadPlatformService;
+    }
+
+    @GET
+    @Path("publickey/{type}")
+    @Consumes({ MediaType.APPLICATION_JSON })
+    @Produces({ MediaType.APPLICATION_JSON })
+    @ApiOperation(value = "Provides RSA public key for {type}", notes = "Key is provided with a version ")
+    @ApiResponses({@ApiResponse(code = 200, message = "RSA public key")})
+    public String getRsaPublicKey(@PathParam("type") @ApiParam(value = "type") final String type) {
+        return this.toApiJsonSerializer.serialize(this.cryptographyReadPlatformService.getPublicRsaKey(type));
+    }
+
+}
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/service/CryptographyReadPlatformService.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/service/CryptographyReadPlatformService.java
@@ -1,0 +1,28 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.infrastructure.crypt.service;
+
+import org.apache.fineract.infrastructure.crypt.data.PublicKeyData;
+
+/**
+ * @author manoj
+ */
+public interface CryptographyReadPlatformService {
+    PublicKeyData getPublicRsaKey(String type);
+}
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/service/CryptographyReadPlatformServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/service/CryptographyReadPlatformServiceImpl.java
@@ -1,0 +1,46 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.infrastructure.crypt.service;
+
+import org.apache.fineract.infrastructure.crypt.data.PublicKeyData;
+import org.apache.fineract.infrastructure.crypt.domain.EncryptionKeyPair;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+/**
+ * @author manoj
+ */
+@Service
+public class CryptographyReadPlatformServiceImpl  implements CryptographyReadPlatformService{
+    private final EncryptionKeyStoreService encryptionKeyStore;
+
+
+    @Autowired
+    public CryptographyReadPlatformServiceImpl(EncryptionKeyStoreService encryptionKeyStore) {
+        this.encryptionKeyStore = encryptionKeyStore;
+
+    }
+
+
+    @Override
+    public PublicKeyData getPublicRsaKey(String type){
+        EncryptionKeyPair keys = this.encryptionKeyStore.retrieveKey(type);
+        return PublicKeyData.instanceWithBase64Encoding(keys.getPublicKey(), keys.getVersion());
+    }
+}
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/service/EncryptionKeyStoreServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/service/EncryptionKeyStoreServiceImpl.java
@@ -1,0 +1,104 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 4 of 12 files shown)


vorburger:
Manoj, how crazy would I sound to you if I suggested that this should have some documentation, somewhere? Ideally in this PR itself, not on some Wiki, or "later"... do you want to update that API doc HTML file, do you know what I mean? Or even .. probably not README, but.. docs/*.md?

fynmanoj:
Sure, @vorburger , I shall add some documentation here

fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/data/CryptData.java
```diff
@@ -0,0 +1,39 @@
...
+ * under the License.
+ */
+package org.apache.fineract.infrastructure.crypt.data;
+
+/**
+ * @author manoj
+ */
+
+
+public class CryptData {
```
vorburger: perhaps this class could have a better name, maybe... PublicKeyWithVersion? Or... StringifiedPublicKeyWithVersion or StringEncodedPublicKeyWithVersion? Someting like that.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/api/CryptographyApiResource.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.fineract.infrastructure.crypt.service.CryptographyReadPlatformService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Scope;
+import org.springframework.stereotype.Component;
+
+
+/**
+ * @author manoj
+ */
+
```
vorburger: remove blank line

fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/data/CryptData.java
```diff
@@ -0,0 +1,39 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.infrastructure.crypt.data;
+
+/**
+ * @author manoj
+ */
+
+
```
vorburger: remove x2 blank lines

fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/service/CryptographyReadPlatformServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.Base64;
+import org.apache.fineract.infrastructure.crypt.data.CryptData;
+import org.apache.fineract.infrastructure.crypt.domain.EncryptionKeys;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+/**
+ * @author manoj
+ */
+
```
vorburger: remove blank line

fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/utils/RSAEncryptionUtils.java
```diff
@@ -0,0 +1,110 @@
...
+            PrivateKey privateKey = keyPair.getPrivate();
+            byte[] publicKeyAsBytes = publicKey.getEncoded();
+            byte[] privateKeyAsBytes = privateKey.getEncoded();
+
+            return new EncryptionKeys(privateKeyAsBytes, publicKeyAsBytes, DateUtils.getLocalDateTimeOfTenant(), generateVersionString());
+
+        } catch (NoSuchAlgorithmException e) {
+           LOG.error("Error occurred", e);
+        }
+        throw new GeneralPlatformDomainRuleException("error.msg.error.in.generating.keys", "Error in generating keys");
```
vorburger: You could avoid this by doing the `keyPairGenerator = KeyPairGenerator.getInstance("RSA");` in the constructor, and let this error propagate there? This would cause it to fail at boot, which would be better. If you do want to keep it, then IMHO it would be better like this, even though it does the exact same thing, just because clearer to understand:

```suggestion
        } catch (NoSuchAlgorithmException e) {
           LOG.error("Error occurred", e);
           throw new GeneralPlatformDomainRuleException("error.msg.error.in.generating.keys", "Error in generating keys");
        }
```

fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/utils/RSAEncryptionUtils.java
```diff
@@ -0,0 +1,110 @@
...
+            if(isBase64Encoded) {
+                encryptedText = Hex.encodeHexString(encryptedText.getBytes( UTF_8));
+                byte[] decryptByteFromUi = decrypt.doFinal(DatatypeConverter.parseHexBinary(encryptedText));
+                decryptedString =  new String(decryptByteFromUi,  UTF_8);
+            } else {
+                byte[] decryptByteFromUi = decrypt.doFinal(encryptedText.getBytes( UTF_8));
+                decryptedString =  new String(decryptByteFromUi, UTF_8);
+            }
+            return decryptedString;
+        }catch (NoSuchAlgorithmException | InvalidKeySpecException | NoSuchPaddingException | InvalidKeyException |
```
vorburger, line 91: as above

fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/api/CryptographyApiResource.java
```diff
@@ -0,0 +1,75 @@
...
+        this.cryptographyReadPlatformService = cryptographyReadPlatformService;
+    }
+
+    @GET
+    @Path("publickey/{type}")
+    @Consumes({ MediaType.APPLICATION_JSON })
+    @Produces({ MediaType.APPLICATION_JSON })
+    @ApiOperation(value = "Provides RSA public key for {type}", notes = "Key is provided with a version ")
+    @ApiResponses({@ApiResponse(code = 200, message = "RSA public key")})
+    public String downloadFile(@PathParam("type") @ApiParam(value = "type") final String type) {
```
vorburger: nit pick: If this `@Produces({ MediaType.APPLICATION_JSON })` then this isn't really a `downloadFile()` but more of a `getPublicKey()`, right?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/domain/EncryptionKeys.java
```diff
@@ -0,0 +1,54 @@
...
+ * under the License.
+ */
+package org.apache.fineract.infrastructure.crypt.domain;
+
+import org.joda.time.LocalDateTime;
+
+/**
+ * @author manoj
+ */
+public final class EncryptionKeys {
```
vorburger: just about the naming - it's confusing (to me) that you name this Keys - in plural. This seems to be A (single, one) key - with it's private and public part. So perhaps better name this a ... PrivateAndPublicKeyWithVersion - kind of thing?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/service/CryptographyReadPlatformServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+    public CryptographyReadPlatformServiceImpl(EncryptionKeyStoreService encryptionKeyStore) {
+        this.encryptionKeyStore = encryptionKeyStore;
+
+    }
+
+
+    @Override
+    public CryptData getPublicRsaKey(String type){
+        EncryptionKeys keys = this.encryptionKeyStore.retrieveKey(type);
+        return CryptData.instance(Base64.getEncoder().encodeToString(keys.getPublicKey()), keys.getVersion());
```
vorburger: perhaps it would be nicer if you made CryptData have a constructor taking an EncryptionKeys argument, and just did this transformation inside there?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/service/CryptographyReadPlatformService.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+package org.apache.fineract.infrastructure.crypt.service;
+
+import org.apache.fineract.infrastructure.crypt.data.CryptData;
+
+/**
+ * @author manoj
+ */
+public interface CryptographyReadPlatformService {
+    CryptData getPublicRsaKey(String type);
```
vorburger: perhaps you would like to consider the fact that this an RSA key an implementation detail, and only name this `getPublicRsaKey()` - because, in the future, there may be other types of keys you want to support?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/service/EncryptionKeyStoreServiceImpl.java
```diff
@@ -0,0 +1,104 @@
...
+            @CacheEvict(value = "encryptionKeys", key = "T(org.apache.fineract.infrastructure.core.service.ThreadLocalContextUtil).getTenant().getTenantIdentifier().concat(#type)")})
+    public void storeKey(String type, EncryptionKeys key){
+        this.keyMap.put(getKeyType(type), key);
+    }
+
+    private EncryptionKeys retrieveValidKey(String type){
+
+        EncryptionKeys keys = getKeys(type);
+        if(keys == null){
+            return null;
```
vorburger, line 64: null are scary bad creatures, best avoided. Would it be clearer to throw some Exception here (include the type, so if it happens, you have context and know why)?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/service/EncryptionKeyStoreServiceImpl.java
```diff
@@ -0,0 +1,104 @@
...
+        if(validUpto.equals(-1)){
+            return keys;
+        } else {
+            Seconds seconds = Seconds.secondsBetween(keys.getCreatedDateTime(), DateUtils.getLocalDateTimeOfTenant());
+            if(seconds.getSeconds() < validUpto) {
+                return keys;
+            }
+
+        }
+        return null;
```
vorburger, line 78: as above... perhaps it's better to "handle" under whatever condition you may end up here?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/utils/RSAEncryptionUtils.java
```diff
@@ -0,0 +1,110 @@
...
+
+
+
+
+
+    private String generateVersionString() {
+        final char[] buf = new char[15];
+        final String alphanum = "ABCDEFGHIJKLMNOPQRSTUVWXYZ" + "0123456789" ;
+        final char[] symbols = alphanum.toCharArray();
+        final Random random= new SecureRandom();
```
vorburger: make this `SecureRandom` a private final field of this Component class instead of a variable in this method

fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/utils/RSAEncryptionUtils.java
```diff
@@ -0,0 +1,110 @@
...
+
+            return new EncryptionKeys(privateKeyAsBytes, publicKeyAsBytes, DateUtils.getLocalDateTimeOfTenant(), generateVersionString());
+
+        } catch (NoSuchAlgorithmException e) {
+           LOG.error("Error occurred", e);
+        }
+        throw new GeneralPlatformDomainRuleException("error.msg.error.in.generating.keys", "Error in generating keys");
+    }
+
+    public String decryptUsingRSA(String encryptedText, final byte[] privateKeyAsBytes, boolean isBase64Encoded) {
```
vorburger, line 76: where is this method used? :smile:  see discussion in FINERACT-1034 ...

vorburger (changes requested):
@fynmanoj interesting PR! Thank You for contribution.

I have some technical feedback, and a larger question about this which I've asked in FINERACT-1034.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/data/CryptData.java
fynmanoj: This class is used for the serialisation only, hence was the carelessness on the name, Thanks for pointing out

fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/api/CryptographyApiResource.java
fynmanoj: done!

fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/utils/RSAEncryptionUtils.java
fynmanoj: Changed this.  I was just being sure that the method return something or an exception(do or die).Yet you are right, there is better ways to do it rather than sending panic

fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/utils/RSAEncryptionUtils.java
fynmanoj: done!

fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/api/CryptographyApiResource.java
fynmanoj: NIce Catch 😄 , yes absolutely, it was a classic copy paste.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/domain/EncryptionKeys.java
fynmanoj: Changed it, i was referring the Key Pair as Keys.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/service/CryptographyReadPlatformService.java
fynmanoj: yes, hence the class name is Cryptography* and I am implementing only RSA now.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/service/CryptographyReadPlatformServiceImpl.java
fynmanoj: It was a single usage of the method so far.
Changed it as you suggested for better reading and maintainability.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/service/EncryptionKeyStoreServiceImpl.java
fynmanoj: This is trying to retrieve a key from the memory. When there is no key available here (that is, when null is returned) the wrapper which calls this method is supposed to create a new key and store it.  do you really think an Exception throw and catch needed in this context? I don't.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/service/EncryptionKeyStoreServiceImpl.java
fynmanoj: same as above

fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/utils/RSAEncryptionUtils.java
fynmanoj: done!

fineract-provider/src/main/java/org/apache/fineract/infrastructure/crypt/utils/RSAEncryptionUtils.java
fynmanoj: This method is meant for future implementations. Lets discuss in FINERACT-1034

awasum:
@fynmanoj , @vorburger .. How do we proceed with this PR? is there more work to be done here? I don;t understand the use case..

vorburger:
@awasum nor do, as per my comments in the JIRA (and I'm therefore still reluctant to LGTM and Merge this).

awasum:
@avikganguly01 , do you have an interest to review this? I have a feeling you might know what this feature is meant for and how it will evolve. Also is there a follow up PR to this which will make it clear to some of us how this feature will be used?

maektwain:
This is completely Ok, but there are some design flaws . 

1. Open to attacks if private keys are stored on server . 
2. Another issue is the API are not  best suitable as of now to handle such leaks, it's just a password guess away. 
3. Generation of keys on the server is an expensive task. 
4. Encryption and Decryption can happen through an identity protocol such as Elliptic Curves etc etc. 

We might need to work on the server side to acknowledge the  keys are valid and then exposing functionality.

vorburger:
@avikganguly01 knowing that @fynmanoj works with you, and assuming that you are still an active committer to this project (are you?), I was wondering if you would like to engage on code review feedback and possible evental merge of PRs such as this one? (I don't want to be a bottleneck on this project, and won't have the spare time to help review changes like this. If there are other active commiters such as @awasum @ptuomola @xurror or others reading this who want to engage on this PR, please don't let me hold you up / wait for me - I'll ignore this one, and try to help on others.)

PR #1037, opened 2020-06-10
FINERACT-942 Added and Enforced IllegalThrows checkstyle

thesmallstar:
Just curious on this change:
IllegalThrows:  Declaring that a method throws java.lang.Error or java.lang.RuntimeException is almost never acceptable.

I was doing a bit of research why this is not a good idea, I came across a contradicting statement on this on [link](https://stackabuse.com/exception-handling-in-java-a-complete-guide-with-best-and-worst-practices/ ) : 

> Since we're throwing a runtime exception, there's no need to include it in the method signature, like in the example above, but it's often considered good practice to do so, at least for the sake of documentation. 

Any comments on this @vorburger @xurror @ptuomola ?

### DIFF
--- a/fineract-provider/config/checkstyle/checkstyle.xml
+++ b/fineract-provider/config/checkstyle/checkstyle.xml
@@ -57,6 +57,7 @@
         </module>
         <module name="EqualsHashCode"/>
         <module name="FallThrough"/>
+        <module name="IllegalThrows" />
         <module name="RedundantImport"/>
         <module name="UnusedImports"/>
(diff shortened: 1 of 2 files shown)


thesmallstar:
Also, I will raise a small PR after this is merged adding this change to the readme.

thesmallstar:
I Have tried to add checkstyles at random lines so as to avoid merge conflicts, if they still arrive please ignore I will fix them ASAP.

vorburger (approved):
LGTM

PR #1020, opened 2020-06-09
FINERACT-1021 - Update to JUnit 5

xurror:
[FINERACT-1021](https://issues.apache.org/jira/browse/FINERACT-1021)

PR #1049, opened 2020-06-13
FINERACT-857: Fixing scheduler integration tests and migrating to executeAndAwaitJob

ptuomola:
## Description
Fixed two bugs that were triggered by scheduler integration tests: 
- Invalid query in SavingsAccountCharge (missing parameters)
- Lazy loading not working for m_savings_account.group_id due to transaction boundary

Also changed triggering of jobs to use executeAndAwaitJob() and fixed logging for exceptions jobs

## Checklist


ptuomola:
OK - looks like turning off lazy loading for group means that we have more than 61 (!) tables in a join in RecurringDeposit, which then breaks... so need to find a way to get the lazy loading to work after all...

fineract-provider/src/main/java/org/apache/fineract/scheduledjobs/service/ScheduledJobRunnerServiceImpl.java
```diff
@@ -229,7 +229,7 @@ public void applyAnnualFeeForSavings() {
                     LOG.error("Apply annual fee failed for account: {} with message {}", savingsAccountReference.getAccountNo(), error);
                 }
             } catch (final Exception ex) {
-                // need to handle this scenario
+                LOG.error("Apply annual fee failed for account: {} with message: {}", savingsAccountReference.getAccountNo(), ex.getMessage());
             }
```
vorburger, line 233: LOG.error(..., e) instead of e.getMessage() will log full cause. Or was this intentional?

fineract-provider/src/main/java/org/apache/fineract/scheduledjobs/service/ScheduledJobRunnerServiceImpl.java
```diff
@@ -307,7 +307,7 @@ public void updateMaturityDetailsOfDepositAccounts() {
                     LOG.error("Update maturity details failed for account: {} with message {}", depositAccount.accountNo(), error.getDeveloperMessage());
                 }
             } catch (final Exception ex) {
-                // need to handle this scenario
+                LOG.error("Update maturity details failed for account: {} with message: {}", depositAccount.accountNo(), ex.getMessage());
```
vorburger: as above

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccount.java
```diff
@@ -2928,6 +2928,7 @@ public void loadLazyCollections() {
         transactions.size();
         charges.size();
         savingsOfficerHistory.size();
+        if(group != null) { group.getOffice(); }
```
vorburger: Is this left over, or for the lazy loading? I fear future automated code clean up may try to remove this as NOOP... maybe put an inline comment to explain it forces lazy loading, if that's what it is? Or even fake store it in a variable used in a useless expression to prevent NOOP clean up? Or remove if this is just a left over and not actually required.

vorburger (changes requested):
Cool!! You are amazing. Some minor feedback..

PS: I'll search in JIRA for the open real bugs you have also fixed here.

fineract-provider/src/main/java/org/apache/fineract/scheduledjobs/service/ScheduledJobRunnerServiceImpl.java
ptuomola: I was thinking that if we have scheduled jobs processing large quantities of accounts, we may not want to have the full stack trace for every account - one line per account with the account number and error message would be enough. But will change this to print the full cause...

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccount.java
ptuomola: This is for lazy loading. And yes, I agree it makes sense to make this explicit with comment and variable assignment (even though hopefully people will work out what's going on from the function name as well). Will update the PR.

fineract-provider/src/main/java/org/apache/fineract/scheduledjobs/service/ScheduledJobRunnerServiceImpl.java
vorburger: I see what you mean, but didn't get it initially. For "consistency", let's keep it like you have fixed it, and deal with it if we have overflowing logs, only.

vorburger (approved):
LGTM! Would love to merge this ASAP.

@ptuomola if you are on and want to do this ASAP, it may avoid new conflicts for #943 which I'd like to do in the coming hours. If I don't hear from you, perhaps I'll even exceptionally resolve the conflict myself just to be able to merge this sooner rather then later - I'm sure you wouldn't mind.

vorburger:
Actually the merge conflict here was due to #1020 and trivial to resolve - I have, for once, just used GitHub's Web UI ... if I didn't screw it up, hopefully this will pass the build and I'll merge it.

PR #1050, opened 2020-06-13
FINERACT-1022 - Update to org.ehcache

xurror:
sues.apache.org/jira/browse/FINERACT-1022

### DIFF
--- a/fineract-provider/build.gradle
+++ b/fineract-provider/build.gradle
@@ -94,7 +94,7 @@ dependencyManagement {
         dependency 'org.flywaydb:flyway-core:6.4.4'
         dependency 'org.quartz-scheduler:quartz:2.3.2'
         dependency 'com.amazonaws:aws-java-sdk-s3:1.11.800'
-        dependency 'net.sf.ehcache:ehcache:2.10.6' // TODO FINERACT-1018
+        dependency 'org.ehcache:ehcache:3.8.1'
         dependency 'com.github.spullara.mustache.java:compiler:0.9.6'
         dependency 'com.jayway.jsonpath:json-path:2.4.0'
         dependency 'org.apache.tika:tika-core:1.24.1'
@@ -110,6 +110,7 @@ dependencyManagement {
         dependency 'io.github.classgraph:classgraph:4.8.85'
         dependency 'org.awaitility:awaitility:4.0.3'
         dependency 'com.github.spotbugs:spotbugs-annotations:4.0.4'
+        dependency 'javax.cache:cache-api:1.1.1'
 
         dependency ('org.dom4j:dom4j:2.1.3') {
             exclude 'relaxngDatatype:relaxngDatatype' // already in com.sun.xml.bind:jaxb-osgi:2.3.0.1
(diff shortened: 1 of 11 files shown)


fineract-provider/build.gradle
```diff
@@ -94,7 +94,7 @@ dependencyManagement {
         dependency 'org.flywaydb:flyway-core:6.4.4'
         dependency 'org.quartz-scheduler:quartz:2.3.2'
         dependency 'com.amazonaws:aws-java-sdk-s3:1.11.800'
-        dependency 'net.sf.ehcache:ehcache:2.10.6' // TODO FINERACT-1018
+        dependency 'org.ehcache:ehcache:3.8.1' // TODO FINERACT-1018
```
vorburger: rm TODO ?

fineract-provider/build.gradle
```diff
@@ -110,6 +110,7 @@ dependencyManagement {
         dependency 'io.github.classgraph:classgraph:4.8.85'
         dependency 'org.awaitility:awaitility:4.0.3'
         dependency 'com.github.spotbugs:spotbugs-annotations:4.0.4'
+        dependency 'javax.cache:cache-api:1.1.1'
```
vorburger, line 113: Isn't that a transitive dependency of ehCache already?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/cache/PlatformCacheConfiguration.java
```diff
@@ -18,6 +18,7 @@
  */
 package org.apache.fineract.infrastructure.cache;
 
+// import javax.cache.CacheManager;
```
vorburger: Don't add new commented out imports, but clean up

fineract-provider/src/main/java/org/apache/fineract/infrastructure/cache/service/RuntimeDelegatingCacheManager.java
```diff
@@ -120,6 +123,11 @@ public Cache getCache(final String name) {
     }
 
     private void clearEhCache() {
-        this.ehcacheCacheManager.getCacheManager().clearAll();
+        // this.jcacheCacheManager.getCacheManager().removeAll();
```
vorburger: rm? Is there no removeAll() on the new standard API?

vorburger (changes requested):
Nice!! Thanks for this. Just some minor feedback.. in general, never comment out, but always just cleanly remove. The history is always there in Git anyway..

fineract-provider/build.gradle
xurror: 😂😂😂 Completely forgot.

fineract-provider/build.gradle
xurror: I wish it were but unfortunately it isn't. I tried without and it didn't work.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/cache/service/RuntimeDelegatingCacheManager.java
xurror: Nopes. Unfortunately there isn't. Just a destroy(still not destroyAll) method which I don't think is what we want to do.

vorburger (approved):
LGTM

vorburger:
FINERACT-1022 says this may be broken, and we thus may revert this again.

PR #1053, opened 2020-06-14
FINERACT-822 enable BadImport error check

percyashu:
FINERACT-822


vorburger:
@percyashu this will need a rebase and conflict resolution, but perhaps it's best if you wait with doing that until I'm done with #943 which I'm doing soon (today), and pick this up tomorrow.

xurror (commented):
Resolve merge conflicts and squash commits and this would be good to go

vorburger (changes requested):
Wait a second, I've only properly reviewed this now.. are we saying that we are forbidding import static with this? That seems weird- import static are nice and useful, no? Or is this preventing some corner case that I'm not getting yet?

percyashu:
@vorburger Sorry this was just for badImport https://errorprone.info/bugpattern/BadImport.
I just added UnnecessaryStaticImport since there wasn't any violation in the code.

awasum:
> Wait a second, I've only properly reviewed this now.. are we saying that we are forbidding import static with this? That seems weird- import static are nice and useful, no? Or is this preventing some corner case that I'm not getting yet?

@vorburger , Looks like other static imports still work on the code base. Also seem @percyashu  just refactored static import which could make the code difficult to read. I dont know if this will make you to change your mind?

awasum (approved):

PR #1054, opened 2020-06-14
add missing useJUnitPlatform() to make Gradle run JUnit 5 [FINERACT-1021]

vorburger:
see #1020 and FINERACT-1021 ...

### DIFF
--- a/fineract-provider/build.gradle
+++ b/fineract-provider/build.gradle
@@ -585,6 +585,7 @@ task integrationTest(type:Test) {
 }
 
 integrationTest {
+	useJUnitPlatform()
 	testLogging { // FINERACT-927
 		events "skipped", "failed"
 		showStandardStreams = false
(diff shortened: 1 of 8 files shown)


vorburger:
This will fail with 14 integration test failures... :cry:  @xurror I doubt you actually ran all tests in #1020.

The failures seems to be due to some parameter order mix up - I'm checking if I can easily fix them myself quickly. If not, I'll (have to..) sink more team into figuring out how to fully revert #1020 cleanly instead (PITA, just fixing is likely faster).

xurror:
> This will fail with 14 integration test failures...  @xurror I doubt you actually ran all tests in #1020.
> 
> The failures seems to be due to some parameter order mix up - I'm checking if I can easily fix them myself quickly. If not, I'll (have to..) sink more team into figuring out how to fully revert #1020 cleanly instead (PITA, just fixing is likely faster).

I ran the tests, some must have excaped my vigilence. Thanks for your followup.

PR #1055, opened 2020-06-14
Spotless (FINERACT-1006)

vorburger:
This PR is #943 from @thesmallstar

plus the automated bulk reformatting (which needs to be done by me to avoid merge conflicts)

FINERACT-1006


thesmallstar:
@vorburger Don't you think it would be a little bit problematic to have ./setting/preferences out of git ignore? If a contributor changes his pref to what he likes, he might add that to the commit, or any small change for instance what he prefers.
Will it not be a problem in this situation? Am I missing anything?

PR #1062, opened 2020-06-14
FINERACT-821 Added and Enforced OperatorWrap Checkstyle

thesmallstar:


vorburger (approved):
The + to concatenate String on the new line makes sense.

PR #1065, opened 2020-06-14
FINERACT-821 Added and Enforced ParenPad Checkstyle

thesmallstar:
Thanks to spotless again ;) #1055


vorburger (changes requested):
Wait a second, these // comments that are "glued" to the code without space look weird, no? Wouldn't this be nicer with a space?

vorburger (changes requested):
Oh this is only for inside a parenthesis... how about just removing the useless comments, and putting the useful ones in the line above instead of inside the parenthesis?

vorburger (approved):
LGTM

PR #1068, opened 2020-06-15
FINERACT-723: Fixed to use the right date in more tests

ptuomola:
## Checklist

### DIFF
--- a/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/ClientSavingsIntegrationTest.java
+++ b/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/ClientSavingsIntegrationTest.java
@@ -37,7 +37,6 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
-import org.apache.fineract.infrastructure.core.service.DateUtils;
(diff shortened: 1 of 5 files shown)


ptuomola:
@vorburger Here are some more date changes to fix timezone issues with integration tests...

vorburger:
This is interesting.. why do we even have those 2 separate date functions? I'm keeping this PR open and not yet merged because I'm hoping to look s little bit more closely at those dates utilities - another day (too tired).

ptuomola:
@vorburger As far as I can see, DateUtils.getDateOfTenant is the "server side" version: it assumes it can find the Tenant object from a ThreadLocal variable initialised on the server. 

org.apache.fineract.integrationtests.common.Utils.getLocalDateOfTenant() is a "client side" version that is used only by the tests. It currently only returns dates based on a hardcoded timezone of "Asia/Kolkata" which matches the default tenant set-up used in the server. 

I think we'll always need a test client and a server side version of this. Of course the latter could be made smarter by e.g. retrieving the timezone from the server. But at the same time, given it's only used in integration tests and all the ITs use the same timezone, I'm not sure this would add a lot of value...

vorburger (approved):

PR #1060, opened 2020-06-14
FINERACT-821 Added and Enforced LeftCurly Checkstyle

thesmallstar:
I had to change the formatter.xml to enforce this change, most of the changes are directly enforced by spotless.

PR #1072, opened 2020-06-16
FINERACT-821 Added Rightcurly checkstyle

thesmallstar:
This PR is in continuation to #1060


thesmallstar:
So I had to modify the checkstyle a little bit:
I have modified from option => ALONE to ALONE_OR_SINGLELINE
Reason: 
Checkstyle would treat: 
function Xyz() {} //violation
I think @vorburger  you would agree to keep it this way rather than
function XyZ(){
}
Spotless love the first one too (With current config). Also if you are wondering that it would also allow: 
function Xyz() {  something } 
Nope, it won't, spotless would take care ;)
This is actually completely automated by spotless, but just an extra check, just in case.

vorburger (approved):
LGTM

PR #1075, opened 2020-06-17
FINERACT-822 Enforce StringSplitter check

percyashu:
FINERACT-822

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/ColumnValidator.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/ColumnValidator.java
@@ -122,6 +122,7 @@ public class ColumnValidator {
         return tableColumnMap;
     }
 
+    @SuppressWarnings("StringSplitter")
     private static Map<String, Set<String>> getTableColumnAliasMap(Set<String> operands) {
         Map<String, Set<String>> tableColumnMap = new HashMap<>();
         for (String operand : operands) {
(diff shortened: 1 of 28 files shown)


vorburger:
@awasum will you be doing a more detailed review of this PR? (Your GSoC student.)

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/ColumnValidator.java
```diff
@@ -122,6 +122,7 @@ public void validateSqlInjection(String schema, String... conditions) {
         return tableColumnMap;
     }
 
+    @SuppressWarnings("StringSplitter")
```
awasum, line 125: Why did you suppress this?

awasum (approved):
This follows the advice as seen here:https://errorprone.info/bugpattern/StringSplitter

The key thing is just looking the way the lists are accessed so make sure they correspond to the index of the arrays been refactored. I think this can be merged except am missing something... atleast the tests have passed.

@percyashu one little thing which was mentioned in the link above:
'`TIP: if you use Splitter, consider extracting the instance to a static final field.`

awasum:
@vorburger merge this when you can. Or maybe you want to take another quick look to make sure am not missing something. @ptuomola @xurror @thesmallstar  take a look.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/ColumnValidator.java
percyashu: Using the Splitter implementation in this method causes an errors in some integration test, I looked into but wasnt making any inroads to find out what was causing this. Will create an issue on issue tracker to remove the annotation.

PR #1079, opened 2020-06-17
Added and Enforced AvoidHidingCauseException Checkstyle (FINERACT-942)

thesmallstar:
FINERACT-942

Edit: For Discussions currently
According to the logging guidelines by @vorburger  I am quoting

> When catching exceptions, either rethrow them, or log them. Either way, include the root cause by using catch (SomeException e) and then either throw AnotherException("..details..", e) or LOG.error("...context...", e).

The AvoidHidingCauseException required that an exception that caused the exception "SomeException" must be thrown and not "AnotherException" as in guidelines.
Eg:
```
try{
throw type(x));
}catch(type(x))
{
throw type(y);
}
```
How should I follow on this?

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/campaigns/email/exception/EmailConfigurationNotFoundException.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/campaigns/email/exception/EmailConfigurationNotFoundException.java
@@ -19,6 +19,7 @@
 package org.apache.fineract.infrastructure.campaigns.email.exception;
 
 import org.apache.fineract.infrastructure.core.exception.AbstractPlatformResourceNotFoundException;
+import org.springframework.dao.EmptyResultDataAccessException;
 
 /**
  * A {@link RuntimeException} thrown when a code is not found.
@@ -28,4 +29,9 @@ public class EmailConfigurationNotFoundException extends AbstractPlatformResourc
     public EmailConfigurationNotFoundException(final String name) {
         super("error.msg.email.configuration.name.not.found", "Email configuration with name " + name + " does not exist", name);
     }
+
+    public EmailConfigurationNotFoundException(String name, EmptyResultDataAccessException e) {
+        super("error.msg.email.configuration.name.not.found", "Email configuration with name " + name + " does not exist", name, e);
+    }
+
 }
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exception/PlatformApiDataValidationException.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exception/PlatformApiDataValidationException.java
@@ -43,6 +43,21 @@ public class PlatformApiDataValidationException extends RuntimeException {
         this.errors = errors;
     }
 
+    public PlatformApiDataValidationException(final String globalisationMessageCode, final String defaultUserMessage,
+            final List<ApiParameterError> errors, Throwable cause) {
+        super(cause);
+        this.globalisationMessageCode = globalisationMessageCode;
+        this.defaultUserMessage = defaultUserMessage;
+        this.errors = errors;
+    }
+
+    public PlatformApiDataValidationException(final List<ApiParameterError> errors, Throwable cause) {
+        super(cause);
+        this.globalisationMessageCode = "validation.msg.validation.errors.exist";
+        this.defaultUserMessage = "Validation errors exist.";
+        this.errors = errors;
+    }
+
     public String getGlobalisationMessageCode() {
         return this.globalisationMessageCode;
     }
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exception/PlatformDataIntegrityException.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exception/PlatformDataIntegrityException.java
@@ -30,18 +30,20 @@ public class PlatformDataIntegrityException extends RuntimeException {
 
     public PlatformDataIntegrityException(final String globalisationMessageCode, final String defaultUserMessage,
             final Object... defaultUserMessageArgs) {
+        super(AbstractPlatformException.findThrowableCause(defaultUserMessageArgs));
         this.globalisationMessageCode = globalisationMessageCode;
         this.defaultUserMessage = defaultUserMessage;
         this.parameterName = null;
-        this.defaultUserMessageArgs = defaultUserMessageArgs;
+        this.defaultUserMessageArgs = AbstractPlatformException.filterThrowableCause(defaultUserMessageArgs);
     }
 
     public PlatformDataIntegrityException(final String globalisationMessageCode, final String defaultUserMessage,
             final String parameterName, final Object... defaultUserMessageArgs) {
+        super(AbstractPlatformException.findThrowableCause(defaultUserMessageArgs));
         this.globalisationMessageCode = globalisationMessageCode;
         this.defaultUserMessage = defaultUserMessage;
         this.parameterName = parameterName;
-        this.defaultUserMessageArgs = defaultUserMessageArgs;
+        this.defaultUserMessageArgs = AbstractPlatformException.filterThrowableCause(defaultUserMessageArgs);
     }
 
     public String getGlobalisationMessageCode() {
--- a/fineract-provider/src/main/java/org/apache/fineract/organisation/provisioning/exception/ProvisioningCriteriaNotFoundException.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/organisation/provisioning/exception/ProvisioningCriteriaNotFoundException.java
@@ -19,10 +19,15 @@
 package org.apache.fineract.organisation.provisioning.exception;
 
 import org.apache.fineract.infrastructure.core.exception.AbstractPlatformResourceNotFoundException;
+import org.springframework.dao.EmptyResultDataAccessException;
 
 public class ProvisioningCriteriaNotFoundException extends AbstractPlatformResourceNotFoundException {
 
     public ProvisioningCriteriaNotFoundException(final Long id) {
         super("error.msg.provisioning.criteria.id.invalid", "Provisioning Criteria with identifier " + id + " does not exist", id);
+    }
+
+    public ProvisioningCriteriaNotFoundException(Long id, EmptyResultDataAccessException e) {
+        super("error.msg.provisioning.criteria.id.invalid", "Provisioning Criteria with identifier " + id + " does not exist", id, e);
     }
 }
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanproduct/service/LoanProductWritePlatformServiceJpaRepositoryImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanproduct/service/LoanProductWritePlatformServiceJpaRepositoryImpl.java
@@ -345,26 +345,25 @@ public class LoanProductWritePlatformServiceJpaRepositoryImpl implements LoanPro
 
             final String externalId = command.stringValueOfParameterNamed("externalId");
             throw new PlatformDataIntegrityException("error.msg.product.loan.duplicate.externalId",
-                    "Loan Product with externalId `" + externalId + "` already exists", "externalId", externalId);
+                    "Loan Product with externalId `" + externalId + "` already exists", "externalId", externalId, realCause);
         } else if (realCause.getMessage().contains("'unq_name'")) {
 
             final String name = command.stringValueOfParameterNamed("name");
             throw new PlatformDataIntegrityException("error.msg.product.loan.duplicate.name",
-                    "Loan product with name `" + name + "` already exists", "name", name);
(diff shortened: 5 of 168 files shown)


vorburger:
Thanks for this Draft. I'll have a closer look at the build failure some time in the coming week, and propose how to address it. (Work on something else in the meantime.) Note to self - see https://github.com/sevntu-checkstyle/sevntu.checkstyle/blob/master/sevntu-checks/src/main/java/com/github/sevntu/checkstyle/checks/coding/AvoidHidingCauseExceptionCheck.java

vorburger:
> I'll have a closer look at the build failure some time in the coming week, and propose how to 

So most of this seems to be related to Fineract's concept of having its own "business" exceptions (like `RoleNotFoundException` and all those other many subclasses of `AbstractPlatformResourceNotFoundException`), and their constructors not accepting any `Throwable cause` argument.

If we do want to introduce this quality check (which I think we should, in order to prevent wrong exception handling for OTHER types of exception than those mentioned above), then I think the easiest way, and most "consistent", would be to simple have even those "business" exceptions accept a `cause` (unrelated to whether they really "need" it or not; I don't really see any "harm"). @ptuomola does this sound reasonable to you?

ptuomola:
@vorburger Makes sense to me - and who knows, there may be cases where having access to the original cause may help with debugging or troubleshooting. At very least, I can't see why it would cause any harm.

thesmallstar:
@vorburger @ptuomola 
So I will be doing this:

```
try{
//intead of throw new Y();
new X()
throw x.initcause(y);
}catch(x){
 //rethrow or do something.  
}
```
Problem:
What if Y is coming from some library function like EmptyResultDataAccessException which comes from JDBCtemplate.queryForObject(), in that case I could not figure out a way to do this. Can you add anything to this?(I hope I am making some sense here)

vorburger:
@thesmallstar no not with `throw x.initcause(y);` but by adding a `, Throwable cause` argument to `RoleNotFoundException` constructor etc.. would it help if I explained it better by raising a first small PR to illustrate what we mean here, which you can then apply everywhere?

vorburger:
> raising a first small PR to illustrate what we mean here, which you can then apply everywhere?

==> #1107

thesmallstar:
@vorburger  @xurror  can you please review this?
I have some points:- 
1. I used suppresswarning at two places( I thought the current approach is correct ).
2. Some Predefined exceptions were of the form: 
          throw new PredefinedException(x.messege()); // x was lost exception
this is converted to
          throw new PredefinedException(x); //x is throwable at the exception accepts it,

This was a a lot of manual work btw :P

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exception/PlatformApiDataValidationException.java
```diff
@@ -43,6 +43,30 @@ public PlatformApiDataValidationException(final String globalisationMessageCode,
         this.errors = errors;
     }
 
+    public PlatformApiDataValidationException(final String globalisationMessageCode, final String defaultUserMessage,
+            final List<ApiParameterError> errors, final Object... defaultUserMessageArgs) {
```
vorburger: In this case, where there is no `Object... defaultUserMessageArgs` in the original far, you should not introduce it. Just have constructors which take an additional last new argument like `, Throwable cause)` - makes sense?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exception/PlatformDataIntegrityException.java
```diff
@@ -30,18 +33,39 @@
...
             final String parameterName, final Object... defaultUserMessageArgs) {
+        super(findThrowableCause(defaultUserMessageArgs));
         this.globalisationMessageCode = globalisationMessageCode;
         this.defaultUserMessage = defaultUserMessage;
         this.parameterName = parameterName;
-        this.defaultUserMessageArgs = defaultUserMessageArgs;
+        this.defaultUserMessageArgs = filterThrowableCause(defaultUserMessageArgs);
+    }
+
+    private static Throwable findThrowableCause(Object[] defaultUserMessageArgs) {
```
vorburger: copy pasting (my) `findThrowableCause()` & `filterThrowableCause()` seems like a shame... always try to avoid copy/paste, as much as you can. How about introducing a new e.g. `public abstract class AbstractPlatformException extends RuntimeException` with suitable protected constructors, and these two methods as protected static methods?

fineract-provider/src/main/java/org/apache/fineract/organisation/provisioning/exception/ProvisioningCriteriaNotFoundException.java
```diff
@@ -19,10 +19,15 @@
...
+import org.springframework.dao.EmptyResultDataAccessException;
 
 public class ProvisioningCriteriaNotFoundException extends AbstractPlatformResourceNotFoundException {
 
     public ProvisioningCriteriaNotFoundException(final Long id) {
         super("error.msg.provisioning.criteria.id.invalid", "Provisioning Criteria with identifier " + id + " does not exist", id);
     }
+
+    public ProvisioningCriteriaNotFoundException(Long id, EmptyResultDataAccessException e) {
+        super("error.msg.provisioning.criteria.id.invalid", "Provisioning Criteria with identifier " + id + " does not exist", id);
```
vorburger: you forgot actually propagating the cause here... ;-)

```suggestion
        super("error.msg.provisioning.criteria.id.invalid", "Provisioning Criteria with identifier " + id + " does not exist", id, e);
```

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanproduct/service/LoanProductWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -357,9 +357,8 @@ private void handleDataIntegrityIssues(final JsonCommand command, final Throwabl
             throw new PlatformDataIntegrityException("error.msg.product.loan.duplicate.short.name",
                     "Loan product with short name `" + shortName + "` already exists", "shortName", shortName);
         } else if (realCause.getMessage().contains("Duplicate entry")) {
-            final Object[] args = null;
             throw new PlatformDataIntegrityException("error.msg.product.loan.duplicate.charge",
-                    "Loan product may only have one charge of each type.`", "charges", args);
+                    "Loan product may only have one charge of each type.`", "charges");
```
vorburger: I'm not 100% sure and haven't looked much at this, but could it make sense to:

```suggestion
                    "Loan product may only have one charge of each type.`", "charges", realCause);
```

vorburger (changes requested):
cool, thanks for this! Some feedback inline.

fineract-provider/src/main/java/org/apache/fineract/organisation/provisioning/exception/ProvisioningCriteriaNotFoundException.java
thesmallstar: nice catch! thanks

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanproduct/service/LoanProductWritePlatformServiceJpaRepositoryImpl.java
thesmallstar: Yes! makes sense :D

thesmallstar: Also adding the same in rest of the throws in the same function.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exception/PlatformApiDataValidationException.java
thesmallstar: okay! noted.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exception/PlatformDataIntegrityException.java
thesmallstar: Agreed! I did not understand do we really need a constructor though(or you mean something else here), an abstract class without any attributes and two static methods will do IMO.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/campaigns/email/exception/EmailConfigurationNotFoundException.java
```diff
@@ -28,4 +29,9 @@
     public EmailConfigurationNotFoundException(final String name) {
```
xurror, line 29: Just curious but wouldn't it be better to add the 3rd param here instead of duplicating the method?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/campaigns/email/exception/EmailConfigurationNotFoundException.java
```diff
@@ -28,4 +29,9 @@
...
         super("error.msg.email.configuration.name.not.found", "Email configuration with name " + name + " does not exist", name);
```
xurror, line 30: Just curious but wouldn't it be better to add the 3rd param here instead of duplicating the method? cause I can see the 2param method will just be history and unused.

thesmallstar: Nope, we cannot change the original implementation. If the cause of the exception is the exception itself, we need this constructor to be used.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exception/PlatformDataIntegrityException.java
thesmallstar: @vorburger  made the change I am unsure though if is this is what you meant.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exception/PlatformApiDataValidationException.java
vorburger: Ok now.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanproduct/service/LoanProductWritePlatformServiceJpaRepositoryImpl.java
vorburger: great, thanks; I've seen it - resolving

fineract-provider/src/main/java/org/apache/fineract/infrastructure/campaigns/email/exception/EmailConfigurationNotFoundException.java
vorburger: @thesmallstar @xurror the real question is really more if there are any (legitimate) remaining usages of the original constructor, without the cause. It seems like this was only used in `EmailConfigurationReadPlatformServiceImpl`, which was changed to use the new constructor, right? Then I actually agree that it would have been nicer and better to just add an additional parameter here... this isn't important enough to justify not finally merging this, so I'll probably go ahead, but I'll leave this open in case you'd like to clean this up in a new PR? (And perhaps in other exceptions where a cause was added - if there is no usage without a cause, it's better to simplify - and force future users to provide a cause.)

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exception/PlatformDataIntegrityException.java
vorburger: It's OK, although arguable now it's perhaps more of a... `PlatformExceptionUtil` than an `AbstractPlatformException`.. I'll merge it anyway, as this isn't super important, and perhaps I'll follow-up with a suggested improvement.

vorburger (approved):
LGTM

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exception/PlatformDataIntegrityException.java
vorburger: @thesmallstar I did this, and much more in #1125

fineract-provider/src/main/java/org/apache/fineract/infrastructure/campaigns/email/exception/EmailConfigurationNotFoundException.java
thesmallstar: Agreed! I will update this

thesmallstar: @vorburger  @xurror  making a new PR for this cleanup.

PR #1080, opened 2020-06-17
FINERACT-821 Added and Enforced FinalClass checkstyle

thesmallstar:
Refer: https://issues.apache.org/jira/browse/FINERACT-821

PR #1082, opened 2020-06-17
FINERACT-821 Added and Enforced ModifierOrder Checkstyle

thesmallstar:

I used the custom spotless code from: https://www.mail-archive.com/commits@geode.apache.org/msg17088.html

@vorburger I think you would love the idea to have the modifier order fixed by spotless as implemented in this, if not we can remove that feature :)


thesmallstar:
Please merge  #1080 before this, it would be easier to deal with merge conflicts(if any).

PR #1083, opened 2020-06-18
FINERACT-822 Enforce SameNameButDifferent check

percyashu:
FINERACT-822
https://errorprone.info/bugpattern/SameNameButDifferent


percyashu:
@awasum LGTY?

awasum:
I think we should leave this check disabled as it may mess with Swagger format especially when using it as a generator for API clients using something like Retrofit. I had an experience on another Fineract fork which when the names of classes are not very descriptive to the last details, one may get confused using these Swagger classes as a generator. I dont know if you understand me? @vorburger  FYI. In addition, this is not a major problem.

awasum:
> I think we should leave this check disabled as it may mess with Swagger format especially when using it as a generator for API clients using something like Retrofit. I had an experience on another Fineract fork which when the names of classes are not very descriptive to the last details, one may get confused using these Swagger classes as a generator. I dont know if you understand me? @vorburger FYI. In addition, this is not a major problem.

@percyashu did you want to do this?

vorburger:
@ptuomola do you have any input/view re. this and Swagger? (Given your Swagger related PR, which BTW I'll review tomorrow.)

ptuomola:
@vorburger Sorry no - I haven't looked into this check, so not sure how it will interact with Swagger. Given @awasum 's concern, perhaps we can first get Swagger to work (including generation of test clients) and then see if this makes sense?

awasum:
On second thought, lets reopen this and merge. When someone gets into the situation I had last year (which was in a private fork), then, they can always reverse this. Sorry for holding this PR. @percyashu .

awasum:
Ohh..Now there are conflicts.. Please @percyashu , will you be able to help me fix them? So sorry for giving you double work

percyashu:
Ok will fix conflict

PR #1092, opened 2020-06-20
FINERACT-822 Editing Checks list

percyashu:
Enable EmptyCatch (already enforced by @thesmallstar) and remove checks that were not include in Error Prone 2.4.0

### DIFF
--- a/fineract-provider/build.gradle
+++ b/fineract-provider/build.gradle
@@ -365,7 +365,6 @@ tasks.withType(JavaCompile) {
         disable(
                 "UnusedVariable",
                 "SameNameButDifferent",
-                "ImmutableEnumChecker",
                 "UnusedMethod",
                 "ModifiedButNotUsed",
                 "TypeParameterUnusedInFormals",
@@ -382,12 +381,10 @@ tasks.withType(JavaCompile) {
                 "ProtectedMembersInFinalClass",
                 "StaticAssignmentInConstructor",
                 "UnescapedEntity",
-                "EmptyCatch",
                 "InvalidBlockTag",
                 "ModifyCollectionInEnhancedForLoop",
                 "NonCanonicalType",
                 "InvalidInlineTag",
-                "InvalidBlockTag"
         )
         //TODO gradually enable these checks
         error(
@@ -419,6 +416,7 @@ tasks.withType(JavaCompile) {
                 "StringSplitter",
                 "AssertThrowsMultipleStatements",
                 "BoxedPrimitiveConstructor",
+                "EmptyCatch",
                 "BoxedPrimitiveEquality",
                 "SynchronizeOnNonFinalField",
                 "WildcardImport",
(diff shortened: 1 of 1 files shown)


fineract-provider/build.gradle
```diff
@@ -382,12 +381,10 @@ tasks.withType(JavaCompile) {
                 "ProtectedMembersInFinalClass",
                 "StaticAssignmentInConstructor",
                 "UnescapedEntity",
-                "EmptyCatch",
```
vorburger, line 385: Wait, what? No, this one I want... :smile: This is related to FINERACT-942 and https://checkstyle.sourceforge.io/config_blocks.html#EmptyCatchBlock @thesmallstar we should be able to enforce this by moving to error, agreed?

fineract-provider/build.gradle
```diff
@@ -444,13 +442,10 @@ tasks.withType(JavaCompile) {
                 "DoubleBraceInitialization",
                 "InconsistentCapitalization",
                 "MissingOverride",
-//                "ReturnMissingNullable",
```
vorburger, line 447: This https://errorprone.info/bugpattern/ReturnMissingNullable could actually be nice, for NPE safety (which is something I was quite passionate about, a while ago, see http://www.lastnpe.org) - but of course depends on the effort... how many violations are there for this?

fineract-provider/build.gradle
```diff
@@ -444,13 +442,10 @@ tasks.withType(JavaCompile) {
...
 //                "InconsistentOverloads",
 //                "MethodCanBeStatic",
 //                "Var",
 //                "ConstantField",
-//                "ConstructorInvokesOverridable",
```
vorburger, line 452: see https://errorprone.info/bugpattern/ConstructorInvokesOverridable .. how many of this are there, is this hard to fix? I think you may be able to easily fix it just by making something methods that are called from constructors `private` and/or `final`...

vorburger (changes requested):

fineract-provider/build.gradle
percyashu: It was removed for version 2.4.0

fineract-provider/build.gradle
percyashu: same as ReturnMissingNullable

fineract-provider/build.gradle
percyashu: Yep will move it

fineract-provider/build.gradle
percyashu: it fails with ConstructorInvokesOverridable is not a valid checker name

fineract-provider/build.gradle
```diff
@@ -382,12 +381,10 @@ tasks.withType(JavaCompile) {
                 "ProtectedMembersInFinalClass",
                 "StaticAssignmentInConstructor",
                 "UnescapedEntity",
-                "EmptyCatch",
                 "InvalidBlockTag",
                 "ModifyCollectionInEnhancedForLoop",
                 "NonCanonicalType",
                 "InvalidInlineTag",
-                "InvalidBlockTag"
```
percyashu, line 390: Seems check was disable in 2.4.0 and must be enabled as if necessary

vorburger (approved):

PR #1094, opened 2020-06-20
increase auto formatting comment line length from 80 to 120

vorburger:
@thesmallstar @ptuomola what's your view on changing `org.eclipse.jdt.core.formatter.comment.line_length` in `config/fineractdev-formatter.xml` from 80 to 120 like this?


thesmallstar:
Yes, this is better :)

PR #1099, opened 2020-06-21
FINERACT-822 Enforce UndefinedEquals check

percyashu:
FINERACT-822
https://errorprone.info/bugpattern/UndefinedEquals


percyashu:
@awasum @vorburger LGTY?

vorburger (commented):
@awasum will you review this one?

fineract-provider/src/main/java/org/apache/fineract/accounting/glaccount/domain/TrialBalance.java
```diff
@@ -100,8 +100,9 @@ public boolean equals(Object obj) {
         }
         TrialBalance other = (TrialBalance) obj;
         return Objects.equals(other.officeId, officeId) && Objects.equals(other.glAccountId, glAccountId)
-                && Objects.equals(other.amount, amount) && Objects.equals(other.entryDate, entryDate)
-                && Objects.equals(other.transactionDate, transactionDate) && Objects.equals(other.closingBalance, closingBalance);
+                && Objects.equals(other.amount, amount) && other.entryDate.compareTo(entryDate) == 0 ? Boolean.TRUE
+                        : Boolean.FALSE && other.transactionDate.compareTo(transactionDate) == 0 ? Boolean.TRUE
```
thesmallstar, line 104: Just curious about this:
How is `other.entryDate.compareTo(entryDate) == 0 ? Boolean.TRUE: Boolean.FALSE`
dIfferent from just  `other.entryDate.compareTo(entryDate) == 0 `  (In terms of end result) of the boolean?
Is it because .equals returns Boolean.TRUE and not primitive data-type boolean?

fineract-provider/src/main/java/org/apache/fineract/portfolio/shareaccounts/domain/ShareAccount.java
```diff
@@ -224,7 +224,7 @@ public boolean setSubmittedDate(final Date submittedDate) {
 
     public boolean setApprovedDate(final Date approvedDate) {
         boolean toReturn = false;
-        if (!this.approvedDate.equals(approvedDate)) {
+        if (this.approvedDate.compareTo(approvedDate) == 0 ? Boolean.FALSE : Boolean.TRUE) {
             this.approvedDate = approvedDate;
```
thesmallstar, line 228: very very minor suggestion, can we not keep it just !=0 (this makes it little confusing IMO).

thesmallstar (approved):
LGTM

fineract-provider/src/main/java/org/apache/fineract/accounting/glaccount/domain/TrialBalance.java
percyashu: compareTo returns an int. If values to be compared are equal, the value returned is zero.

fineract-provider/src/main/java/org/apache/fineract/portfolio/shareaccounts/domain/ShareAccount.java
percyashu: Yeah, but I think this way is clearer.

fineract-provider/src/main/java/org/apache/fineract/accounting/glaccount/domain/TrialBalance.java
```diff
@@ -100,8 +100,9 @@ public boolean equals(Object obj) {
         }
         TrialBalance other = (TrialBalance) obj;
         return Objects.equals(other.officeId, officeId) && Objects.equals(other.glAccountId, glAccountId)
-                && Objects.equals(other.amount, amount) && Objects.equals(other.entryDate, entryDate)
-                && Objects.equals(other.transactionDate, transactionDate) && Objects.equals(other.closingBalance, closingBalance);
+                && Objects.equals(other.amount, amount) && other.entryDate.compareTo(entryDate) == 0 ? Boolean.TRUE
```
vorburger, line 103: @percyashu I don't understand this... given that `entryDate` and `transactionDate` are (both) `java.util.Date`, why is `Objects.equals()` not OK? That does not match what https://errorprone.info/bugpattern/UndefinedEquals seems to be about... what am I missing?

vorburger (changes requested):
let's clarify why Date needs `compareTo` instead of `equals()` before we merge this?

fineract-provider/src/main/java/org/apache/fineract/accounting/glaccount/domain/TrialBalance.java
thesmallstar: Agreed, so if other.entryDate.compareTo(entryDate) == 0 is executed, then 0 == 0 returns true, which is boolean. 
Sorry if I am confusing, I am not able to understand the need for the
 ```?Boolean.TRUE: Boolean.FALSE ``` part.

fineract-provider/src/main/java/org/apache/fineract/accounting/glaccount/domain/TrialBalance.java
percyashu: Error Prone kept complaining java.util.Date equals() doesn't seem to work as expected. This is because from [this](https://docs.oracle.com/javase/6/docs/api/java/util/Date.html#equals(java.lang.Object)) two Date objects are equal if and only if the getTime method returns the same long value for both.

fineract-provider/src/main/java/org/apache/fineract/accounting/glaccount/domain/TrialBalance.java
percyashu: This is just like a Java If statement where other.entryDate.compareTo(entryDate) == 0 is the condition, the first value Boolean.TRUE is return if the condition is true and the second value Boolean.FALSE if it is false. Could have used an If statement but this is shorter. See [this](https://www.baeldung.com/java-ternary-operator)

vorburger (approved):
@percyashu The `Date.equals()` story intrigued me, so I pulled this PR and "undid" one of your change, and indeed, it causes:

```
> Task :compileJava
/home/vorburger/Mifos/fineract/fineract-provider/src/main/java/org/apache/fineract/accounting/glaccount/domain/TrialBalance.java:103: error: [UndefinedEquals] Date does not have well-defined equals behavior.
                && Objects.equals(other.amount, amount) && Objects.equals(other.entryDate, entryDate)
                                                                         ^
    (see https://errorprone.info/bugpattern/UndefinedEquals)
```

https://errorprone.info/bugpattern/UndefinedEquals doesn't explain this, yet - but https://github.com/google/error-prone/commit/9b2ed01e367929a787aa4518a80bd90890fda13d clarifies why. (It's because Date is not `final`, the `equals()` could be wrong - interesting. I've raised https://github.com/google/error-prone/pull/1699 with a minor broken link fixed in the doc.)

I'm now happy to LGTM this! Thanks for your insistence and patience.

Will you resolve merge conflict, so that we can get this in?

fineract-provider/src/main/java/org/apache/fineract/accounting/glaccount/domain/TrialBalance.java
vorburger: @percyashu what @thesmallstar is pointing out here, and I agree with, is that `? Boolean.TRUE: Boolean.FALSE` is not required - as he said, just `other.entryDate.compareTo(entryDate) == 0` yields exactly the same value. The difference between primitive VS non-primitive is hidden by "auto boxing". Do you want to raise a follow-up PR to clean this up, or ignore this?

PR #1100, opened 2020-06-22
FINERACT-1047 Integration test for Audit trails(Create)

thesmallstar:
FINERACT-1047

## Description
Integration Tests are divided into two parts:
1. Creation 
2. Verifying for different routes provided, like: limit,orderby etc. 

This PR deals with the first part.

## Checklist

### DIFF
--- a/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/AuditIntegrationTest.java
+++ b/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/AuditIntegrationTest.java
@@ -1,0 +1,105 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.integrationtests;
+
+import io.restassured.builder.RequestSpecBuilder;
+import io.restassured.builder.ResponseSpecBuilder;
+import io.restassured.http.ContentType;
+import io.restassured.specification.RequestSpecification;
+import io.restassured.specification.ResponseSpecification;
+import java.util.HashMap;
+import java.util.List;
+import org.apache.fineract.integrationtests.common.AuditHelper;
+import org.apache.fineract.integrationtests.common.ClientHelper;
+import org.apache.fineract.integrationtests.common.OfficeHelper;
+import org.apache.fineract.integrationtests.common.Utils;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+/**
+ *
+ * @author Manthan Surkar
+ *
+ */
+public class AuditIntegrationTest {
+
+    private ResponseSpecification responseSpec;
+    private RequestSpecification requestSpec;
+    private ClientHelper clientHelper;
+    private AuditHelper auditHelper;
+
+    /**
+     * Sets up the essential settings for the TEST like contentType, expectedStatusCode. It uses the '@BeforeEach'
+     * annotation provided by jUnit.
+     */
+    @BeforeEach
+    public void setup() {
+        Utils.initializeRESTAssured();
+        this.requestSpec = new RequestSpecBuilder().setContentType(ContentType.JSON).build();
+        this.requestSpec.header("Authorization", "Basic " + Utils.loginIntoServerAndGetBase64EncodedAuthenticationKey());
+        this.responseSpec = new ResponseSpecBuilder().expectStatusCode(200).build();
+        this.auditHelper = new AuditHelper(this.requestSpec, this.responseSpec);
+        this.clientHelper = new ClientHelper(this.requestSpec, this.responseSpec);
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/AuditHelper.java
```diff
@@ -0,0 +1,49 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(AuditHelper.class);
+    private static final String AUDIT_BASE_URL = "/fineract-provider/api/v1/audits?" + Utils.TENANT_IDENTIFIER;
+
+    public static void verifyAuditCreatedOnServer(final RequestSpecification requestSpec, final ResponseSpecification responseSpec,
+            final Integer resourceId, final String actionName, final String entityName) {
+        LOG.info("------------------------------CHECK IF AUDIT CREATED------------------------------------\n");
+        final String AUDIT_URL = "/fineract-provider/api/v1/audits/?" + "entityName=" + entityName + "&resourceId=" + resourceId
+                + "&actionName=" + actionName + "&" + Utils.TENANT_IDENTIFIER;
+        final ArrayList<Integer> responseAuditIDs = Utils.performServerGet(requestSpec, responseSpec, AUDIT_URL, "id");
```
vorburger: very minor feedback FYI, more for your learning: We usually use the abstract `List` instead of the concrete `ArrayList` implementation as type of the variable. Even better, since we are now on Java 11, why not start using `var` instead? Try just `var responseAuditIDs = Utils.performServerGet(requestSpec, responseSpec, AUDIT_URL, "id");` (and don't use `final var`, that's just ugly).

```suggestion
        final List<Integer> responseAuditIDs = Utils.performServerGet(requestSpec, responseSpec, AUDIT_URL, "id");
```

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/AuditIntegrationTest.java
```diff
@@ -0,0 +1,85 @@
...
+     * Here we Create/Update different Entities and verify an audit is generated
+     * for each action. This can be further extened with more entities and
+     * actions in similiar way.
+     */
+    @Test
+    public void auditShouldbeCreated() {
+        this.clientHelper = new ClientHelper(this.requestSpec, this.responseSpec);
+
+        // When Client is created
+        final Integer clientId = ClientHelper.createClient(this.requestSpec, this.responseSpec);
```
vorburger: as we've started discussing in #1019, it's actually kind of very wrong and confusing how classes such as `ClientHelper` and many other existing *Helper, take `this.requestSpec, this.responseSpec` BOTH as constructor arguments AND for their methods - that makes no sense; I should be either/or. That's not really the point of this PR though - just thought I'd let you know here for illustration.

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/AuditIntegrationTest.java
```diff
@@ -0,0 +1,85 @@
...
+        Assertions.assertNotNull(clientId);
+        AuditHelper.verifyAuditCreatedOnServer(this.requestSpec, this.responseSpec, clientId, "CREATE", "CLIENT");
```
vorburger: For a new class like `AuditHelper`, I would go with the pattern of passing `this.requestSpec, this.responseSpec` to the constructor, and then have a method like `verifyAuditCreatedOnServer` not require it. Actually, even better, how about only passing the `requestSpec` and even just hard-code the `responseSpec` without having to pass it. This would make ITs more readable.

```suggestion
        AuditHelper.verifyAuditCreatedOnServer(this.requestSpec, this.responseSpec, clientId, "CREATE", "CLIENT");
```

vorburger (commented):
@thesmallstar I have some minor feedback. This is more about improving style of ITs.

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/AuditIntegrationTest.java
thesmallstar: Yes, I thought about it, but I thought not to mix two PRs since it has to covered in the other PR anyways.

vorburger: SGTM - resolving this conversation, you are absolutely right, fixing that is not part of this PR!

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/AuditIntegrationTest.java
thesmallstar: I was actually going to point this on #1019, HideUtilityClassConstructor checks if it is utility class by its name, to do this(using non-private constructor) we would need to either suppress the warning for that checkstyle then or call this something that does not have Helper or util in its name. What do you suggest?

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/common/AuditHelper.java
thesmallstar: understood!

thesmallstar:
@vorburger  I was a little worried about this approach though and thought of changing this to something more robust.

The problem: 
We are checking count for the entries like this: Get me the count of audit trails whose resource-id is x it was activated and is of type client. Here, the same event can occur twice and hence two audits will be returned, which does not necessarily mean that the test failed(since we are checking count to be equal to one)

The solution:
Calculate the count before the action(create, update, activate) and check if count incremented by one, after the action. 

Why the old approach is still correct?
Two reasons: Every entity is created for the first time, no action is performed twice. 


Just that I thought it was fundamentally wrong to check count to be one(not a general case), I hope I am not too confusing. Which approach will you prefer?

vorburger:
@thesmallstar yes that's a very good point.. you are absolutely right, of course, this test could fail if it runs against a DB that already contains data, for example if it runs after other tests, which makes it potentially flaky and brittle. Change it like you suggested! Maybe instead of only testing that it's +1 more, you could assert on the content of the audit record itself as well?

PS: There's a bigger issue about test isolation underlying this, but we're not going to solve this here.

vorburger (changes requested):
improve as discussed

thesmallstar:
@vorburger I have made the suggested changes. I am not supposed to user var in this case(right?).

vorburger (approved):

PR #1123, opened 2020-06-29
FINERACT-854 Removed string concatenated SQL from CenterReadPlatform

thesmallstar:
Refer: https://issues.apache.org/jira/browse/FINERACT-854 and #725 #723 for background.
The work for this part is completed, but SQLbuilder currently does not support the use of "limit" and "order by" query which I will be adding before this can be merged.

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/group/service/CenterReadPlatformServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/group/service/CenterReadPlatformServiceImpl.java
@@ -28,11 +28,9 @@ import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
-import org.apache.commons.lang3.StringUtils;
 import org.apache.fineract.infrastructure.codes.data.CodeValueData;
 import org.apache.fineract.infrastructure.codes.service.CodeValueReadPlatformService;
 import org.apache.fineract.infrastructure.configuration.domain.ConfigurationDomainService;
-import org.apache.fineract.infrastructure.core.api.ApiParameterHelper;
 import org.apache.fineract.infrastructure.core.data.ApiParameterError;
 import org.apache.fineract.infrastructure.core.data.DataValidatorBuilder;
 import org.apache.fineract.infrastructure.core.data.EnumOptionData;
@@ -46,7 +44,7 @@ import org.apache.fineract.infrastructure.core.service.RoutingDataSource;
 import org.apache.fineract.infrastructure.core.service.SearchParameters;
 import org.apache.fineract.infrastructure.security.service.PlatformSecurityContext;
 import org.apache.fineract.infrastructure.security.utils.ColumnValidator;
-import org.apache.fineract.infrastructure.security.utils.SQLInjectionValidator;
+import org.apache.fineract.infrastructure.security.utils.SQLBuilder;
 import org.apache.fineract.organisation.office.data.OfficeData;
 import org.apache.fineract.organisation.office.service.OfficeReadPlatformService;
 import org.apache.fineract.organisation.staff.data.StaffData;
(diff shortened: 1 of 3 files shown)


fineract-provider/src/main/java/org/apache/fineract/portfolio/group/service/CenterReadPlatformServiceImpl.java
```diff
@@ -122,55 +122,26 @@ public CenterReadPlatformServiceImpl(final PlatformSecurityContext context, fina
...
+        SQLBuilder extraCriteria = new SQLBuilder();
+        extraCriteria.addCriteria("g.level_id =", GroupTypes.CENTER.getId());
         if (searchCriteria != null) {
             String sqlQueryCriteria = searchCriteria.getSqlSearch();
             if (StringUtils.isNotBlank(sqlQueryCriteria)) {
                 SQLInjectionValidator.validateSQLInput(sqlQueryCriteria);
                 sqlQueryCriteria = sqlQueryCriteria.replace(" display_name ", " g.display_name ");
                 sqlQueryCriteria = sqlQueryCriteria.replace("display_name ", "g.display_name ");
-                extraCriteria.append(" and (").append(sqlQueryCriteria).append(") ");
-                this.columnValidator.validateSqlInjection(schemaSl, sqlQueryCriteria);
```
thesmallstar, line 136: Also removing this part is not correct as of NOW, the problem is we are talking SQL query as a query to give users more functionality and certainly that is not supported in SQL builder currently, we probably would need to parse things find relevant queries and then add those to extra criteria still thinking on it. 
Will be looked after limit and order by are supported.

thesmallstar:
Blocked by: https://issues.apache.org/jira/browse/FINERACT-1058

vorburger:
```
CenterIntegrationTest > testListCenters() FAILED
    java.lang.AssertionError: 1 expectation failed.
    Expected status code <200> but was <500>.```

File: fineract-provider/src/main/java/org/apache/fineract/portfolio/group/service/CenterReadPlatformServiceImpl.java (hunk 1 above)
Comment (thesmallstar) [reply fineract-provider/src/main/java/org/apache/fineract/portfolio/group/service/CenterReadPlatformServiceImpl.java:136]: @vorburger  WDYT on this?

**Comment by thesmallstar** [comment]
Failed due to https://issues.apache.org/jira/browse/FINERACT-1016
Looking into it.

**Comment by thesmallstar** [comment]
@vorburger @awasum if you like this approach Before we merge this I will add the tests for the same, if not do you have any suggestions?

**Review by vorburger — CHANGES_REQUESTED** [review]
I'm OK with something like this, but have several feedbacks:

1. I don't know if this actually works (PreparedStatement), so will have to trust you that you test that it really does.

2. The current implementation seems to assume that `setLimit()` & Co. would only ever be called after `addCriteria()` - right? That's... not a good idea. Imagine some dumb developer coming along in a few months, not knowing how you implemented this internally (and correctly so, they should not have to). So they do `new SQLBuilder().addOrderBy().addCriteria(...).getSQLTemplate()` - and that would create invalid SQL - agreed?

3. `SQLBuidlerTest` must be extended to cover these new methods.

I'll re-review this week after next (vacation).

File: fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/SQLBuilder.java [hunk 2]
```diff
@@ -110,6 +111,39 @@ public void addNonNullCriteria(String criteria, Object argument) {
...
+        args.add(offset);
+    }
+
+    /**
+     * Adds the orderBy statement in the end of WHERE clause
+     *
+     * @param orderBy
+     *            The value that will be used as orderBy
+     */
+    public void addOrderBY(Object orderBy) {
```
vorburger: lower-case _y_:

```suggestion
    public void addOrderBy(Object orderBy) {
```

fineract-provider/src/main/java/org/apache/fineract/portfolio/group/service/CenterReadPlatformServiceImpl.java
vorburger: @thesmallstar I'm not sure I fully understand what you mean here, but starting to think about adding parsing things sounds like the wrong direction - try not to have to do that (it will be a mess - trust me).

thesmallstar:
> 1. I don't know if this actually works (PreparedStatement), so will have to trust you that you test that it really does.

-> Yes it works! I have tested it :) 
> 2. The current implementation seems to assume that `setLimit()` & Co. would only ever be called after `addCriteria()` - right? That's... not a good idea. Imagine some dumb developer coming along in a few months, not knowing how you implemented this internally (and correctly so, they should not have to). So they do `new SQLBuilder().addOrderBy().addCriteria(...).getSQLTemplate()` - and that would create invalid SQL - agreed?

->Agreed I am making the change. 

> 3. `SQLBuidlerTest` must be extended to cover these new methods.

->Extending, I was waiting for the approach to be marked correct :)

thesmallstar:
failed due to connection error. Closing and opening this PR to retest.

thesmallstar:
I am removing the part to add limit and orderby to SQL builder separate from this PR, so as to review this properly, and also we can merge this quickly then and not keep it blocked by the work and testing on that part.

fineract-provider/src/main/java/org/apache/fineract/portfolio/group/service/CenterReadPlatformServiceImpl.java
vorburger: @thesmallstar sorry for the huge delay in getting back to you on this one. I've re-read it again now (took me a minute to get back into it). Are you suggesting that we merge this now? Because, unless I misunderstand, this would break the currently existing functionality for these sqlQueryCriteria, agreed? But they really are a problem, huh? I need to dig more into the code to understand where this is coming from..

vorburger (changes requested):

fineract-provider/src/main/java/org/apache/fineract/portfolio/group/service/CenterReadPlatformServiceImpl.java
vorburger: We can't (shouldn't) merge this PR as is, because it will break support for `sqlSearch`. We need to EITHER support it here, OR (my preference) should just cleanly remove it all together - as (now) suggested in https://issues.apache.org/jira/browse/FINERACT-1095.

xurror:
@thesmallstar any updates on this?

thesmallstar:
This build will fail now, will pass only after rebase from #1171

thesmallstar:
@vorburger review/merge this?
Removed SQLsearch
Fixed small error in groups
SQL concatenation remove ;)

vorburger (approved):
LGTM

PR #1131, opened 2020-06-30
FINERACT-822 Enforcing Checks

percyashu:
FINERACT-822
Enforce StaticAssignmentInConstructor check
Enable ProtectedMembersInFinalClass check

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/domain/Report.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/domain/Report.java
@@ -111,7 +111,7 @@ public final class Report extends AbstractPersistableCustom {
         return new Report(reportName, reportType, reportSubType, reportCategory, description, useReport, reportSql, reportTypes);
     }
 
-    protected Report() {
+    Report() {
         //
     }
 
--- a/fineract-provider/src/test/java/org/apache/fineract/common/RestAssuredFixture.java
+++ b/fineract-provider/src/test/java/org/apache/fineract/common/RestAssuredFixture.java
@@ -32,14 +32,16 @@ import io.restassured.specification.ResponseSpecification;
  */
 public class RestAssuredFixture {
 
-    private final int httpPort = 8443;
+    private final int httpPort;
 
     private ResponseSpecification responseSpec;
     private RequestSpecification requestSpec;
 
-    public RestAssuredFixture() {
+    public RestAssuredFixture(int http) {
         super();
+        this.httpPort = http;
         Utils.initializeRESTAssured();
+        Utils.initializeRESTAssuredPort(http);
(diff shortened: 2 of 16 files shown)


fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/domain/Report.java
```diff
@@ -111,7 +111,7 @@ public static Report fromJson(final JsonCommand command, final Collection<String
         return new Report(reportName, reportType, reportSubType, reportCategory, description, useReport, reportSql, reportTypes);
     }
 
-    protected Report() {
+    Report() {
```
vorburger, line 114: @percyashu couldn't this just be removed, instead of made package local?

fineract-provider/src/test/java/org/apache/fineract/common/RestAssuredFixture.java
```diff
@@ -33,16 +32,14 @@
...
 
     private ResponseSpecification responseSpec;
     private RequestSpecification requestSpec;
 
-    public RestAssuredFixture(int httpPort) {
+    public RestAssuredFixture() {
         super();
-        this.httpPort = httpPort;
         Utils.initializeRESTAssured();
-        RestAssured.port = httpPort;
```
vorburger, line 45: I was quite surprised that you were able to remove this - how come? Is RestAssured's default 8443? Isn't it better to keep this explicit?

vorburger (changes requested):
Some Feedback. Also IT Test Failure FYI just due to FINERACT-1016.

fineract-provider/src/test/java/org/apache/fineract/common/RestAssuredFixture.java
percyashu: RestAssured.port is a static field  hence picked up by StaticAssignmentInConstructor check. 
I removed  that line because Utils.initializeRESTAssured() sets the port to 8443.
I could try creating a method in Utils to initialize RestAssured with specific port or just revert this leave the check disabled.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/domain/Report.java
percyashu: Sure will remove them

percyashu: I look at these and JPA requires entities to have a no-args constructo .

fineract-provider/src/test/java/org/apache/fineract/common/RestAssuredFixture.java
percyashu: Does this makes it more explicit

vorburger (approved):
LGTM

PR #1149, opened 2020-07-04
FINERACT-821 Added and Enforced LocalVariableName Checkstyle

thesmallstar:
I have also changed the regex in checkstyle.xml to support single-letter variables.
Since:
int i; // was not supported :(

### DIFF
--- a/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/AccountingScenarioIntegrationTest.java
+++ b/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/AccountingScenarioIntegrationTest.java
@@ -802,10 +802,10 @@ public class AccountingScenarioIntegrationTest {
 
         float totalInterest = (float) loanSchedule.get(1).get("interestOriginalDue");
         DecimalFormat numberFormat = new DecimalFormat("#.00", new DecimalFormatSymbols(Locale.US));
-        float INTEREST_4_DAYS = totalInterest / totalDaysInPeriod * 4;
-        INTEREST_4_DAYS = Float.valueOf(numberFormat.format(INTEREST_4_DAYS));
+        float interest4Days = totalInterest / totalDaysInPeriod * 4;
+        interest4Days = Float.valueOf(numberFormat.format(interest4Days));
 
-        this.loanTransactionHelper.checkAccrualTransactionForRepayment(getDateAsLocalDate(currentDate), INTEREST_4_DAYS, FEE_PORTION,
+        this.loanTransactionHelper.checkAccrualTransactionForRepayment(getDateAsLocalDate(currentDate), interest4Days, FEE_PORTION,
                 PENALTY_PORTION, loanID);
 
     }
@@ -906,15 +906,15 @@ public class AccountingScenarioIntegrationTest {
 
         float totalInterest = (float) loanSchedule.get(1).get("interestOriginalDue");
         DecimalFormat numberFormat = new DecimalFormat("#.00", new DecimalFormatSymbols(Locale.US));
-        float INTEREST_3_DAYS = totalInterest / totalDaysInPeriod * 3;
-        INTEREST_3_DAYS = Float.valueOf(numberFormat.format(INTEREST_3_DAYS));
-        this.loanTransactionHelper.checkAccrualTransactionForRepayment(getDateAsLocalDate(runOndate), INTEREST_3_DAYS, FEE_PORTION,
+        float interest3Days = totalInterest / totalDaysInPeriod * 3;
+        interest3Days = Float.valueOf(numberFormat.format(interest3Days));
+        this.loanTransactionHelper.checkAccrualTransactionForRepayment(getDateAsLocalDate(runOndate), interest3Days, FEE_PORTION,
                 PENALTY_PORTION, loanID);
 
         runOndate = dateFormat.format(todayDate.getTime());
 
         this.periodicAccrualAccountingHelper.runPeriodicAccrualAccounting(runOndate);
-        float interestPerDay = (totalInterest / totalDaysInPeriod * 4) - INTEREST_3_DAYS;
+        float interestPerDay = (totalInterest / totalDaysInPeriod * 4) - interest3Days;
         interestPerDay = Float.valueOf(numberFormat.format(interestPerDay));
         this.loanTransactionHelper.checkAccrualTransactionForRepayment(getDateAsLocalDate(runOndate), interestPerDay, NEXT_FEE_PORTION,
                 NEXT_PENALTY_PORTION, loanID);
--- a/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/ClientLoanIntegrationTest.java
+++ b/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/ClientLoanIntegrationTest.java
@@ -3195,8 +3195,8 @@ public class ClientLoanIntegrationTest {
         String prepayAmount = String.valueOf(prepayDetail.get("amount"));
         validateNumberForEqualWithMsg("verify pre-close amount", "3551.93", prepayAmount);
         todaysDate = Calendar.getInstance(Utils.getTimeZoneOfTenant());
-        String LOAN_REPAYMENT_DATE = dateFormat.format(todaysDate.getTime());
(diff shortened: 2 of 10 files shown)


thesmallstar:
failed due to connection error, closing and reopening this.

thesmallstar:
I know it is not a very nice idea to close and reopen again :( but build failed again due to connection error.

xurror:
No problem. Resolve merge conflicts and I can merge this.

xurror (approved):

thesmallstar:
> I have also changed the regex in checkstyle.xml to support single-letter variables.
> Since:
> int i; // was not supported :(

-> @xurror  @awasum  any contradicting thoughts on this?

thesmallstar:
@xurror  donot merge this yet, another way to fix this violation was to add a final keyword, which would fix most of this violations without change in variable name.

On the other hand, AbbreviationAsWordInName did not ignore final variable and did not allow XYZ_ABC format.

Currently I am looking into AbbreviationAsWordInName's property to ignore final and then we can decide on what would be the best thing to do.

IMO, we should instead just add the final keyword which would fix nearly all this violations and this variables dont change the value once assigned.

thesmallstar:
@xurror  ready for review.

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/ClientSavingsIntegrationTest.java
```diff
@@ -257,22 +257,22 @@ public void testSavingsAccount_WITH_ENFORCE_MIN_BALANCE() {
 
         DateFormat dateFormat = new SimpleDateFormat("dd MMMM yyyy");
         Calendar todaysDate = Calendar.getInstance();
-        final String TRANSACTION_DATE = dateFormat.format(todaysDate.getTime());
+        final String transactionDateValue = dateFormat.format(todaysDate.getTime());
```
xurror: Did this really need to be renamed? Seems like FOO_BAR format worked pretty well on other local variables.
There are other variables you've renamed here too. I think the changes are cool. Maybe just clarify me on this rename.

thesmallstar: This probably is a side-effect of mass find-replace, thank you for finding this, updating.

vorburger: I actually think this was right (good), because all upper case variable names (on Java), by convention, are ONLY for true static final constants (which to this is not)..

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/AccountingScenarioIntegrationTest.java
```diff
@@ -802,10 +802,10 @@ public void checkPeriodicAccrualAccountingTillCurrentDateFlow() throws Interrupt
 
         float totalInterest = (float) loanSchedule.get(1).get("interestOriginalDue");
         DecimalFormat numberFormat = new DecimalFormat("#.00", new DecimalFormatSymbols(Locale.US));
-        float INTEREST_4_DAYS = totalInterest / totalDaysInPeriod * 4;
-        INTEREST_4_DAYS = Float.valueOf(numberFormat.format(INTEREST_4_DAYS));
+        float interest4Days = totalInterest / totalDaysInPeriod * 4;
+        interest4Days = Float.valueOf(numberFormat.format(interest4Days));
```
vorburger, line 806: I don't understand what this does (why it's needed) - but it has nothing to do with the Checkstyle change..

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/ClientSavingsIntegrationTest.java
thesmallstar: Edit: I am speaking wrong here, please ignore. -> See next comment. 

@vorburger  in that case:
This would clash with AbbreviationAsWordInName checkstyle which only allows up to 4 capital letters in final variables. 
What should we go with?

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/ClientLoanIntegrationTest.java
```diff
@@ -3195,7 +3195,7 @@ public void testLoanScheduleWithInterestRecalculation_WITH_REST_SAME_AS_REPAYMEN
         String prepayAmount = String.valueOf(prepayDetail.get("amount"));
         validateNumberForEqualWithMsg("verify pre-close amount", "3551.93", prepayAmount);
         todaysDate = Calendar.getInstance(Utils.getTimeZoneOfTenant());
-        String LOAN_REPAYMENT_DATE = dateFormat.format(todaysDate.getTime());
+        final String LOAN_REPAYMENT_DATE = dateFormat.format(todaysDate.getTime());
```
vorburger: I would actually make this lower case, because it's not a constant.

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/ClientLoanIntegrationTest.java
```diff
@@ -4843,7 +4843,7 @@ public void testLoanRefundByTransferCashBasedAccounting() {
         // refund 20 means paid: principal 1980, interest 240, fees 50, penalty
         // 0
 
-        Float TRANSFER_AMOUNT = 20f;
+        Float transferAmountValue = 20f;
```
vorburger, line 4846: If this is a real constant, make it final and upper case?

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/AccountingScenarioIntegrationTest.java
thesmallstar: LocalVariableName would only allow the name (local var,unless final) to have following type of regex: ^[a-z][a-z0-9][a-zA-Z0-9]*$|[a-z]

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/ClientLoanIntegrationTest.java
thesmallstar: No it is not, it is changed a few lines later 
https://github.com/apache/fineract/blob/acd9de407f29cf1ba90f682b13fc03c55916f4e2/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/ClientLoanIntegrationTest.java#L4887

vorburger: Ah ok fine, then ignore this (I'm doing some late night review on my mobile phone)

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/AccountingScenarioIntegrationTest.java
vorburger: No I just meant the Float value format thing in the 2nd line that looks weird, to me.

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/ClientSavingsIntegrationTest.java
thesmallstar: Ok agreed, making the changes.

thesmallstar:
@vorburger What have I done here is:
1. If a variable is not changed after initialisation I have made it final, and then it allows XYZ_ABC type of naming style.
2. If a variable is changed and we cannot make it final, then we need to have to follow camelcasing (abcXyz) . 

So, if a var is not changed I have made it final. Else I have forced it to follow the required regex. 

Since we want only static final to have XYZ_ABC naming, I will make every name ( final/non-final) to follow camelcasing.

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/ClientLoanIntegrationTest.java
thesmallstar: agreed, changing.

vorburger (approved):

PR #1156, opened 2020-07-05
Fineract-1058 added support for limit,offset and orderBy in SQLbuilder

thesmallstar:
Fineract-1058

Requires more research and Testing.
If anyone can help me verify if prepared statements are allowed for limit, order by and offset? I have seen various comments with different opinions and no mention in any official documentation.

I tested this approach with an example in the CentersReadPlatform and it worked as it should, but I am separating this from that PR so as to not to block it.

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/SQLBuilder.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/SQLBuilder.java
@@ -48,6 +48,12 @@ public class SQLBuilder {
     // args
     private final ArrayList<String> crts = new ArrayList<String>();
 
+    // These variables holds the value for limit, orderBy and offset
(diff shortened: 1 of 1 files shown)


thesmallstar:
I will first complete #1171 #1123 and then try to implement this feature in this those two files (in this PR), so that we can review this independently.

vorburger:
The PRs #1171 and #1123 are blocked by FINERACT-1095. You could still continue here already, if you like - just for the future.

But didn't you loose the `SQLBuidlerTest` you originally had in #1123 when you "split" your work? Hope you still have it..

xurror:
@thesmallstar this is getting stale. Any updates?

vorburger:
@thesmallstar remind us where we are with this one? Both #1171 and #1123 (mentioned above) are merged now.

vorburger:
@thesmallstar ping on this?

PR #1159, opened 2020-07-06
FINERACT-1066 Fixes: NullPointerException at SQLInjectionValidator.va…

thesmallstar:
This happened due to 
```
  if (searchParameters.isOrderByRequested()) {
                sqlBuilder.append(" order by ").append(searchParameters.getOrderBy()).append(' ').append(searchParameters.getSortOrder());
                this.columnValidator.validateSqlInjection(sqlBuilder.toString(), searchParameters.getOrderBy(),
                        searchParameters.getSortOrder());
            }
```
-> here, we checked if OrderBY was not null but did not check if sortorder was "not null".

Instead of fixing it here, it is better to fix it in the called function (IMO).
Since the function SQLvalidateinput can be called either from columnValidator or independently, I have added check for not null at both the places.

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/ColumnValidator.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/ColumnValidator.java
@@ -29,6 +29,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.fineract.infrastructure.core.service.RoutingDataSource;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/ColumnValidator.java
```diff
@@ -91,6 +92,9 @@ private void validateColumn(Map<String, Set<String>> tableColumnMap) {
 
     public void validateSqlInjection(String schema, String... conditions) {
         for (String condition : conditions) {
+            if (!StringUtils.isNotBlank(condition)) {
```
vorburger: Why the "double negation" and not just if isBlank() ?

vorburger (changes requested):

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/ColumnValidator.java
thesmallstar: Opps, updating.

vorburger (approved):

awasum:
This is actually a fix for https://issues.apache.org/jira/browse/FINERACT-1066 as @thesmallstar pointed out on JIRA

PR #1162, opened 2020-07-08
FINERACT-1044 Added redirects in swagger-ui

thesmallstar:
Refer: https://issues.apache.org/jira/browse/FINERACT-1044


vorburger (approved):
Not manually tested, but looks about right..

PR #1166, opened 2020-07-09
FINERACT-1054 fixed sql grammar at loan repayment

francisguchie:
## Description
see https://issues.apache.org/jira/browse/FINERACT-1054

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanReadPlatformServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanReadPlatformServiceImpl.java
@@ -2197,6 +2197,14 @@ public class LoanReadPlatformServiceImpl implements LoanReadPlatformService {
             sqlBuilder.append(" (select min(ls.duedate) datedue,ls.loan_id from m_loan_repayment_schedule ls  ");
             sqlBuilder.append(" where ls.loan_id = ? and  ls.completed_derived = 0)");
             sqlBuilder.append(" )asq on asq.loan_id = ls.loan_id and asq.datedue = ls.duedate");
+            sqlBuilder.append(" GROUP BY ls.duedate");
+            sqlBuilder.append(",ls.principal_amount,ls.principal_completed_derived,ls.principal_writtenoff_derived");
+            sqlBuilder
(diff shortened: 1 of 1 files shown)


francisguchie:
@awasum 

This one passed

vorburger:
This change looks OK to me, if you say that it fixes the error you described in the JIRA.. I'm just curious why none of our ITs (tests) hit this- is this not covered by any tests? Or does the problem which this fixes only appear on some (old?) MySQL version?

francisguchie:
> This change looks OK to me, if you say that it fixes the error you described in the JIRA.. I'm just curious why none of our ITs (tests) hit this- is this not covered by any tests? Or does the problem which this fixes only appear on some (old?) MySQL version?

This appears on mysSQL 5.7 and above in which full_group_by comes as a default for 5.7 and above
This is a "systematic" error in the code  the SQL used before works well for MySQL < 5.7 

there is need to make compatibility tests especially where there is change in version especially the database engine. 
Backward compatibility could also become an issue. a number of times, newer SQL versions come with newer keywords which when used the  older versions would not recognize

awasum (approved):
This does not look harmful. just wondering if this API actually has an integration test which could have caught this? @francisguchie , will you like to investigate and create and issue and send a PR for the test if non is available?

@vorburger , I will let you merge this one if @francisguchie already answered your question.

Thank you @francisguchie for this work. Keep making improvements like this and Fineract will get better.

francisguchie:
@awasum, it might take me longer than expected to come up with an integration test and raise a PR for this test. I would not want to delay the team

vorburger:
@ptuomola do you have any interest in reviewing this? My SQL Foo is... limited. Feel free to ignore this request and I'll just merge on my next round.

ptuomola:
@vorburger Unfortunately I'm not familiar with the business functionality here - i.e. in what scenarios will we end up with multiple rows for the same due date that need to be grouped - and when we do, how do these rows differ? 

I agree that this change should fix the SQL error. But just not sure if there are any cases where we would end up with different behaviour than before: i.e. previously we would have had one row returned with values picked at random from the rows being grouped, and now we would end up with multiple rows being returned. If we know that in all cases the rows with the same due date will also have the same other values, then this will naturally never happen...

Having said that, at the moment it clearly does not work at all but throws an SQL error - so this is definitely an improvement regardless!

PR #1173, opened 2020-07-12
FINERACT-821 Added and enforced ParameterName checkstyle

thesmallstar:
I have changed the regex used to also support single character names.

### DIFF
--- a/fineract-provider/config/checkstyle/checkstyle.xml
+++ b/fineract-provider/config/checkstyle/checkstyle.xml
@@ -46,6 +46,11 @@
 
 
     <module name="TreeWalker">
+    <module name="ParameterName">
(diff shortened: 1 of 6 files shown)


fineract-provider/config/checkstyle/checkstyle.xml
```diff
@@ -46,6 +46,11 @@
 
 
     <module name="TreeWalker">
+    <module name="ParameterName">
+            <property name="format" value="^[a-z][a-z0-9][a-zA-Z0-9]*$|[a-z]"/>
```
vorburger, line 50: @thesmallstar I was surprised that the RegExp change was required.. * is zero or many (+ is "at least one"), are you 100% sure that this is required? I'm just going to merge this anyway as-is, but perhaps you'd like to retry after if this can't be changed back?

vorburger (approved):
LGTM

PR #1177, opened 2020-07-14
FINERACT-1090 Removed top bar and Added Search (swagger)

thesmallstar:
The final layout looks like this:
image


vorburger (approved):
LGTM. @ptuomola I'm assuming/ am sure that this is fine with you.

PR #1180, opened 2020-07-16
Update openjpa to v3.1.2 also in dependencyManagement not just buildscript

vorburger:
This is a follow-up to #1170 which upgrade only one but not the other occurrence of the OpenJPA version.


vorburger:
@rarkins I thought that it may interest you to look into on your end why #1170 upgrade only 1 occurrence, and missed this one.

awasum (approved):

PR #1184, opened 2020-07-16
FINERACT-1089 removing the rs.previous()

percyashu:
## Description
Describe the changes made and why they were made. Ignore if these details are present on the associated Jira ticket https://issues.apache.org/jira/browse/FINERACT-1089

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/DepositAccountInterestRateChartReadPlatformServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/DepositAccountInterestRateChartReadPlatformServiceImpl.java
@@ -293,8 +293,8 @@ public class DepositAccountInterestRateChartReadPlatformServiceImpl implements D
             Long interestRateChartSlabId = null;
             int ircIndex = 0;// Interest rate chart index
             int ircdIndex = 0;// Interest rate chart Slabs index
-            rs.previous();
-            while (rs.next()) {
+
+            do {
                 Long tempIrcdId = rs.getLong("ircdId");
                 if (interestRateChartSlabId == null || interestRateChartSlabId.equals(tempIrcdId)) {
                     if (chartSlabData == null) {
@@ -305,12 +305,8 @@ public class DepositAccountInterestRateChartReadPlatformServiceImpl implements D
                     if (incentiveData != null) {
                         chartSlabData.addIncentives(incentiveData);
                     }
-                } else {
-                    rs.previous();
-                    break;
                 }
-
-            }
+            } while (rs.next());
             return chartSlabData;
         }
     }


percyashu:
@ptuomola @vorburger LGTY?

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/DepositAccountInterestRateChartReadPlatformServiceImpl.java
```diff
@@ -305,12 +305,8 @@ public DepositAccountInterestRateChartSlabData extractData(ResultSet rs) throws
                     if (incentiveData != null) {
                         chartSlabData.addIncentives(incentiveData);
                     }
-                } else {
-                    rs.previous();
```
ptuomola, line 309: OK so looks like we have two calls to rs.previous(), not just one. 

As far as I can see, this function is trying to combine all rows with the same interestRateChartSlabId to a single chartSlabData. So when you get a row with a different ID, then it backs the cursor by one to ensure the calling function gets the right row when it calls next(). Just removing this rs.previous() will not work, as then the cursor is in the wrong position. 

We need to think how to fix this. Personally I think this "manipulating ResultsSet in different functions and hoping that each function returns it in the right state" is not a very clean approach, but I'm not sure it's worth a complete rewrite - probably just easier to find some quick fix.

ptuomola: Having looked at this some more... I think the easiest thing would be just to "collapse" the two functions into a single one: i.e. given the extractData only returns one record, why not just make it a RowMapper and call it in the right place in the calling function? The logic to move in the ResultSet would then be in the calling function only, not split across two functions.

I'm not really sure about the business logic here in the first place: despite the whole rs.next() logic, the method only returns a single record. So in fact all rows except the last one are ignored as far as I can see. Is that really correct: in what cases will we have multiple rows, and how would we expect them to be handled?

To change this, I think the first step would be to set up a test case with known output. We can then refactor the code and ensure the output stays unchanged. I don't think we have any test cases for this at the moment?

edcable:
@percyashu are you still addressing comments from @ptuomola - looked like next step was a test case with known output.

percyashu:
@edcable am still addressing it, had some difficulties in understanding the appropriate test case for the extractData and the subsequent mapRow replacement.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/DepositAccountInterestRateChartReadPlatformServiceImpl.java
percyashu: @ptuomola concerning making extractData in to mapRow does this entails removing whole the rs.next() and using the ircIndex of the calling function as row number of the two RowMappers incentiveMapper and chartSlabsMapper.

ptuomola: @percyashu Yes as far as I can see, the easiest solution would be to have all the logic advancing in the record set in the calling function - so not to move anywhere (forward or backward) in the called functions. That would mean changing them to mapRow (as they are operating on this one row only) rather than extractData (which operates on the whole resultset).

But my problem in the previous comment is: I don't really understand why there's logic to move forward / backward anyway? Typically that's there to e.g. combine multiple rows into one. But in this case it returns just a single row and seems to ignore all the others? Is this just to ensure the last record is returned or what's the intention? 

So my suggestion would be to set up a test that checks the results being returned (and covers this scenario where rs.next()/rs.previous() is triggered) - and then refactor the method, with the test ensuring that the result remains the same. Of course that assumes the current logic is actually correct... which would be great for someone with more functional knowledge of the data model to confirm!

vorburger: @percyashu will you have time to do that @ptuomola suggests above? /Cc @awasum FYI.

percyashu: @vorburger Yeah I will make time, just the issues of setting up a test case to ensure the output stays unchanged and if you can confirm the functional knowledge of the data model.

vorburger: Nope, I unfortunately don't have any functional knowledge about this either, so a non regression test case indeed seems like our best bet here.

PR #1186, opened 2020-07-17
FINERACT-1052 global config for Loan OverPayment

francisguchie:
## Description
https://issues.apache.org/jira/browse/FINERACT-1052

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanAccountDomainServiceJpa.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanAccountDomainServiceJpa.java
@@ -30,7 +30,9 @@ import java.util.Map;
 import java.util.Set;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.fineract.accounting.journalentry.service.JournalEntryWritePlatformService;
+import org.apache.fineract.infrastructure.configuration.data.GlobalConfigurationPropertyData;
 import org.apache.fineract.infrastructure.configuration.domain.ConfigurationDomainService;
+import org.apache.fineract.infrastructure.configuration.service.ConfigurationReadPlatformService;
 import org.apache.fineract.infrastructure.core.data.ApiParameterError;
 import org.apache.fineract.infrastructure.core.data.CommandProcessingResultBuilder;
 import org.apache.fineract.infrastructure.core.data.DataValidatorBuilder;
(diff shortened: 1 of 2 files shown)


francisguchie:
@nikpawar  and @awasum 

this PR is failing at line 712 
org.apache.fineract.integrationtests.AccountingScenarioIntegrationTest.checkPeriodicAccrualAccountingFlow_OVER_PAYMENT(AccountingScenarioIntegrationTest.java:712)
while testing 
AccountingScenarioIntegrationTest > checkPeriodicAccrualAccountingFlow_OVER_PAYMENT() 
for details look at lines 467 to 511 of the test at https://travis-ci.org/github/apache/fineract/builds/709311599


if over_payment is disabled, then this test should be skipped please advise

vorburger:
> if over_payment is disabled, then this test should be skipped please advise

@francisguchie it's your responsibility to make this PR pass the build.

I don't understand enough about the functionality here, but are we changing the default here? Should the SQL script insert TRUE instead of FALSE to preserve current behaviour, so that this becomes an explicit "opt-in"? Alternatively, should (will YOU) change the test, as part of this PR?

@ptuomola don't know if you have any interest to engage on this one, some time (no rush).

francisguchie:
> Alternatively, should (will YOU) change the test, as part of this PR?
Yes @vorburger  i have to do changes on the test to cater for the "disabled overpayment" 

@ptuomola it would be nice to get some help on this when you have time

ptuomola:
Hi

My suggestion on this would be:
- The current behaviour should be kept as the default - i.e. without a configuration present in the database, the same behaviour should remain (overpayments allowed)
- The default configuration added to this should also keep the current behaviour as-is - i.e. the default Flyway script should create the configuration in such way that it allows overpayments
- The existing integration tests should be kept-as is and they should continue to be used to test the current behaviour (i.e. overpayment allowed)
- We should create a new integration test (or multiple tests) to test the scenario where overpayments are not allowed. These tests should first set the flag to disable overpayments, then carry out the tests to confirm that overpayments fail in different scenarios, and then reset the flag back to enabling overpayments

That way we a) preserve the current functionality for those who are relying on it, b) preserve the tests that test the current functionality, and c) add the ability to disable the overpayments and the relevant tests for this

Personally I'm a bit surprised that we want to do this as a global configuration - wouldn't it make more sense to configure this per-product? But the original JIRA / requirements docs are very focused on making this a global setting...

edcable:
@bharathc27 could you do a functional review on this in line with comments from @ptuomola

PR #1194, opened 2020-07-20
FINERACT-971 Added and enforced AvoidNoArgumentSuperConstructorCall Checkstyle

thesmallstar:
FINERACT-971

### DIFF
--- a/fineract-provider/config/checkstyle/checkstyle.xml
+++ b/fineract-provider/config/checkstyle/checkstyle.xml
@@ -192,6 +192,7 @@
             <message key="name.invalidPattern"
              value="Local variable name ''{0}'' must match pattern ''{1}''."/>
         </module>
+        <module name="AvoidNoArgumentSuperConstructorCall"/>
 <!-- TODO Enable many more checks (go about this one by one, step by step, raise separate PRs fixing and then enforcing):
 
         <module name="HiddenField">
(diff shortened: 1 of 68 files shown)


vorburger (changes requested):
@thesmallstar I like the removal of the `super()`, but why did you also make few classes `final`? Especially for JPA Entities like `Teller extends AbstractPersistableCustom` I'm not sure that's a good idea, with all the magic bytecode enhancement from OpenJPA, and the planned move to EclipseLink.

vorburger (changes requested):
@thesmallstar I like the removal of the `super()`, but why did you also make few classes `final`? Especially for JPA Entities like `Teller extends AbstractPersistableCustom` I'm not sure that's a good idea, with all the magic bytecode enhancement from OpenJPA, and the planned move to EclipseLink.

thesmallstar:
@vorburger  that had to be done due to other checkstyle. Since I removed the public constructor, I will instead only remove super() so that they don't have to be made final.

vorburger:
@thesmallstar how is finishing this up coming along?

thesmallstar:
@vorburger I am about to make a PR for sqlSearch and will fix this today itself :)

vorburger (approved):
LGTM

PR #1207, opened 2020-07-28
FINERACT-1095 Added status parameter in Clients API

thesmallstar:
This PR adds a new parameter in client - onlyPending.
The reason for adding onlyPending and not isPending is because 
we are trying to remove the use of
"c.status_enum= 100"
Now that means the status should be "only pending".

Just to remember how we are fixing this:
1. This PR is merged
2. @luckyman20 makes changes in community App
3. We wait for confirmation from the community and remove sqlSearch param in the client.

Next up I will take up the loan module.
After which, if the community agrees we will completely remove the sqlSearch param in all APIs.

### DIFF
--- a/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/ClientTest.java
+++ b/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/ClientTest.java
@@ -18,12 +18,16 @@
  */
 package org.apache.fineract.integrationtests;
 
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+
 import io.restassured.builder.RequestSpecBuilder;
 import io.restassured.builder.ResponseSpecBuilder;
 import io.restassured.http.ContentType;
 import io.restassured.specification.RequestSpecification;
 import io.restassured.specification.ResponseSpecification;
 import java.util.HashMap;
+import java.util.List;
+import java.util.Random;
 import org.apache.fineract.integrationtests.common.ClientHelper;
 import org.apache.fineract.integrationtests.common.Utils;
 import org.junit.jupiter.api.Assertions;
@@ -42,12 +46,12 @@ public class ClientTest {
         this.requestSpec = new RequestSpecBuilder().setContentType(ContentType.JSON).build();
         this.requestSpec.header("Authorization", "Basic " + Utils.loginIntoServerAndGetBase64EncodedAuthenticationKey());
         this.responseSpec = new ResponseSpecBuilder().expectStatusCode(200).build();
-
+        this.clientHelper = new ClientHelper(this.requestSpec, this.responseSpec);
     }
 
     @Test
     public void testClientStatus() {
-        this.clientHelper = new ClientHelper(this.requestSpec, this.responseSpec);
+
         final Integer clientId = ClientHelper.createClient(this.requestSpec, this.responseSpec);
         Assertions.assertNotNull(clientId);
 
@@ -80,7 +84,6 @@ public class ClientTest {
     @Test
     public void testClientAsPersonStatus() {
 
-        this.clientHelper = new ClientHelper(this.requestSpec, this.responseSpec);
         final Integer clientId = ClientHelper.createClientAsPerson(this.requestSpec, this.responseSpec);
         Assertions.assertNotNull(clientId);
 
@@ -113,7 +116,6 @@ public class ClientTest {
     @Test
     public void testClientAsEntityStatus() {
 
-        this.clientHelper = new ClientHelper(this.requestSpec, this.responseSpec);
         final Integer clientId = ClientHelper.createClientAsEntity(this.requestSpec, this.responseSpec);
         Assertions.assertNotNull(clientId);
 
@@ -141,6 +143,38 @@ public class ClientTest {
         clientStatusHashMap = this.clientHelper.withdrawClient(clientId);
         ClientStatusChecker.verifyClientWithdrawn(clientStatusHashMap);
 
+    }
+
+    @SuppressWarnings("unchecked")
+    @Test
+    public void testPendingOnlyClientRequest() {
+
+        Random rand = new Random();
+        // Add a few clients to the server and activate a random amount of them
+        for (int i = 0; i < 15; i++) {
+            final Integer clientId = ClientHelper.createClientAsEntity(this.requestSpec, this.responseSpec);
+            if (rand.nextInt(10) % 2 == 0) {
+                // Takes Client to pending status
+                this.clientHelper.closeClient(clientId);
+                this.clientHelper.reactivateClient(clientId);
+            }
+            // Other clients stay in Active status
+        }
+        List<HashMap<String, Object>> clientsRecieved = (List<HashMap<String, Object>>) clientHelper.getClientWithStatus(50, "pending");
+        assertNotEquals(clientsRecieved.size(), 0);
+        for (int i = 0; i < clientsRecieved.size(); i++) {
+            HashMap<String, Object> clientStatus = ClientHelper.getClientStatus(requestSpec, responseSpec,
+                    String.valueOf(clientsRecieved.get(i).get("id")));
+            ClientStatusChecker.verifyClientPending(clientStatus);
+        }
+
+        clientsRecieved = (List<HashMap<String, Object>>) clientHelper.getClientWithStatus(50, "active");
+        assertNotEquals(clientsRecieved.size(), 0);
+        for (int i = 0; i < clientsRecieved.size(); i++) {
+            HashMap<String, Object> clientStatus = ClientHelper.getClientStatus(requestSpec, responseSpec,
+                    String.valueOf(clientsRecieved.get(i).get("id")));
+            ClientStatusChecker.verifyClientIsActive(clientStatus);
+        }
     }
 
 }
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/ClientStatus.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/ClientStatus.java
@@ -18,6 +18,8 @@
  */
 package org.apache.fineract.portfolio.client.domain;
 
+import org.springframework.util.StringUtils;
+
 /**
  * Enum representation of client status states.
  */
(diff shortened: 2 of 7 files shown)


thesmallstar:
Just remembered, @vorburger  A bug fixed without integration tests is not a bug fixed. 
Adding them ;)

thesmallstar:
Updated with integration tests, ready for review.

ptuomola:
Hi - great stuff. 

One question: rather than having a boolean flag to check for status 100 only, would it make sense to make this more generic by making this parameter an int and allowing the caller of the API to specify any status code as a search parameter? 

Then you could get the same behaviour by calling the API with status = 100, but you could also use the same parameter for any other status if you wanted? Otherwise I think we run the risk of over time having to introduce a new boolean for every status type...

thesmallstar:
The reason is we are using an enum, then we would also need to specify to the user(developer) what enum we are using for each type will it be a good idea?

luckyman20:
I second with @ptuomola

It will provide more options to dev if wants to make further changes in frontend.

thesmallstar:
@ptuomola  @vorburger  @luckyman20  instead of taking enum code we can take a string and then map them to integers, so that the API user don't need to worry about the numbers.

I hope you would agree, using this.

thesmallstar:
@luckyman20  @ptuomola  I have updated the API and tests.
so now this accepts a parameter status which can take value active, closed, withdrawn, pending, etc :) 
Please review this.

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/service/ClientReadPlatformServiceImpl.java
```diff
@@ -258,6 +261,19 @@ private String buildSqlStringFromClientCriteria(String schemaSql, final SearchPa
...
         }
 
+        if (status != null) {
+            final String lowerCaseStatus = status.toLowerCase();
+            Map<String, Integer> statusNumber = new HashMap<String, Integer>();
+            statusNumber.put("active", 300);
+            statusNumber.put("withdrawn", 800);
+            statusNumber.put("closed", 600);
+            statusNumber.put("rejected", 700);
+            statusNumber.put("pending", 100);
```
ptuomola: If you look at org.apache.fineract.portfolio.client.domain.ClientStatus, there seems to be more values for this enum than shown here. Was there a reason why only this subset is relevant for this?

Also for mapping from string to value,  would it make sense to put this into the ClientStatus class and make it reusable? The same way as for example for GLAccountType.fromString() or CalendarFrequencyType.fromString()? That way we would be maintaining the logic in one place only...

thesmallstar: @ptuomola updated!

vorburger (approved):
I'm marking this Approved to indicate that this looks roughly about right to me, but I've not reviewed it in full details, seeing that @ptuomola is on it - thank you! I'll let you drive the review merge whenever you're good with it, but won't merge it myself.

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/ClientTest.java
```diff
@@ -143,4 +143,36 @@ public void testClientAsEntityStatus() {
...
+            final Integer clientId = ClientHelper.createClientAsEntity(this.requestSpec, this.responseSpec);
+            if (rand.nextInt(10) % 2 == 0) {
+                // Takes Client to pending status
+                this.clientHelper.closeClient(clientId);
+                this.clientHelper.reactivateClient(clientId);
+            }
+            // Other clients stay in Active status
+        }
+        List<HashMap<String, Object>> clientsRecieved = (List<HashMap<String, Object>>) clientHelper.getClientWithStatus(50, "pending");
+
```
ptuomola: Should we not check that we receive at least some clients? As far as I can see, this test will still pass if the search doesn't actually work and returns nothing... 

Ideally we should check that we receive some of the same clients that we set to status pending above - but at the very least we should check that we've received something...

fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/ClientTest.java
```diff
@@ -143,4 +143,36 @@ public void testClientAsEntityStatus() {
...
+        for (int i = 0; i < clientsRecieved.size(); i++) {
+            HashMap<String, Object> clientStatus = ClientHelper.getClientStatus(requestSpec, responseSpec,
+                    String.valueOf(clientsRecieved.get(i).get("id")));
+            ClientStatusChecker.verifyClientPending(clientStatus);
+        }
+
+        clientsRecieved = (List<HashMap<String, Object>>) clientHelper.getClientWithStatus(50, "active");
+
+        for (int i = 0; i < clientsRecieved.size(); i++) {
+            HashMap<String, Object> clientStatus = ClientHelper.getClientStatus(requestSpec, responseSpec,
```
ptuomola, line 174: Same here as above - shouldn't we check we received at least one active client?

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/ClientStatus.java
```diff
@@ -18,17 +18,17 @@
...
-    INVALID(0, "clientStatusType.invalid"), //
-    PENDING(100, "clientStatusType.pending"), //
-    ACTIVE(300, "clientStatusType.active"), //
-    TRANSFER_IN_PROGRESS(303, "clientStatusType.transfer.in.progress"), //
-    TRANSFER_ON_HOLD(304, "clientStatusType.transfer.on.hold"), //
-    CLOSED(600, "clientStatusType.closed"), REJECTED(700, "clientStatusType.rejected"), WITHDRAWN(800, "clientStatusType.withdraw");
+    INVALID(0, "clientStatusType.invalid"), PENDING(100, "clientStatusType.pending"), ACTIVE(300,
+            "clientStatusType.active"), TRANSFER_IN_PROGRESS(303, "clientStatusType.transfer.in.progress"), TRANSFER_ON_HOLD(304,
+                    "clientStatusType.transfer.on.hold"), CLOSED(600, "clientStatusType.closed"), REJECTED(700,
+                            "clientStatusType.rejected"), WITHDRAWN(800, "clientStatusType.withdraw");
```
ptuomola: Hmm... couldn't spot any difference except removal of the formatting (one status per line) - was there something changed here?

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/ClientStatus.java
```diff
@@ -63,6 +63,33 @@ public static ClientStatus fromInt(final Integer statusValue) {
         return enumeration;
     }
 
+    public static ClientStatus fromString(final String clientString) {
```
ptuomola, line 68: Hmmm... how does this actually work now? Here we are comparing to the strings as defined in the enum (e.g. clientStatusType.active) but the test is passing in just "active". So it should not match / return anything. What did I miss here?

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/service/ClientReadPlatformServiceImpl.java
```diff
@@ -258,6 +259,13 @@ private String buildSqlStringFromClientCriteria(String schemaSql, final SearchPa
             extraCriteria += " and c.display_name like ? ";
         }
 
+        if (status != null) {
+            ClientStatus clientStatus = ClientStatus.fromString(status);
+            if (clientStatus.getValue() != 0) {
```
ptuomola: So this would mean you can't search for clients with status = INVALID. Is this correct behaviour - shouldn't we still allow searching even if the value you pass in is INVALID? I.e. to find any INVALID entries (if any)...

At the same time it would be good to validate the provided input string to ensure it has actually mapped to a valid value, and throw PlatformValidationException if it doesn't. 

So you should be able to search for any invalid entries by sending in clientStatusType.invalid - but if you send in clientStatusType.blah then you should get a validation exception.

ptuomola (changes requested):

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/ClientStatus.java
thesmallstar: I removed the useless comments, that made it happen. 
Weird, we might need to change some spotless rules.

thesmallstar:
@ptuomola I am collectively answering to your doubts here:
1. I  am actually adding clients, so there would be some clients in the database for sure. (Check the first for loop).
2. I am adding 15 clients and each time I am assigning him pending or active status randomly, that means

The probability that there are no active/pending clients = 1/pow(2,15) = 0.00003051757, I hope you will agree we can ignore this and consider that there is at least one active/pending client. 

3. The client status never actually becomes invalid, it just means that there is no such status. In that case, we as you said we can either give an exception or show all entries. 
@ptuomola as you said, should we instead throw an exception? still confused or we should instead show all the clients if it does not map to any of them. 

FYI @vorburger

ptuomola:
> @ptuomola I am collectively answering to your doubts here:
> 
> 1. I  am actually adding clients, so there would be some clients in the database for sure. (Check the first for loop).
> 2. I am adding 15 clients and each time I am assigning him pending or active status randomly, that means
> 
> The probability that there are no active/pending clients = 1/pow(2,15) = 0.00003051757, I hope you will agree we can ignore this and consider that there is at least one active/pending client.

No that wasn't my concern - I know you create active and pending clients. My concern was: what if the query selection doesn't work as expected, and instead returns an empty list without matching any clients? I don't see the test checking for the case where there is no clients returned - it just checks that whatever clients are returned are in the correct status. 

So something like assertNotEquals(list.size(), 0) for both cases...

> 1. The client status never actually becomes invalid, it just means that there is no such status. In that case, we as you said we can either give an exception or show all entries.
>    @ptuomola as you said, should we instead throw an exception? still confused or we should instead show all the clients if it does not map to any of them.

My personal view would be that we should throw an exception if the caller sends a status that is not a valid value for ClientStatus. Otherwise the caller thinks that there are no clients matching this status and never finds out that they actually sent the wrong value. Empty list should mean that there are no matches.

thesmallstar:
@ptuomola thank you so much for the quick response!
I got it now, updating it now.

thesmallstar:
@ptuomola I have updated this as you suggested, the tests are running for hours now idk why. Would you review this please?

thesmallstar:
looking into the failed test case.

thesmallstar:
@ptuomola ready for your review :D

ptuomola:
@thesmallstar thanks - will take a look tomorrow morning!

ptuomola:
Sorry for the delay in looking into this - have been a bit busy! Will try to look at this tonight...

ptuomola:
@thesmallstar Finally had a chance to look at this and to test it out locally. The functionality itself works as expected - great work! 

The only concern I have is: I really don't like the fact that we need to maintain two lists of ClientStatus values - one in the enum itself, another one in the validator. It's really easy for these to get out of synch: in fact they have already done so, as there's a spelling mistake in the validator ("transfer_in_progreess") which doesn't match the validator.

Given this, maybe the right thing to do after all is to use the ClientStatus enum itself to validate? I.e. call the fromString method to validate, and throw an error if you get ClientStatus.INVALID? I know that doesn't allow you to search for values that actually have the status INVALID, but I don't think that's a valid status in the first place - so we should not have any such records anyway.

Does that make sense? That would allow us to remove the arrayList of valid values, but otherwise keep the functionality intact. What do you think?

thesmallstar:
@ptuomola agreed and good point! updating this.

thesmallstar:
@ptuomola done, Let's see if the test passes.

ptuomola (approved):

ptuomola:
Looks great to me! Thanks for your work on this @thesmallstar

PR #1218, opened 2020-07-31
FINERACT-1098 - Catch JpaSystemException in jpa transactions

xurror:
https://issues.apache.org/jira/browse/FINERACT-1098

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/accounting/closure/service/GLClosureWritePlatformServiceJpaRepositoryImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/accounting/closure/service/GLClosureWritePlatformServiceJpaRepositoryImpl.java
@@ -41,6 +41,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.dao.DataIntegrityViolationException;
+import org.springframework.dao.NonTransientDataAccessException;
+import org.springframework.orm.jpa.JpaSystemException;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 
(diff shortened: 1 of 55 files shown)


fineract-provider/src/main/java/org/apache/fineract/accounting/closure/service/GLClosureWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -142,8 +144,7 @@ public CommandProcessingResult deleteGLClosure(final Long glClosureId) {
      * @param command
      * @param dve
      */
-    private void handleGLClosureIntegrityIssues(final JsonCommand command, final DataIntegrityViolationException dve) {
-        final Throwable realCause = dve.getMostSpecificCause();
+    private void handleGLClosureIntegrityIssues(final JsonCommand command, final Throwable realCause, final Exception dve) {
```
vorburger: @xurror it's not super important, but I thought perhaps this code review feedback could interest you: Instead of changing this like you have, what you really want to do in a case like this is just use the common super class of `JpaSystemException` and `DataIntegrityViolationException` like suggested below, does this make sense? Do you want to change this like that? It makes the code easier to read.

```suggestion
    private void handleGLClosureIntegrityIssues(final JsonCommand command, final NestedRuntimeException dve) {
        final Throwable realCause = dve.getMostSpecificCause();
```

vorburger (changes requested):

fineract-provider/src/main/java/org/apache/fineract/accounting/closure/service/GLClosureWritePlatformServiceJpaRepositoryImpl.java
xurror: @vorburger instead NestedRuntimeException I used NonTransientDataAccessException which is a more suited super class.
You can check this and if it's LGTY you can merge.

xurror:
Merging this so I can continue with migration.

PR #1206, opened 2020-07-26
FINERACT-838 Added Swagger Codegen Support

Grandolf49:
FINERACT-838

## Description
Integrated Swagger Codegen support for automatically generating client code from OpenAPI Specification

## Checklist
Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Commit message starts with the issue number from https://issues.apache.org/jira/projects/FINERACT/. Ex: FINERACT-646 Pockets API.

- [x] Coding conventions at https://cwiki.apache.org/confluence/display/FINERACT/Coding+Conventions have been followed.

- [x] API documentation at fineract-provider/src/main/resources/static/api-docs/apiLive.htm has been updated with details of any API changes.

- [x] Integration tests have been created/updated for verifying the changes made.

- [x] All Integrations tests are passing with the new commits.

- [x] Submission is not a "code dump".  (Large changes can be made "in repository" via a branch.  Ask on the list.)

Our guidelines for code reviews is at https://cwiki.apache.org/confluence/display/FINERACT/Code+Review+Guide

PR #1226, opened 2020-08-01
Use java8 instead of joda as dateLibrary in Swagger Client (FINERACT-838)

vorburger:
see FINERACT-838, this is a #1206 follow-up

In light of @percyashu #1179 for FINERACT-826, IMHO we should promote Java 8 native JSR-310 and not Joda Time even for clients.

@Grandolf49 @thesmallstar @ptuomola OK for you? Merge.

I've NOT actually tested (run) this, it's just based on https://openapi-generator.tech/docs/generators/java/.


ptuomola:
LGTM

PR #1228, opened 2020-08-01
FINERACT-838 TravisCI Setup | Updated swagger-codegen-cli version

Grandolf49:
## Description
- Updated the `swagger-codegen-cli` version to `3.0.18` since all greater versions throw `Missing OpenApi File` error. https://github.com/swagger-api/swagger-codegen/issues/10157 . (A bug in swagger-codegen-cli)
- Fixed Base URL in [fineract-input.yml](https://github.com/Grandolf49/fineract/blob/swagger-codegen/fineract-provider/config/swagger/fineract-input.yaml) for Swagger Codegen since build for generated code was failing due to incorrect Base URL.
- Added "Generate and Build" commands in [.travis.yml](https://github.com/Grandolf49/fineract/blob/swagger-codegen/.travis.yml)

## Checklist

### DIFF
--- a/fineract-provider/config/swagger/fineract-input.yaml
+++ b/fineract-provider/config/swagger/fineract-input.yaml
@@ -22,7 +22,7 @@ info:
     name: Apache 2.0
     url: 'http://www.apache.org/licenses/LICENSE-2.0.html'
 servers:
-- url: /fineract-provider/api/v1
+- url: https://demo.openmf.org/fineract-provider/api/v1
 components:
   securitySchemes:
     basicAuth:
--- a/fineract-provider/config/swagger/fineract-input.yaml
+++ b/fineract-provider/config/swagger/fineract-input.yaml
@@ -22,7 +22,7 @@ info:
     name: Apache 2.0
     url: 'http://www.apache.org/licenses/LICENSE-2.0.html'
 servers:
-- url: https://demo.openmf.org/fineract-provider/api/v1
+- url: https://demo.fineract.dev/fineract-provider/api/v1
 components:
   securitySchemes:
     basicAuth:
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,4 +80,5 @@ script:
 
   # Generate and Build Swagger Code
   - ls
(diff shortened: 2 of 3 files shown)


Grandolf49:
@vorburger Upon further local testing, I noticed that the build for generated code was failing due to incomplete base URL in [fineract.yml](https://github.com/apache/fineract/blob/c5c62de491a1ca2d26457f99c49e48d0cca7c0bc/fineract-provider/config/swagger/fineract-input.yaml#L25) file and the version of `swagger-codegen-cli`.

Now, generated code builds successfully on Java 8 with these changes.

fineract-provider/config/swagger/fineract-input.yaml
```diff
@@ -22,7 +22,7 @@ info:
     name: Apache 2.0
     url: 'http://www.apache.org/licenses/LICENSE-2.0.html'
 servers:
-- url: /fineract-provider/api/v1
+- url: https://demo.openmf.org/fineract-provider/api/v1
```
vorburger: This currently appears to be down? But even if it was up, I would much prefer this to use `fineract.dev` instead:

```suggestion
- url: https://demo.fineract.dev/fineract-provider/api/v1
```

@Grandolf49 see https://www.fineract.dev for background... OK for you?

vorburger: @Grandolf49 in the interest of moving forward with this, I'll add this to your branch by self-accepting this suggestion to your PR, and if the build still passes, I'll merge this. Please comment here if you disagree with changing this URL, and we can discuss it further. Whatever server we use server is IMHO "wrong" anyway - we should use the LOCAL server - otherwise this will never work for new APIs being added... I'll comment about this on  FINERACT-838.

vorburger:
Sh#* I actually meant to "squash" this, as we usually do... too late. Anyway.

fineract-provider/config/swagger/fineract-input.yaml
Grandolf49: Yeah sure! This works for me!

PR #1231, opened 2020-08-03
FINERACT-1104 - adjust JPQL query @likelihoodRepository

xurror:
https://issues.apache.org/jira/browse/FINERACT-1104


vorburger (approved):
LGTM. @xurror if you can wait for peer reviews by other active committers before self merging, it's better... ;-)

PR #1238, opened 2020-08-05
Removed MissingSwitchDefault Checkstyle from TODO

thesmallstar:
Please check the discussion on: https://github.com/apache/fineract/pull/1006

### DIFF
--- a/fineract-provider/config/checkstyle/checkstyle.xml
+++ b/fineract-provider/config/checkstyle/checkstyle.xml
@@ -29,6 +29,7 @@
     <!--
         You can learn more about individual checkstyles at: https://checkstyle.sourceforge.io/
         Checks related to LineLength and Indentation are only handled by Spotless
+        Checks included in MissingSwitchDefault are handled by ErrorProne
     -->
 
     <module name="FileTabCharacter">
@@ -223,7 +224,7 @@
 
 
 
-        <module name="MissingSwitchDefault"/>
+
 
 
 


vorburger (approved):

vorburger (commented):
@thesmallstar I'm curious, if you just enable this check instead, won't it just work now? IMHO it's OK to have something covered by BOTH EP and CS - it doesn't "hurt", and it's better to ensure future non-regression e.g. if we ever we're to disable one or the other if these tools again in future years. Of course, if the tools "disagree" with each other and it's a PITA, then it's probably not worth it. But for a simple verification like switch case... just try enabling it- does it work?

thesmallstar:
Take this code for example:
https://github.com/apache/fineract/blob/2323450e71d1508e6ff218d48b856aefa04542d0/fineract-provider/src/main/java/org/apache/fineract/accounting/journalentry/service/JournalEntryRunningBalanceUpdateServiceImpl.java#L243-L269

here, it wants us to add a default statement and from what I have understood from https://github.com/apache/fineract/pull/1006#issuecomment-641687223 by @ptuomola  that

1. If we are adding new value to the enum, error-prone directly asks us to cover that in the switch case.
2. If there is no common way to handle things there is no need for a default statement.

In other words, in all of these cases all the values that are covered by the enum are always included thanks to error-prone, and also we don't have any common way of handling things, that means we would end up adding an incorrect/blank default statement that brings no benefit.

would you agree on this? @vorburger  @ptuomola @percyashu ?

ptuomola:
@thesmallstar that makes sense to me. Either we have a default case to cover all the cases not covered, or we require all cases to be covered explicitly. As far as I can see, they are mutually exclusive...

ptuomola:
@vorburger would you be OK if we merge this?

ptuomola:
No objection from @vorburger so far - I will merge this so that it makes the 1.4.0 release. 

If there's any concern from anyone, we can easily add this back to the todos.

vorburger:
Yes I agree, that makes perfect sense; Thank You for the detailed explanation about this check.

PR #1242, opened 2020-08-08
FINERACT-821 Added and Enforced MemberName Checkstyle(2)

thesmallstar:
I have suppressed the checkstyle where I thought it is not a good idea to change how we are naming, like database Schemas in swagger.

### DIFF
--- a/fineract-provider/config/checkstyle/checkstyle.xml
+++ b/fineract-provider/config/checkstyle/checkstyle.xml
@@ -183,6 +183,11 @@
              <message key="ws.notPreceded"
              value="WhitespaceAround: ''{0}'' is not preceded with whitespace."/>
         </module>
+        <module name="MemberName">
+            <property name="format" value="^[a-z][a-z0-9][a-zA-Z0-9]*$"/>
+            <message key="name.invalidPattern"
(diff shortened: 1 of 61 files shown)


ptuomola:
Looking at this, the changes seem to fall into a couple of different categories:

- ones where it makes sense - e.g. is_active to isActive

- ones where it in my view doesn't really make sense - e.g. S3, SMS etc are acronyms that are capitalised, so to make them lowercase just because they are in a variable name seems strange

- ones where we are taking a fully capitalised variable and making that camelcase. I'm wondering if these should actually be static constants rather than variables in the first place?

thesmallstar:
@ptuomola I thought not to make them static since they are used in that class itself, I thought there is no use to make them static since they are not used elsewhere if you think the static makes more sense I will change it ;)

Should I instead suppress the checkstyle for point "2" then?

ptuomola:
@thesmallstar Thanks for your reply! I think it's all just a matter of preference and what you're used to seeing :-) 

But I think consistency is more important than trying to adhere to anyone's specific preferences. 

So to make things consistent re member naming, let's merge your change as proposed.

ptuomola (approved):

PR #1251, opened 2020-08-17
FINERACT-1095 Added status parameter in Loans API

thesmallstar:
This is in continuation to #1207
1. support for new parameter "status" to loans is added.
2. It can have multiple values example: staus=active&status=approved
3. Validations are also added.

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanStatus.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanStatus.java
@@ -18,6 +18,8 @@
  */
 package org.apache.fineract.portfolio.loanaccount.domain;
 
+import org.springframework.util.StringUtils;
+
 /**
  * Enum representation of loan status states.
  */
@@ -78,6 +80,42 @@ public enum LoanStatus {
             break;
         }
         return enumeration;
+    }
+
+    public static LoanStatus fromString(final String statusString) {
+
+        LoanStatus enumeration = LoanStatus.INVALID;
+
+        if (StringUtils.isEmpty(statusString)) {
+            return enumeration;
+        }
+
+        if (statusString.equalsIgnoreCase(LoanStatus.SUBMITTED_AND_PENDING_APPROVAL.toString())) {
+            enumeration = LoanStatus.SUBMITTED_AND_PENDING_APPROVAL;
+        } else if (statusString.equalsIgnoreCase(LoanStatus.ACTIVE.toString())) {
+            enumeration = LoanStatus.ACTIVE;
+        } else if (statusString.equalsIgnoreCase(LoanStatus.APPROVED.toString())) {
+            enumeration = LoanStatus.APPROVED;
+        } else if (statusString.equalsIgnoreCase(LoanStatus.WITHDRAWN_BY_CLIENT.toString())) {
+            enumeration = LoanStatus.WITHDRAWN_BY_CLIENT;
+        } else if (statusString.equalsIgnoreCase(LoanStatus.OVERPAID.toString())) {
+            enumeration = LoanStatus.OVERPAID;
+        } else if (statusString.equalsIgnoreCase(LoanStatus.TRANSFER_ON_HOLD.toString())) {
+            enumeration = LoanStatus.TRANSFER_ON_HOLD;
+        } else if (statusString.equalsIgnoreCase(LoanStatus.REJECTED.toString())) {
+            enumeration = LoanStatus.REJECTED;
+        } else if (statusString.equalsIgnoreCase(LoanStatus.CLOSED_RESCHEDULE_OUTSTANDING_AMOUNT.toString())) {
+            enumeration = LoanStatus.CLOSED_RESCHEDULE_OUTSTANDING_AMOUNT;
+        } else if (statusString.equalsIgnoreCase(LoanStatus.CLOSED_OBLIGATIONS_MET.toString())) {
+            enumeration = LoanStatus.CLOSED_OBLIGATIONS_MET;
+        } else if (statusString.equalsIgnoreCase(LoanStatus.CLOSED_WRITTEN_OFF.toString())) {
+            enumeration = LoanStatus.CLOSED_WRITTEN_OFF;
+        } else if (statusString.equalsIgnoreCase(LoanStatus.TRANSFER_IN_PROGRESS.toString())) {
+            enumeration = LoanStatus.TRANSFER_IN_PROGRESS;
+        }
+
+        return enumeration;
+
     }
 
     LoanStatus(final Integer value, final String code) {
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanReadPlatformServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanReadPlatformServiceImpl.java
@@ -36,8 +36,10 @@ import org.apache.fineract.accounting.common.AccountingRuleType;
 import org.apache.fineract.infrastructure.codes.data.CodeValueData;
 import org.apache.fineract.infrastructure.codes.service.CodeValueReadPlatformService;
 import org.apache.fineract.infrastructure.configuration.domain.ConfigurationDomainService;
+import org.apache.fineract.infrastructure.core.data.ApiParameterError;
 import org.apache.fineract.infrastructure.core.data.EnumOptionData;
 import org.apache.fineract.infrastructure.core.domain.JdbcSupport;
+import org.apache.fineract.infrastructure.core.exception.PlatformApiDataValidationException;
 import org.apache.fineract.infrastructure.core.service.DateUtils;
 import org.apache.fineract.infrastructure.core.service.Page;
 import org.apache.fineract.infrastructure.core.service.PaginationHelper;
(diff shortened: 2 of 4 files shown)


thesmallstar:
@vorburger  @ptuomola  @luckyman20 
As soon as this is merged, we can completely remove sqlsearch and the PRs #1171 and #1123 can be merged with minor changes.
And we will be done with removing all SQL string concatenations from our code :D

ptuomola:
@thesmallstar the code looks great to me - thanks!

Can you add some intergration tests as well so we have this feature covered and know it works? e.g. search with one status, search with multiple statuses, and search with invalid status value?

luckyman20:
@thesmallstar It is not possible in community-app that we can send the same parameter more than once with the request, please look into this.

vorburger:
@thesmallstar are you on this? :smiling_imp:

thesmallstar:
@vorburger  yes yes, I am in conversion with @luckyman20 we are trying to resolve the frontend part. We should confirm it by today, And then find the best way for this.

thesmallstar:
@vorburger  @ptuomola  taking exceptionally long time to write tests since it is really hard to understand how loans work :(

vorburger:
> @vorburger @ptuomola taking exceptionally long time to write tests since it is really hard to understand how loans work :(

if you have a more specific question, then perhaps @ptuomola (likely better than me!) can answer.

but if it's just writing integration tests for this that's very significant additional work, than perhaps it wouldn't be the end of the world to get this merged without additional new ITs already (but it can break existing ITs, of course), and then perhaps later add more test coverage for it, when (if) you have more time.

thesmallstar:
@vorburger It would be nice if we could merge this rn, Since this is blocking the work for removing SQL String concatenation, I will have enough time and I am making an issue so that I don't forget the IT for this ;) 

Edit: Don't MERGE THIS RN :(

thesmallstar:
@luckyman20 and I, are still not able to figure out how to do this on the frontend, I will try to get it done tomorrow, after which we can merge this.

ptuomola:
> @vorburger @ptuomola taking exceptionally long time to write tests since it is really hard to understand how loans work :(

@thesmallstar Hopefully you can reuse the code from ClientLoanIntegrationTest? That creates and checks lots of loans in different statuses and has utils for doing that - so creating additional loans and putting them to different statuses and then testing searching should just be a matter of some copy & paste from existing tests.

thesmallstar:
@ptuomola  @vorburger , From my discussions with @luckyman20 we came to the conclusion that we cannot send two parameters from the community-app, do you see any other solution to solve this one? 
We need our API to accept a list of strings...

ptuomola:
@thesmallstar did you see my earlier suggestion in the jira ticket on how to do this?

vorburger:
@thesmallstar the suggestion from @ptuomola in the last comment on FINERACT-1095 to use e.g. `"loanStatus": ["all","active","overpaid"], "loanProducts": ["65","97","73","82"], "offices": ["1","10","100"]` syntax makes a lot of sense, to me.. will you do the needful here? It would be cool if we could wrap up FINERACT-1095!

thesmallstar:
@vorburger  yup I will fix this in a day or two. 
@ptuomola  thanks for the suggestion!!

vorburger:
@thesmallstar friendly ping, update? (as well as now have to rebase and fix merge conflict)

thesmallstar:
@vorburger sorry for the delay, I am in between of my college mid-semester exams, I will continue working as soon as I am done with it(hopefully no sooner than 19th Sept).

vorburger:
@thesmallstar hope your exams went well! Whenever you have time to pick this up, we appreciate your contributions, very much!

thesmallstar:
@vorburger  yes already working on it!

thesmallstar:
@vorburger @ptuomola @luckyman20 Thank you for waiting a long time for these here is the update:

1. As mentioned by @ptuomola  for using the implementation ["a","b"] that cannot be really used here because the request type is not JSON, but we can do something similar.
2. The API now takes one string and works for both: examples:
                    a. "active"
                    b. "active,approved"
Example URLS: 
https://localhost:8443/fineract-provider/api/v1/loans?status=active,approved&tenantIdentifier=default&pretty=true

3. We will merge this now, I will write the Integration tests and I have also made an issue so that we don't forget to do it. https://issues.apache.org/jira/browse/FINERACT-1160

vorburger:
@thesmallstar sounds like a plan, and OK for me to add the IT in a follow up. But the build on this is not passing, did you see? BTW watch the related discussion over on https://github.com/openMF/community-app/issues/3275.

thesmallstar:
@vorburger oops, I missed to ping you after I pushed new changes.
I think we can merge this now?

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanStatus.java
```diff
@@ -80,6 +82,42 @@ public static LoanStatus fromInt(final Integer statusValue) {
...
 
+    public static LoanStatus fromString(final String statusString) {
+
+        LoanStatus enumeration = LoanStatus.INVALID;
+
+        if (StringUtils.isEmpty(statusString)) {
+            return enumeration;
+        }
+
+        if (statusString.equalsIgnoreCase(LoanStatus.SUBMITTED_AND_PENDING_APPROVAL.toString())) {
```
vorburger, line 93: @thesmallstar I'm probably missing something, but isn't this just `LoanStatus.valueOf(statusString)`? If you're doing this only because of `equalsIgnoreCase()` then wouldn't a `LoanStatus.valueOf(statusString.toUpperCase())` suffice and have the same effect? If you want an invalid status String to return LoanStatus.INVALID, you may have to `catch (IllegalArgumentException e)` .. (I think; not 100% sure, do check). Actually that is debatable.. wouldn't you rather that an invalid status filter passed through the REST API from a client lead to an error, than be silently ignored?

vorburger (changes requested):

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanReadPlatformServiceImpl.java
```diff
@@ -288,6 +290,19 @@ public LoanScheduleData retrieveRepaymentSchedule(final Long loanId,
...
+                if (LoanStatus.fromString(statusValues.get(i)) == LoanStatus.INVALID) {
+                    final List<ApiParameterError> dataValidationErrors = new ArrayList<>();
+                    final String defaultUserMessage = "The Status value '" + statusValues.get(i) + "' is not supported.";
+                    final ApiParameterError error = ApiParameterError.parameterError("validation.msg.loan.status.value.is.not.supported",
+                            defaultUserMessage, "status", statusValues.get(i));
+                    dataValidationErrors.add(error);
+                    throw new PlatformApiDataValidationException(dataValidationErrors);
+                }
+            }
+        }
```
thesmallstar, line 305: @vorburger  here, we are checking for all strings that were given to us and make sure none of them returns an Invalid status if it does we return a PlatformApiDataValidationException, does this answer your question? (I might have missed something?)

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanStatus.java
thesmallstar: For the first part, yes, they would have the same effect. I copied this implementation from some other enum and hence the code looks like this and not how you suggested. Do you see any advantage of one over the other?

2nd part answer in a comment below.

ptuomola: Just another point: given this validation seems to be a common pattern in the enums, I've added a method in the DataValidatorBuilder (isOneOfEnumValues()) to check if the value provided is valid for a specific enum (and report back with correct PlatformApiDataException if it isn't).

So one way to handle these is in the validation phase check the values using that, after which it should be safe to use valueOf() without getting an exception. 

So as far as I can see there are three ways to do the same thing:

- custom valueOf method with error handling to avoid an exception (as currently in many places in the code), and then check for invalid status afterwards - i.e. the current logic in this PR

- call valueOf and catch the exception if the value is invalid, and rethrow as PlatformApiDataException

- call DataValidatorBuilder.isOneOfEnumValues() to validate and then valueOf to convert

Not sure there's a significant advantage of one way over another...

vorburger:
@thesmallstar @ptuomola where were we with this one? Good to merge as-is, or updates needed? (Perhaps best to rebase it now.)

ptuomola:
@vorburger based on the above discussions, my suggestion would still be to change the validation to isOneOfEnumValues() and the conversion to be done with LoanStatus.valueOf(statusString.toUpperCase()) as you suggest. That would avoid a lot of "boilerplate" code.

thesmallstar:
@ptuomola @vorburger thanks, on it :) 
Sorry for delaying this long, I have been juggling between college exams, assignments and health :( 
I will be done with my finals on 28th and have a month-long holiday, and I would have plenty of time to commit daily!!
PS: I miss the summer days and lot of "productive" discussions with @vorburger @ptuomola and @xurror  :D

vorburger:
> Sorry for delaying this long, I have been juggling between college exams, assignments and health :(

Don't say Sorry! You have no "obligation" to continue contributing to this project (after your GSoC) - no stress! Best wishes for health.

> I will be done with my finals on 28th and have a month-long holiday, and I would have plenty of time to commit daily!!

But we are of course excited to hear from you! If you can make the time, we always love to hear from you.

PR #1264, opened 2020-08-20
FINERACT-1052 global config loan overpayment leaving situation as is

francisguchie:
## Description
https://github.com/apache/fineract/pull/1186

Addressing the following comments from @ptuomola 

**Addressing these comments below**  
The current behaviour should be kept as the default - i.e. without a configuration present in the database, the same behaviour should remain (overpayments allowed)
The default configuration added to this should also keep the current behaviour as-is - i.e. the default Flyway script should create the configuration in such way that it allows overpayments
The existing integration tests should be kept-as is and they should continue to be used to test the current behaviour (i.e. overpayment allowed)

**but Not able to answer this**
We should create a new integration test (or multiple tests) to test the scenario where overpayments are not allowed. These tests should first set the flag to disable overpayments, then carry out the tests to confirm that overpayments fail in different scenarios, and then reset the flag back to enabling overpayments

## Checklist

### DIFF
--- a/fineract-provider/config/swagger/fineract-input.yaml
+++ b/fineract-provider/config/swagger/fineract-input.yaml
@@ -1,0 +1,37 @@
+openapi: 3.0.3
+info:
+  version: release-1.4.0-561-g9bad487-dirty
+  title: Apache Fineract
+  description: |-
+    Apache Fineract is a secure, multi-tenanted microfinance platform
+
+    The goal of the Apache Fineract API is to empower developers to build apps on top of the Apache Fineract Platform<br>The [reference app](https://cui.fineract.dev) (username: mifos, password: password) works on the same demo tenant as the interactive links in this documentation
+
+    - The API is organized around [REST](https://en.wikipedia.org/wiki/Representational_state_transfer)
+    - Find out more about Apache Fineract [here](/fineract-provider/api-docs/apiLive.htm#top)
+    - You can [Try The API From Your Browser](/fineract-provider/api-docs/apiLive.htm#interact)
+    - The Generic Options are available [here](/fineract-provider/api-docs/apiLive.htm#genopts)
+    - Find out more about [Updating Dates and Numbers](/fineract-provider/api-docs/apiLive.htm#dates_and_numbers)
+    - For the Authentication and the Basic of HTTP and HTTPS refer [here](/fineract-provider/api-docs/apiLive.htm#authentication_overview)
+    - Check about ERROR codes [here](/fineract-provider/api-docs/apiLive.htm#errors)
+
+    Please refer to the [old documentation](/fineract-provider/api-docs/apiLive.htm) for any documentation queries
+  contact:
+    email: dev@fineract.apache.org
+  license:
+    name: Apache 2.0
+    url: 'http://www.apache.org/licenses/LICENSE-2.0.html'
+servers:
+- url: https://demo.fineract.dev/fineract-provider/api/v1
+components:
+  securitySchemes:
+    basicAuth:
+      type: http
+      scheme: basic
+    tenantid:        
+      type: apiKey
+      in: header       
+      name: fineract-platform-tenantid 
+security:
+  - basicAuth: []
+    tenantid: []
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanAccountDomainServiceJpa.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanAccountDomainServiceJpa.java
@@ -30,7 +30,9 @@ import java.util.Map;
 import java.util.Set;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.fineract.accounting.journalentry.service.JournalEntryWritePlatformService;
+import org.apache.fineract.infrastructure.configuration.data.GlobalConfigurationPropertyData;
 import org.apache.fineract.infrastructure.configuration.domain.ConfigurationDomainService;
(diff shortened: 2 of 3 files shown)


vorburger:
@francisguchie https://github.com/apache/fineract#pull-requests

fineract-provider/config/swagger/fineract-input.yaml
```diff
@@ -0,0 +1,37 @@
+openapi: 3.0.3
+info:
+  version: release-1.4.0-561-g9bad487-dirty
+  title: Apache Fineract
```
ptuomola, line 4: We should not check in this file as it is autogenerated

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanAccountDomainServiceJpa.java
```diff
@@ -168,6 +180,16 @@ public LoanTransaction makeRepayment(final Loan loan, final CommandProcessingRes
         final List<Long> existingReversedTransactionIds = new ArrayList<>();
 
         final Money repaymentAmount = Money.of(loan.getCurrency(), transactionAmount);
+
+        // customization configure ability to overpay loan
+        final Money outstandingBalance = Money.of(loan.getCurrency(), loan.getSummary().getTotalOutstanding());
+        if (outstandingBalance.minus(repaymentAmount).isLessThanZero() && !isLoanOverpaymentEnabled) {
+            baseDataValidator.reset().failWithCode("Repayment exceeding outstanding balance");
```
ptuomola, line 187: failWithCode should take as parameter the error code property, not freetext - you can take a look at the other calls to it in this file

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanAccountDomainServiceJpa.java
```diff
@@ -168,6 +180,16 @@ public LoanTransaction makeRepayment(final Loan loan, final CommandProcessingRes
...
+            if (!dataValidationErrors.isEmpty()) {
+                throw new PlatformApiDataValidationException(dataValidationErrors);
+            }
```
ptuomola, line 190: To be consistent with the other exceptions thrown in this class, there are a couple of other parameters that should be passed to PlatformApiDataValidationException - take a look at the other throws

fineract-provider/src/main/resources/sql/migrations/core_db/V362__Loan_Overpayment_Global_Configuration.sql
```diff
@@ -0,0 +1,23 @@
...
+-- "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+-- KIND, either express or implied. See the License for the
+-- specific language governing permissions and limitations
+-- under the License.
+--
+
+--adding new configuration to enable/disable loan over payment
+
+INSERT INTO `c_configuration` (`name`, `enabled`,`description`)
+VALUES ('Enable-Loan-Overpayment',TRUE, 'configuration to enable loan payment beyond outstanding balance');
```
ptuomola, line 23: As per my comments, shouldn't we have this as FALSE by default to retain the current behaviour as-is - unless configured otherwise

ptuomola (changes requested):
Some review comments - please have a look at let me know your thoughts. Also would be good to have some tests for this before merging - maybe you could add a few?

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanAccountDomainServiceJpa.java
```diff
@@ -150,6 +155,13 @@ public LoanTransaction makeRepayment(final Loan loan, final CommandProcessingRes
             final LocalDate transactionDate, final BigDecimal transactionAmount, final PaymentDetail paymentDetail, final String noteText,
             final String txnExternalId, final boolean isRecoveryRepayment, boolean isAccountTransfer, HolidayDetailDTO holidayDetailDto,
             Boolean isHolidayValidationDone, final boolean isLoanToLoanTransfer) {
+
+        final List<ApiParameterError> dataValidationErrors = new ArrayList<>();
+        final DataValidatorBuilder baseDataValidator = new DataValidatorBuilder(dataValidationErrors).resource("loan.transaction");
+        final GlobalConfigurationPropertyData configuration = this.configurationReadPlatformService
+                .retrieveGlobalConfiguration("Enable-Loan-Overpayment");
```
ptuomola, line 162: To make the current behaviour default, would it not make sense to have a configuration value that works the other way around - i.e. call it "block-loan-overpayment", and if it is present & enabled then overpayments are blocked.

vorburger:
@francisguchie will you update this PR based on @ptuomola code review feedback? If you don't, it will auto-close soon.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanAccountDomainServiceJpa.java
francisguchie: thanks @ptuomola i did not think of this suggested approach. let me try just that

francisguchie:
> @francisguchie will you update this PR based on @ptuomola code review feedback? If you don't, it will auto-close soon.

Yes @vorburger, will do this weekend

francisguchie:
@awasum, @vorburger , @ptuomola 

I am going to make sure this will be done asap.

francisguchie:
new pull request made

https://github.com/apache/fineract/pull/1536

PR #1278, opened 2020-08-25
Cherry-pick "Update dependency gradle to v6.6.1" to 1.4.0 also

vorburger:
back-port #1277 because https://github.com/gradle/gradle/releases/v6.6.1 says _We recommend that you use Gradle 6.6.1 over the initial release of Gradle 6.6._

@vidakovic if OK for you, do you want to merge this after the build passed?


vidakovic:
@vorburger yes, can do

PR #1279, opened 2020-08-25
FINERACT-854 Removed string concatenated SQL in ClientReadPlatform

thesmallstar:
Refer: https://issues.apache.org/jira/browse/FINERACT-854 and #725 #723 for background.

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/SQLBuilder.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/SQLBuilder.java
@@ -85,7 +85,8 @@ public class SQLBuilder {
                     "criteria cannot contain more than 1 space (between column name and operator): " + trimmedCriteria);
         }
         if (!operator.equals("=") && !operator.equals("<") && !operator.equals(">") && !operator.equals("<=") && !operator.equals(">=")
-                && !operator.equals("<>") && !operator.equals("LIKE") && !operator.equals("like")) {
+                && !operator.equals("<>") && !operator.equals("LIKE") && !operator.equals("like")
+                && !operator.toLowerCase().equals("not in")) {
             // add support for SQL's BETWEEN and IN, if/when ever needed.. (it's
             // a little more than just adding above, as it can have multiple
             // arguments)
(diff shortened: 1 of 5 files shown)


fineract-provider/src/main/java/org/apache/fineract/portfolio/client/service/ClientReadPlatformServiceImpl.java
```diff
@@ -196,25 +197,20 @@ public ClientData retrieveTemplate(final Long officeId, final boolean staffInSel
...
         // underHierarchySearchString = searchParameters.getHierarchy() + "%";
         // }
-        List<Object> paramList = new ArrayList<>(Arrays.asList(underHierarchySearchString, underHierarchySearchString));
         final StringBuilder sqlBuilder = new StringBuilder(200);
         sqlBuilder.append("select SQL_CALC_FOUND_ROWS ");
         sqlBuilder.append(this.clientMapper.schema());
-        sqlBuilder.append(" where (o.hierarchy like ? or transferToOffice.hierarchy like ?) ");
+        final SQLBuilder extraCriteria = buildSqlStringFromClientCriteria(this.clientMapper.schema(), searchParameters);
+        sqlBuilder.append(" where (o.hierarchy like " + underHierarchySearchString + " or transferToOffice.hierarchy like "
+                + underHierarchySearchString + " ) ");
```
thesmallstar, line 205: @vorburger  can you please see this part? 
I am not sure how to handle this or condition in our SQL builder.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/SQLBuilder.java
```diff
@@ -122,6 +122,18 @@ public String getSQLTemplate() {
...
+     *
+     * @return added criterias
+     */
+    public String getCriteria() {
+        if (sb.length() > 0) {
+            return sb.toString();
+        }
+        return "";
+    }
+
```
thesmallstar, line 137: I thought we would need this, initially, since the "where" was already written in a SQL query which I was not able to replace with sqlBuilder. Please look into other comments, if solved I think this would resolve (needed/to be removed) itself :)

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/service/ClientReadPlatformServiceImpl.java
```diff
@@ -196,25 +197,20 @@ public ClientData retrieveTemplate(final Long officeId, final boolean staffInSel
...
 
         if (searchParameters != null) {
             if (searchParameters.isSelfUser()) {
-                sqlBuilder.append(
-                        " and c.id in (select umap.client_id from m_selfservice_user_client_mapping as umap where umap.appuser_id = ? ) ");
-                paramList.add(appUserID);
+                extraCriteria.addNonNullCriteria(
+                        " c.id in (select umap.client_id from m_selfservice_user_client_mapping as umap where umap.appuser_id = ",
+                        appUserID + ")");
```
thesmallstar, line 211: (definitely wrong current implementation), How to handle nested select queries where the inner query has the parameter? The normal way should not work IMO? Any ideas?

PR #1286, opened 2020-08-26
Removed string concatenated SQL & sqlSearch from Staff (Fineract-854)

thesmallstar:
Removed string concatenated SQL and use of SQLseach parameter from Staff,
I have looked closely in what I am changing but a detailed review would help since we probably don't have good tests coverage on fineract yet.

@awasum  how do you think this as the project idea for outreachy this winter? WE NEED MORE TESTS!!

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/organisation/staff/api/StaffApiResource.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/organisation/staff/api/StaffApiResource.java
@@ -118,7 +118,6 @@ public class StaffApiResource {
             @ApiResponse(responseCode = "200", description = "OK", content = @Content(array = @ArraySchema(schema = @Schema(implementation = StaffApiResourceSwagger.GetStaffResponse.class)))),
             @ApiResponse(responseCode = "200", description = "GET https://DomainName/api/v1/staff?status={ACTIVE|INACTIVE|ALL}", content = @Content(schema = @Schema(implementation = StaffApiResourceSwagger.GetStaffResponse.class))) })
     public String retrieveStaff(@Context final UriInfo uriInfo,
-            @QueryParam("sqlSearch") @Parameter(description = "sqlSearch") final String sqlSearch,
             @QueryParam("officeId") @Parameter(description = "officeId") final Long officeId,
             @DefaultValue("false") @QueryParam("staffInOfficeHierarchy") @Parameter(description = "staffInOfficeHierarchy") final boolean staffInOfficeHierarchy,
             @DefaultValue("false") @QueryParam("loanOfficersOnly") @Parameter(description = "loanOfficersOnly") final boolean loanOfficersOnly,
@@ -130,7 +129,7 @@ public class StaffApiResource {
         if (staffInOfficeHierarchy) {
             staff = this.readPlatformService.retrieveAllStaffInOfficeAndItsParentOfficeHierarchy(officeId, loanOfficersOnly);
         } else {
-            staff = this.readPlatformService.retrieveAllStaff(sqlSearch, officeId, loanOfficersOnly, status);
+            staff = this.readPlatformService.retrieveAllStaff(officeId, loanOfficersOnly, status);
         }
 
         final ApiRequestJsonSerializationSettings settings = this.apiRequestParameterHelper.process(uriInfo.getQueryParameters());
--- a/fineract-provider/src/main/java/org/apache/fineract/organisation/staff/service/StaffReadPlatformServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/organisation/staff/service/StaffReadPlatformServiceImpl.java
@@ -24,12 +24,12 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
-import org.apache.commons.lang3.StringUtils;
 import org.apache.fineract.infrastructure.core.domain.JdbcSupport;
 import org.apache.fineract.infrastructure.core.exception.UnrecognizedQueryParamException;
 import org.apache.fineract.infrastructure.core.service.RoutingDataSource;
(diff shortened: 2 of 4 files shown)


awasum:
> Removed string concatenated SQL and use of SQLseach parameter from Staff,
> I have looked closely in what I am changing but a detailed review would help since we probably don't have good tests coverage on fineract yet.
> 
> @awasum how do you think this as the project idea for outreachy this winter? WE NEED MORE TESTS!!

Improving test overage will be a good Outreachy project. First thing, we need to identify scenarios which require unit or integration test on Fineract. Are there open JIRA issue for test coverage? If not, create one and begin preparing test cases and scenarios so that it will be easy for an intern to work on.. THat is if you are willing to mentor during the upcoming Outreachy session. 

@thesmallstar also engage on the Outreachy thread on the Dev list so we have all these ideas in one place and also create awareness as Outreachy and GSoC are two ways we get new contributors and new features.bug fixes on Fineract. CC @vorburger @ptuomola @xurror @percyashu

thesmallstar:
Argh! failed due to connection error, closing and opening this PR.

thesmallstar:
Failed again due to connection error.

fineract-provider/src/main/java/org/apache/fineract/organisation/staff/api/StaffApiResource.java
```diff
@@ -118,7 +118,6 @@ public StaffApiResource(final PlatformSecurityContext context, final StaffReadPl
             @ApiResponse(responseCode = "200", description = "OK", content = @Content(array = @ArraySchema(schema = @Schema(implementation = StaffApiResourceSwagger.GetStaffResponse.class)))),
             @ApiResponse(responseCode = "200", description = "GET https://DomainName/api/v1/staff?status={ACTIVE|INACTIVE|ALL}", content = @Content(schema = @Schema(implementation = StaffApiResourceSwagger.GetStaffResponse.class))) })
     public String retrieveStaff(@Context final UriInfo uriInfo,
-            @QueryParam("sqlSearch") @Parameter(description = "sqlSearch") final String sqlSearch,
             @QueryParam("officeId") @Parameter(description = "officeId") final Long officeId,
```
ptuomola, line 121: Are we sure that no client is calling this - i.e. will this change make an impact to one of the UIs?

fineract-provider/src/main/java/org/apache/fineract/organisation/staff/service/StaffReadPlatformServiceImpl.java
```diff
@@ -159,7 +159,18 @@ public StaffData mapRow(final ResultSet rs, @SuppressWarnings("unused") final in
...
+        SQLBuilder extraCriteria = new SQLBuilder();
+        extraCriteria.addCriteria(" office_id = ", officeId);
+        extraCriteria.addCriteria(" is_loan_officer= = ", 1);
+        final String hierarchy = this.context.authenticatedUser().getOffice().getHierarchy() + "%";
+        extraCriteria.addCriteria(" o.hierarchy like ", hierarchy);
+
+        sql += " " + extraCriteria.getSQLTemplate();
+        sql = sql + " order by s.lastname ";
+
+        return this.jdbcTemplate.query(sql, rm, extraCriteria.getArguments());
```
ptuomola: Would we not want to do this "magic" of adding the condition for hierarchy and then executing the JDBC template only in one place? I.e. would it not be better to keep this code still in retrieveAllStaff and just call retrieveAllStaff with the right SQLBuilder from here? Or is there a reason why we want to duplicate the code here?

fineract-provider/src/main/java/org/apache/fineract/organisation/staff/service/StaffReadPlatformServiceImpl.java
```diff
@@ -204,49 +215,38 @@ public StaffData retrieveStaff(final Long staffId) {
...
+        return retrieveAllStaff(extraCriteria);
     }
 
-    private Collection<StaffData> retrieveAllStaff(final String extraCriteria, Long officeId) {
+    private Collection<StaffData> retrieveAllStaff(final SQLBuilder extraCriteria) {
 
         final StaffMapper rm = new StaffMapper();
         String sql = "select " + rm.schema();
-        final String hierarchy = this.context.authenticatedUser().getOffice().getHierarchy() + "%";
-        if (StringUtils.isNotBlank(extraCriteria)) {
```
ptuomola, line 218: I think it would have actually been good to leave looking up hierarchy and adding the condition for it here. That way, regardless of how you call retrieveAllStaff, you can't retrieve things you are not allowed to see. Otherwise we are relying on the caller to add the condition for hierarchy - which they may forget...

fineract-provider/src/main/java/org/apache/fineract/organisation/staff/api/StaffApiResource.java
thesmallstar: Yes, we already confirmed that.
It was only used in loans and clients, for which I added extra status param,

thesmallstar:
@ptuomola thanks for a great review, I have updated this!

fineract-provider/src/main/java/org/apache/fineract/organisation/staff/service/StaffReadPlatformServiceImpl.java
```diff
@@ -257,14 +252,7 @@ private String getStaffCriteria(final String sqlSearch, final Long officeId, fin
         // employee who does not belong to his office or a sub office for his
         // office.
```
ptuomola, line 254: I think this comment is now in the wrong place and should be moved to retrieveAllStaff(), right?

thesmallstar:
@ptuomola  updated, thanks for pointing that out.

ptuomola (approved):

PR #1289, opened 2020-08-26
FINERACT-1128: Switching build script from httpie to curl to allow retries

ptuomola:
## Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,8 +30,7 @@ services:
 addons:
   apt:
     packages:
-    # https://httpie.org is used by the docker-compose test below
-    - httpie
+    - curl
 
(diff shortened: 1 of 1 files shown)


xurror (approved):
LGTM

ptuomola:
@vorburger here's a PR to use curl with retries instead of httpie. If the connection failures are due to the container not coming up quickly enough, then I hope this will address this. I have another theory that they might instead be due to Fineract coming up faster than MySQL - if that's the case then this will not help, but let's start with the easy approach and see if that works.

xurror:
The error generally arise when pulling the community-app which has no direct relation with whether MySQL is pulled or not. I want to believe this will resolve the difficulties pulling the community-app image. Otherwise my theory will be to instead checkout the community-app's image or the way it's built. Let's merge this first and watch what happens.

PR #1292, opened 2020-08-26
FINERACT-1129: Gradle task to create release package

vidakovic:
## Description
see: https://issues.apache.org/jira/browse/FINERACT-1129. Needs to be backported to 1.4.0.

## Checklist


README.md
```diff
@@ -353,6 +353,41 @@ Releasing
...
 [How to Release Apache Fineract](https://cwiki.apache.org/confluence/x/DRwIB) documents the process how we make the source code that is available here in this Git repository into a binary release ZIP available on http://fineract.apache.org.
 
+Before you use Gradle to create a release you need to make sure that your GPG is properly setup and that you have set the following properties in your ~/gradle/gradle.properties:
+```
+signing.gnupg.keyName=7890ABCD
+signing.gnupg.passphrase=secret
+```
+
+IMPORTANT: Do not set your GPG secrets in one of the project gradle.properties and double check that you are not accidentally committing them to Git.
+
```
ptuomola, line 365: To avoid exactly this, is it possible to instruct people instead to pass these as command line parameters when they run the build?

README.md
```diff
@@ -353,6 +353,41 @@ Releasing
...
+NOTE: Let's assume your GPG key ID would be "ABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890ABCD" then you have to use the last 8 characters (i. e. "7890ABCD") for the signing plugin property "signing.gnupg.keyName". 
+
+Execute the following task to create a distribution with an ASCII armored signature (.asc) and a SHA512 checksum file (.sha512):
+```
+./gradlew -Pfineract.release clean build 
+```
```
ptuomola, line 369: Would it be cleaner to have a specific distribution task that does all the distribution activities (package and signing etc), rather than relying on the build task and a new profile?

README.md
```diff
@@ -353,6 +353,41 @@ Releasing
...
+Above task will create the following files in folder build/distributions:
+
+- distribution file: apache-fineract-1.4.0-binary.tar.gz
+- ASCII armored signatures: apache-fineract-1.4.0-binary.tar.gz.asc
+- SHA512 checksum: apache-fineract-1.4.0-binary.tar.gz.sha512
+
+Verify the validity of the release distribution with:
+```
+gpg --verify build/distributions/apache-fineract-1.4.0-binary.tar.gz.asc
+```
```
ptuomola, line 389: Would it make sense to include this in the distribution task as well - so that this gets automatically ran and the user just has to review the output?

fineract-provider/gradle.properties
```diff
@@ -16,5 +16,5 @@
 # specific language governing permissions and limitations
 # under the License.
 #
-releaseVersion=1.0.0
+releaseVersion=1.4.0
 buildType=BUILD
```
ptuomola, line 20: Here we have the version hardcoded - for the Swagger documentation we pick it automatically from the git describe output. We should probably make this consistent and use one or the other...

README.md
vidakovic: You can do this anyway. Personally I prefer to set this up with my personal gradle.properties in my home folder, but it's not required; you can just pass the required parameters with "-P" on the command line and it'll work as well.

README.md
vidakovic: I had that conversation with  @edcable and it was preferred to include this "naturally" (i. e. without a specific task) in the build process. Note: the release artifacts are only created when "-Pfineract.release" is provided on the command line which is pretty explicit. For the next release I intend to add even more features here (publishing client jar on Maven Central, uploading distribution to Apache, maybe publishing release Confluence page).

README.md
vidakovic: Good point, I'll add this.

fineract-provider/gradle.properties
vidakovic: I saw that you use a Gradle version plugin to do this and I actually prefer that way. For this release I'll leave it to the hard coded version number, but would suggest your method for the whole project. Note: the version of the client JAR in the release branch is not correct (shows something like "client-1.3.0-xyz-dirty.jar"; I do understand that it creates that based on the last release tag; we just have to make sure that this works correctly when the release build is running (release tag 1.4.0 has to exist).

README.md
ptuomola: Thanks. I know you can do this either way, I just wondered if it would be better in the README to suggest that way - rather than suggest that people put it in gradle.properties AND warn them not to check it in. If the suggested way is to use command line there's no risk that people will check anything in.

fineract-provider/gradle.properties
ptuomola: Yes - good point, please let's make sure it works correctly when you do the release build. If not happy to help fix it if needed.

README.md
ptuomola: OK - if it's been decided, then that's fine. Personally I would have preferred that all different activities are triggered as tasks: it's a bit confusing now that all other activities are triggered as tasks and only these with a profile. 

Also as far as I know, you can only have one profile active - correct? So if in the future we want to have different types of builds activated as profiles, you can't create distributions of them.

README.md
vidakovic: I've added a clarification to the README describing both ways to provide the necessary parameters (either in your home gradle.properties or via command line).

vidakovic: This section is anyway only intended for a limited audience (I think there are 3 or 4 people right now that can sign the releases). But you have a point here. Happy to change the default recommendation for the next release. Maybe you can raise that issue then on the mailing list to see what others have to contribute on that subject?

README.md
vidakovic: My first "reflex" was also to add an explicit task, but have to say that I'm not necessarily opposed to the way it's now (aka as a side effect of the distribution task). Having said that: given the short time frame I made a couple of lone decisions so I think it makes a lot of sense to present this to a broader audience (mailing list) and let them have a say how this should look like for the next release.

For now I just wanted to have something that helps me during the release.

I'm not sure how this (creating the release via a command line parameter) would impede different profiles? Let's discuss this on the mailing list or maybe in a separate Jira ticket?

vidakovic:
@ptuomola @vorburger do you have anything else that's urgent to add to this improvement? If not I'd like to merge this to develop and port it back to branch 1.4.0.

Grandolf49 (approved):

PR #1284, opened 2020-08-26
add Merge Strategy section to README

vorburger:
@ptuomola @xurror @awasum @avikganguly01 @thesmallstar @percyashu @fynmanoj @vidakovic @conradsp @Grandolf49 how about (something like) this to document the recommended _Merge Strategy_ for this project? Your review feedback, if any, most welcome, of course.

PR #1296, opened 2020-08-27
FINERACT-1135: Fix SQL error at loan repayment template

francisguchie:
## Description
https://issues.apache.org/jira/browse/FINERACT-1135

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanReadPlatformServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanReadPlatformServiceImpl.java
@@ -431,7 +431,7 @@ public class LoanReadPlatformServiceImpl implements LoanReadPlatformService {
         this.context.authenticatedUser();
 
         RepaymentTransactionTemplateMapper mapper = new RepaymentTransactionTemplateMapper();
-        String sql = "select " + mapper.schema() + " where l.id =?";
+        String sql = "select " + mapper.schema();
         LoanTransactionData loanTransactionData = this.jdbcTemplate.queryForObject(sql, mapper, LoanTransactionType.REPAYMENT.getValue(),
                 loanId, loanId);
         final Collection<PaymentTypeData> paymentOptions = this.paymentTypeReadPlatformService.retrieveAllPaymentTypes();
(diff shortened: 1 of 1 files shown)


xurror (approved):
Looks OK to me.

awasum:
Minor feedback for now and the future, Always use the issue number in all caps when using it in a commit message: `FINERACT-1135: Fix Error at Loan Repayment` will be commit message or `Fix Error at Loan Repayment (FINERACT-1135) ` as in upcoming: #1284

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanReadPlatformServiceImpl.java
```diff
@@ -431,7 +431,7 @@ public LoanTransactionData retrieveLoanTransactionTemplate(final Long loanId) {
         this.context.authenticatedUser();
 
         RepaymentTransactionTemplateMapper mapper = new RepaymentTransactionTemplateMapper();
-        String sql = "select " + mapper.schema() + " where l.id =?";
+        String sql = "select " + mapper.schema();
         LoanTransactionData loanTransactionData = this.jdbcTemplate.queryForObject(sql, mapper, LoanTransactionType.REPAYMENT.getValue(),
```
ptuomola, line 435: If we are removing one parameter from the query, should we not also pass one less parameter when we execute the query?

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanReadPlatformServiceImpl.java
```diff
@@ -2176,35 +2176,36 @@ public CurrencyData mapRow(ResultSet rs, @SuppressWarnings("unused") int rowNum)
...
+                    "ls.fee_charges_amount - IFNULL(ls.fee_charges_completed_derived,0) - IFNULL(ls.fee_charges_writtenoff_derived,0) - IFNULL(ls.fee_charges_waived_derived,0) as feeDue, ");
             sqlBuilder.append(
-                    "ls.penalty_charges_amount - IFNULL(ls.penalty_charges_completed_derived,0) - IFNULL(ls.penalty_charges_writtenoff_derived,0) - IFNULL(ls.penalty_charges_waived_derived,0) as penaltyDue,");
+                    "ls.penalty_charges_amount - IFNULL(ls.penalty_charges_completed_derived,0) - IFNULL(ls.penalty_charges_writtenoff_derived,0) - IFNULL(ls.penalty_charges_waived_derived,0) as penaltyDue, ");
             sqlBuilder.append(
-                    " l.currency_code as currencyCode, l.currency_digits as currencyDigits, l.currency_multiplesof as inMultiplesOf, rc.`name` as currencyName, ");
-            sqlBuilder.append(" rc.display_symbol as currencyDisplaySymbol, rc.internationalized_name_code as currencyNameCode ");
-            sqlBuilder.append(" FROM m_loan l");
```
ptuomola, line 2191: Is there a reason for removing all these spaces from beginning of the strings? The reason for them, as far as I can see, is avoiding a situation where someone adds before a string that doesn't have a space at the end, and causes the next string to therefore generate broken SQL. I don't think they are harmful and can save us from some accidental illegal SQL...

francisguchie:
> Minor feedback for now and the future, Always use the issue number in all caps when using it in a commit message: `[FINERACT-1135](https://issues.apache.org/jira/browse/FINERACT-1135): Fix Error at Loan Repayment` will be commit message or `Fix Error at Loan Repayment ([FINERACT-1135](https://issues.apache.org/jira/browse/FINERACT-1135)) ` as in upcoming: #1284

thanks @awasum

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanReadPlatformServiceImpl.java
francisguchie: I did not remove the parameter i only put it in its rightful place 
            sqlBuilder.append(" GROUP BY ls.duedate");	            sqlBuilder.append("WHERE l.id = ? ");
just before the groub by object

look at line 2200  

But still the sql runs even without this paramater is it is already called in other places like line 2198, line 2200 is just a "fail safe" parameter just in case there could be more than one result generated by the SQL script ( you never know)

ptuomola (approved):
LGTM

vidakovic:
@francisguchie any chance to get this backported to the release branch 1.4.0? The Jira ticket for that is still open... would appreciate if we could do this soon to push the release over the finish line. Thanks.

godiraonedipuo:
Good day, am failing to import LoanRepayment on fineract using template. This is the screenshot of my template download after it fails to upload the records. 

Monosnap 7d0ab762-b2e1-42f7-86d4-974a626f23f6 xls

Thank you, your support is highly appreciated.

PR #1309, opened 2020-08-30
FINERACT-1140 - Adopt GenerationType.TABLE as ID generation strategy [Discussion]

xurror:
Considering the discussion we had on the mailing list about the way to set up this generation strategy. 
I tried setting up an AbstractPersistable class for this but it didn't work out because the IDs for all ~200 entities where being generated by one table and thus the IDs were not controlled, causing even way more trouble than what this change was meant to resolve.
I know there was alot of comments about this being confusing but I am still trying to picture how bad it is to have this applied to just one entity.

### DIFF
--- a/fineract-provider/build.gradle
+++ b/fineract-provider/build.gradle
@@ -130,6 +130,7 @@ dependencyManagement {
         dependency 'org.mock-server:mockserver-junit-jupiter:5.11.0'
         dependency 'org.webjars.npm:swagger-ui-dist:3.32.5'
         dependency 'org.webjars:webjars-locator-core:0.46'
+        dependency 'org.gaul:modernizer-maven-annotations:2.1.0'
 
         dependency ('org.dom4j:dom4j:2.1.3') {
             exclude 'relaxngDatatype:relaxngDatatype' // already in com.sun.xml.bind:jaxb-osgi:2.3.0.1
--- a/fineract-provider/dependencies.gradle
+++ b/fineract-provider/dependencies.gradle
@@ -80,7 +80,8 @@ dependencies {
 
             'com.github.spotbugs:spotbugs-annotations',
             'io.swagger.core.v3:swagger-annotations',
-            'org.webjars:webjars-locator-core'
+            'org.webjars:webjars-locator-core',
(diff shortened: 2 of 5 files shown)


xurror:
Hey @ptuomola @vorburger @awasum any input on this?

ptuomola:
Sorry @xurror - not my area of expertise! It's not clear to me why we would need to change the GenerationType for this table - i.e. why doesn't the previous setting work. Do you know? 

And re the GenerationType.TABLE - all I can find on the web is recommendations never to use it. Seems like this is slow (which makes sense, you have to keep updating the table) and SEQUENCE does a much better job. Was TABLE the only one that worked for you, and any idea what the performance overhead is?

vorburger:
My concern here is that, apparently, we would have to do this for EACH
Entity (Table), separately? I doubt that's a good idea. If I understood it
correctly , this comes out of a problem encountered during the EclipseLink
migration. Couldn't we solve whatever problem that was, instead of having
to switch GenerationType, per table?

>

xurror:
Here comes my concern exactly.
I wasn’t able to resolve the problem with eclipselink cause turns out in
certain cases IDs were not generated fast enough so I constantly got null
pointers exception when trying to persist the entity.
Of all my attempts, having a table of pre-generated IDs was what did the
trick.

ptuomola:
Hi 

Quick search for people who have had a similar problem comes up with this:

https://stackoverflow.com/questions/12429237/eclipselink-returns-entity-with-null-primary-key

In this case it was resolved by setting eclipselink.weaving.internal to "false"

Is this something you've already tried? If not, would it be worth giving it a go?

xurror:
I can try it out.

xurror:
> Hi
> 
> Quick search for people who have had a similar problem comes up with this:
> 
> https://stackoverflow.com/questions/12429237/eclipselink-returns-entity-with-null-primary-key
> 
> In this case it was resolved by setting eclipselink.weaving.internal to "false"
> 
> Is this something you've already tried? If not, would it be worth giving it a go?

Unfortunately, that didn't fix but I'm looking at other solutions. I'm drafting this for now.

PR #1320, opened 2020-09-10
Change OAuth doc in apiLive.htm (FINERACT-629)

vorburger:
see FINERACT-629

Nota bene that I HAVE NOT actually tested this recommended JavaScript - I've just made it up, based on a (very) brief look at JQuery doc; perhaps someone would like to actually try out if this works, or someone with more JS experience could squint at it to see if it looks about right?


vorburger:
Self merged documentation only PR after (almost) 2 weeks of no reaction from any other committer.

ptuomola:
@vorburger I saw your PR, but unfortunately I have no experience with OAuth - so I thought maybe someone who has used it could comment...

PR #1366, opened 2020-10-02
FINERACT-982 - Completely ditch use of Drizzle JDBC Driver

xurror:
FINERACT-982

### DIFF
--- a/README.md
+++ b/README.md
@@ -381,6 +381,12 @@ Our `ClasspathHellDuplicatesCheckRuleTest` detects classes that appear in more t
 Releasing
 ---------
 
+PRE-RELEASE NOTICE: Please before starting any release process, ensure that the you have replaced mysql connector package from mariadb connector/j back to drizzle driver. Files requiring changes are:
+* docker-compose.yml
+* fineract-provider/build.gradle
+* fineract-provider/src/main/resources/META-INF/spring/jdbc.properties
+* kubernetes/fineract-server-deployment.yml
+
 [How to Release Apache Fineract](https://cwiki.apache.org/confluence/x/DRwIB) documents the process how we make the source code that is available here in this Git repository into a binary release tar.gz available on http://fineract.apache.org.
 
 Before you use Gradle to create a release you need to make sure that you provide the proper GPG parameters. You have to options:
--- a/fineract-provider/build.gradle
+++ b/fineract-provider/build.gradle
@@ -276,23 +276,24 @@ import groovy.sql.Sql
 
 project.ext.mysqlUser='root'
 project.ext.mysqlPassword='mysql'
+project.ext.driver='org.mariadb.jdbc.Driver'
 
 configurations {
-    driver
+    connector
 }
 dependencies {
-    driver 'org.drizzle.jdbc:drizzle-jdbc:1.4'
+    connector "${mysqlConnector}:${mysqlConnectorVersion}"
 }
 
 URLClassLoader loader = GroovyObject.class.classLoader
-configurations.driver.each {File file ->
+configurations.connector.each {File file ->
     loader.addURL(file.toURL())
 }
 
 task createDB {
     description= "Creates the Database. Needs database name to be passed (like: -PdbName=someDBname)"
     doLast {
-        def sql = Sql.newInstance( 'jdbc:mysql:thin://localhost:3306/', mysqlUser, mysqlPassword, 'org.drizzle.jdbc.DrizzleDriver' )
+        def sql = Sql.newInstance( 'jdbc:mysql://localhost:3306/', mysqlUser, mysqlPassword, driver )
         sql.execute( 'create database '+"`$dbName`" )
     }
 }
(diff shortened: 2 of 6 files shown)


xurror:
I won't bet on this passing the build. I ran tests locally 3 times but some schedular jobs test kept failing. I guess schedular jobs will just do their thing.
But if this passes then it's cool

vidakovic:
I would externalize all references to specific drivers. What I mean with that is: instead of hard-coding the driver's class names and connection strings I would replace it with environment variables. Defaults for those variables would be the Drizzle related settings (for license compatibility), but anyone running Fineract in production (I guess that's the concern here) would then overwrite those environment variables either in the docker-compose.yml file or on the command line. That way no tricks have to be applied during releases and we don't have to reference the LGPL driver in this repository... which - as much as I understand it - we actually can't (LGPL being incompatible with Apache License 2)... unless I'm missing anything here?

xurror:
@vidakovic  I think we also have a challenge using the current config in development as it causes challenges during testing. You can follow discussion here on why these changes are the way they are: https://issues.apache.org/jira/browse/FINERACT-982 https://issues.apache.org/jira/browse/FINERACT-980.
But I like your suggestions very much, just think it will be a PITA to be setting env variables each time we trying to run tests. Makes it hard for us and for new contributors.
What do you think?

FYI: @vorburger @ptuomola @awasum

vidakovic:
In the end there are various mechanisms that we could use here to set different properties (e. g. Spring profiles)... agreed, tests would need maybe a bit of thinking... 

The one thing that seems to me unsolvable (disclaimer: I'm not a lawyer) is the dependency reference to the LGPL jar. And if we can't include that then the devs have to do manual fiddling anyway.

Easiest in terms of contributors would be (but these are more long-term - if it all - solutions that are more intrusive):

*   fix Drizzle and bring it to the same quality level as the "official" drivers (I think that's going to take a while)
*   find another Apache License 2 compatible driver (searched already extensively; I'd say there is non available)
*   switch to PostgreSQL as the default database (but you would still have a lot of people who want/need MySQL; so that wouldn't solve their problem)

Just spitballing...

xurror:
Adopting profiles was my first option too. Perhaps we could get more inputs how to go about this from the community.

vidakovic:
:+1:

vidakovic:
Ok, I think I have a bit more info how other Apache projects deal with the issue of (L)GPL licensed libraries. See e. g. https://github.com/apache/camel/blob/master/components/camel-debezium-mysql/pom.xml. The important part to note here is the dependency scope ("provided") which means they don't include the dependency in the final build artifact. As far as I know the closest equivalent in Gradle should be compileOnly. That means that at least from the IDE perspective this issue should be solved (e. g. IntelliJ doesn't make a distinction between these scopes). Not sure if that applies also for running unit tests on the console, would need to test. It will not be included in the release package or Docker images (automatically). The current Dockerfile does download mysql-connector-j from Maven Central with a wget command and the jar file is then included to the Fineract startup with "java -cp"... I guess that means the Docker images are not considered release artifacts.

Just FYI

xurror:
Nice digging. I'll try it out.

README.md
```diff
@@ -379,6 +379,12 @@ Our `ClasspathHellDuplicatesCheckRuleTest` detects classes that appear in more t
 Releasing
 ---------
 
+PRE-RELEASE NOTICE: Please before starting any release process, ensure that the you have replaced mysql connector package from mariadb connector/j back to drizzle driver. Files requiring changes are:
+* docker-compose.yml
+* fineract-provider/build.gradle
+* fineract-provider/src/main/resources/META-INF/spring/jdbc.properties
+* kubernetes/fineract-server-deployment.yml
+
```
vorburger, line 389: Nah, this will be super confusing... we cannot (should not) have the release be anything different than what we work with during development. Let's figure out a way to avoid this. My proposal is that we simply do not ship any JDBD driver in the Fineract distribution at all, ever.

Instead, why don't we just include a 2-3 lines trivial small fineract.sh script illustrating how to launch `java ...` with both `fineract.jar` and `mariadb-java-client-2.7.0.jar` on the CP? And document that in https://github.com/apache/fineract#instructions-to-build-the-jar-file as well (and also a word about having to add it to Tomcat when using the WAR), as part of this PR. And then we could entirely remove this section for "special instructions" for Releasing. WDYT?

As for actually getting the `mariadb-java-client-2.7.0.jar`, it must be available somewhere under `.gradle/` or so? I'm not just where you can find it... or we could just `wget` (or `curl` - whatever) the MariaDB Connector/J?

fineract-provider/dependencies.gradle
```diff
@@ -80,7 +79,8 @@ dependencies {
 
             'com.github.spotbugs:spotbugs-annotations',
             'io.swagger.core.v3:swagger-annotations',
-            'org.webjars:webjars-locator-core'
+            'org.webjars:webjars-locator-core',
+            mysqlConnector
```
vorburger: So this, unfortunately, as I understand it, is No Go. This would mean that we include a LGPL into the distro. Or that we would have to make special handstands to remove it again when we release to distribute, which I fear will be too confusing.

What we CAN do, as far as my understanding goes, is to have a test only scope dependency. So for unit tests, it's on the classpath, but it's never "packaged" and distributed. We currently only have one Spring Integration Test, but I would still do add it, with another scope than `implementation`, but some... test only something (whatever the right name is).

fineract-provider/build.gradle
```diff
@@ -29,6 +29,8 @@ repositories {
 }
 
 project.ext.jerseyVersion = '1.19.4'
+project.ext.mysqlConnector = 'org.mariadb.jdbc:mariadb-java-client'
```
vorburger: how about ditching the _mysql_ prefix from the name here, and just calling this (and the following)... maybe `jdbcArtifact` and `jdbcVersion`?

vorburger (changes requested):
NOK, as is - but close! This is cool, and long overdue.. hope you can finish this to a state where it's OK to merge!

vorburger:
> Not sure if that applies also for running unit tests on the console, would need to test.

So for `./gradlew integrationTest` to work with what I have proposed above, where I suggest that we do NOT include any JDBC driver JAR into the WAR, we would just have to add it (a JDBC driver) to the external Tomcat which we start for running the ITs. This, hopefully, shouldn't actually be that hard - note the [`extraClasspath` and `sharedClasspath` properties](https://github.com/bmuschko/gradle-cargo-plugin#extension-properties), which we should be able to use somewhere in [our Cargo plugin configuration](https://github.com/apache/fineract/blob/develop/fineract-provider/build.gradle#L682)?

xurror:
Thanks for the input. I’ll look into it as soon as I can.

vorburger:
BTW we'll probably also have to do something to add the JDBC driver to the
classpath for ./gradlew bootRun.

PR #1368, opened 2020-10-04
Improve error handling if reporting service implementation is missing [FINERACT-1173]

vorburger:
see FINERACT-1173 (for FINERACT-1127)

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/api/RunreportsApiResource.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/api/RunreportsApiResource.java
@@ -25,10 +25,6 @@ import io.swagger.v3.oas.annotations.media.Schema;
 import io.swagger.v3.oas.annotations.responses.ApiResponse;
 import io.swagger.v3.oas.annotations.responses.ApiResponses;
 import io.swagger.v3.oas.annotations.tags.Tag;
-import java.io.File;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Set;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.DefaultValue;
 import javax.ws.rs.GET;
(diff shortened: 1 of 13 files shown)


fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/api/RunreportsApiResource.java
```diff
@@ -124,12 +125,13 @@ public Response runReport(@PathParam("reportName") @Parameter(description = "rep
             parameterTypeValue = "report";
             String reportType = this.readExtraDataAndReportingService.getReportType(reportName, isSelfServiceUserReport);
             ReportingProcessService reportingProcessService = this.reportingProcessServiceProvider.findReportingProcessService(reportType);
-            if (reportingProcessService != null) {
-                return reportingProcessService.processRequest(reportName, queryParams);
+            if (reportingProcessService == null) {
+                throw new PlatformServiceUnavailableException("err.msg.report.service.implementation.missing",
+                        ReportingProcessServiceProvider.SERVICE_MISSING + reportType, reportType);
```
ptuomola: Unfortunately based on my read of the code, I don't think this is quite right:

After this change, we are bailing out if we can't find a reporting process service for the report type. 
Previously, if we couldn't find one, we would continue to execute the "hardcoded" reporting process service in the same function below. 

I think continuing execution is the expected behaviour: for any report that is not of type Pentaho or SMS, the report_type is set to Table. But there is no ReportingProcessService with type Table, so therefore the logic in this function below is used to simply prepare the CSV / PDF file. 

In other words: if we check this in, as far as I can see it would break any report that doesn't have a specific report handler configured. 

Now having said that, a good idea would be to create a "Table" type report handler and move the code from this function to that report handler. And after that we could assume that there's a handler for each type of report, and bail out if we can't find one.

ptuomola (changes requested):

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/api/RunreportsApiResource.java
vorburger: Oh... wow, very well spotted. I'll have a go at the re-factoring you're suggesting, that would make a lot of sense, yes.

But before I get to that, I'm struggling with more fundamental classloading issues... urgh!! More soon on JIRA, hopefully.

vorburger: @ptuomola I'm finally (7 weeks later, phew!) picking this up, and I'm struggling to understand how the `reportType: SMS` is intended to work... @vidakovic do you understand that? Can those be run on the `/runreports` API? I think what I'll do is just write a `DefaultReportingProcessService` with the current code to which this will delegate to for both `Table` and `SMS`.. I'll therefore make the `type` on `org.apache.fineract.infrastructure.report.annotation.ReportService` a multivalue. Does that sounds reasonable to you guys?

vidakovic: Also not sure how this SMS thing is supposed to work... will have a look at this later today.

vorburger: Actually I guess that one would then more appropriately be our `DatatableReportingProcessService` (in `infrastructure.dataqueries.service`) instead of a `DefaultReportingProcessService` in  `infrastructure.report.service.ReportingProcessService`.... watch for an update on this PR! :smile_cat:

vorburger:
@ptuomola re-review this, better now? @vidakovic do you also want to co-pair-review this? Let's give Petri a chance to look it over, so don't merge, just LGTM; this one is a big enough change to warrant 4 extra eyes.. this was supposed to be "just adding a check" but turned into a fairly big refactoring! :smile:

vidakovic (approved):
LGTM

ptuomola (approved):
LGTM... sorry for the delay, moving across the world with family in the times of a pandemic is not easy!

PR #1383, opened 2020-10-10
Remove mavenCentral(), because jcenter() includes it

vorburger:


vidakovic (approved):
LGTM

vorburger:
Self merged based on 1 LGTM, this change is trivial enough that I don't think we need several LGTM.

PR #1387, opened 2020-10-10
Merge Release 1.4.0 branch back to develop (FINERACT-1154)

ptuomola:
see FINERACT-1154


vorburger:
@ptuomola re. our discussion (together with @vidakovic) in FINERACT-1154, are you sure this is right, as is? If I'm following along correctly here, you have created [a `release-1.4.0` branch](https://github.com/apache/fineract/compare/release-1.4.0), presumably from [the 1.4.0 tag](https://github.com/apache/fineract/commits/1.4.0), and then in ffed594 _Merge branch 'develop' into release-1.4.0_ (don't we want to do the opposite?), and now this PR wants to merge 16 commits from that `release-1.4.0` branch back into `develop` - which is a NOOP, with no diff.

The way I understood what you were after in FINERACT-1154 would be more something like this:

```
git checkout develop
git checkout -b FINERACT-1154_merge-1.4.0
git merge -s recursive -Xignore-all-space 1.4.0
```

This showed me 3 conflicts, which were relatively trivial to resolve; the result is https://github.com/apache/fineract/compare/develop...vorburger:FINERACT-1154_merge-1.4.0?expand=1. It's very similar to this - except for the last merge commit... your _Merge branch 'develop' into release-1.4.0_ has parents ee99f30 + e985597, whereas my _Merge tag '1.4.0' into develop (FINERACT-1154)_ has parents 3ee46b0 + ee99f30 - the "other way around" (and a later point on develop). Which is right/better? I'm honestly :smile_cat: asking, as I'm not 100% sure what we are after here. Whatever we conclude, we should then document it on https://cwiki.apache.org/confluence/display/FINERACT/How+to+Release+Apache+Fineract.

ptuomola:
Hi @vorburger - I'm almost sure I'm **not** doing this right :-) So really appreciate your input...

As far as I can see, what we need to achieve is to have the tagged 1.4.0 release commit (ee99f30) reachable from the develop branch - so that "git describe" and jgitver etc can "see" it in commit history of develop. 

As far as I can see, either approach will achieve that. The only reason resolved the merge conflict "the wrong way" (i.e. into the branch rather than into develop) was because the Github UI made me do it. I agree with you that doing it "the other way" makes more sense - but as far as I can see, the outcome is the same regardless.

But I'm definitely not an expert on this, so would be very keen to understand pros / cons of either approach!

vorburger:
> As far as I can see, what we need to achieve is to have the tagged 1.4.0 release commit (ee99f30) reachable from the develop branch - so that "git describe" and jgitver etc can "see" it in commit history of develop.

Can I admit that I've only really fully understand now :smiling_imp: that that is the real reason for why you want to do this? I've played a bit with both `git describe` (and https://jgitver.github.io / https://github.com/jgitver/jgitver), and now see what you mean! I'll leave some related notes in FINERACT-1154.

> As far as I can see, either approach will achieve that.

Yeah, I think you're probably right - the order of the parents in a merge commit doesn't actually seem to really matter.

> I agree with you that doing it "the other way" makes more sense - but as far as I can see, the outcome is the same regardless.

OK, I've raised PR #1390 from my branch, where I have just updated the commit message, and we merge that one, just for clarity? Close this?

PR #1389, opened 2020-10-11
Remove license header from .dockerignore (FINERACT-1195)

vorburger:
see FINERACT-1195

PR #1390, opened 2020-10-11
 Merge git tag '1.4.0' into branch 'develop' (FINERACT-1154)

vorburger:
see related discussions in FINERACT-1154 and #1387, as per commit message on the merge commit:

    git checkout develop
    git checkout -b FINERACT-1154_merge-1.4.0
    git merge -s recursive -Xignore-all-space 1.4.0

> see discussion in https://issues.apache.org/jira/browse/FINERACT-1154 for the full background about why we are doing this; the TL;DR is that this makes `git describe` work correctly, which is important for automation.

@ptuomola would you be willing to merge this as an alternative to your #1387 ?

@vidakovic FYI


ptuomola:
@vorburger I'm assuming the right thing to do now is to just do a "merge pull request" - i.e. not a squash, or rebase? Just confirming before I mess something up...

vorburger:
> @vorburger I'm assuming the right thing to do now is to just do a "merge pull request" - i.e. not a squash, or rebase? Just confirming before I mess something up...

Yup! Note that you actually cannot use _Rebase and merge_ (see that it's "greyed out" on the UI?). And _Squash and merge_, while available (green) and "possible", as far as I understand things actually wouldn't do at all what you/we are after here - I think that would create 1 "empty" commit, without the merge commit that's really what this is all about (`git describe`) - pointless. So TL;DR, yeah, click _Merge pull request_ which _Create a merge commit_. My hope here is that GitHub is somehow smart enough to see that this PR ALREADY ends with a Merge Commit (the 09e08fdc71e86d0133ebf0c27f829d07738208af), and doesn't add ANOTHER one. If it does, then IMHO that's a bit "ugly" - but not the end of the world either. Go for it! :smiling_imp:

ptuomola:
It seems to have worked! Great stuff. And now "git describe" also says something useful.

vorburger:
> My hope here is that GitHub is somehow smart enough to see that

This didn't work - we now have 2 merge commits about this (the original 09e08fdc71e86d0133ebf0c27f829d07738208af AND 5365da4d243da92f4aa6a2edd17adf40d5c55950). I suspect that this most likely is because we merged #1389 as e978971c679d36705e38f3f4d0afe129ba7bd2de - if we had done it in the opposite order, perhaps GitHub would not have added that 2nd merge commit? Anyway, doesn't matter!

PR #1398, opened 2020-10-13
Minor: Re-use Random in ContentRepositoryUtils instead of creating new one

vorburger:
@ptuomola does this make sense to you?


ptuomola:
As implementation makes a lot more sense than creating a new instance every time.

But I do wonder why the length of the random string needs to be random. Why not just have a fixed length random string? Interesting...

ptuomola (approved):
LGTM

PR #1402, opened 2020-10-14
FINERACT-1189: fineract client as a separate module

vidakovic:
## Description

Some pointers for your reviews:

1. fineract-client uses the official OpenAPI Gradle plugin to generate sources; I've removed the plugin that was previously used from fineract-provider; the CLI based code generation is also removed (including spawning a separate Gradle build); just to say fineract-client is a true module now
2. the "allprojects" section contains now more plugin configurations that can be applied to all modules; I've moved Spring dependency management, Rat and License check there, because like this we only have to configure them once
3. there is a new section (after "allprojects") that contain every plugin configuration common to Java modules. Source sets, compiler arguments, JDK compatibility as well as Spotless Java related configurations are done in one place here. We can move more stuff here, e. g. Errorprone, Jacoco, Git properties, Checkstyle, Modernizer, Spotbugs, project report; these plugins are not fineract-provider specific, but can be applied to any Java module (fineract-client included); I just wanted to limit this first set of changes (already 25+ files changed)
4. please look at line 26 of the build.gradle file in the root folder to see how the selection of Java projects works; this allows us to use one common configuration for Java projects (see line 333)
5. fineract-client/src/main/templates contains OpenAPI/Swagger code generator template files that fix a couple of style errors or just add licenses; just to have the bare minimum there; these sources are still excluded from Spotless and other checks, because... well, generated code.
6. in the current upstream develop branch the Swagger generated code is built on Travis with "mvn test"; I would be surprised if we have any tests there... I remember there's a flag to generate test stubs, but how useful are these skeleton tests then? Have to check this again
7. I still have to decide how to handle the Swagger file ("fineract.yaml"). In my first prototype I've copied it from fineract-provider to fineract-client, because I couldn't convince Gradle to build the modules in a certain order (fineract-provider -> fineract-client); it's not really a good practice to do this anyway in Gradle, but still there are a few mechanics that should help (like declaring a project dependency or something along those lines). If I manage to get this working then we don't need to Git manage fineract.yaml under fineract-client (we can just pick it up in fineract-provider/build). Just FYI

## Checklist


build.gradle
```diff
@@ -85,51 +218,235 @@ allprojects  {
...
+                )
             }
         }
 
+        // TODO: @vidakovic why do we need this again? Spotless is checking anyway
+        // if (project.hasProperty("automatedBuild")) {
+        //     dependsOn(spotlessCheck)
+        // } else {
+        //     dependsOn(spotlessApply)
+        // }
```
ptuomola: There was a bit of a debate about this earlier. Initially we just had spotlessCheck which complained about any formatting issues, and then people had to manually run spotlessApply to fix them. There was a fair bit of dissatisfaction about this - and hence this was introduced: if you run build manually it will automatically fix any spotless issues with spotlessApply - but when you run the build in Travis, it just checks and doesn't modify the code. If we want to change that behaviour now that should be a different JIRA & PR up for discussion...

build.gradle
```diff
@@ -46,6 +56,7 @@ buildscript {
 plugins {
     id 'io.spring.dependency-management' version '1.0.10.RELEASE'
     id 'com.diffplug.spotless' version '5.6.1'
+    id 'org.openapi.generator' version '4.3.1'
```
ptuomola: Hmm... I thought we wanted to only have "common" plugins in the main Gradle file. But the OpenAPI one is only needed / used on the client project. Is there any way to make it specific for that subproject only?

build.gradle
```diff
@@ -63,14 +74,136 @@ allprojects  {
...
+            dependency 'io.github.classgraph:classgraph:4.8.90'
+            dependency 'org.awaitility:awaitility:4.0.3'
+            dependency 'com.github.spotbugs:spotbugs-annotations:4.1.3'
+            dependency 'javax.cache:cache-api:1.1.1'
+            dependency 'org.mock-server:mockserver-junit-jupiter:5.11.1'
+            dependency 'org.webjars.npm:swagger-ui-dist:3.35.1'
+            dependency 'org.webjars:webjars-locator-core:0.46'
+
+            // fineract client dependencies
+            dependency "com.squareup.retrofit2:retrofit:$retrofitVersion"
```
ptuomola, line 129: It's going to be pretty big list to have all dependencies in a single list covering all possible future subprojects. Is there any way to make these lists subproject-specific, and only keep any common dependencies here?

fineract-client/src/main/templates/java/retrofit2/ApiClient.mustache
```diff
@@ -0,0 +1,468 @@
+{{>licenseInfo}}
+package {{invokerPackage}};
```
ptuomola: These files look like boilerplate... have we modified them? If not, is there a possibility to have them as dependency we download rather than something we check into git?

ptuomola (commented):
In addition to my other comments: I don't think it's a good idea to regress the automatic generation of fineract.yaml and check it into git again. We should find a way to continue with the automatic generation as part of Gradle build.

fineract-client/src/main/templates/java/retrofit2/ApiClient.mustache
vidakovic: We could leave them out; the templates are already available as a dependency... but: I thought we want at least the license info included and this is the only way to do it. And in the current upstream client generation there are also a handful of templates tweaking things.

For me both approaches are OK; if we just use the dependencies then one thing less to maintain.

build.gradle
vidakovic: Yes, use BOMs. I went through that phase with my old projects already. Switching to BOMs makes life so much easier and ensures that mechanics (e. g. Spring - Spring Boot - Spring Cloud) are working well together. Nowadays a handful of BOMs cover 99.9% of all dependencies that you might ever use. And they are really up to date, even for the transitive dependencies. I used to have a weekly task that would list me everything that is new and then I would update that manually. Not worth it, BOMs are your friends.

build.gradle
vidakovic: This is just to define the plugin dependencies. We could also do something like this:

```
...
id 'org.openapi.generator' version '4.3.1' apply false
...
```

Then this is really just declaring a dependency (much like the dependencyManagement section) and the respective plugins would only be activated with (same example):

```
apply plugin: 'org.openapi.generator'
```

vidakovic: On that note: all those plugin dependencies declared like this in the buildscript section:

```
    dependencies {
        classpath 'com.bmuschko:gradle-cargo-plugin:2.7.1'
        classpath 'org.zeroturnaround:gradle-jrebel-plugin:1.1.10'
        classpath 'org.springframework.boot:spring-boot-gradle-plugin:2.3.4.RELEASE'
        classpath 'gradle.plugin.com.hierynomus.gradle.plugins:license-gradle-plugin:0.15.0'
        classpath 'org.apache.openjpa:openjpa:3.1.2' // when upgrading, also change OpenJPA version repeated in fineract-provider/build.gradle!
        classpath 'com.radcortez.gradle:openjpa-gradle-plugin:3.1.0'
        classpath 'org.nosphere.apache:creadur-rat-gradle:0.7.0'
        classpath "gradle.plugin.com.github.spotbugs.snom:spotbugs-gradle-plugin:4.5.1"
        classpath "gradle.plugin.com.github.andygoossens:gradle-modernizer-plugin:1.3.0"
        classpath "gradle.plugin.com.gorylenko.gradle-git-properties:gradle-git-properties:2.2.3"
        classpath "net.ltgt.gradle:gradle-errorprone-plugin:1.2.1"
        classpath "io.swagger.core.v3:swagger-gradle-plugin:2.1.5"
        classpath "gradle.plugin.org.hidetake:gradle-swagger-generator-plugin:2.18.2"
    }
```

... are considered deprecated Gradle style (will probably not be supported in Gradle 7).

build.gradle
vidakovic: Ok, understood. Then I have to move this most likely somewhere else. As is, it complains about spotlessApply not being available.

vidakovic: Concerning the automatic yaml file generation: the intention is _not_ to regress. Right now unfortunately Gradle is resisting all attempts to order the builds. But sooner or later I'll find a solution for that. So, we are on the same page here.

On that note: there's also a little bit of a chicken and egg problem left there concerning the distribution task. We need to build fineract-provider first which gives us the fineract.yaml file and then we can build fineract-client without the need to hold a hard copy. But as it is right now fineract-provider is also responsible to create the distribution which should include the client library... which would create in this current configuration a circular dependency. Solution: a separate module (fineract-distribution) should be responsible just for creating the distribution file (and signing it etc.). That way we can avoid the circular dependency. Later PR.

vidakovic: Found a solutions:

*   automatic spotlessApply works again as before
*   I found a way to express the dependency of fineract-client on fineract-provider; works like a charm; no hard copy of fineract.yaml in fineract-client needed anymore

build.gradle
vidakovic: I've added "apply false" in the plugin section; that way we are sure that those plugins are not activated globally (with potential side effects). FYI

fineract-provider/build.gradle
```diff
@@ -302,140 +125,6 @@ configurations {
...
-        "-Werror"
-    ]
-    // TODO FINERACT-959 (gradually) enable -Xlint:all (see "javac -help -X")
-
-    options.deprecation = true
-    options.errorprone {
-        disableWarningsInGeneratedCode = true
-        //Todo Remove necessary checks from this list
-        disable(
-                "UnusedVariable",
```
ptuomola, line 340: As part of this refactoring, we seem to have lost this ErrorProne configuration. At least I now get 20 pages of errorprone warnings about all these things that used to be disabled...

ptuomola (changes requested):
Looks good otherwise, except the error prone config... can you see if that could be fixed? Apart from that, I think this is ready to go!

fineract-provider/build.gradle
vidakovic: Thanks for catching that one... will fix that.

vidakovic: @ptuomola Found and fixed it... I was a bit too eager when I moved the Java compiler configuration to the root build.gradle. I verified locally, no more Errorprone warnings.

fineract-client/src/main/templates/java/retrofit2/ApiClient.mustache
vorburger: If it's just for the license headers, then yes I would remove this, to reduce ma maintenance. If we need to improve or fix things, then we should, ideally, contribute that upstream... Remove?

.travis.yml
```diff
@@ -72,6 +72,6 @@ script:
...
+  - ./gradlew --console=plain -PautomatedBuild=true licenseMain licenseTest licenseIntegrationTest check  &&  ./gradlew --console=plain -PautomatedBuild=true integrationTest --fail-fast  &&  sudo service mysql stop  &&  docker-compose build  &&  docker-compose up -d  &&  sleep 60s  && curl -f -k --retry 5 --retry-connrefused --connect-timeout 30 --retry-delay 30 https://localhost:8443/fineract-provider/actuator/health  &&  (( $(curl -f -k --retry 5 --retry-connrefused --connect-timeout 30 --retry-delay 30 https://localhost:8443/fineract-provider/actuator/info | wc --chars) > 100 ))
```
vorburger, line 75: How does Travis cover for the client now? Just curious.

build.gradle
```diff
@@ -20,6 +20,16 @@
 buildscript {
     ext {
         jacocoVersion = '0.8.5'
+        retrofitVersion = '2.9.0'
+        okhttpVersion = '4.7.2'
+        oltuVersion = '1.0.1'
```
vorburger, line 25: This seems curious, to me.. isn't this RUNTIME related, not BUILD time? Why is e.g. okhttp required to generate code? There's probably a good reason. But we didn't seem to have it before, so I thought I could ask.

vorburger (changes requested):
Let's remove the templates? (And 2 Qs.)

Also @Grandolf49 this could interest you? Feel free to chime in here, if any input.

.travis.yml
vidakovic: The modified templates only contain additional license headers (thought this was a requirement). If that doesn't matter we can leave out, easy. Just let me know what how you decide.

build.gradle
vidakovic: OKHttp doesn't generate anything, but we are still building the client JAR... so unless I'm missing something here seems to be normal to me...

Before it was not defined in Fineract's dependency management section, because the build was driven by a totally independent - single project - Maven pom.xml that had no ties to the fineract project setup (in terms of dependencies).

These dependencies are just a consequence of having fineract-client as a module.

gradlew.bat
```diff
@@ -1,3 +1,22 @@
+@REM
+@REM Licensed to the Apache Software Foundation (ASF) under one
+@REM or more contributor license agreements. See the NOTICE file
```
ptuomola: Hmm... are we changing the license that is on gradlew.bat? Surely we should retain the existing license that it is provided with

vidakovic: That one was not intended... have to adjust the license excludes... fixing.

vidakovic: Yup, as expected was the license plugin configuration that just included a bit too much (\*\*/\*.bat). Fixed.

vidakovic:
@ptuomola @vorburger So... how do you guys want to decide on the Mustache templates in fineract-client? Without? I think that's the last open question here before we can merge...

ptuomola:
> @ptuomola @vorburger So... how do you guys want to decide on the Mustache templates in fineract-client? Without? I think that's the last open question here before we can merge...

I agree with @vorburger - if the only change to the templates is the license, let's not check them in but include them as dependency instead.

vidakovic:
@ptuomola @vorburger Swagger templates are gone.

ptuomola (approved):
LGTM. @vorburger did your comments also get addressed? If you're also OK with this, then let's merge it..

Except just noticed there's a merge conflict. Should be easy to fix though...

Grandolf49:
This is great!

vorburger:
I'm re-reviewing it now.. give me a moment to pull it down and actually try it out as well. This is giving me some ideas.. watch for new JIRA! :sweat_smile: 

@vidakovic did you notice it has a conflict again, in `fineract-provider/build.gradle` - will you resolve? #1408 may have (not checked) caused the new conflict.. as it looks like we are almost there with this one now, I'll hold merging other PRs until we merged this :smiling_imp: sorry about that.

.travis.yml
vorburger: @vidakovic wait, what does this Q have to do with the (now removed) license headers? I may be missing something. What I was asking about here is why we're removing `./gradlew generateSwaggerCode` from Travis.. I've just pulled this PR and tried building it locally, and noticed that Gradle task doesn't seem to exist anymore now, so.. do we not need to run this on Travis anymore now, because... Swagger code generation now runs as part of the multi module build in `fineract-client`?

vidakovic:
@vorburger Let the new ideas come... looking forward :smile:

Checking now for the conflict... brb.

Update: ... and fixed.

config/fineractdev-cleanup.xml
```diff
@@ -1,4 +1,24 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!--
```
vorburger, line 2: Not sure why this is suddenly needed here, and what it has to do with FINERACT-1189, but doesn't really hurt much either, not a reason to request removing it. I do expect that likely will be a small PITA for anyone ever making changes to this file again - because it's exported from Eclipse, not manually edited, and when Eclipse rewrites it, it will most likely loose this ASL header again, so whoever makes such changes will have to manually put it back (if we're enforcing this now, else we'll just loose it again sooner or later).

fineract-client/build.gradle
```diff
@@ -0,0 +1,84 @@
...
+ */
+description = 'Fineract Client'
+
+ext['swaggerFile'] = "$rootDir/fineract-provider/build/classes/java/main/static/swagger-ui/fineract.yaml".toString()
+
+apply plugin: 'org.openapi.generator'
+
+apply from: 'dependencies.gradle'
+
+// TODO: @vidakovic we should publish this lib to Maven Central; do in separate PR
```
vorburger, line 27: FYI we have FINERACT-1102 about that - do you want to "take" it and assign it to yourself?

fineract-client/build.gradle
```diff
@@ -0,0 +1,84 @@
...
+    generateModelTests = false
+    generateApiTests = false
+    // trick to make sure fineract.yaml is generated first
+    dependsOn = [
+        ':fineract-provider:compileJava'
+    ]
+    finalizedBy = [licenseFormat]
+}
+
+// TODO: @vidakovic we could provide even more client libs in different languages (Go, Ruby, Swift etc.)
```
vorburger, line 66: IMHO much more interesting and directly useful than _Go, Ruby, Swift_ would be a TS (better; or JS) one.. UIs could use!

fineract-provider/build.gradle
```diff
@@ -18,15 +18,11 @@
  */
 description = 'Fineract Provider'
 
-apply plugin: "org.nosphere.apache.rat"
```
vorburger, line 21: FTR: we're not actually remove RAT (or `hierynomus.license`) here, as they were already copied to `build.gradle` in a previous PR, so this is just clean up. (@vidakovic perhaps in future build refactorings we can have the move more clearly in a single PR.)

fineract-client/build.gradle
vorburger: see FINERACT-1210

vorburger:
> @vorburger Let the new ideas come... looking forward smile

FINERACT-1209 is the big one, that could be so cool; FINERACT-1210 a more medium/longer later idea; FINERACT-1100 minor.

fineract-client/build.gradle
vidakovic: On it.

PR #1416, opened 2020-10-17
FINERACT-1190: Add fineract-api module

vidakovic:
## Description

This PR adds a new module to expose Fineract extension points:

1. I've added an initial ReportingProcessService interface that is modelled after "fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/api/RunreportsApiResource.java"; I removed all references to JAX-RS packages and classes, because I think we should not make any implementation detail assumptions and should try as much as possible to just rely on JDK runtime APIs
2. above interface is not considered final; please chime in to discuss what we need to expose here
3. as a consequence of Nr. 2 I have not integrated this in fineract-provider yet; probably this will be a breaking change for everything that is relying on Pentaho
4. as soon as we settle on a reporting API we should create maybe a Github repo "mifos/fineract-contributions" where we could add a module that ports the Pentaho reporting stuff and makes it available as a JAR that Fineract users can include at runtime
5. now that we have multiple Java modules I've moved all Gradle plugin configurations that are useful "up" to the root build.gradle file; that way we don't have to configure this multiple times and fineract-provider gets a lot leaner; these plugins are: Checkstyle, Jacoco, Errorprone, Spotbugs, Modernizer

Note: maybe we could defer any decision how that ReportingProcessService interface should look like, leave the stuff here as is and do the real implementation with a separate PR as part of FINERACT-1127. Just to avoid overloading this PR.

## Checklist


vorburger (changes requested):
> as soon as we settle on a reporting API we should create maybe a Github repo "mifos/fineract-contributions" where we could add a module that ports the Pentaho reporting stuff and makes it available as a JAR that Fineract users can include at runtime

That's what https://github.com/vorburger/fineract-pentaho is, have you seen that? (The way I see us going forward is not to have a "mifos/fineract-contributions" monorepo, but to encourage "decentralized" plugins - and simply have a MD doc with links.)

> above interface is not considered final; please chime in to discuss what we need to expose here

Understood. So what I suggest is to go about this a little differently - instead of starting with a `fineract-api` with an "example" interface that isn't usable / used, let's actually make it a real actual used think already. So NOT a "standalone" `fineract-api`, but from the start, make `fineract-provider` depend on `fineract-api` straight way... so I would go about this "the other way around" - a pair of refactoring PRs that goes together in `apache/fineract` and `vorburger/fineract-pentaho`. Do you see what I mean / makes sense?

> now that we have multiple Java modules I've moved all Gradle plugin configurations that are useful "up" to the root build.gradle file; that way we don't have to configure this multiple times and fineract-provider gets a lot leaner; these plugins are: Checkstyle, Jacoco, Errorprone, Spotbugs, Modernizer

If you like to get this in faster, you could already have a PR with JUST that, as a first step, for now?

vidakovic:
@vorburger Alright, was not aware of your repo; will have a look at it and adapt. As for the decentralized plugins: understood, good idea.

I'll pick up the interface you used in your implementation and make it as close as possible (I would still suggest avoiding any external dependencies if possible; like the JAX-RS stuff in the original interface in fineract-provider).

When we've all signed of on the reporting API I'll send a PR in your repo for Pentaho.

vidakovic:
@vorburger I've made this more concrete now. As per your suggestion I've added fineract-api as a dependency to fineract-provider. I've also replaced the existing reporting interface with the new one in:

*   org.apache.fineract.infrastructure.report.provider.ReportingProcessServiceProvider
*   org.apache.fineract.infrastructure.reportmailingjob.service.ReportMailingJobWritePlatformServiceImpl
*   org.apache.fineract.infrastructure.dataqueries.api.RunreportsApiResource

Note that there is a new utility class "org.apache.fineract.infrastructure.report.util.ParameterUtil" with static function "toSingleValueMap" that I use to translate the JAX-RS multi-value maps to single value/normal maps. I've checked the details of the Pentaho implementation, something similar is happening there anyway (see function "getReportParams" in class "org.apache.fineract.infrastructure.report.service.PentahoReportingProcessServiceImpl" in vorburger/fineract-pentaho). Just to say: we are not losing any information here. And I still think it's important to keep the API interfaces free from any technology decisions like JAX-RS.

I've also allowed myself to reduce the code in org.apache.fineract.infrastructure.report.provider.ReportingProcessServiceProvider. We can use a "feature" of Spring dependency injection to make this a lot easier: see the new constructor where we inject a list of all available implementations of org.apache.fineract.api.ReportingProcessService currently available in the application context. We add those references then to a hash map and can use this then immediately without any further initialization or a direct reference to application context. This also removes any need for the "ReportService" annotation, because we have already enough information to retrieve the service instances.

vorburger:
I'll review this together with and once I get to https://github.com/vorburger/fineract-pentaho/pull/5 (which may take a few days).

vidakovic:
All good. I'll see your comments if you need anything else here. I might once in a while just rebase with upstream if merge conflicts pop up, but don't plan any functional changes until you got some time to check it out.

vorburger:
> > now that we have multiple Java modules I've moved all Gradle plugin configurations that are useful "up" to the root build.gradle file; that way we don't have to configure this multiple times and fineract-provider gets a lot leaner; these plugins are: Checkstyle, Jacoco, Errorprone, Spotbugs, Modernizer
> 
> If you like to get this in faster, you could already have a PR with JUST that, as a first step, for now?

I'll try to raise a PR do do this for you to help get this in sooner rather than later (because I'm also interested in that part).

fineract-api/src/main/java/org/apache/fineract/api/ReportingProcessService.java
```diff
@@ -0,0 +1,55 @@
...
+    String PARAM_USER_HIERARCHIE = "userhierarchy";
+    String PARAM_USERNAME = "username";
+    String PARAM_PASSWORD = "password";
+    String PARAM_USER_ID = "userid";
+    String PARAM_DATABASE_ULR = "databaseUrl";
+    String PARAMETER_PREFIX = "R_";
+
+    void process(String name, Map<String, String> parameter, OutputStream os);
+
+    String getType();
```
vorburger, line 54: Why add this instead of just using the `org.apache.fineract.infrastructure.report.annotation.ReportService` annotation with the `type()` as it's currently?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/report/provider/ReportingProcessServiceProvider.java
```diff
@@ -20,66 +20,39 @@
...
-import org.springframework.context.ApplicationContext;
-import org.springframework.context.ApplicationContextAware;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Scope;
 import org.springframework.stereotype.Component;
 
 @Component
 @Scope("singleton")
-public class ReportingProcessServiceProvider implements ApplicationContextAware {
+public class ReportingProcessServiceProvider {
```
vorburger, line 34: I have integrated the changes you proposed for `ReportingProcessServiceProvider` into my #1368 already, so that in this PR you don't have to change this class anymore, hope you don't mind? (It's not a verbatim copy/paste but inspired by your changes here.)

fineract-api/src/main/java/org/apache/fineract/api/ReportingProcessService.java
```diff
@@ -0,0 +1,55 @@
...
+    String PARAM_LOCALE = "locale";
+    String PARAM_TENANT_URL = "tenantUrl";
+    String PARAM_USER_HIERARCHIE = "userhierarchy";
+    String PARAM_USERNAME = "username";
+    String PARAM_PASSWORD = "password";
+    String PARAM_USER_ID = "userid";
+    String PARAM_DATABASE_ULR = "databaseUrl";
+    String PARAMETER_PREFIX = "R_";
+
+    void process(String name, Map<String, String> parameter, OutputStream os);
```
vorburger, line 52: As mentioned in FINERACT-1264, I would suggest that we use a signature like this instead (we should definitely get rid of the `javax.ws.rs.core.Response` which we currently use, that's too much "web specific", and doesn't belong here):

```suggestion
    ByteSource process(String name, MediaType format, Map<String, String> parameters);
```

fineract-api/src/main/java/org/apache/fineract/api/ReportingProcessService.java
```diff
@@ -0,0 +1,55 @@
...
+        ReportType(String contentType) {
+            this.contentType = contentType;
+        }
+
+        private final String contentType;
+
+        public String getContentType() {
+            return contentType;
+        }
+    }
```
vorburger, line 40: I'm not sure we should make this part of the Reporting plugin API. Isn't this better up to each Plugin? If anything, we could simply have a `String contentType` to indicate what the user of this API wants to receive in response? We could also use a strong type than `String`, there are about 27 types we can choose from, e.g. Spring's `MimeType`, note their `MimeTypeUtils`, or Spring's `MediaType`, or JAX RS' `MediaType` (but I think this API should NOT be tied to JAX RS) or Guava's `MediaType` (I personally actually prefer that last one, Guava).

fineract-api/src/main/java/org/apache/fineract/api/ReportingProcessService.java
```diff
@@ -0,0 +1,55 @@
...
+
+    String PARAM_OUTPUT_TYPE = "output-type";
+    String PARAM_LOCALE = "locale";
+    String PARAM_TENANT_URL = "tenantUrl";
+    String PARAM_USER_HIERARCHIE = "userhierarchy";
+    String PARAM_USERNAME = "username";
+    String PARAM_PASSWORD = "password";
+    String PARAM_USER_ID = "userid";
+    String PARAM_DATABASE_ULR = "databaseUrl";
+    String PARAMETER_PREFIX = "R_";
```
vorburger, line 50: I'm not sure we should make all these part of the Reporting API. Definitely not as String constants. Perhaps as strongly typed arguments? Better in a `ReportOptions` kind of wrapper? We should carve out what (1) the first external https://github.com/vorburger/fineract-pentaho/blob/develop/src/main/java/org/apache/fineract/infrastructure/report/service/PentahoReportingProcessServiceImpl.java, (2) the internal `DatatableReportingProcessService` (new in #1368) and (3) other ones you may have in mind really require... some thoughts:

* The output-type IMHO should be replaced by the `MediaType` argument 
* The Locale should better be in some sort of new `UserContext` kind of object, what to you think?
* userhierarchy could be part of such a `UserContext` perhaps? Also a `userID` I guess.
* What is PARAMETER_PREFIX = "R_" for now? Won't we pass those as part of the `Map<String, String> parameters` arg?
* About the TenantDB connection details: We should find some appropriate existing JDBC class which we can "pass through" for all these. What's type in JDBC best encapsulates such connection details? I wish we could just pass an actual `Connection`, but some of these reporting tools will probably want to connect to the DB themselves? (This is actually not great at all from a security point of view, I mean the reporting plugin can just query, and write, whatever it wants - but that's not the point of this PR.)

fineract-api/src/main/java/org/apache/fineract/api/exception/AbstractApiException.java
```diff
@@ -0,0 +1,86 @@
...
+ */
+package org.apache.fineract.api.exception;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Exception with internationalization support (it's message can be translated).
+ */
+public abstract class AbstractApiException extends RuntimeException {
```
vorburger, line 27: Copying (forking) instead of just moving existing Exception types is going to cause issues. I don't think this is a good idea. I would EITHER just move them, keeping the same package, or (better?) not have them in the API, for now. The `PentahoReportingProcessServiceImpl` can just throw some more generic exception?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/reportmailingjob/service/ReportMailingJobWritePlatformServiceImpl.java
```diff
@@ -447,7 +450,18 @@ private StringBuilder generateReportOutputStream(final ReportMailingJob reportMa
                     .findReportingProcessService(reportType);
 
             if (reportingProcessService != null) {
-                final Response processReport = reportingProcessService.processRequest(reportName, reportParams);
+                final Map<String, String> params = toSingleValueMap(reportParams);
```
vorburger, line 453: This is copy/pasted from above... :crying_cat_face: let's find a better solution together?

vorburger (changes requested):
@vidakovic some thoughts on this one, hope helpful.

PS: I'll try to make progress on FINERACT-1177 soon.

PR #1417, opened 2020-10-17
FINERACT-1102: Swagger Codegen Client JAR should be available on a Ma…

vidakovic:
FINERACT-1102

## Description

The following things need to be solved before the publishing can work:

1. It's probably best to create an account at https://bintray.com. They make publishing on Maven Central so much easier.
2. We need to upload a GPG signing key to Bintray once we have an account.
3. We will also need a user token from Sonatype to be able to publish to Maven Central. Details see here: https://stackoverflow.com/questions/38294001/how-to-sync-from-bintray-to-maven-central-without-knowing-maven
4. We have to decide if we want to list any developers in the generated pom.xml that gets published. There is a TODO in the root build.gradle
5. I've created the Maven publish configuration already in a way so that it can be applied to multiple modules without major edits. Note: next one to publish will be fineract-api.
6. There's a sample configuration to show how publishing on Github Packages would work. This just FYI and will be removed once we are done here. I think Maven Central is the best choice.

## Checklist

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -29,7 +29,11 @@ buildscript {
                 'fineract-client'
             ].contains(it.name)
         }
-
+        fineractPublishProjects = subprojects.findAll{
+            [
+                'fineract-client'
+            ].contains(it.name)
+        }
     }
     repositories {
         jcenter()
@@ -452,5 +456,46 @@ configure(project.fineractJavaProjects) {
         } else {
             dependsOn(spotlessApply)
         }
+    }
+}
+
+configure(project.fineractPublishProjects) {
+    apply plugin: 'maven'
+    apply plugin: 'maven-publish'
+
+    publishing {
+        publications {
+            mavenJava(MavenPublication) {
+                from components.java
+
+                versionMapping {
+                    usage('java-api') {
+                        fromResolutionOf('runtimeClasspath')
+                    }
+                    usage('java-runtime') {
+                        fromResolutionResult()
+                    }
+                }
+
+                pom {
+                    name = 'Fineract'
+                    description = 'A secure, multi-tenanted microfinance platform'
+                    url = 'https://fineract.apache.org'
+                    licenses {
+                        license {
+                            name = 'The Apache License, Version 2.0'
+                            url = 'http://www.apache.org/licenses/LICENSE-2.0.html'
+                        }
+                    }
+                    scm {
+                        connection = 'scm:git:https://github.com/apache/fineract.git'
+                        developerConnection = 'scm:git:git://github.com:apache/fineract.git'
+                        url = 'https://fineract.apache.org/'
+                    }
+                }
+            }
+        }
+
+        // TODO FINERACT-1102: Actually use this to deploy to ASF (Apache Software Foundation) Nexus Maven Repository
     }
 }


build.gradle
```diff
@@ -454,3 +459,89 @@ configure(project.fineractJavaProjects) {
...
+                    name = 'Fineract'
+                    description = 'A secure, multi-tenanted microfinance platform'
+                    url = 'https://fineract.apache.org'
+                    licenses {
+                        license {
+                            name = 'The Apache License, Version 2.0'
+                            url = 'http://www.apache.org/licenses/LICENSE-2.0.html'
+                        }
+                    }
+                    // TODO: should we only list the Fineract contributors here? Or everyone that sent a PR (ca. 150; a bit much)? Or should we just omit this?
```
vorburger: Let's simply omit this entire POM section entirely - remove?

vorburger (dismissed):
Interesting! The real Q here is from which CI system we want to push to what repository. As is, this would presumably attempt to push every build (PRs **and** `develop`, so No Go, as is) to BinTray - yet another external service someone needs to manage credential for and all. I've lefft a comment in FINERACT-1102 for a possible alternative.

build.gradle
vidakovic: As far as I'm aware you have to provide a POM... otherwise it won't work on Maven Central... unless you meant something different?

We could reduce the information though and just have name, license, Git repo URL.

vidakovic: As far as I'm aware you have to provide a POM... otherwise it won't work on Maven Central... unless you meant something different?

We could reduce the information though and just have name, license, Git repo URL.

vidakovic: As far as I'm aware you have to provide a POM... otherwise it won't work on Maven Central... unless you meant something different?

We could reduce the information though and just have name, license, Git repo URL.

vorburger: Yeah sorry I was responding to your TODO, and meant specifically and only not listing developers at all, but keeping the rest of the POM. BTW how does this actually work - will the POM include the required `<dependencies>`? It, ideally, should, right?

vidakovic: Yeah, the developer section didn't make too much sense to me neither. People can see this - if they are interested - here on Github a lot better.

As for the dependencies: yes, this is done automatically by the "maven" Gradle plugin. You get proper pom.xml with dependencies per module.

vidakovic: ... and it's gone

vorburger:
@vidakovic as in the FINERACT-1102 discussion we are now converging towards using ASF's Nexus, I would LGTM and merge this only if you remove the GitHub Packages and BinTray lines from the `build.gradle`. I fear it could create more confusion _(Noob Contributor: "is this project deploying to GitHub Packages and BinTray? But how?")_ than add value to merge this as-is. And inline `TODO FINERACT-1102: Actually use this to deploy to ASF (Apache Software Foundation) Nexus Maven Repository` also couldn't hurt. But if this PR is "just" WIP from you that you don't want to see merged just yet, please mark it as Draft for now.

vidakovic:
@vorburger I've removed any references to Github Packages and Bintray. Todo as you suggested placed instead. I'd say let's merge (it won't hurt anyone; this task is similar to release, just for a very small subset of Fineract devs will use it) and I'll circle back to this when I've published an artifact successfully. BTW: was able to login on Apache's Nexus admin app; as soon as I have something working I'll send a new PR with updates to the README.

build.gradle
```diff
@@ -454,3 +458,45 @@ configure(project.fineractJavaProjects) {
         }
     }
 }
+
+configure(project.fineractPublishProjects) {
+    apply plugin: 'maven'
+    apply plugin: 'maven-publish'
+    apply plugin: 'com.jfrog.bintray'
```
vorburger: You probably want to remove this? FYI Travis build failed due to `Plugin with id 'com.jfrog.bintray' not found.`

vidakovic: Arrgh! ;-)

vidakovic: The maven stuff can be left there (then we don't have to remove the whole block). And as said, it won't hurt anyway. If someone does accidentally "./gradlew publish" then it will just do nothing.

vorburger (approved):
LGTM I'll merge this when the build passed.

PR #1424, opened 2020-10-17
Clean up ImageData (FINERACT-1201)

vorburger:
FINERACT-1201


vorburger:
Draft until I double check if the changes here are covered by an integration test.

vorburger:
Maybe it's not worth to have this small initial change as a separate PR.. probably simpler to "fold" this into the overall #1425.

PR #1425, opened 2020-10-18
Introduce ImageResizer to simplify ImageData (FINERACT-1201)

vorburger:
FINERACT-1201

PR #1426, opened 2020-10-18
remove weird ImageData.available (FINERACT-1201)

vorburger:
I need to clean this up (remove it) so that I can proceed with significant refactorings I'm doing in the context of FINERACT-1201.

I don't understand what this really does, or how to test it. It seems to have been original introduced for FINERACT-288 in PR #259, but the description there doesn't clearly enough explain the problem.

Much related code is about to be substantially refactored, and whatever problem this originally worked around may no longer be present when the ongoing refactoring is fully completed.  If it still is, then there will likely be a better way to address it.

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/documentmanagement/data/ImageData.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/documentmanagement/data/ImageData.java
@@ -25,7 +25,6 @@ import java.awt.image.BufferedImage;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.FileInputStream;
-import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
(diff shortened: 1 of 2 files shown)


vorburger:
@vidakovic would you be willing to review this PR from me, and merge it if you LGTM?

vorburger:
FYI I've locally pulled and combined this PR's main code change with the improved test code from #1427, and seen that the expanded `org.apache.fineract.integrationtests.StaffImageApiTest.deleteStaffImage()` still works even with this change.

vidakovic:
@vorburger yup, will be soon back to review and merge

ptuomola:
@vorburger I tried to understand what this code was originally introduced for. As far as I can see, the code in retrieveImage() used to do result.getContent() != null to see if it actually got an image. The problem with that is that calling getContent() reads the full input stream to EoS - so the image content is then not available later when the code tries to actually get it with getContent() when building the response to client. So they coded available() as sort of "peek" to determine whether there's data to be returned in a "non-invasive" way.

So as you say the risk that this change runs is that there's some way for the image not to be available that the original code was build to cater for. It's clearly not the scenario where the file is missing altogether - but perhaps there's some case where the file is deemed present but no content is returned? But if that's the case then would be interesting to understand when it arises...

In summary I think this is OK to merge in my view as long as we keep our eyes open for some failure scenario with missing content / image data...

ptuomola (approved):
As per my comment - looks good and OK to merge, but let's watch out for some weird failure scenario that this may have been introduced to address...

vorburger:
> as we keep our eyes open for some failure scenario with missing content / image data...

FTR: This change will get "compensated" by the (currently) 4th commit on #1425 which _replaces use of InputStream by ByteSource in ImageData._

PR #1431, opened 2020-10-18
Create CONTRIBUTING.md

vorburger:
as per https://docs.github.com/en/free-pro-team@latest/github/building-a-strong-community/setting-guidelines-for-repository-contributors from https://github.com/apache/fineract/community.

PR #1428, opened 2020-10-18
Add FineractClientTest for Fineract SDK REST Client (FINERACT-1216)

vorburger:
and introduce handy `Calls` utilitiy

see also discussion in FINERACT-1216

PR #1432, opened 2020-10-19
FINERACT-1216: Add FineractClientTest for Fineract SDK REST Client (fix)

vidakovic:
@ptuomola @vorburger Could one of you quickly have a look at this and merge please? There are 2 small issues with Checkstyle that this PR fixes.
Builds are currently broken because of this one here.

### DIFF
--- a/fineract-client/src/main/java/org/apache/fineract/client/util/Calls.java
+++ b/fineract-client/src/main/java/org/apache/fineract/client/util/Calls.java
@@ -27,7 +27,7 @@ import retrofit2.Response;
  *
  * @author Michael Vorburger.ch
  */
-public class Calls {
+public final class Calls {
 
     private Calls() {}
 
--- a/fineract-client/src/main/java/org/apache/fineract/client/util/FineractClient.java
+++ b/fineract-client/src/main/java/org/apache/fineract/client/util/FineractClient.java
@@ -26,7 +26,7 @@ import org.apache.fineract.client.auth.ApiKeyAuth;
  *
  * @author Michael Vorburger.ch
  */
-public class FineractClient {
+public final class FineractClient {
 
     private final ApiClient apiClient;
 
@@ -42,7 +42,7 @@ public class FineractClient {
         return apiClient.createService(serviceClass);
     }
 
-    public static class FineractClientBuilder {
+    public static final class FineractClientBuilder {
 
         private String baseURL;
         private String tenant;


vorburger (changes requested):
Wait a second, then why was #1428 green?! That's super weird, no? (Or was it not?)

https://travis-ci.org/github/apache/fineract/branches says we're currently green on develop

Do you have a link to where the build broken how, so that we can understand what is going on here?

fineract-client/src/main/java/org/apache/fineract/client/util/Calls.java
```diff
@@ -27,7 +27,7 @@
  *
  * @author Michael Vorburger.ch
  */
-public class Calls {
+public final class Calls {
```
vorburger, line 30: This makes sense, because it's a utility class with a private constructor... (and I vaguely remember that we have a Checkstyle or ErrorProne or SpotBugs check to enforce this - just want to find out why builds don't show up as red - or where they do)

fineract-client/src/main/java/org/apache/fineract/client/util/FineractClient.java
```diff
@@ -26,7 +26,7 @@
  *
  * @author Michael Vorburger.ch
  */
-public class FineractClient {
+public final class FineractClient {
```
vorburger, line 29: .. but this shouldn't be required? Especially since in #1433 I'm making `createService()` a protected method, so this class should remain _open_ (extends-ible).

fineract-client/src/main/java/org/apache/fineract/client/util/FineractClient.java
```diff
@@ -42,7 +42,7 @@ public static FineractClientBuilder builder() {
         return apiClient.createService(serviceClass);
     }
 
-    public static class FineractClientBuilder {
+    public static final class FineractClientBuilder {
```
vorburger, line 45: as above (I think); best to see failing build and check if all 3 or only 1st (I think) need to be fixed?

vorburger (changes requested):

vorburger:
Did you locally see a failure (running what?) for a task which we are missing on Travis?

I'm curious if e.g. #1431 or #1433 will fail to build on Travis...

vorburger:
> I'm curious if e.g. #1431 or #1433 will fail to build on Travis...

They passed, so what does this fix? :smiling_imp: But I have a suspicion that we may have a larger problem.. we don't have the same quality controls in `fineract-client` that we have in `fineract-provider` - do we?

vidakovic:
Already taken care of with #1447

PR #1441, opened 2020-10-19
Reduce fineract-client dependencies (FINERACT-1216)

vorburger:
All the dependencies removed here are not used required (proven by test still works).

see FINERACT-1216

### DIFF
--- a/fineract-client/dependencies.gradle
+++ b/fineract-client/dependencies.gradle
@@ -19,28 +19,22 @@
 dependencies {
     // project(':fineract-provider')
     implementation(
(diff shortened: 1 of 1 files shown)


ptuomola:
This what I was hoping for in my earlier comment - i.e. that most of the dependencies would not be needed... so looks good to me!

vorburger:
> This what I was hoping for in my earlier comment - i.e. that most of the dependencies would not be needed... so looks good to me!

Thanks! @vidakovic OK for you to merge?

vidakovic (approved):
LGTM

PR #1444, opened 2020-10-19
Introduce first "pure Spring" integration test, without Cargo (FINERACT-764)

vorburger:
The new SpringBootLoginTest is an illustration of what we could do more of.

NOTE: Use only https://:8443 instead of http://:7070 (redirects don't work)

see FINERACT-764


vorburger:
@ptuomola @vidakovic how do you like this? Where could & should we take this to next?

vidakovic (approved):
Looks good... I think this here should be also running faster than the Testcontainers approach I suggested on Jira

fineract-provider/src/test/java/org/apache/fineract/common/RestAssuredFixture.java
```diff
@@ -34,14 +34,12 @@
...
-    private ResponseSpecification responseSpec;
-    private RequestSpecification requestSpec;
+    private final ResponseSpecification responseSpec;
+    private final RequestSpecification requestSpec;
 
-    public RestAssuredFixture(int http) {
-
-        this.httpPort = http;
+    public RestAssuredFixture() {
+        this.httpPort = 844;
```
awasum, line 41: @vorburger Just curious and catching up to all the brilliant development done the past few weeks. Is this suppose to be port: 8443? or something new which I dont understand?

vorburger: @awasum well spotted, yes this looks wrong. It's surprising that it still worked - I think this line can probably be removed. I could do that, but I think there is a larger question of what direction we want to take re ITs - I'll comment on the main comment instead of inline here about that.

vorburger:
@ptuomola @vidakovic I'm undecided / unclear where we should take this to.. thoughts? I guess we could... EITHER (probably best AFTER #1465 will have gone in):

a) move towards entirely ditching the Cargo plugin and not start the WAR but just use this new style to fire up Spring "in-process" for (all) ITs

b) forget about this approach and abandon this PR and just close FINERACT-764 (I don't actually really mind either)

c) have some mixed form - keep testing the WAR with the Cargo plugin for some basic tests, but move other ITs to this style.

vorburger:
BTW #1422 has some related thoughts re. our testing strategy, I'll comment over there as well.

vidakovic:
> @ptuomola @vidakovic I'm undecided / unclear where we should take this to.. thoughts? I guess we could... EITHER (probably best AFTER #1465 will have gone in):
> 
> a) move towards entirely ditching the Cargo plugin and not start the WAR but just use this new style to fire up Spring "in-process" for (all) ITs

a) I'd prefer to go that way. The Cargo plugin requires the WAR artifact and is in the end an external process... just to say: a lot of moving parts to take care of. During a normal build it maybe doesn't make much of a difference for devs, but if you execute e. g. "./gradlew :integration-tests:test" then Cargo doesn't shutdown for some reason and leaves a zombie process (just found that out yesterday). And I have a hunch that Spring tests would run faster.

> b) forget about this approach and abandon this PR and just close [FINERACT-764](https://issues.apache.org/jira/browse/FINERACT-764) (I don't actually really mind either)
> 
> c) have some mixed form - keep testing the WAR with the Cargo plugin for some basic tests, but move other ITs to this style.

c) Realistically, we will probably take a moment or two to refactor all integration tests. So for a while we would be probably running in a "hybrid" mode. Maybe we could do this migration together with moving towards Fineract Client usage.

xurror:
I support ditching cargo, we could save us a lot of time running tests.

vorburger:
It makes more sense to pick this one up only AFTER #1465 is merged.

xurror:
Also, I noticed `./gradlew integrationTest` is no longer a thing, I don't know since when that task was removed, I just think you might want to update the readme on how to run tests now.

xurror:
Also, I noticed `./gradlew integrationTest` is no longer a thing and the readme was not updated to that regards, I don't know since when that task was removed, I just think you might want to update the readme on how to run tests now.

vorburger:
> Also, I noticed `./gradlew integrationTest` is no longer a thing and the readme was not updated to that regards, I don't know since when that task was removed, I just think you might want to update the readme on how to run tests now.

That's since #1465; @vidakovic do you want to fix up the README? @xurror you could just raise a PR for it yourself as well? ;-)

PR #1445, opened 2020-10-19
Add initial Fineract SDK Java API Client documentation (FINERACT-1216)

vorburger:
@vidakovic @ptuomola @Grandolf49 how about this, as initial end-user doc?

This will conclude FINERACT-1216, for now (from my side).


vidakovic (approved):
I think this is easy to understand and good enough for now. Let's see what questions (if any) first users of the SDK will have and improve based on those.

vidakovic:
@vorburger simple Spotless error?

PR #1447, opened 2020-10-19
Apply Checkstyle, Errorprone etc. for all Java modules (FINERACT-1171)

vorburger:
This is a subset (extract) from #1416

Attribution to Aleks is preserved!

for FINERACT-1171

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,7 +208,7 @@ Messaging configuration is disabled by default. If you want to enable it and reg
 Checkstyle and Spotless
 ============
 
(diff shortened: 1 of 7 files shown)


vorburger:
@vidakovic as per comment in #1416, how about we already do (just) this, and hold off with the `fineract-api` a little? Would you LGTM this? Note how while this PR appears to be from me, which doesn't really matter, the commit itself is still attributed (authored) by you!

@ptuomola do you want to review and LGTM this?

vorburger:
FTR: I've stared at this a bit, and it's just moves (I've actually verified we don't "loose" anything) - so LGTM from me on this!

vidakovic (approved):
LGTM

PR #1446, opened 2020-10-19
Limit fineract-client to Java 8 not 11 (FINERACT-1214)

vorburger:
see discussion in FINERACT-1214

PR #1448, opened 2020-10-20
Fix broken build (var & Java 8/11 from FINERACT-1214)

vidakovic:
Removed usage of "var" in FineractClient which is incompatible with the JDK8 build.

FINERACT-1219

### DIFF
--- a/fineract-client/src/main/java/org/apache/fineract/client/util/FineractClient.java
+++ b/fineract-client/src/main/java/org/apache/fineract/client/util/FineractClient.java
@@ -428,7 +428,7 @@ public final class FineractClient {
                         }
                     };
 
-                    var sslContext = SSLContext.getInstance("SSL");
+                    SSLContext sslContext = SSLContext.getInstance("SSL");
                     sslContext.init(null, new TrustManager[] { insecureX509TrustManager }, new SecureRandom());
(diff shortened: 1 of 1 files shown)


vorburger (approved):
LGTM. Feel free to self merge this trivial change when build passes (based on this LGTM).

PS: How come the build didn't fail for me?! How can we prevent this in the future?

vidakovic:
Not sure if we can define this kind of thing in Checkstyle? I should have seen it yesterday, but my brain skipped the fact that we have that new JDK8 compiler flag set in fineract-client. Maybe we set that flag in a different PR? Don't remember.

Will merge after the build.

vorburger:
> Not sure if we can define this kind of thing in Checkstyle? I should have seen it yesterday, but my brain skipped the fact that we have that new JDK8 compiler flag set in fineract-client. Maybe we set that flag in a different PR? Don't remember.

Checkstyle can't catch this, but I would have expected #1446 to fail.. `sourceCompatibility = JavaVersion.VERSION_1_8` should not compile `var`? Something doesn't seem quite right if we didn't automatically catch this. Anyway.

vorburger:
> Something doesn't seem quite right if we didn't automatically catch this. Anyway.

Oh! FINERACT-1226... :imp:

PR #1449, opened 2020-10-20
Fix possible NPE when uploading Documents & Images (FINERACT-1036)

vorburger:
FINERACT-1036

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exceptionmapper/ValidationExceptionExceptionMapper.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exceptionmapper/ValidationExceptionExceptionMapper.java
@@ -1,0 +1,48 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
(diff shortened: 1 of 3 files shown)


vorburger:
@ptuomola (or @vidakovic) I'm sure this is OK for you? Asking more pro-forma; I'll self merge this one in 24h if you don't get to it.

vidakovic (approved):
LGTM

ptuomola:
Hi @vorburger - is there a reason why you didn't use the "usual" DataValidatorBuilder/BaseDataValidator pattern for this check? I'm sure this works - but that would have been more consistent with the rest..

vorburger:
> is there a reason why you didn't use the "usual" DataValidatorBuilder/BaseDataValidator pattern for this check?

Good point. I can't find `BaseDataValidator` (:question:), but the `DataValidatorBuilder` seems to be more used in `.services.` and `.command.`.. I can't find a single instance of it in an `.api.` - but I guess there is no reason why not to? I'll leave this PR as is, and will raise a separate alternative one, for comparison. Who knows, I may even use this as the opportunity to write DOC - imagine that, an open source project with real technical framework documentation!! :smile:

vorburger:
#1456

PR #1450, opened 2020-10-20
Introduce CallFailedRuntimeException in client

vorburger:


vorburger:
@ptuomola @vidakovic would this be OK for you? As I'm using the `FineractClient` to write tests, I'm already tired of adding `throws IOException` to every `@Test`.

vorburger:
FYI the build on this actually passed and was fully green, and _continuous-integration/travis-ci/pr_ still is, but _continuous-integration/travis-ci/push_ restarted because of an experiment I did (I've also pushed the same branch to the `apache/fineract` instead of only `vorburger/fineract`, to try out something).. please ignore, and consider this green.

vidakovic (approved):
Looks good... wrapping as RuntimeException makes totally sense.

PR #1451, opened 2020-10-20
Do not expose getApiClient() in FineractClient (FINERACT-1220)

vorburger:
Because FineractClient is about to not use Swagger's ApiClient anymore.


vorburger:
@vidakovic OK for you? This is more minor, so I'll self merge this ASAP if you don't have time (no stress).

There will be a follow-up PR which will include this commit, I just raised it separately first for both PRs to be easier to review.

vidakovic (approved):
LGTM

vidakovic:
FYI: I see no reason why the build should fail where it did on Travis... I think this might have been just a resource problem and re-building would just succeed

vorburger:
The rebuild failed due to FINERACT-1167 (but I saw it green before that). I'll go ahead and self merge this, so that an upcoming new PR which builds on this is easier to review.

PR #1456, opened 2020-10-21
Fix possible NPE when uploading Documents & Images (FINERACT-1036)

vorburger:
FINERACT-1036

### DIFF
--- a/docs/developers/validation.md
+++ b/docs/developers/validation.md
@@ -1,0 +1,32 @@
+# Validation
+
+## Programmatic
+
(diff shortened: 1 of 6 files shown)


ptuomola:
@vorburger wow that ended up being a lot bigger change than the 4 lines of code I expected! 

What was your thinking around changing the variables from specific types to var? I suppose the same change could be applied to pretty much any line of code in Fineract... :-)

vorburger:
> What was your thinking around changing the variables from specific types to var?
> I suppose the same change could be applied to pretty much any line of code in Fineract... :-)

Fair point! Now reverted from this PR. Let's discuss further separately over in FINERACT-1237?

ptuomola (approved):
LGTM

PR #1458, opened 2020-10-21
Disable frequently failing SchedulerJobsTest.testFlippingSchedulerStatus (FINERACT-1167)

vorburger:
see FINERACT-1167

### DIFF
--- a/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/SchedulerJobsTest.java
+++ b/fineract-provider/src/integrationTest/java/org/apache/fineract/integrationtests/SchedulerJobsTest.java
@@ -33,6 +33,7 @@ import org.apache.fineract.integrationtests.common.SchedulerJobHelper;
 import org.apache.fineract.integrationtests.common.Utils;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
 public class SchedulerJobsTest {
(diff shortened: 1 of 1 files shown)


ptuomola:
Hi @vorburger - do we know why this fails? When I looked at this at one point it was actually due to a timeout - i.e. some of the scheduled jobs took too long to complete. If that's the case then fix would be easy by just increasing the timeout. Maybe to understand why this fails we need to first fix the logging for integration tests...

xurror:
I just feel increasing timeouts won't cut it. Doesn't sound like a real solution. Unfortunately, I don't understand much about schedular jobs to have a proper say in this but I think we have some performance issues. Need to figure out how to deal with that.

ptuomola:
@xurror fully agree that if the problem is timeouts, we should fix the performance issue and not increase the timeout. But I wasn't sure whether that's actually the cause in this case: that's one reason why I've seen this test fail, but that has mainly been when you run it on a test database that has a lot of data in it. I have never seen a timeout failure on an empty database (as in our integration tests), so we should first establish why this fails - unless someone already knows?

But this may then be linked to improving logging on integration test failures which could be achieved by using Spring Boot to run them directly without Cargo... so maybe we should first focus on those in order to have the data to debug this with.

vorburger:
It's probably not really a timeout, but re-enabling the scheduler seems to sometimes fail.

It just happened again e.g. in https://travis-ci.org/github/apache/fineract/builds/740634171 for #1487. It's frequent enough to be a PITA.

Why not just merge this, until someone wants to investigate fixing that flaky test? It's what https://github.com/apache/fineract#pull-requests says we do...

PR #1454, opened 2020-10-21
client: serialize LocalDate as 2009-01-01 (FINERACT-1234)

vorburger:
FINERACT-1234

PR #1460, opened 2020-10-21
Add new DocumentTest (FINERACT-1218)

vorburger:
This depends on #1454 and needs (tests) #1456 and will fail until those are in.

see FINERACT-1218


vorburger:
The failing `DocumentTest > createDocumentBadArgs()` here requires #1456 to be merged first, and deployed to Fineract.dev.

Which of course is all wrong (that it's testing Fineract.dev instead of localhost latest code), but to fix that, we need FINERACT-1209.

Until then, we could also just merge #1456, wait 15', and then this will pass (against Fineract.dev)... :smile:

vorburger:
@ptuomola thanks for previous merges, that makes this one now ready as the next one in my "chain"... how does this one look to you?

fineract-client/src/test/java/org/apache/fineract/integrationtests/newstyle/DocumentTest.java
```diff
@@ -0,0 +1,114 @@
...
+        GetEntityTypeEntityIdDocumentsResponse doc = ok(fineract().documents.getDocument("clients", clientId, documentId));
+        assertThat(doc.getName()).isEqualTo(newName);
+        assertThat(doc.getDescription()).isEqualTo(newDescription);
+    }
+
+    @Test
+    @Order(99)
+    void deleteDocument() throws IOException {
+        ok(fineract().documents.deleteDocument("clients", clientId, documentId));
+    }
```
ptuomola, line 108: Probably would make sense to check that get doesn't return the deleted document?

fineract-client/src/test/java/org/apache/fineract/integrationtests/newstyle/DocumentTest.java
```diff
@@ -0,0 +1,114 @@
...
+    final File testFile = new File(getClass().getResource("/michael.vorburger-crepes.jpg").getFile());
+
+    Long clientId = new ClientTest().getClientId();
+    Long documentId;
+
+    @Test
+    @Order(1)
+    void retrieveAllDocuments() throws IOException {
+        assertThat(ok(fineract().documents.retreiveAllDocuments("clients", clientId))).isNotNull();
+    }
```
ptuomola, line 48: Spelling mistake - should be retrieve not retreive

fineract-client/src/main/java/org/apache/fineract/client/services/DocumentsApiFixed.java
```diff
@@ -0,0 +1,147 @@
...
+import retrofit2.Call;
+import retrofit2.http.DELETE;
+import retrofit2.http.GET;
+import retrofit2.http.POST;
+import retrofit2.http.PUT;
+
+// This class was originally generated by OpenAPI Generator (https://openapi-generator.tech),
+// but then had to be manually edited by Michael Vorburger.ch to manually fix https://issues.apache.org/jira/browse/FINERACT-1227.
+// If we could fix our OpenAPI / Swagger YAML generation from the JAX RS and OpenAPI annotation to have the correct notation for binary document files and images, then this could be removed again.
+public interface DocumentsApiFixed {
```
ptuomola, line 36: Clearly would be great not to have to create this file but to have it autogenerated. I should have some time this weekend so can try to see if I can find a way to get this to work...

fineract-client/src/test/java/org/apache/fineract/integrationtests/newstyle/DocumentTest.java
```diff
@@ -0,0 +1,114 @@
...
+        assertThat(doc.getDescription()).isNull();
+        assertThat(doc.getId()).isEqualTo(documentId);
+        assertThat(doc.getParentEntityType()).isEqualTo("clients");
+        assertThat(doc.getParentEntityId()).isEqualTo(clientId);
+        // TODO huh?! It's more than uploaded file; seems like a bug - it's including create body, not just file size
+        assertThat(doc.getSize()).isEqualTo(testFile.length() + 385);
+        // TODO huh?! MIME is always text/plain instead of image/jpeg... :(
+        assertThat(doc.getType()).isEqualTo("text/plain");
+        // TODO doc.getStorageType() shouldn't be exposed by the API?!
+    }
```
ptuomola, line 79: Brilliant to see how many bugs writing a decent test uncovers! :-)

fineract-client/src/main/java/org/apache/fineract/client/services/DocumentsApiFixed.java
vorburger: If you do, take FINERACT-1227. (I'm not entirely sure if this is completely trival; curious myself how you would fix that.)

fineract-client/src/test/java/org/apache/fineract/integrationtests/newstyle/DocumentTest.java
vorburger: done, added: `assertThat(fineract().documents.getDocument("clients", clientId, documentId)).hasHttpStatus(404);`

fineract-client/src/main/java/org/apache/fineract/client/services/DocumentsApiFixed.java
vorburger: Agreed, see FINERACT-1227.

fineract-client/src/test/java/org/apache/fineract/integrationtests/newstyle/DocumentTest.java
vorburger: fixed (in 2 places)

fineract-client/src/test/java/org/apache/fineract/integrationtests/newstyle/DocumentTest.java
vorburger: Indeed. As well as how easy it now is to write decent tests, using the shiny new client API! PS: The MIME thing is a bug in the test util, not core - fix for that coming up as soon as this PR is merged... :smile_cat:

vorburger:
@ptuomola thanks for (as always very prompt, much appreciated) review! Have addressed your feedback. Want to merge this? Let's not hold this up with FINERACT-1227; if you do get to doing something about that, just wipe my `DocumentsApiFixed` again in a follow-up PR, but after this is merged? I would be obviously have no objection at all to see it gone - if this test still passes! :smile:

ptuomola:
Thanks @vorburger - happy to merge, except looks like the build is failing - something cosmetic by the looks of it...

vorburger:
@ptuomola build passed now!

PR #1465, opened 2020-10-22
FINERACT-1209: Integration Tests using new Swagger Client API

vidakovic:
FINERACT-1209

Still work in progress, the module layout is as it should be, but suddenly Spring Boot dependencies can't be resolved anymore without adding a BOM.
All integration test code moved in separate module as discussed.
Currently one problem left: compiled test classes can't be found although we are using standard dirs for eveything.
I have a feeling that we are defining somewhere paths that we shouldn't...


build.gradle
```diff
@@ -416,9 +408,14 @@ configure(project.fineractJavaProjects) {
...
     // https://github.com/diffplug/spotless/tree/main/plugin-gradle
+    dependencies {
+        spotbugsPlugins 'jp.skypencil.findbugs.slf4j:bug-pattern:1.5.0@jar'
+        spotbugs 'com.github.spotbugs:spotbugs:4.1.4'
+    }
+
     spotless {
         java {
-            targetExclude '**/build/**', '**/bin/**', '**/out/**'
+            targetExclude '**/build/**', '**/bin/**', '**/out/**', '**/integration-tests/**'
```
vorburger: why exclude integration-tests from Spotless now? Or are we already not covering them? Then we should separately fix that, first?

config/checkstyle/suppressions.xml
```diff
@@ -27,4 +27,6 @@
 
 <suppressions>
     <suppress files="[/\\]build[/\\]generated[/\\]" checks=".*"/>
+    <!-- TODO: fix integration test formatting -->
+    <suppress files="[/\\]integration-tests[/\\]src[/\\]test[/\\]main[/\\]" checks=".*"/>
```
vorburger, line 31: I'm guessing we currently haven't covered integration-tests? We should perhaps try to fix that, first? Or leave as is?

vorburger (commented):
Looks pretty cool! I have faith you'll be able to work out the problem.. :smiley_cat:  And confirming the direction / goal; yes this is what we're after, assuming that this will allow us to then add a dependency to `fineract-client` to the new `integration-tests` - that's the plan here, right? PS: Are our current `integrationTest` in `fineract-provider` not covered by Spotless and Checkstyle? We should probably fix that.. separately, first? Spotless should be trivial, change build and build apply - no? Checkstyle... more work? Do you want help with that?

config/checkstyle/suppressions.xml
vidakovic: For now I'd leave the exclusion. Checkstyle reports an avalanche of 1500 errors. I'd say we can fix this when we migrate step by step to to fineract-client usage.

build.gradle
vidakovic: True... was a bit late and was not sure where all the warnings/errors were coming from. I'll revert.

vidakovic:
@vorburger Can't believe it... it's alive! Was almost giving up hope on this one... but in the end it kind of made sense that it failed. The problem was the test classpath. I thought including fineract-provider with "testImplementation project(':fineract-provider')" would be enough, but didn't take into account that we are not creating a "normal" JAR artifact in that module! The repackaged Spring Boot JAR can't be used as a dependency, because the test classloader doesn't know how to retrieve the fineract-provider classes from there.

Solution: point test dependencies directly to fineract-provider/build/classes/java/main... voila!

vidakovic:
FYI: don't think this is related to my changes; before a run was failing on a different line (creating Docker images). Locally the integration tests worked ok; maybe the Travis configuration needs some tweaking.

vorburger (changes requested):
@vidakovic check out build failure, and merge conflict?

vorburger:
> FYI: don't think this is related to my changes; before a run was failing on a different line (creating Docker images). Locally the integration tests worked ok; maybe the Travis configuration needs some tweaking.

@vidakovic I had a quick closer look, and created FINERACT-1244. Always follow https://github.com/apache/fineract/#pull-requests in such cases.

build.gradle
vorburger: Thanks! I have checked and see that this was done, so Resolving this point.

config/checkstyle/suppressions.xml
vorburger: OK. Let's follow-up on that in FINERACT-1246. I'll leave this review comment open.

vorburger:
FYI build failure here should be addressed by my #1474. I'll rebase this when I got that in, because I'd like to continue this work.. wwe'll still need to move the new API client tests in the new module; I'm hoping that will be relatively trivial once this goes in, and am happy to do that last step, unless you wanted to do so?

vorburger:
@vidakovic actually I should go get some fresh air... :smile: will you rebase this? I'd love to review and hopefully merge it when it's green. FYI fhe Merge conflict on `RecurringDepositTest` is re. FINERACT-1248 - will you add that `@Disabled` to the moved file?

vidakovic:
@vorburger yup, will do... enjoy your walk 🙂

vorburger:
@vidakovic I've resolved the merge conflict (due to Spring upgrade), using GitHub's Merge UI thus there's a Merge commit, let's see if it builds now.

vorburger:
@vidakovic it failed an integration test, see FINERACT-1250.

vorburger (changes requested):
NOK as as FINERACT-1250.

vidakovic:
@vorburger Looks like the tests are running twice for some reason... have to look into this... probably a "dependsOn" too many or something along those lines.

integration-tests/build.gradle
```diff
@@ -0,0 +1,73 @@
...
+cargoStartLocal.mustRunAfter 'testClasses'
+
+test {
+    useJUnitPlatform()
+    testLogging {
+        // FINERACT-927
+        events 'skipped', 'failed'
+        showStandardStreams = false
+        exceptionFormat 'full'
+    }
```
vorburger, line 70: @vidakovic my Gradle foo isn't as strong as yours :smile: but isn't this "inherited" from the root `build.gradle`, after (my) #1477? It's not a big enough problem to not merge this ASAP anyway (I'm reviewing this right now..), but if you agree, then perhaps you want to raise a minor clean-up PR as follow-up?

integration-tests/dependencies.gradle
```diff
@@ -0,0 +1,45 @@
...
+    testImplementation ('org.mock-server:mockserver-junit-jupiter') {
+        exclude group: 'com.sun.mail', module: 'mailapi'
+        exclude group: 'javax.servlet', module: 'javax.servlet-api'
+        exclude group: 'javax.validation'
+    }
+    testImplementation ('io.rest-assured:rest-assured') {
+        exclude group: 'commons-logging'
+        exclude group: 'org.apache.sling'
+        exclude group: 'com.sun.xml.bind'
+    }
```
vorburger, line 41: @vidakovic given that this is here now, we should remove it from `fineract/fineract-provider/dependencies.gradle`, no? Again not a big enough problem to not merge this ASAP anyway (I'm reviewing this right now..), but if you agree, then perhaps you want to raise a minor clean-up PR as follow-up?

.travis.yml
```diff
@@ -72,6 +72,6 @@ script:
...
-  - ./gradlew --console=plain -PautomatedBuild=true licenseMain licenseTest licenseIntegrationTest check  &&  ./gradlew --console=plain -PautomatedBuild=true build integrationTest --fail-fast  &&  sudo service mysql stop  &&  docker-compose build  &&  docker-compose up -d  &&  sleep 60s  && curl -f -k --retry 5 --retry-connrefused --connect-timeout 30 --retry-delay 30 https://localhost:8443/fineract-provider/actuator/health  &&  (( $(curl -f -k --retry 5 --retry-connrefused --connect-timeout 30 --retry-delay 30 https://localhost:8443/fineract-provider/actuator/info | wc --chars) > 100 ))
+  - ./gradlew --console=plain -PautomatedBuild=true licenseMain licenseTest check -x test  &&  ./gradlew --console=plain -PautomatedBuild=true build test --fail-fast  &&  sudo service mysql stop  &&  docker-compose build  &&  docker-compose up -d  &&  sleep 60s  && curl -f -k --retry 5 --retry-connrefused --connect-timeout 30 --retry-delay 30 https://localhost:8443/fineract-provider/actuator/health  &&  (( $(curl -f -k --retry 5 --retry-connrefused --connect-timeout 30 --retry-delay 30 https://localhost:8443/fineract-provider/actuator/info | wc --chars) > 100 ))
```
vorburger, line 75: FTR: FINERACT-1250 had a suggestion to simplify this, which we should try in a follow-up PR (I'll raise one).

vorburger (approved):
LGTM! I'll merge this. -- Three minor suggestions for follow-up PRs.

integration-tests/build.gradle
vidakovic: @vorburger correct... will create a PR for that... saw this during a rebase, but forgot to clean it up here.

.travis.yml
vorburger: #1489

integration-tests/build.gradle
vorburger: #1490

integration-tests/dependencies.gradle
vorburger: #1491

PR #1473, opened 2020-10-31
Revert "FINERACT-1241-elastic-web-hook"

vorburger:
Reverts apache/fineract#1466 due to broken build, see details in FINERACT-1241.

PR #1466, opened 2020-10-24
FINERACT-1241-elastic-web-hook

fynmanoj:
FINERACT-1241

## Description

Describe the changes made and why they were made.

Ignore if these details are present on the associated [Apache Fineract JIRA ticket](https://github.com/apache/fineract/pull/1284).

## Checklist

Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Write the commit message as per https://github.com/apache/fineract/#pull-requests

- [x] Acknowledge that we will not review PRs that are not passing the build _("green")_ - it is your responsibility to get a proposed PR to pass the build, not primarily the project's maintainers.

- [ ] Create/update unit or integration tests for verifying the changes made.

- [x] Follow coding conventions at https://cwiki.apache.org/confluence/display/FINERACT/Coding+Conventions.

- [ ] Add required Swagger annotation and update API documentation at fineract-provider/src/main/resources/static/api-docs/apiLive.htm with details of any API changes

- [x] Submission is not a "code dump".  (Large changes can be made "in repository" via a branch.  Ask on the developer mailing list for guidance, if required.)

FYI our guidelines for code reviews are at https://cwiki.apache.org/confluence/display/FINERACT/Code+Review+Guide.

PR #1474, opened 2020-10-31
Disable failing RecurringDepositTest + ImageTest + Revert "FINERACT-1241-elastic-web-hook (#1466)"

vorburger:
This is #1473, the revert of #1466 due to broken build (see details in FINERACT-1241) + a `@Disabled` on `ImageTest` due to FINERACT-1247.


vorburger:
This STILL failed, now due to FINERACT-1248, so I've added a 3rd commit here to disable that test as well.. will it finally pass? :imp:

vorburger:
As this combination of 3 (!) problems is now finally green built, I've taken the liberty to exceptionally self merge this to unblock everyone.

PR #1477, opened 2020-10-31
Move Gradle test exceptionFormat up (FINERACT-1247)

vorburger:
see FINERACT-1247

@vidakovic would you like to review and merge this one?

PR #1490, opened 2020-11-02
Remove redundant testLogging etc. (FINERACT-1209)

vorburger:
because https://github.com/apache/fineract/pull/1477/files
moved this from fineract-provider/build.gradle to the root ./build.gradle
so it's inherited.

### DIFF
--- a/integration-tests/build.gradle
+++ b/integration-tests/build.gradle
@@ -61,13 +61,6 @@ cargoStartLocal.dependsOn ':fineract-provider:war'
 cargoStartLocal.mustRunAfter 'testClasses'
 
 test {
-    useJUnitPlatform()
-    testLogging {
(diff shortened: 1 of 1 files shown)


vorburger:
@vidakovic my Gradle foo isn't as strong as yours smile but isn't this from 1465 "inherited" from the root build.gradle, after (my) #1477?

Nota bene I have **NOT** actually tested this out - proposing it for your review, **_IFF_** my understanding is correct.

vidakovic:
@vorburger I might have created a duplicate PR #1492. At least partially... I've also cleaned up the test dependencies in fineract-provider. So maybe we can merge it after this one. FYI

vidakovic (approved):
LGTM

PR #1491, opened 2020-11-02
Remove un-used code & testImplementation dependencies from provider (FINERACT-1209)

vorburger:
FINERACT-1209 related clean up follow-up to #1465.


vorburger:
@vidakovic how about this?

vidakovic:
@vorburger I just ran tests locally for fineract-provider and it was complaining about the missing restassured dependencies. Other than that I think this is good.

Update: Ah, just saw that you removed those classes that referenced restassured. Ignore above then.

vidakovic (approved):
LGTM

PR #1495, opened 2020-11-07
Log only WARN & ERR but not INFO anymore (re. FINERACT-1168)

vorburger:
FINERACT-1295

PS: Originally related to (but doesn't solve) FINERACT-1168, but can really be independently reviewed and merged - or not.

### DIFF
--- a/fineract-provider/src/main/resources/logback-spring.xml
+++ b/fineract-provider/src/main/resources/logback-spring.xml
@@ -1,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements. See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership. The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License. You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied. See the License for the
+    specific language governing permissions and limitations
+    under the License.
+
+-->
+<configuration>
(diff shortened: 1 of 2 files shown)


ptuomola:
Just wondering... do we want to make this the default behaviour? Or just something that is used when running integration tests? I personally think seeing the INFO would be OK as normal behaviour (as that's what people would expect to see from other Spring Boot apps) - and we would just turn it off for Integration Tests. But not sure how easy that is to achieve?

vorburger:
I thought about this, and while using it actually started liking it better like this, it's suddenly so... "quiet, and clean"! ;-) It also "feels" like it's "lighter" and"starts faster", although that's just "perception", of course. But YMMV. It's a bit like Maven (which verbosely logs everything it's doing) VS Gradle (which spews less logs). This also feels like it's more inline with our https://github.com/apache/fineract#logging-guidelines, notably the point about log levels. So my vote is on just changing this generally, not just for ITs, but I'm interested in how others feel about it.

vorburger:
What do other recently active commiters think re this - pro or con? @vidakovic @avikganguly01 @fynmanoj @percyashu  @awasum

xurror:
I understand this does not resolve https://issues.apache.org/jira/browse/FINERACT-1168 but I think this change only looks good if we can get https://issues.apache.org/jira/browse/FINERACT-764. So, just a thought instead having a foreplay with this logging why not just package it with https://github.com/apache/fineract/pull/1444 or bring it in after.
But overall, I think it's a good fit though.

avikganguly01:
@vorburger : Logstash scripts in some older implementations will break if we upgrade but that's fine.

vorburger:
@avikganguly01 do you think it's useful to have INFO logs and has any real world in your logstash? My experience in running Fineract.dev and using Google Error Reporting has been that all I ever look at and seem to care about is WARN and ERROR... (BTW note all the issues linked from FINERACT-932 which I've been able to find! Do you want to contribute fixes for some of them?)

ptuomola:
@vorburger just to be clear, I'm happy either way... haven't tried this to see how much cleaner it is, but at the same time I don't think there's a huge amount of useful information in the INFO logging either.

vorburger:
FINERACT-1295

vorburger:
I'm "self merging" this, given that nobody objected on https://lists.apache.org/thread.html/r9e3de9996a7f9f1bd49615b5a7d44f6c8daeca5cbccdac2967233bbd%40%3Cdev.fineract.apache.org%3E

PR #1494, opened 2020-11-06
Move all SDK client tests into integrationtests (FINERACT-1209)

vorburger:
see FINERACT-1209; this now finally runs those tets against localhost instead of https://fineract.dev

PR #1498, opened 2020-11-07
Fix Bulk Import failing for office names containing hyphens (FINERACT-1256)

vorburger:
see FINERACT-1256


vorburger:
The same commit is also included as part of #1494, so I can close this if that gets merged.

ptuomola (approved):
Looks great to me!

PR #1502, opened 2020-11-18
docs: Fix broken build

vorburger:
I've tried out `./gradlew :fineract-doc:asciidoctor` for the first time today, and found that it actually had errors, which this fixes.

This also makes the build fail if new such errors would be re-introduced again later, which seems like a useful thing to do, to me.


vorburger:
@vidakovic can you review this and merge when the build passes? Hope you'll like... :smile:

vidakovic (approved):
LGTM

PR #1507, opened 2020-11-26
FINERACT-1198 - Revert spotlessApply and use only spotlessCheck

xurror:
https://issues.apache.org/jira/browse/FINERACT-1198


ptuomola (approved):
LGTM

PR #1508, opened 2020-11-26
FINERACT-27 - Extend holiday period to 30 days

xurror:
https://issues.apache.org/jira/browse/FINERACT-27


awasum (approved):
LGTM. I wonder if any domain experts will have problems with this? Is it worth sending a message on the list? But looks ok to be merged just from technical side of things.

awasum:
I will merge this one. Is a test specifically testing this case?

PR #1513, opened 2020-11-28
chore: Switch GitHub rebase action checkout@master to @v2

vorburger:
Using a fixed version here is more stable (otherwise future changes in actions/checkout@v3 could theoretically break this for us), and in line with https://github.com/cirrus-actions/rebase/commit/51854755db80c104a3f4be4b617988933209b5a1, which I've stumbled upon from #1509. This is also consistent with our other GitHub actions.


awasum:
This makes sense. Just waiting for routine checks to pass on Travis. Thanks @vorburger

PR #1531, opened 2020-12-02
Fix missing SMTP Server Port in GmailBackedPlatformEmailService (FINERACT-1070)

vorburger:
see FINERACT-1070


ptuomola:
@vorburger - now that we turned off the automagic spotlessApply, looks like you need to run it manually for this one...

PR #1541, opened 2020-12-11
Better fix for SMTP Server Port (FINERACT-1070)

vorburger:
FINERACT-1070


vorburger:
@thesmallstar would you like to code review this, and merge it of you think that it's fine? It's time to inverse our roles now! ;-)

thesmallstar:
From the recent comment of  @francisguchie  on [FINERACT-1070 ](https://issues.apache.org/jira/browse/FINERACT-1070)  it seems like the entire system is still not working correctly. The problem however I feel is related to the frontend (and pre-existed as from comment from awasum) and the original problem related to FINERACT-1070 is solved here. Hence, merging this.

PR #1562, opened 2021-01-05
Fix - use expected disbursement for ideal disbursement to fix wrong interest

eitanfr:
**ISSUE**
My issue/bug happens when I am setting "First repayment on" field I get a repayment schedule with a wrong interest rate for the first installment.
Example: When I set "First repayment on" to 01/02/21 and expected disbursement is 27/1/21(Frist image) then I get the wrong interest.The first installment interest is 21,232.88$ (second Image) - calculated for the entire month (31 days) instead of 5 days (the time from the disbursement to the first repayment), the right interest should be 3424.75$.

**After debugging:**
In AbstractLoanScheduleGenerator there is a function `calculateInterestStartDateForPeriod` - it uses the `idealDisbursment` for setting the interest start date for the first installment, The `idealDisbursment` is calculated by `idealDisbursementDateBasedOnFirstRepaymentDate` which does [`firstRepaymentDate` - 1 repayment Period Frequency] aka, for 01/02/21 with months as repyment frequency it will return 01/01/21.

So when we dont set the "First repayment on" the `idealDisbursment` will be the expected disbursement date because the first repayment is always `expectedDisbursment`+1. But if we set the  "First repayment on" we get the wrong value for the `idealDisbursment` - in my example we got 01/0/21 and this is why the first installment interest was calculated for an entire month instead of  5 days.

**Is it a bug or not?**
I am not sure why `idealDisbursementDateBasedOnFirstRepaymentDate` exists, the simple solution is to do  idealDisbursment=expectedDisbursmentDate.I tried it locally and looks like it works, I will be happy to make a PR, but want to make sure indeed that the `idealDisbursementDateBasedOnFirstRepaymentDate` is redundant.

@vorburger 

image

image

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/AbstractLoanScheduleGenerator.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/AbstractLoanScheduleGenerator.java
@@ -123,12 +123,11 @@ public abstract class AbstractLoanScheduleGenerator implements LoanScheduleGener
         boolean isFirstRepayment = true;
         LocalDate firstRepaymentdate = this.scheduledDateGenerator
                 .generateNextRepaymentDate(loanApplicationTerms.getExpectedDisbursementDate(), loanApplicationTerms, isFirstRepayment);
-        final LocalDate idealDisbursementDate = this.scheduledDateGenerator.idealDisbursementDateBasedOnFirstRepaymentDate(
-                loanApplicationTerms.getLoanTermPeriodFrequencyType(), loanApplicationTerms.getRepaymentEvery(), firstRepaymentdate,
-                loanApplicationTerms.getLoanCalendar(), loanApplicationTerms.getHolidayDetailDTO(), loanApplicationTerms);
+
+        final LocalDate idealDisbursementDate = loanApplicationTerms.getExpectedDisbursementDate();
 
         if (!scheduleParams.isPartialUpdate()) {
-            // Set Fixed Principal Amount
+            // Set Fixed Principal  Amount
             updateAmortization(mc, loanApplicationTerms, scheduleParams.getPeriodNumber(), scheduleParams.getOutstandingBalance());
 
             if (loanApplicationTerms.isMultiDisburseLoan()) {


vorburger:
@eitanfr my functional understanding is actually not deep enough to be able to easily review this. Wondering if perhaps @ptuomola or @vidakovic or @awasum would like to chime in here? Otherwise this probably a good kind of topic to take the mailing list. Having said all that, you appear to have a good grasp of what's going here, so I would merge this PR once it passes the build, and if you could please create a bug about it with the description above on https://issues.apache.org/jira/ - we also operate on "trust" in this project, so I'm pretty open to merging a PR like this.

PS: All PR builds are currently broken due to FINERACT-924, just FYI.

PPS: Sorry it took me a while to respond here; been busy at my day job work.

vorburger:
https://lists.apache.org/thread.html/raf85df08b65ced14ae6294ea5af0695c5124d196389c9cccf388b848%40%3Cdev.fineract.apache.org%3E thread seems to related to this PR? Given @bharathc27 (?) reply there, would you still like this to be merged?

eitanfr:
> https://lists.apache.org/thread.html/raf85df08b65ced14ae6294ea5af0695c5124d196389c9cccf388b848%40%3Cdev.fineract.apache.org%3E thread seems to related to this PR? Given @bharathc27 (?) reply there, would you still like this to be merged?

As I understood Bharath mentioned a way for a workaround around the issue, but the issue still exists... 
I think its need to be merged, and it works fine when I tested it, With that said I want to mention that I am still a Fineract beginner :)


I will rebase soon,

vorburger:
@eitanfr I would merge this PR if you could please create a bug about it with the description above on https://issues.apache.org/jira/ ? Only so that we have Release Notes for it in the upcoming 1.5.0 (we derive those from JIRA, not the commits).

vorburger:
And you should amend the git commit to refer to that JIRA - see the convention documented on https://github.com/apache/fineract#pull-requests.

eitanfr:
Hi, 
done and done: https://issues.apache.org/jira/browse/FINERACT-1307

awasum:
@eitanfr  Travis fails this PR...Do you want to take and look and see if you can fix the issue?

vorburger:
> @eitanfr Travis fails this PR...Do you want to take and look and see if you can fix the issue?

@eitanfr it's just because of a Spotless violation. Do you know how to find the log? It says what you have to do...

bharathcgowda:
@eitanfr
Just wanted to mention the functional scenarios again here

Scenario 1
when "**interest calculation period**" is set to "**daily**" then interest has to calculate from the **Disbursement date**

Scenario 2
when "**interest calculation period**" is set to "**same as repayment**" and  "Calculate interest for exact days in partial period" checkbox is **selected** then interest has to calculate from the **Disbursement date**

Scenario 3
when "**interest calculation period**" is set to "**same as repayment**" and  "Calculate interest for exact days in partial period" checkbox is **not selected** then interest has to calculate for that month.

As long as this fix doesn't break these 3 scenarios, it should be fine. 

Also, I would suggest a thorough test with different configurations before making any changes to the interest calculation functionality.

edcable:
@Nayan could Pramod Sharma from your team help in doing the functional review requested above by Bharath. 

@eitanfr have you had a chance to test against these scenarios yet?

eitanfr:
@bharathcgowda @edcable 
Hi, as you guys requested :)

**First Scenario:**
when "interest calculation period" is set to "daily" then interest has to calculate from the Disbursement date.
Expected:
image

Works as Expected:
image

**Second Scenario:**
when "interest calculation period" is set to "same as repayment" and "Calculate interest for exact days in partial period" checkbox is selected then interest has to calculate from the Disbursement date
Expected:
image

Works as Expected:
image

**Third Scenario**
when "interest calculation period" is set to "same as repayment" and "Calculate interest for exact days in partial period" checkbox is not selected then interest has to calculate for that month.
Expected:
image

Works as Expected:
image

PR #1575, opened 2021-01-15
clean up (mechanical) SchedulerJobsTestResults

vorburger:
re. FINERACT-924; but not actually fixing that

Not tested, needs rebase and testing after FINERACT-924 is fixed


vorburger:
This may still fail due to a missing `./gradlew spotlessApply` - I'll do that when I'll rebase this after FINERACT-924 is fixed.

vorburger:
Given that the test still passes after what should be my only "mechanical" clean up changes to the test (only; no production code changed), I'll exceptionally "self merge" this hopefully harmless and trivial change, instead of waiting for reviews.

PR #1582, opened 2021-01-18
remove un-used class Tenant

vorburger:


davidyaha:
Hey @vorburger, I think that this should be used in the future to expose tenants api

vorburger:
@davidyaha then we should put it back in future, when it's used. Until then, it's "dead code", which can and should be removed. - Certain companies I can't publicly tell you about ;) even do this kind of thing automatically by bots!

PR #1583, opened 2021-01-18
Fix MYSQL Access Denied Issue on restart

BLasan:
## Description

See [FINERACT-1294
](https://issues.apache.org/jira/browse/FINERACT-1294)

## Checklist

### DIFF
--- a/kubernetes/fineractmysql-deployment.yml
+++ b/kubernetes/fineractmysql-deployment.yml
@@ -96,11 +96,7 @@ spec:
                   key: password
           livenessProbe:
             exec:
-              command:
-                - mysqladmin
-                - ping
-                - -h
-                - localhost
+              command: ["sh","-c","mysqladmin ping -h localhost -uroot -p${MYSQL_ROOT_PASSWORD}"]
             failureThreshold: 10
             timeoutSeconds: 10
           ports:
--- a/README.md
+++ b/README.md
@@ -178,6 +178,12 @@ To shutdown and reset your cluster, run:
 
     ./kubectl-shutdown.sh
 
+To shutdown and reset your cluster, run:
+
+    minikube ssh
+
+    sudo rm -rf /mnt/data/
+
 We have [some open issues in JIRA with Kubernetes related enhancement ideas](https://jira.apache.org/jira/browse/FINERACT-783?jql=labels%20%3D%20kubernetes%20AND%20project%20%3D%20%22Apache%20Fineract%22%20) which you are welcome to contribute to.
 
 


kubernetes/kubectl-startup.sh
```diff
@@ -19,7 +19,7 @@
 #
 
 echo "Setting Up Fineract service configuration..."
-kubectl create secret generic fineract-tenants-db-secret --from-literal=username=root --from-literal=password=$(head /dev/urandom | tr -dc A-Za-z0-9 | head -c 16)
+kubectl create secret generic fineract-tenants-db-secret --from-literal=username=root --from-literal=password=root
```
vorburger: Wait a second... doesn't this hardcode it to"root"?! That's..  probably not such a good idea? 😎

vorburger (changes requested):

kubernetes/kubectl-startup.sh
BLasan: Actually these usernames and passwords will be saved as secrets.  But anyway you can get those values by running the command `kubectl get secrets/fineract-tenants-db-secret --template={{.data.password}} | base64 --decode` . And the problem in here is that when we use a random password, after restarting the container we'll be getting access denied issue.

BLasan: I do have another solution for this. Not sure whether it'll help to sort out this issue. Let me try that way as well :)

BLasan: I have used an initcontainer instead of hardcoding the password. So before the container creation it will bind /var/lib/mysql directory which could help us to get rid of this "Access Denied" issue after restarting the container

BLasan: @vorburger Please review the new changes

BLasan:
@vorburger  @xurror  This issue will occur if we do not remove volume mounted inside minikube node when cluster resetting. I've updated REDME.md and changed the liveness probe command which needs to run with root user credentials. Please review

kubernetes/fineractmysql-deployment.yml
```diff
@@ -96,11 +96,7 @@ spec:
                   key: password
           livenessProbe:
             exec:
-              command:
-                - mysqladmin
-                - ping
-                - -h
-                - localhost
+              command: ["sh","-c","mysqladmin ping -h localhost -uroot -p${MYSQL_ROOT_PASSWORD}"]
```
vorburger, line 99: @BLasan sorry I'm not getting something here- where/by what is this MYSQL_ROOT_PASSWORD defined? Tx. /Cc @xurror

BLasan:
@vorburger Sir could you please take a look?

vorburger (changes requested):

kubernetes/fineractmysql-deployment.yml
vorburger: @BLasan oh wait I am dumb, this `MYSQL_ROOT_PASSWORD` **IS** set by us, in https://github.com/apache/fineract/blob/develop/kubernetes/fineractmysql-deployment.yml#L92, now I get how this works! Then this change makes perfect sense - LGTM and I'm mergint it now. FYI @xurror

PR #1584, opened 2021-01-20
Add init-container to check MYSQL server deployment (FINERACT-1300)

BLasan:
## Description

See[ FINERACT-1300](https://issues.apache.org/jira/browse/FINERACT-1300)

## Checklist

### DIFF
--- a/kubernetes/fineract-server-deployment.yml
+++ b/kubernetes/fineract-server-deployment.yml
@@ -54,6 +54,10 @@ spec:
         app: fineract-server
         tier: backend
     spec:
+      initContainers:
+        - name: init-mydb
+          image: busybox:1.28
+          command: ['sh', '-c', 'echo -e "Checking for the availability of MYSQL server deployment"; while ! nc -z "fineractmysql" 3306; do sleep 1; printf "-"; done; echo -e " >> MYSQL server has started";']
       containers:
       - name: fineract-server
         image: apache/fineract:latest


BLasan:
@xurror @vorburger SIr Please review

kubernetes/fineractmysql-deployment.yml
```diff
@@ -96,11 +96,7 @@ spec:
                   key: password
           livenessProbe:
             exec:
-              command:
-                - mysqladmin
-                - ping
-                - -h
-                - localhost
+              command: ["sh","-c","mysqladmin ping -h localhost -uroot -p${MYSQL_ROOT_PASSWORD}"]
```
vorburger: NB this change here was already in #1583 (for FINERACT-1294), so this diff should disappear on a rebase.

kubernetes/fineract-server-deployment.yml
```diff
@@ -54,6 +54,10 @@ spec:
         app: fineract-server
         tier: backend
     spec:
+      initContainers:
+        - name: init-mydb
+          image: busybox:1.28
+          command: ['sh', '-c', 'echo -e "Checking for the availability of MYSQL server deployment"; while ! nc -z "fineractmysql" 3306; do sleep 1; printf "-"; done; echo -e " >> MYSQL server has started";']
```
vorburger, line 60: @BLasan a thought: instead of using `busybox` and `nc`, why not just re-use `image: mysql:5.7` with `mysqladmin` ?

vorburger (changes requested):
feedback proposing a slightly different / better way to do this

PS: Our `/rebase` comment bot apparently doesn't actually seem to work anymore (https://github.com/apache/fineract/actions says _Startup failure_); so please rebase this PR when you implemented the change above, so that diff in `kubernetes/fineract-server-deployment.yml` disappears, now that  #1583 has been merged.

kubernetes/fineract-server-deployment.yml
BLasan: Sir, in here we need to just check whether the MYSQL server (remote server) is up and running. By using busybox image we can achieve that easily (from the bash command). mysql:5.7 image will need some extra configurations to establish the connection and listen to a remote server.(setting up credentials etc.. which is unnecessary) Simply the idea here is to verify that mysql server is running properly on the correct port and then FINERACT server could properly connect to mysql server and do the required tasks. Otherwise if we get a problem with mysql server, fineract server will go in to a CrashLoopBackOff state which is unnecessary.

BLasan:
@vorburger @xurror  please review sir

kubernetes/fineract-server-deployment.yml
vorburger: > Simply the idea here is to verify that mysql server is running properly on the correct port and then FINERACT server could properly connect to mysql server and do the required tasks. Otherwise if we get a problem with mysql server, fineract server will go in to a CrashLoopBackOff state which is unnecessary.

> we need to just check whether the MYSQL server (remote server) is up and running. 

Understod, and I think that's a great idea!

> By using busybox image we can achieve that easily (from the bash command). 

`nc` just makes sure there's a connection on the port. 

`mysqladmin ping` may do more? I' haven't checked. It should make sure server is really up. 

`mysqladmin status` maybe better.

> mysql:5.7 image will need some extra configurations to establish the connection and listen to a remote server.(setting up credentials etc.. which is unnecessary) 

It shouldn't, hopefully just `-h fineractmysql` should suffice? (But I haven't tested it myself.)

But you know how the saying "done is better than perfect"? 

I'll merge this anyway, and leave it up to you if you want to improve it in a follow-up PR, or leave it as is, that's fine too; it's not super important.

BLasan: Yeah, but sir that's why we have added liveness probe for mysql deployment. And from fineract point of view, it only needs to check whether the particular server and port up and running. Your point is correct and it's fulfilled by liveness probe inside mysql container.

PR #1623, opened 2021-02-16
feat: Helm chart

Yann-J:
## Description

This is a proper Helm chart to run it on Kubernetes, as a more flexible alternative to the existing deployment definitions and setup scripts. It uses the bitnami mysql chart as a dependency to install MySQL, so all these values can be overridden.

To install, just run:

```sh
# Install mysql chart dependency
helm dependency update

# Install chart
helm upgrade --install --set mysql.auth.rootPassword=xxxx my-fineract helm/fineract
```

## Checklist

### DIFF
--- a/helm/fineract/templates/fineract-secret.yaml
+++ b/helm/fineract/templates/fineract-secret.yaml
@@ -1,0 +1,8 @@
+kind: Secret
+apiVersion: v1
+metadata:
+  name: {{ .Release.Name }}
+  namespace: {{ .Release.Namespace }}
+data:
+  {{- include "secrets" . | nindent 2 }}
+type: Opaque
--- a/helm/fineract/templates/fineract-secret.yaml
+++ b/helm/fineract/templates/fineract-secret.yaml
@@ -1,3 +1,20 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
 kind: Secret
 apiVersion: v1
 metadata:
--- a/helm/fineract/templates/fineract-secret.yaml
+++ b/helm/fineract/templates/fineract-secret.yaml
@@ -13,7 +13,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-#
+{{- $unused := required "You must set a MySQL root password in mysql.auth.rootPassword" .Values.mysql.auth.rootPassword }}
+{{- $unused := required "You must set a MySQL user password in mysql.auth.password" .Values.mysql.auth.password }}
 
 kind: Secret
 apiVersion: v1
--- a/helm/fineract/templates/fineract-server-deployment.yml
+++ b/helm/fineract/templates/fineract-server-deployment.yml
@@ -38,11 +38,13 @@ spec:
         tier: backend
         instance: {{ .Release.Name }}
       annotations:
+        # Force a pod restart if secret values change
         checksum/secrets: {{ include "secrets" . | sha256sum }}
     spec:
       initContainers:
         - name: wait-db
           image: jwilder/dockerize
+          # We don't care about having the latest...
           imagePullPolicy: IfNotPresent
           args:
           - -wait
@@ -64,10 +66,11 @@ spec:
               value: {{ include "mysql_user" . }}
             - name: FINERACT_DEFAULT_TENANTDB_UID
               value: {{ include "mysql_user" . }}
-            {{ if .Values.fineractServer.extraEnv }}
+            {{- if .Values.fineractServer.extraEnv }}
             {{- range $key, $value := .Values.fineractServer.extraEnv }}
             - name: {{ $key | quote}}
-              value: {{ $value | quote }}
+              value: |-
+                {{- tpl $value $ | nindent 16 }}
             {{- end }}
             {{- end }}
           ports:
@@ -75,13 +78,15 @@ spec:
               name: http
           readinessProbe:
             httpGet:
-              path: /fineract-provider/actuator/health
+              path: /fineract-provider/actuator/health/readiness
               port: 8080
-            initialDelaySeconds: 15
+            initialDelaySeconds: 90
             timeoutSeconds: 5
           livenessProbe:
             httpGet:
-              path: /fineract-provider/actuator/health
+              path: /fineract-provider/actuator/health/liveness
               port: 8080
-            initialDelaySeconds: 300
+            initialDelaySeconds: 90
             timeoutSeconds: 10
+          resources:
+            {{- .Values.fineractServer.resources | toYaml | nindent 12 }}
--- a/helm/fineract/templates/fineract-server-deployment.yml
+++ b/helm/fineract/templates/fineract-server-deployment.yml
@@ -1,0 +1,77 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements. See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership. The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License. You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ .Release.Name }}-backend
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: fineract
+    tier: backend
+spec:
+  replicas: {{ .Values.replicas }}
+  selector:
+    matchLabels:
+      app: fineract-backend
+      tier: backend
+      instance: {{ .Release.Name }}
+  strategy:
+    type: Recreate
+  template:
+    metadata:
+      labels:
+        app: fineract-backend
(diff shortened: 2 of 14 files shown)


edcable:
@Yann-J It was nice speaking to you the other day. @vidakovic our release manager would love to include this in the forthcoming Fineract 1.5 release - are you able to address the failing build?

helm/fineract/values.yaml
```diff
@@ -0,0 +1,83 @@
...
+# Any other parameter from https://artifacthub.io/packages/helm/bitnami/mysql
+mysql:
+  enabled: true
+  # Required because db driver doesn't support mysql 8...
+  image:
+    tag: '5.7'
+  auth:
+    # Please change these...
+    rootPassword: ozIRdODs8Jvs3BzZywgK
+    password: ihcqsCcsX6
```
vorburger: These default password here scare me, a lot... people will NOT change these - they never do.

There must a better way to do this, with Helm; what would be the Helm equivalent of what we're doing in https://github.com/apache/fineract/blob/develop/kubernetes/kubectl-startup.sh#L22 with `kubectl create secret generic fineract-tenants-db-secret --from-literal=username=root --from-literal=password=$(head /dev/urandom | tr -dc A-Za-z0-9 | head -c 16)` ?

vorburger (changes requested):
No Go from me for hard-coded default passwords - there must be a better solution?

More minor: How about move this into `kubernetes/helm/` instead of (root) `helm/`?

Lastly, could you include the how to use doc in the README instead of PR desc?

vorburger:
@BLasan have you noticed this PR? It looks like something you could be interested in helping to review... 😄 

I was thinking about something when reviewing #1653: That (resource limits) should be done here as well? More importantly, we're now maintaining Kubernetes YAML in 2 places (original raw Kube YAML, and Helm chart templates). That's perhaps not ideal... you guys should figure out how you want to do this going forward? Keep both? Replace the original YAML with the Helm templates, and subsequently remove the originals? Up to you - work together.

BLasan:
> @BLasan have you noticed this PR? It looks like something you could be interested in helping to review... 
> 
> I was thinking about something when reviewing #1653: That (resource limits) should be done here as well? More importantly, we're now maintaining Kubernetes YAML in 2 places (original raw Kube YAML, and Helm chart templates). That's perhaps not ideal... you guys should figure out how you want to do this going forward? Keep both? Replace the original YAML with the Helm templates, and subsequently remove the originals? Up to you - work together.

It would be great if we could add resource limits in order to run the image using minimum number of resources in the cluster. (Cluster resources will not be over eaten by then) If someone wants to run the product in k8s without cloning it, they could use the helm charts directly (after releasing this) What if we could move these helm implementations to a new repository and maintain it there? Don't know whether this is a good idea. But we could maintain this and can do releases separately.

helm/fineract/templates/fineract-server-deployment.yml
```diff
@@ -0,0 +1,87 @@
...
+            {{ if .Values.fineractServer.extraEnv }}
+            {{- range $key, $value := .Values.fineractServer.extraEnv }}
+            - name: {{ $key | quote}}
+              value: {{ $value | quote }}
+            {{- end }}
+            {{- end }}
+          ports:
+            - containerPort: 8080
+              name: http
+          readinessProbe:
```
BLasan, line 79: You can use separate paths to check liveness and readiness probes. We have added configurations in to spring boot. See https://github.com/apache/fineract/pull/1649

helm/fineract/templates/fineract-server-deployment.yml
```diff
@@ -0,0 +1,87 @@
...
+            - name: fineract_tenants_uid
+              value: {{ include "mysql_user" . }}
+            - name: FINERACT_DEFAULT_TENANTDB_UID
+              value: {{ include "mysql_user" . }}
+            {{ if .Values.fineractServer.extraEnv }}
+            {{- range $key, $value := .Values.fineractServer.extraEnv }}
+            - name: {{ $key | quote}}
+              value: {{ $value | quote }}
+            {{- end }}
+            {{- end }}
```
BLasan, line 75: I think you have missed this env.
```
        - name: FINERACT_DEFAULT_TENANTDB_PWD
          valueFrom:
            secretKeyRef:
              name: fineract-tenants-db-secret
              key: password
```

helm/fineract/values.yaml
BLasan: I don't know whether the following line would generate a proper random password. You can try this to generate the mysql password @Yann-J 

`password: {{ randAlphaNum 64 | quote }}`

Or else we can point a particular secret name to the get password. But in order to do that we need to create that secret first before installing helm chart
Refer: https://stackoverflow.com/questions/56170052/how-not-to-overwrite-randomly-generated-secrets-in-helm-templates

BLasan (changes requested):

helm/fineract/values.yaml
Yann-J: Hello there, sorry for my delay here!

We can't use `randAlphaNum` inside the chart because that would cause the value to be re-generated each time we deploy. However there's no problem to keep that value empty and implement a runtime warning to instruct users to provide a value.

I'm not so sure that "people will NOT change these" is really true any more with Helm since it's just so easy to do... but for extra protection it makes sense to force it, I can do that.

BLasan: Agree with that. I think there was the same issue in mysql deployment in local cluster. I've fixed that. This issue was raised in a previous PR https://github.com/apache/fineract/pull/1583

helm/fineract/templates/fineract-server-deployment.yml
Yann-J: I don't think so, this one is handled with:

```
          envFrom:
            - secretRef:
                name: {{ .Release.Name | quote }}
```

helm/fineract/templates/fineract-secret.yaml
```diff
@@ -0,0 +1,25 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+kind: Secret
+apiVersion: v1
+metadata:
+  name: {{ .Release.Name }}
```
BLasan, line 22: Use `{{ .Release.Name }}-secret`.

helm/fineract/templates/fineract-server-deployment.yml
```diff
@@ -0,0 +1,87 @@
...
+          - -wait
+          - tcp://{{ include "mysql_host" . }}:3306
+          - -timeout
+          - 300s
+      containers:
+        - name: fineract-backend
+          image: "{{ .Values.fineractServer.image.name }}:{{ .Values.fineractServer.image.tag }}"
+          envFrom:
+            - secretRef:
+                name: {{ .Release.Name | quote }}
```
BLasan, line 59: Use the secret name (this is bit confusing)

BLasan (changes requested):

helm/fineract/templates/fineract-server-deployment.yml
Yann-J: 👍 done!

helm/fineract/templates/fineract-secret.yaml
Yann-J: Actually, it's already Kind = Secret so I think that might be a bit redundant...

Yann-J:
OK, I've made an update that addresses most comments above (I think), in particular avoid setting default passwords, and raising an error if they're empty.

Some tuning along the way, in particular all extra variables (Secret or not) are now evaluated as templates - don't know if that's useful here but it doesn't hurt, and I've seen many cases where it is. They're also all supporting multiline values. Again, not sure if it's super useful here but you never know.

Yann-J:
> @BLasan have you noticed this PR? It looks like something you could be interested in helping to review... 😄
> 
> I was thinking about something when reviewing #1653: That (resource limits) should be done here as well? More importantly, we're now maintaining Kubernetes YAML in 2 places (original raw Kube YAML, and Helm chart templates). That's perhaps not ideal... you guys should figure out how you want to do this going forward? Keep both? Replace the original YAML with the Helm templates, and subsequently remove the originals? Up to you - work together.

Regarding the scripts vs Helm, I would say that Helm is probably the de-facto standard nowadays for deploying to k8s, but not everyone will have it, so it could make sense to keep supporting both.

However, I would say that in the long run, once it's quite stable, the Helm chart probably should be in a separate repo so it can be versioned independently from the application, and then published into https://artifacthub.io/

helm/fineract/values.yaml
Yann-J: Yeah... I would argue that the situation is a bit different in a Helm context since overriding parameters is dead simple (`helm install --set mysql.auth.password=...`), and really part of the Helm workflow, so the resistance is probably a bit lower there...

Anyway my latest commit is addressing this, and provides an error message in case the values aren't provided.

Yann-J:
Note that there would be a lot more we could do... in particular the next item on my list is to implement a `helm test` hook, as well as making the `mysql` dependency optional, in case you already have an external instance running.

This is really only a quickstart...

BLasan:
> Note that there would be a lot more we could do... in particular the next item on my list is to implement a `helm test` hook, as well as making the `mysql` dependency optional, in case you already have an external instance running.
> 
> This is really only a quickstart...



> > @BLasan have you noticed this PR? It looks like something you could be interested in helping to review... smile
> > I was thinking about something when reviewing #1653: That (resource limits) should be done here as well? More importantly, we're now maintaining Kubernetes YAML in 2 places (original raw Kube YAML, and Helm chart templates). That's perhaps not ideal... you guys should figure out how you want to do this going forward? Keep both? Replace the original YAML with the Helm templates, and subsequently remove the originals? Up to you - work together.
> 
> Regarding the scripts vs Helm, I would say that Helm is probably the de-facto standard nowadays for deploying to k8s, but not everyone will have it, so it could make sense to keep supporting both.
> 
> However, I would say that in the long run, once it's quite stable, the Helm chart probably should be in a separate repo so it can be versioned independently from the application, and then published into https://artifacthub.io/

+1 @vorburger I think this is ideal as we can maintain the helm chart implmentations separately and do releases accordingly

BLasan:
> Note that there would be a lot more we could do... in particular the next item on my list is to implement a `helm test` hook, as well as making the `mysql` dependency optional, in case you already have an external instance running.
> 
> This is really only a quickstart...

Great. Shall we wait until @vorburger and @edcable  's reviews ?

helm/fineract/templates/fineract-secret.yaml
BLasan: Yuh, but you have used the prefix `{{ .Release.Name }}` for other resources. So instead of having just the `{{ .Release.Name }}` I would suggest to use `{{ .Release.Name }}-secret` as the secret name. That'll be something meaningful. But it's not a requirement with high priority. Have you done a release for this implementation? Hope this works fine :) The configurations you've done look good to me

Yann-J:
Is there anything else you folks need from me on this one?

awasum:
@Yann-J  There are some Apache Rat failures in the build on Travis: https://travis-ci.com/github/apache/fineract/builds/220288308

You could either add the licenses or ignore the files. After that @vorburger may give his go ahead or more feedback...

PR #1650, opened 2021-03-01
FINERACT-982 - Completely ditch use of Drizzle JDBC Driver

xurror:
https://github.com/apache/fineract/pull/1366


xurror:
Cargo is failing to start. Probably cause it can't find the driver.

xurror:
I've added connector as an external library for cargo. It should be OK now.

xurror:
@ptuomola please can I have your support on this?
I've been trying a couple of hacks here with cargo. I managed to externalize our JDBC driver but then I noticed cargo works in containerized environment so I tried to add the jar to its classpath with a download task. I tested it on my local and it works but fails everytime with travis. Maybe you have an idea why it's happening.

xurror:
@ptuomola please can I have your support on this?
I've been trying a couple of hacks here with cargo. I managed to externalize our JDBC driver but then I noticed cargo works in containerized environment so I tried to add the jar to its classpath with a download task. I tested it on my local and it works but fails evrytime with travis.

PR #1882, opened 2021-09-23
FINERACT-982 - Completely ditch use of Drizzle JDBC Driver after all

IOhacker:
## Description

Changes were done for replacing the Drizzle JDBC Driver as per issue reported at FINERACT-982 - Completely ditch use of Drizzle JDBC Driver after all

### DIFF
--- a/fineract-provider/src/main/resources/sql/migrations/core_db/V119__add_template_table.sql
+++ b/fineract-provider/src/main/resources/sql/migrations/core_db/V119__add_template_table.sql
@@ -20,8 +20,8 @@
 DROP TABLE IF EXISTS `m_template`;
 CREATE TABLE `m_template` (
   `id` BIGINT NOT NULL AUTO_INCREMENT,
-  `name` varchar(255) COLLATE utf8mb4_unicode_ci NOT NULL,
-  `text` longtext COLLATE utf8mb4_unicode_ci NOT NULL,
+  `name` varchar(255) COLLATE utf8mb4_general_ci NOT NULL,
+  `text` longtext COLLATE utf8mb4_general_ci NOT NULL,
   `entity` INT DEFAULT NULL,
   `type` INT DEFAULT NULL,
   PRIMARY KEY (`id`),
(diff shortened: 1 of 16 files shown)


fineract-provider/src/main/resources/sql/migrations/core_db/V119__add_template_table.sql
```diff
@@ -31,9 +31,9 @@ CREATE TABLE `m_template` (
 DROP TABLE IF EXISTS `m_templatemappers`;
 CREATE TABLE `m_templatemappers` (
   `id` BIGINT NOT NULL AUTO_INCREMENT,
-  `mapperkey` varchar(255) COLLATE utf8mb4_unicode_ci DEFAULT NULL,
+  `mapperkey` varchar(255) COLLATE utf8mb4_general_ci DEFAULT NULL,
```
awasum, line 34: What happens to old deployments when u update and previous migration file? Can we instead update in a newer migration script?

awasum (changes requested):

IOhacker:
Hi Awasum,

I will fix it adding the update script. Also I have noticed an error while
running the integraron test in the Pentaho reports.

Regards

francisguchie:
@awasum , I have gone through the discussion on [FINERACT-982](https://issues.apache.org/jira/browse/FINERACT-982), does the PR cater for the non-tech users?

IOhacker:
@awasum 

This will work? (of course with the tables affected)

fineract/fineract-provider/src/main/resources/sql/migrations/core_db/Vxxx__alter_m_template_for_utf8mb4_general_ci.sql

Also I would like to propose to set this at DB level, like this

- ./fineract-db/server_collation.cnf:/etc/mysql/conf.d/server_collation.cnf

The file will contain:

fineract-db/server_collation.cnf 
[mysqld]
character_set_server=utf8mb4
collation_server=utf8mb4_general_ci

Any thoughs/suggestions?

IOhacker:
I will change the approach, not changing anything on the table collation. I will submit another PR

PR #1883, opened 2021-09-24
FINERACT-982 - Completely ditch use of Drizzle JDBC Driver after all

IOhacker:
## Description

Changes were done for replacing the Drizzle JDBC Driver as per issue reported at FINERACT-982 - Completely ditch use of Drizzle JDBC Driver after all. Now using MariaDB with charset utf8mb4 and collation utf8_unicode_ci

## Checklist

- [ X] Write the commit message as per https://github.com/apache/fineract/#pull-requests

- [ X] Acknowledge that we will not review PRs that are not passing the build _("green")_ - it is your responsibility to get a proposed PR to pass the build, not primarily the project's maintainers.

- [ X] Follow coding conventions at https://cwiki.apache.org/confluence/display/FINERACT/Coding+Conventions.

### DIFF
--- a/LICENSE_RELEASE
+++ b/LICENSE_RELEASE
@@ -243,9 +243,9 @@ This product bundles DOM4J, Copyright 2001-2005 (C) MetaStuff, Ltd.,
 which is available under a BSD license.
 For details, see licenses/binary/Dom4J.BSD
 
-This product bundles JDBC driver for MySQL and Drizzle v1.3, Copyright (c)
-2009-2013, Marcus Eriksson which is available under a BSD license.
-For details, see licenses/binary/Drizzle.BSD
+This product bundles JDBC driver for MariaDB v2.7.3, Copyright (c)
+which is available under a LGPL license.
+For details, see https://mariadb.com/kb/en/about-mariadb-connector-j/#license
 
 This product bundles Ical4J v1.0.4, Copyright (c) 2012, Ben Fortuna,
 which is available under a BSD license.
--- a/build.gradle
+++ b/build.gradle
@@ -115,7 +115,7 @@ allprojects  {
             dependency 'com.google.truth.extensions:truth-java8-extension:1.1.3'
             dependency 'org.apache.commons:commons-email:1.5'
             dependency 'commons-io:commons-io:2.11.0'
-            dependency 'org.drizzle.jdbc:drizzle-jdbc:1.4'
+            dependency 'org.mariadb.jdbc:mariadb-java-client:2.7.3'
             dependency 'com.github.librepdf:openpdf:1.3.26'
             dependency 'org.mnode.ical4j:ical4j:3.1.0'
             dependency 'org.quartz-scheduler:quartz:2.3.2'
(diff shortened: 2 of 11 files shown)


avikganguly01:
@IOhacker : Can you please test if MariaDB driver will work with AWS RDS (and if possible with Azure's managed MySQL database)?

IOhacker:
I can, or what about testing this JDBC Driver, which is for AWS
https://github.com/awslabs/aws-mysql-jdbc

LICENSE_RELEASE
```diff
@@ -243,9 +243,9 @@ This product bundles DOM4J, Copyright 2001-2005 (C) MetaStuff, Ltd.,
 which is available under a BSD license.
 For details, see licenses/binary/Dom4J.BSD
 
-This product bundles JDBC driver for MySQL and Drizzle v1.3, Copyright (c)
-2009-2013, Marcus Eriksson which is available under a BSD license.
-For details, see licenses/binary/Drizzle.BSD
+This product bundles JDBC driver for MariaDB v2.7.3, Copyright (c)
+which is available under a LGPL license.
+For details, see https://mariadb.com/kb/en/about-mariadb-connector-j/#license
```
ptuomola, line 249: If my understanding is correct, I don't think we are allowed to bundle anything with a LGPL license. We can make the JDBC driver a dependency so that is is downloaded at build / run time - that way the users can choose to use another, non-LGPL'd, JDBC driver. But we should not bundle anything in our release.

build.gradle
```diff
@@ -115,7 +115,7 @@ allprojects  {
             dependency 'com.google.truth.extensions:truth-java8-extension:1.1.3'
             dependency 'org.apache.commons:commons-email:1.5'
             dependency 'commons-io:commons-io:2.11.0'
-            dependency 'org.drizzle.jdbc:drizzle-jdbc:1.4'
+            dependency 'org.mariadb.jdbc:mariadb-java-client:2.7.3'
             dependency 'com.github.librepdf:openpdf:1.3.26'
```
ptuomola, line 119: Given that we can't bundle a LGPL driver, we need to find some way to a) use it as a dependency in the classpath but b) not bundle it in the release JAR. I'd suggest experimenting with the different gradle classpaths to do this (e.g. is it adding this to runtimeClasspath rather than implementationClasspath?)

ptuomola (changes requested):

build.gradle
avikganguly01: @IOhacker @ptuomola : In terms of bundling it as a dependency as an optional build time flag instead of putting it in the release JAR, there is a complication : -
There are places in the code where return type casting is based on Drizzle (search Drizzle in codebase). There is no consistency between JDBC drivers when it comes to some return types so this needs some work in those source files as well.

BLasan:
@IOhacker Please resolve the conflicts and make the requested changes by @avikganguly01 @ptuomola

build.gradle
ptuomola: @IOhacker @avikganguly01 I wonder if we need the equivalent code if we are using MariaDB? A lot of the code is to handle Drizzle's peculiarities re return types. Now OpenJPA supports MariaDB "out of the box" so there may not be any need to customise the JDBC behaviour. 

Also interesting... if we've dropped all dependencies to Drizzle but the code is doing casting based on the Drizzle classes, how does this still compile? I'd expect to get a compilation error.

ptuomola:
@IOhacker I think we should also update README.md to give people instructions to use MariaDB rather than MySQL, and .travis.yml to use MariaDB.

ptuomola:
Hi @IOhacker - will you have time to finish this one off? I can give it a go if you're busy - would be great to get this implemented...

ptuomola:
Changes incorporated in #1990

PR #1888, opened 2021-10-08
[FINERACT-1404] fix loan provisioning entry deduplication

jamesidw:
## Description

The previous/current approach to removing duplicates from the `LoanProductProvisioningEntry` collection generated is to check for equality between two objects and add the reserved amount of one to the other if they are the same. This doesn't take into account the possibility that the new total amount may make the object similar to another object in the collection. 

The change I'm proposing is to group the similar entries before considering the reserved amount. So if the only difference between two entries is the reserve amount, we should sum up their values. Any other entries that remain will have at least some other field that is different and shouldn't cause duplicates. 

See details in [FINERACT-1404](https://issues.apache.org/jira/browse/FINERACT-1404).

## Checklist


ptuomola (approved):
LGTM - definitely makes sense not to consider the amount as part of the similarity check, as this will result in strange behaviour...

PR #1934, opened 2021-10-26
FINERACT-1418: Upgrade to Jersey 2

ptuomola:
## Description

## Checklist

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -131,7 +131,6 @@ allprojects  {
             dependency 'jakarta.jms:jakarta.jms-api:2.0.3'
             dependency 'jakarta.xml.bind:jakarta.xml.bind-api:2.3.3'
             dependency 'org.glassfish.jaxb:jaxb-runtime:2.3.5'
-            dependency 'jakarta.validation:jakarta.validation-api:3.0.0'
             dependency 'org.apache.activemq:activemq-broker:5.16.3'
             dependency 'org.apache.bval:org.apache.bval.bundle:2.0.5'
 
@@ -191,16 +190,6 @@ allprojects  {
                 exclude 'pull-parser:pull-parser'
             }
 
-            dependencySet(group: 'com.sun.jersey', version: jerseyVersion) {
-                entry 'jersey-core'
-                entry 'jersey-servlet'
-                entry 'jersey-server'
-                entry 'jersey-json'
-            }
-            dependencySet(group: 'com.sun.jersey.contribs', version: jerseyVersion) {
-                entry 'jersey-spring'
-                entry 'jersey-multipart'
-            }
             dependencySet(group: 'org.apache.poi', version: '4.1.2') {
                 entry 'poi'
                 entry 'poi-ooxml'
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/data/ApiGlobalErrorResponse.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/data/ApiGlobalErrorResponse.java
@@ -18,10 +18,9 @@
  */
 package org.apache.fineract.infrastructure.core.data;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
 import java.util.ArrayList;
 import java.util.List;
-import javax.xml.bind.annotation.XmlElement;
-import javax.xml.bind.annotation.XmlElementWrapper;
 
 /**
  *
@@ -53,6 +52,7 @@ public class ApiGlobalErrorResponse {
      * A list of zero or more of the actual reasons for the HTTP error should they be needed. Typically used to express
      * data validation errors with parameters passed to API.
      */
+
     private List<ApiParameterError> errors = new ArrayList<>();
 
     public static ApiGlobalErrorResponse unAuthenticated() {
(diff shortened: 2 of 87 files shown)


ptuomola:
Found the missing line - now single calls work...

ptuomola:
Now trying to find some way to fix this classpath nightmare

ptuomola:
Ok - hopefully have now fixed the classpath issues. Let's see if this works.

vidakovic:
@ptuomola ... please check this for the Jersey configuration: https://github.com/FITER1/fineract-ng/blob/feature/01-migration-eclipse-link/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/boot/JerseyConfiguration.java

vidakovic:
I'll have also a look at the classpath issues... I remember that I fixed something similar by cleaning up all the dependencies (that was before the major improvements happened after version 1.3)... so, the Gradle file in that branch is very different from what we have now upstream, but maybe it contains part of the solution... just as a caveat.

ptuomola:
Thanks! I think I actually found a simpler solution: if you just do package() with the root package then all the classes get automatically found. 

Also fixed the classpath issues. Now just need to debug the failing integration tests...

ptuomola:
This now works for me locally - just need to see if the tests pass here as well.

vidakovic:
@ptuomola what was the problem? Anything I can help with?

ptuomola:
It all works for me locally - needed to rewrite the error handling for Interop as new Jersey did not serialise the HttpClientErrors correctly. Now test is failing with the Docker step - I'll try to see what's wrong with that tonight...

ptuomola:
Build finally passes! @vidakovic can you review and merge once ok? After that, I'll take a look at rewriting the Spring config

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/data/ApiGlobalErrorResponse.java
```diff
@@ -194,8 +194,7 @@ public ApiGlobalErrorResponse(final List<ApiParameterError> errors) {
         this.errors = errors;
     }
 
-    @XmlElementWrapper(name = "errors")
-    @XmlElement(name = "errorResponse")
+    @JsonProperty("errors")
```
vidakovic, line 197: @ptuomola isn't this changing the structure of the error response? As I see it from the changes here:

Before:

```
<errors>
  <errorResponse>
    ...
  </errorResponse>
  <errorResponse>
    ...
  </errorResponse>
</errors>
```

After:

```
<errors>
  <apiParameterError>
    ...
  </apiParameterError>
  <apiParameterError>
    ...
  </apiParameterError>
</errors>
```
... if I'm missing something here and the errors still serialize to the same structure then please ignore.

build.gradle
```diff
@@ -131,7 +131,6 @@ allprojects  {
             dependency 'jakarta.jms:jakarta.jms-api:2.0.3'
             dependency 'jakarta.xml.bind:jakarta.xml.bind-api:2.3.3'
             dependency 'org.glassfish.jaxb:jaxb-runtime:2.3.5'
-            dependency 'jakarta.validation:jakarta.validation-api:3.0.0'
```
vidakovic, line 134: @ptuomola Any reason why we would replace the newer Jakarta validation lib with the older javax.* stuff? This dependency results in a couple of import statement changes further down... unless there is no strict requirement by Jersey 2.x I would revert this change.

vidakovic:
@ptuomola ... only those 2 small comments above to resolve... I saw that you also replaced the ancient "ord.codehouse.jackson.*" references with the latest Jackson packages... great! The rest of your changes look good to me.

build.gradle
ptuomola: I tried first retaining the validation-api 3.0.0. The problem is that it is incompatible with about half a dozen libraries included in spring-boot-starter-jersey, and for many of which no version exists that is compatible for 3.0.0. So I had to backtrack and downgrade validation API to 2.x. 

I think overall it would make sense for us to try to remove as many of the specific library dependencies and instead use the Spring Boot bundles wherever they are available - that way we don't have to worry about compatibility of every individual library with everything else.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/data/ApiGlobalErrorResponse.java
ptuomola: This code is actually used to serialise to JSON, not to XML. And with the new version of Jackson,  it seems the behaviour was changed. In old Jackson, this would serialise to:

errors: [ "error1", "error2" ] 

ie take the name from the @XmlElementWrapper

In the new version of Jackson, the annotation @XmlElementWrapper is ignored, and instead the existing code serialises to:

errorResponse: [ "error1", "error2 ]

Which resulted in a lot of integration tests failing

So to restore the existing behaviour I changed the annotation to @JsonProperty (as this is anyway used for Json, not XML) - and for JSON there is no concept of a wrapper anyway. 

You can find discussion about this issue here:
https://github.com/FasterXML/jackson-modules-base/issues/45

Now if the same class is somewhere serialised to XML then I fully agree that this will introduce a regression. But I wasn't able to find any such cases - just the serialisation for JSON, for which this change actually retains the current behaviour...

vidakovic:
@ptuomola ... thanks for the clarifications... I agree with not micro managing the dependencies... let Jersey fix the dependencies, eventually they'll trickle down to us. Concerning the JAXB annotations: actually I'm glad that they disappeared and you used JsonProperty... and if there's no other references to that then the small structure change should be OK.

vidakovic (approved):
LGTM

PR #1984, opened 2021-11-25
FINERACT-1431:PassportDropDown-in-ClientIdentifier

rrpawar96:
## Description

## Checklist


francisguchie:
@rrpawar96 @ptuomola @awasum  this i have tested and works for me

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/ClientIdentifier.java
```diff
@@ -58,20 +62,29 @@
...
+
         final String documentKey = command.stringValueOfParameterNamed("documentKey");
         final String description = command.stringValueOfParameterNamed("description");
         final String status = command.stringValueOfParameterNamed("status");
-        return new ClientIdentifier(client, documentType, documentKey, status, description);
+        String documentTypeValue = null;
+        if (documentType.label().equals("Passport")) {
+            documentTypeValue = command.stringValueOfParameterNamed("documentTypeValue");
+            if ("".equals(documentTypeValue)) {
+                throw new PlatformDataIntegrityException("Selection of Country is Mandatory", "error.msg.selection.of.country.mandatory");
```
fynmanoj: Please move this validation to Validator class.

fynmanoj:
@rrpawar96 , @francisguchie as per the Screenshots attached to JIRA ticket the new field appears to be a dropdown. where does the dropdown values are fetched from? Also if the dropdown values are predefined, why are we not using ID instead of String Value ?

francisguchie:
@fynmanoj  
This list is going to be hard coded i presume the number of counties in the world does not change that much and can be easily found here https://www.worldometers.info/geography/alphabetical-list-of-countries/

fynmanoj:
@francisguchie isn't it a good idea to fetch  the predefined data set from the platform itself?, especially when it is a mandatory field? That would help others to upgrade effortlessly.

rrpawar96:
@fynmanoj , please re-review this PR, now we are fetching the countries list from the database so it can be modified directly from the database.
The documentTypeValue param is now used for passport(countries) and passed in String so that in future enhancement, the same param (documentTypeValue) can be used to pass values for dropdowns other than passports

fynmanoj:
@rrpawar96 , this looks fine. Thanks for that change.
Could you please move the validation from domain object to validator class?

 Also I am just curious,is there a specific reason for not using code-code value table for new drop down?

francisguchie:
> @rrpawar96 , this looks fine. Thanks for that change. Could you please move the validation from domain object to validator class?> Also I am just curious, is there a specific reason for not using code-code value table for new drop down?

@rrpawar96 have you addressed @fynmanoj  request. On my side the functional test is working fine

rrpawar96:
Hi @fynmanoj, @francisguchie  comment has been addressed now, validation has moved to validator class.
We didn't use a code-value table since it won't be seen as good to keep more than 250 lists of countries in the code value table which we use generally for all codes, that's why we kept the table separate.

fynmanoj:
@rrpawar96 , thanks for addressing the change request.

rrpawar96:
Travis fixed..

ptuomola:
Two comments from my side:

1. If the column documentTypeValue is meant to capture the issuing country for the document (which actually can make sense for all kinds of documents, not just passports), would it not make sense to call it that? So how about calling the column e.g. document_issue_country_id rather than document_type_value? Value seems to suggest it's the counterpart for the document_type_id, which is not the case...

2. Why do we need to create a new table for country codes? Can't we just treat it just like any other code value and use the existing code value infrastructure...

francisguchie:
> 2. Why do we need to create a new table for country codes? Can't we just treat it just like any other code value and use the existing code value infrastructure...

@ptuomola , kindly allow me to comment on having an extra table. 
Countries information is almost "static" changes are not made all the time. And if there is a new country, we can create a PR for it like we do for currencies. I think (i stand to be corrected) we get more validation / control options when we add an extra table.

francisguchie:
> 1. If the column documentTypeValue is meant to capture the issuing country for the document (which actually can make sense for all kinds of documents, not just passports), would it not make sense to call it that? So how about calling the column e.g. document_issue_country_id rather than document_type_value? Value seems to suggest it's the counterpart for the document_type_id, which is not the case...

@ptuomola & @rrpawar96 , 
The reason why only passport is needed for this function is that, as is in (Apache fineract) the validation is such that the document id is supposed to be unique in the database irrespective of the document selected. For example, if user selects drivers license, and enters 123456 for client X then later selects passport and enters 123456 then a validation errors shows up. 
because 123456 has been used already.

Come to reality- In Africa where i have worked most of my life, as countries are fighting to reduce the financial exclusion percentages **(some countries are at 80% financial exclusion)**, **new / stringent KYC rules are being put in place**. For example in East and West Africa the central banks are pushing for one document only and that is the National ID. All other IDs are not allowed as KYC documents. Meanwhile, for Foreigners, only the Passport is recognized when opening a "bank" account in any bank of the hosting country.

This flexibility of various documents is being reduced to just National Id and Passports

rrpawar96:
> 2. Why do we need to create a new table for country codes? Can't we just treat it just like any other code value and use the existing code value infrastructure...

Hi @ptuomola, we have not created codes for Country, we are just using class/objects of it to fetch the values from the new table.

rrpawar96:
> 1. If the column documentTypeValue is meant to capture the issuing country for the document (which actually can make sense for all kinds of documents, not just passports), would it not make sense to call it that? So how about calling the column e.g. document_issue_country_id rather than document_type_value? Value seems to suggest it's the counterpart for the document_type_id, which is not the case...

Comment has been addressed change the variableName from documentTypeValue to documentIssueCountry and not Id because we are saving countries name in String directly.

francisguchie:
> > 1. If the column documentTypeValue is meant to capture the issuing country for the document (which actually can make sense for all kinds of documents, not just passports), would it not make sense to call it that? So how about calling the column e.g. document_issue_country_id rather than document_type_value? Value seems to suggest it's the counterpart for the document_type_id, which is not the case...
> 
> Comment has been addressed change the variableName from documentTypeValue to documentIssueCountry and not Id because we are saving countries name in String directly.

@ptuomola  Kindly review

rrpawar96:
> 2. Why do we need to create a new table for country codes? Can't we just treat it just like any other code value and use the existing code value infrastructure...

@ptuomola, Please review, Comment has been addressed, the existed code-value infrastructure is been used.

francisguchie:
@fynmanoj , @ptuomola and @vidakovic  kindly help on reviewing this PR

vidakovic:
@rrpawar96 @francisguchie ... most of it looks ok for me... but the database migration scripts need to be rewritten for Liquibase... Flyway not used anymore.

francisguchie:
@vidakovic you so much, @rrpawar96  please address this. Reach out to me if you need my help

rrpawar96:
@galovics @vidakovic  Please review the migration script written for Liquibase.

francisguchie:
> @rrpawar96 @francisguchie ... most of it looks ok for me... but the database migration scripts need to be rewritten for Liquibase... Flyway not used anymore.

@vidakovic have you had the time to look through and review changes made by @rrpawar96

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/command/ClientIdentifierCommand.java
```diff
@@ -50,18 +54,30 @@ public String getDocumentKey() {
...
 
         final DataValidatorBuilder baseDataValidator = new DataValidatorBuilder(dataValidationErrors).resource("clientIdentifier");
 
         baseDataValidator.reset().parameter("documentTypeId").value(this.documentTypeId).notNull().integerGreaterThanZero();
         baseDataValidator.reset().parameter("documentKey").value(this.documentKey).notBlank();
 
+        final Long documentTypeId = command.longValueOfParameterNamed("documentTypeId");
+        if (documentTypeId == 1) {
+            final String documentIssueCountry = command.stringValueOfParameterNamed("documentIssueCountry");
+            if ("".equals(documentIssueCountry)) {
```
galovics: StringUtils.isBlank could make more sense here. Thoughts?

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/command/ClientIdentifierCommand.java
```diff
@@ -50,18 +54,30 @@ public String getDocumentKey() {
...
+    public void validateForCreate(final JsonCommand command) {
         final List<ApiParameterError> dataValidationErrors = new ArrayList<>();
 
         final DataValidatorBuilder baseDataValidator = new DataValidatorBuilder(dataValidationErrors).resource("clientIdentifier");
 
         baseDataValidator.reset().parameter("documentTypeId").value(this.documentTypeId).notNull().integerGreaterThanZero();
         baseDataValidator.reset().parameter("documentKey").value(this.documentKey).notBlank();
 
+        final Long documentTypeId = command.longValueOfParameterNamed("documentTypeId");
+        if (documentTypeId == 1) {
```
galovics: Can we somehow make this documentTypeId a non-magic number? Like what does 1 represent here?

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/ClientIdentifier.java
```diff
@@ -61,17 +64,24 @@ public static ClientIdentifier fromJson(final Client client, final CodeValue doc
         final String documentKey = command.stringValueOfParameterNamed("documentKey");
         final String description = command.stringValueOfParameterNamed("description");
         final String status = command.stringValueOfParameterNamed("status");
-        return new ClientIdentifier(client, documentType, documentKey, status, description);
+
+        String documentIssueCountry = null;
+        if (documentType.label().equals("Passport")) {
```
galovics: Any chance we can extract this magic String to somewhere?

fineract-provider/src/main/resources/db/changelog/tenant/parts/0006_client_identifier_passport_countrieslist.xml
```diff
@@ -0,0 +1,1901 @@
...
+    </changeSet>
+    <changeSet author="fineract" id="2">
+        <insert tableName="m_code">
+            <column name="code_name" value="Countries"/>
+            <column name="is_system_defined" valueBoolean="true"/>
+        </insert>
+    </changeSet>
+    <changeSet author="fineract" id="3">
+        <insert tableName="m_code_value">
+            <column name="code_id" valueNumeric="35"/>
```
galovics: I assume this refers to the inserted m_code row right before this changeset. If so, then let's define the id of that to make sure it really gets id 35.

galovics (changes requested):

francisguchie:
@rrpawar96 @galovics  I see the fineract gradle build PostgreSQL is failing. What is the remedy here?

rrpawar96:
@francisguchie @galovics, after running the latest build on an empty database, I have noticed one thing is that there are no permissions created in the m_permission table, probably once we get that permission data should fix this test case as well.
I am not quite sure about this but will be better if @galovics confirms the m_permission table data issue.

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/command/ClientIdentifierCommand.java
```diff
@@ -50,18 +57,29 @@ public String getDocumentKey() {
...
-    public void validateForCreate() {
+    public void validateForCreate(final CodeValue documentType, final JsonCommand command) {
         final List<ApiParameterError> dataValidationErrors = new ArrayList<>();
 
         final DataValidatorBuilder baseDataValidator = new DataValidatorBuilder(dataValidationErrors).resource("clientIdentifier");
 
         baseDataValidator.reset().parameter("documentTypeId").value(this.documentTypeId).notNull().integerGreaterThanZero();
         baseDataValidator.reset().parameter("documentKey").value(this.documentKey).notBlank();
 
+        if (documentType.label().equals(ClientApiConstants.clientIdentifierPassportParamName.toString())) {
```
galovics: Is there a need to actually call the toString() on that constant? It would also even better if we could switch the order of the equality check to be null-safe.
Thoughts?

fineract-provider/src/main/resources/db/changelog/tenant/parts/0002_initial_data.xml
```diff
@@ -1916,7 +1916,7 @@
             <column name="code" value="CRC"/>
             <column name="decimal_places" valueNumeric="2"/>
             <column name="currency_multiplesof"/>
-            <column name="display_symbol" value="₡"/>
+            <column name="display_symbol"/>
```
galovics: Let's not change anything in the existing changelog files. If you would like to remove the display_symbol for these currencies, let's add a new changelog file with the changes needed.
Btw, any reason you decided to remove these?

fineract-provider/src/main/resources/db/changelog/tenant/parts/0002_initial_data.xml
```diff
@@ -2204,7 +2204,7 @@
             <column name="code" value="INR"/>
             <column name="decimal_places" valueNumeric="2"/>
             <column name="currency_multiplesof"/>
-            <column name="display_symbol" value="₹"/>
+            <column name="display_symbol"/>
```
galovics: Same as above.

fineract-provider/src/main/resources/db/changelog/tenant/parts/0002_initial_data.xml
```diff
@@ -2672,7 +2672,7 @@
             <column name="code" value="PYG"/>
             <column name="decimal_places" valueNumeric="0"/>
             <column name="currency_multiplesof"/>
-            <column name="display_symbol" value="₲"/>
+            <column name="display_symbol"/>
```
galovics: Same as above.

galovics (changes requested):

galovics:
@rrpawar96 Judging by the test error in the console, it's failing due to the fact that you removed some displaySymbols:
```
java.lang.NullPointerException: Cannot invoke "String.equals(Object)" because "this.displaySymbol" is null
      at org.apache.fineract.integrationtests.CurrenciesTest.testUpdateCurrencySelection(CurrenciesTest.java:92)
```

rrpawar96:
Hi @galovics, there is a character set issue while deploying the latest branch in local. So I was made the changes in the display_symbol table.  Now created new ticket for it https://issues.apache.org/jira/browse/FINERACT-1516
Once the build passes I will create a new PR for this ticket.

galovics:
@rrpawar96 are you using MySQL locally? If so, make sure you set up collation for the server in its entirety. The Liquibase changeset you introduce will break existing develop deployments since you've changed the migration script in the middle. Plus the change you introduced is only MySQL compatible.

Please try to set up the collation for your MySQL installation first.

rrpawar96:
> @rrpawar96 are you using MySQL locally? If so, make sure you set up collation for the server in its entirety. The Liquibase changeset you introduce will break existing develop deployments since you've changed the migration script in the middle. Plus the change you introduced is only MySQL compatible.
> 
> Please try to set up the collation for your MySQL installation first.

Thanks.

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/command/ClientIdentifierCommand.java
rrpawar96: Addressed,

galovics:
@rrpawar96 I just looked at the last changes and I still have the question standing. Don't you think it would be better to use country ISO codes instead of the actual English country names in the DB?
My idea would be, store the ISO country codes in the DB and create a properties file with mapping the ISO country code to an English country name - if needed at all.
What do you think?

rrpawar96:
> @rrpawar96 I just looked at the last changes and I still have the question standing. Don't you think it would be better to use country ISO codes instead of the actual English country names in the DB? My idea would be, store the ISO country codes in the DB and create a properties file with mapping the ISO country code to an English country name - if needed at all. What do you think?

Hi @galovics, comment addressed, stored country in codes, and used Locale Library for conversion.

rrpawar96:
Hi @galovics, I have created the countries.properties file for mapping but using it in this way, is taking 3 secs of delay for loading the countries' names on UI, wouldn't be fine if we consider taking direct English country names?

galovics:
Hi @rrpawar96 , yeah, I can see why. You're going to read the properties file for each SQL query, that's definitely not the way.
What we should do instead:

* Upon application startup, load the properties file and keep the mapping in memory
* Upon the need to translate country codes to country names, use the mapping in the memory to resolve the country name

Also, please make sure we're following the ResourceBundle naming convention, i.e. let's call the file `countries_en.properties` instead of `countries.properties`.

Note: for the country translation, you could introduce a brand new service class to deal with the preloading and doing the actual translation.

fineract-provider/src/main/resources/countries.properties
```diff
@@ -0,0 +1,245 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+AF=AFGHANISTAN
```
galovics, line 20: Doesn't it make sense to not have the country names in full uppercase but as a readable format, like `Albania` instead of `ALBANIA`?

francisguchie:
> Hi @rrpawar96 , yeah, I can see why. You're going to read the properties file for each SQL query, that's definitely not the way. What we should do instead:
> 
> * Upon application startup, load the properties file and keep the mapping in memory
> * Upon the need to translate country codes to country names, use the mapping in the memory to resolve the country name
> 
> Also, please make sure we're following the ResourceBundle naming convention, i.e. let's call the file `countries_en.properties` instead of `countries.properties`.
> 
> Note: for the country translation, you could introduce a brand new service class to deal with the preloading and doing the actual translation.


@galovics  
As a lay man, who might be using apache-fineract in a place with little to no internet, will this kind of complex update be able to work without internet?

galovics:
@francisguchie would you elaborate more on this please? I'm not really sure I'm able to follow the "no internet" aspect of this change or Fineract for that matter.

rrpawar96:
closing this PR, opened up a new cleaned PR: https://github.com/apache/fineract/pull/2190

PR #1986, opened 2021-11-28
FINERACT-1446: Docker image build with Jib and JDK 17 fixes

vidakovic:
## Description

- proper Docker image build with reasonable startup parameters (the ones that were in place until now were not really great)
- using the latest JDK 17 LTS release everywhere (Docker image and workflows)
- Travis builds were failing, some regressions fell through the cracks (the biggest: GSON serializer was missing an adapter; without it was crashing with security exceptions as reflection is not generally allowed anymore); see: "org.apache.fineract.infrastructure.core.serialization.GoogleGsonSerializerHelper"
- moved the WAR build into a separate module; it's really unnecessarily hard to make everything work in one build script without making it overly complex; main reason for move: Jib (the Docker build module) detects if a project is of type WAR and starts to do all kind of automatic WAR related configuration; the WAR stuff is not our main deliverable; maybe we can drop it in the future; anyway... still there, just different directory
- I've made some small adjustments in the docker-compose.yml file; this thing is not used anymore for building the Docker images (with this change Dockerfile, entrypoint.sh and .dockerignore also disappeared)
- the integration tests run now with Docker Compose instead of Cargo; no more brittle Mysql configuration and startup; this will run no matter the OS or the environment; now there is only one mechanism that handles everything (via a Docker Compose Gradle plugin)
- with the move to JDK 17 the Travis builds were not working anymore; the main reason was missing JVM arguments in the gradle.properties to make the Gradle Spotless plugin properly work

## Checklist


.travis.yml
```diff
@@ -57,13 +48,8 @@ install:
...
   - sudo apt-get -q update
-  - sudo apt-get -yq install zulu17-jdk
-  - export JAVA_HOME=/usr/lib/jvm/zulu17
-  - echo "USE mysql;\nALTER USER 'root'@'localhost' IDENTIFIED BY 'mysql';\n" | mysql -u root
-  - mysql -u root -pmysql -e 'CREATE DATABASE IF NOT EXISTS `fineract_tenants`;'
-  - mysql -u root -pmysql -e 'CREATE DATABASE IF NOT EXISTS `fineract_default`;'
-# Hardcoding the time zone is a temporary fix for https://issues.apache.org/jira/browse/FINERACT-723
-  - export TZ=Asia/Kolkata
+  - sudo apt-get -yq install zulu11-jdk
+  - export JAVA_HOME=/usr/lib/jvm/zulu11
```
ptuomola, line 53: This looks like it's still pointing to JDK 11?

fineract-provider/build.gradle
```diff
@@ -274,10 +252,50 @@ bootJar {
...
+            '-XshowSettings:vm',
+            '-XX:+UseContainerSupport',
+            '-XX:+UseStringDeduplication',
+            '-XX:MinRAMPercentage=25',
+            '-XX:MaxRAMPercentage=80'
+        ]
+        args = [
+            '-Dspring.profiles.active=basicauth',
+            '-Dfile.encoding=UTF-8',
+            '-Duser.timezone=UTC',
```
ptuomola, line 284: At least previously the integration tests required timezone Asia/Kalkuta to pass... does this work?

fineract-provider/build.gradle
```diff
@@ -274,10 +252,50 @@ bootJar {
...
+            '-Xmx1G',
+            '-Xms1G',
+            '-XshowSettings:vm',
+            '-XX:+UseContainerSupport',
+            '-XX:+UseStringDeduplication',
+            '-XX:MinRAMPercentage=25',
+            '-XX:MaxRAMPercentage=80'
+        ]
+        args = [
+            '-Dspring.profiles.active=basicauth',
```
ptuomola, line 282: Could we make this env parameter for the image, but default to basicauth? That way we could use the same image for all the different profiles (twofactor, oauth etc) and also run tests for each of them?

fineract-provider/build.gradle
```diff
@@ -289,7 +307,6 @@ distributions {
         distributionBaseName = "apache-fineract-binary"
         contents {
             from bootJar
-            from war
```
ptuomola, line 292: Out of interest, does the distribution task actually work for you? I was yesterday working on another PR (automate updating of all the dependent licenses into the distribution) but could not get distTar etc to do anything.

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/AccountNumberGenerator.java
```diff
@@ -45,6 +48,12 @@
     private static final String LOAN_PRODUCT_SHORT_NAME = "loanProductShortName";
     private static final String SAVINGS_PRODUCT_SHORT_NAME = "savingsProductShortName";
     private static final String SHARE_PRODUCT_SHORT_NAME = "sharesProductShortName";
+    private final ConfigurationReadPlatformService configurationReadPlatformService;
+
+    @Autowired
+    public AccountNumberGenerator(final ConfigurationReadPlatformService configurationReadPlatformService) {
+        this.configurationReadPlatformService = configurationReadPlatformService;
+    }
```
ptuomola: I think these changes are from another PR that was already merged? Can you rebase your PR to develop?

ptuomola (changes requested):
This looks great! And definitely a great improvement to get rid of Cargo and actually test our "real" distribution. Provided some comments - also will look at the logs to see what sort of output we get now that we're running the tests against a Docker container...

.travis.yml
vidakovic: ah... didn't see that... that stuff is not necessary at all

fineract-provider/build.gradle
vidakovic: Well... we should override such test specific settings in the Docker Compose file now... UTC as a default makes sense I think... users need to override this one anyway in prod.
In terms of unit tests: I think less than a handful depend on the timezone.

fineract-provider/build.gradle
vidakovic: You got me ;-) ... that one would need rework too if we want to maintain the WAR files. I have no sense how often the WAR files are still used. I'd argue that it's outdated and that those who want it could easily build from distribution sources. But that's just my 2 cents. If a wider audience still needs this then I'll figure something out.

fineract-provider/build.gradle
ptuomola: Yes - I didn't mean changing the default if it's possible for us to override it for the integration tests. Unfortunately I think there's around 20-30 tests that fail at different times of day (some before midnight, some after midnight etc) if the timezone is not set to Asia/Kalkuta

fineract-provider/build.gradle
ptuomola: Sorry - I didn't mean we should distribute the WAR file - I don't think that's necessary. But when I tried to use distTar I could not get it to do anything at all - not even build a tar file with the bootJar in it. Does it work for you?

vidakovic:
... not sure if my comment made it to GH... so here again:

We can do that (environment variables to configure basic auth vs. oauth)... could we just do this in another iteration/PR? Just to avoid piling up too many changes here. I'm working in parallel with @hiteshprajapati on a follow up PR that will take full advantage of your improvements (aka Java config). I promise that this will cover your request... the build time copy and paste of different application.properties files is anyway not a best practice.

.travis.yml
vidakovic: Fixed

fineract-provider/build.gradle
vidakovic: Alright... I'll set this in the Docker Compose file, no problem.

vidakovic: Done. The complete entry point is now available in the Docker Compose file. Serves also as an example for other configuration (e. g. mem reservation etc.)

fineract-provider/build.gradle
vidakovic: Ah, that... it should do something... good catch, I'll look at that.

vidakovic: ... actually it still does something (even creates the SHA512 file). The command is:

```
./gradlew -Pfineract.release clean build binaryDistTar
```

My plan was to replace that stitched together process with https://jreleaser.org/ ... less moving parts, proper Gradle plugin.

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/AccountNumberGenerator.java
vidakovic: Done.

fineract-provider/build.gradle
ptuomola: Ok great - let me try that command, I was assuming the build would automatically be taken care of as a dependency. Will update the README if it works.

vidakovic:
@ptuomola ... well... step by step sometimes doesn't work. Docker Compose on Github Actions is just not working, so I replaced the integration test setup as well (which was anyway my ultimate goal):

image

Locally it's already working like a charm. My stretch goal of running tests in parallel doesn't work just yet, but that wasn't there with Tomcat Cargo either, so no problem.

My changes are quite minimal (just added one test base class and one annotation per test class). This solution is using Testcontainers.

Let's see if Github Actions runs this now (it should; Testcontainers are also on GH and run their tests there).

So much easier now...

ptuomola:
@vidakovic I know you've been busy, but there's lots of great stuff here - hope you have some time to get these ready for merging! Perhaps it would be possible to break this to smaller incremental fixes - e.g. one PR for the Java 17 support, another one to move WAR to a separate module, another PR to fix the Docker parameters etc...?

vidakovic:
@ptuomola I'll take care of this today and tomorrow... will take advantage that other things are calming down... at least for a bit :-) ... will try to break it up into smaller pieces.

vidakovic:
This PR is obsolete and was split up in smaller chunks.

PR #1994, opened 2021-12-04
Implement integration tests for twofactor profile

ptuomola:
## Description

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/configuration/service/ExternalServicesPropertiesReadPlatformServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/configuration/service/ExternalServicesPropertiesReadPlatformServiceImpl.java
@@ -83,7 +83,7 @@ public class ExternalServicesPropertiesReadPlatformServiceImpl implements Extern
 
                 if (ExternalServicesConstants.SMTP_USERNAME.equalsIgnoreCase(name)) {
                     username = value;
-                } else if (ExternalServicesConstants.SMTP_PORT.equalsIgnoreCase(port)) {
+                } else if (ExternalServicesConstants.SMTP_PORT.equalsIgnoreCase(name)) {
                     port = value;
                 } else if (ExternalServicesConstants.SMTP_PASSWORD.equalsIgnoreCase(name)) {
                     password = value;
--- a/fineract-provider/src/main/resources/sql/migrations/core_db/V265__modify_external_service_schema.sql
+++ b/fineract-provider/src/main/resources/sql/migrations/core_db/V265__modify_external_service_schema.sql
@@ -53,9 +53,9 @@ insert into c_external_service_properties (`name`, `value`, `external_service_id
 
 insert into c_external_service_properties (`name`, `value`, `external_service_id`) values('password', 'support80', (select id from c_external_service where name = 'SMTP_Email_Account'));
 
-insert into c_external_service_properties (`name`, `value`, `external_service_id`) values('host', 'smtp.gmail.com', (select id from c_external_service where name = 'SMTP_Email_Account'));
+insert into c_external_service_properties (`name`, `value`, `external_service_id`) values('host', 'localhost', (select id from c_external_service where name = 'SMTP_Email_Account'));
 
-insert into c_external_service_properties (`name`, `value`, `external_service_id`) values('port', '587', (select id from c_external_service where name = 'SMTP_Email_Account'));
+insert into c_external_service_properties (`name`, `value`, `external_service_id`) values('port', '3025', (select id from c_external_service where name = 'SMTP_Email_Account'));
 
 insert into c_external_service_properties (`name`, `value`, `external_service_id`) values('useTLS', 'true', (select id from c_external_service where name = 'SMTP_Email_Account'));
 
--- a/twofactor-tests/build.gradle
+++ b/twofactor-tests/build.gradle
@@ -1,0 +1,67 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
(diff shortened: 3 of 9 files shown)


fineract-provider/src/main/java/org/apache/fineract/infrastructure/configuration/service/ExternalServicesPropertiesReadPlatformServiceImpl.java
```diff
@@ -83,7 +83,7 @@ public SMTPCredentialsData extractData(final ResultSet rs) throws SQLException,
 
                 if (ExternalServicesConstants.SMTP_USERNAME.equalsIgnoreCase(name)) {
                     username = value;
-                } else if (ExternalServicesConstants.SMTP_PORT.equalsIgnoreCase(port)) {
+                } else if (ExternalServicesConstants.SMTP_PORT.equalsIgnoreCase(name)) {
```
vorburger, line 86: @ptuomola this change looks curious - is this intentional?

vorburger: sorry yeah no this is obviously right (it makes more sense looking at the full file then the diff here), nice catch!

fineract-provider/src/main/resources/sql/migrations/core_db/V265__modify_external_service_schema.sql
```diff
@@ -53,9 +53,9 @@ insert into c_external_service_properties (`name`, `value`, `external_service_id
 
 insert into c_external_service_properties (`name`, `value`, `external_service_id`) values('password', 'support80', (select id from c_external_service where name = 'SMTP_Email_Account'));
 
-insert into c_external_service_properties (`name`, `value`, `external_service_id`) values('host', 'smtp.gmail.com', (select id from c_external_service where name = 'SMTP_Email_Account'));
+insert into c_external_service_properties (`name`, `value`, `external_service_id`) values('host', 'localhost', (select id from c_external_service where name = 'SMTP_Email_Account'));
 
-insert into c_external_service_properties (`name`, `value`, `external_service_id`) values('port', '587', (select id from c_external_service where name = 'SMTP_Email_Account'));
+insert into c_external_service_properties (`name`, `value`, `external_service_id`) values('port', '3025', (select id from c_external_service where name = 'SMTP_Email_Account'));
```
vorburger, line 58: Traditionally we tried to avoid changing existing migration files (because then the DB after a migration and a fresh install is not the same anymore) so perhaps you would consider putting this into a new file?

vorburger:
Looks like this implements https://issues.apache.org/jira/browse/FINERACT-1335! NICE.

twofactor-tests/build.gradle
```diff
@@ -0,0 +1,67 @@
+/**
```
vorburger, line 1: copy/paste of https://github.com/apache/fineract/blob/develop/integration-tests/build.gradle ... I don't want to be a PITA, but obviously have to ask if there could be a nice way with a bit of Gradle foo to avoid this?

twofactor-tests/src/test/java/org/apache/fineract/twofactortests/common/Utils.java
```diff
@@ -0,0 +1,267 @@
+/**
```
vorburger: Copy/paste of https://github.com/apache/fineract/blob/develop/integration-tests/src/test/java/org/apache/fineract/integrationtests/common/Utils.java ... hm. Too much of a PITA to centralized this into a common integration-test-lib kinda module, maybe? Or... just not write the `TwoFactorAuthenticationTest` with this but e.g. using the `fineract-client`?

fineract-provider/src/main/resources/sql/migrations/core_db/V265__modify_external_service_schema.sql
ptuomola: The reason why I chose this "unorthodox" approach is that we don't want to overwrite people's existing SMTP configurations. We only want to change the default for new systems that is used in testing. So we don't want to migrate/change any existing setups (as adding a new migration script would do), we want to just change the value that is used when setting up a new system...

Another approach would be to have some custom SQL overwriting the initial values in testing. But as no one should be using the initial values anyway in their systems, why not just change them...

twofactor-tests/build.gradle
ptuomola: I tried to initially do something along those lines i.e. have one master cargo task and then inherit from that. I couldn't get it to work so thought it's easier just to have two parallel projects with no dependencies between them. But I'm not very good with Gradle so very happy if someone can do something more elegant...

twofactor-tests/src/test/java/org/apache/fineract/twofactortests/common/Utils.java
ptuomola: I'm only using two methods from that so can inline them into the other class and delete this. 

I agree that it would be good to move away from restassured etc.  But I haven't spent much time with fineract-client yet, and just wanted to quickly come up with something that stops us from accidentally breaking non-basicauth profiles - as we had recently done without noticing. Very happy if someone wants to do this better ie with fineract-client etc...

vorburger: SGTM. Do you want to do that (inline) and I'll merge this?

twofactor-tests/build.gradle
vorburger: SGTM. Maybe this is a clean up that e.g. @vidakovic who has great Gradle foo ;) wants to do as a follow up after this is merged.

fineract-provider/src/main/resources/sql/migrations/core_db/V265__modify_external_service_schema.sql
vorburger: SGTM, then we'll do this as an "exception to the rule".

twofactor-tests/src/test/java/org/apache/fineract/twofactortests/common/Utils.java
ptuomola: Done... took me a bit longer than I expected, as I found a time rounding error in the code as well

PR #2002, opened 2021-12-25
FINERACT-1393:RandomAccountNumber

rrpawar96:
## Description

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/AccountNumberGenerator.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/AccountNumberGenerator.java
@@ -18,8 +18,10 @@
  */
 package org.apache.fineract.portfolio.client.domain;
 
+import java.math.BigInteger;
 import java.util.HashMap;
 import java.util.Map;
+import org.apache.commons.lang3.RandomStringUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.fineract.infrastructure.accountnumberformat.domain.AccountNumberFormat;
 import org.apache.fineract.infrastructure.accountnumberformat.domain.AccountNumberFormatEnumerations.AccountNumberPrefixType;
@@ -29,7 +31,9 @@ import org.apache.fineract.infrastructure.configuration.data.GlobalConfiguration
 import org.apache.fineract.infrastructure.configuration.service.ConfigurationReadPlatformService;
 import org.apache.fineract.portfolio.group.domain.Group;
 import org.apache.fineract.portfolio.loanaccount.domain.Loan;
+import org.apache.fineract.portfolio.loanaccount.domain.LoanRepository;
 import org.apache.fineract.portfolio.savings.domain.SavingsAccount;
+import org.apache.fineract.portfolio.savings.domain.SavingsAccountRepository;
 import org.apache.fineract.portfolio.shareaccounts.domain.ShareAccount;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
@@ -44,6 +48,7 @@ public class AccountNumberGenerator {
     private static final int maxLength = 9;
 
     private static final String ID = "id";
+    private static final String ENTITY_TYPE = "entityType";
     private static final String CLIENT_TYPE = "clientType";
     private static final String OFFICE_NAME = "officeName";
     private static final String LOAN_PRODUCT_SHORT_NAME = "loanProductShortName";
@@ -52,18 +57,26 @@ public class AccountNumberGenerator {
     private static final String PREFIX_SHORT_NAME = "prefixShortName";
     private final AccountNumberFormatRepository accountNumberFormatRepository;
     private final ConfigurationReadPlatformService configurationReadPlatformService;
+    private final ClientRepository clientRepository;
+    private final LoanRepository loanRepository;
+    private final SavingsAccountRepository savingsAccountRepository;
 
     @Autowired
     public AccountNumberGenerator(final ConfigurationReadPlatformService configurationReadPlatformService,
-            final AccountNumberFormatRepository accountNumberFormatRepository) {
+            final AccountNumberFormatRepository accountNumberFormatRepository, final ClientRepository clientRepository,
+            final LoanRepository loanRepository, final SavingsAccountRepository savingsAccountRepository) {
         this.configurationReadPlatformService = configurationReadPlatformService;
         this.accountNumberFormatRepository = accountNumberFormatRepository;
+        this.clientRepository = clientRepository;
+        this.loanRepository = loanRepository;
+        this.savingsAccountRepository = savingsAccountRepository;
     }
 
     public String generate(Client client, AccountNumberFormat accountNumberFormat) {
         Map<String, String> propertyMap = new HashMap<>();
         propertyMap.put(ID, client.getId().toString());
         propertyMap.put(OFFICE_NAME, client.getOffice().getName());
+        propertyMap.put(ENTITY_TYPE, "client");
         CodeValue clientType = client.clientType();
         if (clientType != null) {
             propertyMap.put(CLIENT_TYPE, clientType.label());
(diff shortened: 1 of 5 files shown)


fineract-provider/src/main/resources/sql/migrations/core_db/V379__configuration_for_random_account_number.sql
```diff
@@ -0,0 +1,20 @@
...
+--
+-- Unless required by applicable law or agreed to in writing,
+-- software distributed under the License is distributed on an
+-- "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+-- KIND, either express or implied. See the License for the
+-- specific language governing permissions and limitations
+-- under the License.
+--
+
+INSERT INTO c_configuration ( name, description) VALUES ( 'random-account-number', 'if enabled, the savings account will be created with Random Savings Account Number');
```
ptuomola: This only talks about savings accounts but the code seems to impact clients and loans as well?

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanRepository.java
```diff
@@ -67,10 +67,12 @@
 
     String DOES_PRODUCT_HAVE_NON_CLOSED_LOANS = "select case when (count (loan) > 0) then 'true' else 'false' end from Loan loan where loan.loanProduct.id = :productId and loan.loanStatus in (100,200,300,303,304,700)";
 
-    String FIND_BY_ACCOUNT_NUMBER = "select loan from Loan loan where loan.accountNumber = :accountNumber and loan.loanStatus in (100,200,300,303,304)";
+    String FIND_NON_CLOSED_BY_ACCOUNT_NUMBER = "select loan from Loan loan where loan.accountNumber = :accountNumber and loan.loanStatus in (100,200,300,303,304)";
```
ptuomola, line 70: So this allows reusing an account number from a closed loan. Is that really OK? Would it not make more sense to ensure loan account numbers are unique as well?

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/AccountNumberGenerator.java
```diff
@@ -108,7 +124,14 @@ private String generateAccountNumber(Map<String, String> propertyMap, AccountNum
             }
         }
 
-        String accountNumber = StringUtils.leftPad(propertyMap.get(ID), accountMaxLength, '0');
+        final GlobalConfigurationPropertyData randomAccountNumber = this.configurationReadPlatformService
+                .retrieveGlobalConfiguration("random-account-number");
```
ptuomola, line 128: Does it make sense to have this as a single global configuration property so that then all numbers are random (client, loan, savings account)? Or would it make more sense to store this as part of the AccountNumberFormat as a parameter so that people can enable this for different entities if they'd like to? Then we could also store the length of the random number to be created as part of the AccountNumberFormat.

ptuomola (changes requested):

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/AccountNumberGenerator.java
```diff
@@ -154,6 +182,36 @@ private String generateAccountNumber(Map<String, String> propertyMap, AccountNum
...
+    private String randomNumberGenerator(int accountMaxLength, Map<String, String> propertyMap) {
+        String randomNumber = RandomStringUtils.random(accountMaxLength, false, true);
+        String accountNumber = randomNumber.substring(0, accountMaxLength);
+        BigInteger accNumber = new BigInteger(accountNumber);
+
+        if (!accNumber.equals(BigInteger.ZERO)) { // to avoid account no. 00 in randomisation
+            String entityType = propertyMap.get(ENTITY_TYPE);
+            if (entityType.equals("client")) { // avoid duplication it will loop until it finds new random account no.
+
+                Client client = this.clientRepository.getClientByAccountNumber(accountNumber);
```
ptuomola: What if the account numbers have prefixes / shortnames added to them? Then this will never return a match even though there may be matching ones after we append the prefix...

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/AccountNumberGenerator.java
```diff
@@ -143,8 +166,13 @@ private String generateAccountNumber(Map<String, String> propertyMap, AccountNum
             }
             if (accountNumberPrefixType.getValue().equals(AccountNumberPrefixType.PREFIX_SHORT_NAME.getValue())) {
                 Integer prefixLength = prefix.length();
-                Integer numberLength = accountMaxLength - prefixLength;
-                accountNumber = StringUtils.leftPad(propertyMap.get(ID), numberLength, '0');
+
+                if (randomAccountNumber.isEnabled()) {
+                    accountNumber = accountNumber.substring(prefixLength);
```
ptuomola, line 171: How do we ensure the substring is unique? What if we had generated a string that only differs from existing ones by the positions that are now being removed? Would it not make more sense to do the full generation of account ID (including any prefix etc) and only after that check for duplicates? Or some other similar way to ensure uniqueness...

ptuomola (changes requested):

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanRepository.java
rrpawar96: Yes, we are not using FIND_NON_CLOSED_BY_ACCOUNT_NUMBER, I have just changed the variableName here to avoid confusion, We are using a new SQL statement that I have created, FIND_BY_ACCOUNT_NUMBER which takes all accounts.

rrpawar96:
@ptuomola, Thanks for the review, now we are validating the existing randomAccountNumber after adding the prefix.

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/AccountNumberGenerator.java
```diff
@@ -143,17 +166,60 @@ private String generateAccountNumber(Map<String, String> propertyMap, AccountNum
...
     }
 
+    private void randomNumberValidation(Map<String, String> propertyMap, AccountNumberFormat accountNumberFormat, String accountNumber) {
+
+        String entityType = propertyMap.get(ENTITY_TYPE);
+        if (entityType.equals("client")) { // avoid duplication it will loop until it finds new random account no.
+
+            Client client = this.clientRepository.getClientByAccountNumber(accountNumber);
+            if (client != null) {
+                accountNumber = generateAccountNumber(propertyMap, accountNumberFormat);
```
ptuomola: Did you test this? I don't think this will work... in case of a conflict, you're assigning a new value to accountNumber here. But this does not change the value returned by the calling function - the calling function never sees the new value of accountNumber and therefore will return the originally generated number. Or am I missing something here? 

Maybe easiest solution is to change this function to e.g. checkAccountNumberConflict() and return true if conflict exists. Then the calling function can check the return value and recursively call itself in case there's a conflict.

ptuomola (changes requested):

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/AccountNumberGenerator.java
rrpawar96: @ptuomola, comment has been addressed. I have tested it, it is working fine, if the randomNumberConflict is true then it will re-call the function and create a random number until it finds a new one and stores it in the accountNumber variable.

rrpawar96:
I have rebased the code, not sure why the Actuator status is unhealthy.

ptuomola:
@rrpawar96 the reason why this does not build is that in the mean time another PR was merged that introduced a database migration with version "V379". Can you rename your migration to "V380" and then it should hopefully work again...

rrpawar96:
Thanks, @ptuomola.

ptuomola (approved):

PR #2004, opened 2021-12-28
Make entrypoint.sh respect FINERACT_DEFAULT_TENANTDB* (fixes FINERACT-1454)

vorburger:
see FINERACT-1454

I have locally verified that the `entrypoint.sh` does what it needs to with these changes, and that it actually works with busybox's `sh` which is used in our container image based on `azul/zulu-openjdk-alpine:17`.

@IOhacker @ptuomola @vidakovic @awasum

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -18,14 +18,16 @@
 # under the License.
 #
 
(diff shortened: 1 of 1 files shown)


ptuomola (approved):
LGTM

PR #2005, opened 2021-12-30
FINERACT-1457: Change Kubernetes set-up to use MariaDB

ptuomola:
## Description

## Checklist

### DIFF
--- a/kubernetes/fineract-mifoscommunity-deployment.yml
+++ b/kubernetes/fineract-mifoscommunity-deployment.yml
@@ -21,15 +21,15 @@ apiVersion: v1
 kind: Service
 metadata:
   labels:
-    app: fineract-mifoscommunity
-  name: fineract-mifoscommunity
+    app: mifos-community
+  name: mifos-community
 spec:
   ports:
   - protocol: TCP
-    port: 80
+    port: 9090
     targetPort: 80
   selector:
-    app: fineract-mifoscommunity
+    app: mifos-community
     tier: frontend
   type: LoadBalancer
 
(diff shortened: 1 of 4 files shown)


ptuomola:
@vorburger at least for me this fixes the Kubernetes set-up. Had to make some other minor changes as well to get this to run on Rancher/k3s on OSX as for some reason it did not like some of the config / scripts..

awasum:
I am wondering if we should support and test on MySQL 8.x vs MariaDD as our user base are more familiar with MySQL than mariadb. Mariadb might be a drop in replacement but I wonder if we are making a conscious decision that we want to be testing and developing on mariadb exclusively? Do you see some issue going forward. I was looking at this: https://mariadb.com/kb/en/mariadb-vs-mysql-compatibility/ and having a few doubts based on the face that the Fineract Community has always use MySQL.

vorburger:
I'm personally actually generally in favor of "pushing" for MariaDB. I also
believe that it is (probably) much more about "recent versions" - of either
one. We didn't break MySQL - I bet it also still works just fine - just not
on an ancient 5.6 version. But it's just easier for the project to stick to
recommending just one, and I'm personally in favor of that being MariaDB
instead of MySQL.

francisguchie:
@vorburger  I have not used MariaDB that much but the last time i did (2017-2018), fineract "misbehaved" and being new to production usage of Apache-Fineract, i chose to stick with what every one was using. But then i learnt of the acquisition of MySQL https://en.wikipedia.org/wiki/MariaDB. 

And am thinking this might lead us into a situation like the one that forced us into a situation of  ["Fineract-Pentaho"](https://issues.apache.org/jira/browse/FINERACT-1127) due to licensing issues 

**That said, I will vote for "pushing" for MariaDB. I like the comfort of knowing am using one of the "best tools available in town" and not worrying about having to change that "tool" in future**

ptuomola:
FWIW - a lot of large companies have switched / are switching from MySQL to MariaDB due to the licensing concerns. I know many banks running thousands of MariaDB instances in production. So I think it's very much "ready for prime time". 

In terms of the compatibility issues, I haven't come across anything major assuming we don't use the latest and greatest features (which exist in MariaDB but not in MySQL). But I don't think we use (or plan to use) any such features in Fineract.

awasum:
Thank you all for your comments. I agree, lets use MariaDB and make it the default going forward. We will deal with whatever issues if any arise. Thanks @ptuomola for doing this.

PR #2009, opened 2022-01-02
Replace command line Docker image build with Gradle Jib plugin

vidakovic:
## Description

Introduce Gradle Jib to build Docker images in preparation of build performance improvements.

NOTE: "war" task had to be removed from "fineract-provider". WAR build is taken care of in PR #2008

PR #2012, opened 2022-01-02
Simplify build for different auth strategies

vidakovic:
## Description

The current process of selecting the authentication strategy for Fineract is unnecessarily complicated. Spring profiles are not the best way to enable/disable these features. By using "@ConditionalOnProperty" we only need one build and can enable the required authentication scheme with environment variables. This makes maintenance of the application.properties file also easier (just one instead of 4 different variants), and in turn makes the build script simpler.


vidakovic:
@ptuomola just FYI: please approve and merge #2009 first

.github/workflows/build.yml
```diff
@@ -47,5 +56,24 @@ jobs:
         run: |
             sudo apt-get update
             sudo apt-get install ghostscript -y
-      - name: Build & Test
+
+      - name: Basic Auth Build & Test
+        env:
+          - FINERACT_SECURITY_BASICAUTH_ENABLED: true
+          - FINERACT_SECURITY_OAUTH_ENABLED: false
```
ptuomola: Do these variables actually do anything? It seems that we are hardcoding the same parameters in the command line when starting the server for oauth2-test and basictest etc? So it would seem that these are redundant... correct?

oauth2-tests/build.gradle
```diff
@@ -45,7 +45,7 @@ cargo {
         }
         startStopTimeout = 240000
         containerProperties {
-            property 'cargo.start.jvmargs', '-Dspring.profiles.active=oauth'
+            property 'cargo.start.jvmargs', '-Dfineract.security.basicauth.enabled=false -Dfineract.security.oauth.enabled=true -Dfineract.security.2fa.enabled=false'
```
ptuomola, line 48: Ie the environment is hardcoded here and these are the values used - correct?

ptuomola (changes requested):
Two further questions/comments:
- I think the README and oauth etc documentation should also be updated to reflect this new set-up? 
- Some of the settings are mutually exclusive.. i.e. you can't have both basicauth and oauth enabled at the same time. Should we detect this and fail the start-up?

oauth2-tests/build.gradle
vidakovic: Correct, exactly the same behavior as before, just with properties instead of profiles.

.github/workflows/build.yml
vidakovic: Hm, now that you mention it... actually no... I was starting with a different idea and then I guess I forgot to remove them. Good catch!

vidakovic:
I'll update the README. Good idea with the startup validation, I like that idea.

vidakovic:
@ptuomola ... alright, added documentation and configuration checks... and did the requested cleanups.

vidakovic:
@ptuomola Hi Petri... finally... I think this should run through now properly... added a couple of last minute fixes to make the unit tests run again properly, do some minor cleanup (e. g. `TwoFactorUtil` not used anymore). I've also rebased with the latest in `develop`. As soon as this one is in I'll start the migration to Testcontainers. Would be great if you could approve and merge.

ptuomola (approved):
LGTM - great job!

PR #2020, opened 2022-01-06
FINERACT-1481: Fix version info

vidakovic:
## Description

Align Spring Boot actuator version info, Docker image tag and Gradle project version.


vidakovic:
@ptuomola FYI: I've also prepared proper output of the app version in the Spring Boot console banner; it's not yet displayed, because of the current launch class (the current one doesn't initialize the required properties). Didn't want to touch that part right now; was also not displayed before, so no harm. The point is that all displays of versions are now aligned.

vidakovic:
@ptuomola ... one more notable thing I just remembered: this change also removes an explicitly dependency between the Swagger YAML prepare task and the Git properties plugin; not much, but one less that potentially gets in the way of proper build caching.

ptuomola (approved):
LGTM

PR #2023, opened 2022-01-17
FINERACT-1490: SSL configuration based on application.properties

vidakovic:
## Description

SSL and Tomcat (performance) configuration via environment variables and without recompiling.


vidakovic:
@ptuomola FYI: #2024 will fix this one, too or at least it will make the fix easier.

SubhamPramanik:
@vidakovic Currently Fineract API doesn't really work without HTTPS (Through port 8080).
Does this PR removes the dependency of HTTPS and all APIs work normally through 8080?
Or does this PR only makes it easier to configure the SSL settings but APIs still require HTTPS?

ptuomola:
> @vidakovic Currently Fineract API doesn't really work without HTTPS (Through port 8080). Does this PR removes the dependency of HTTPS and all APIs work normally through 8080? Or does this PR only makes it easier to configure the SSL settings but APIs still require HTTPS?

I think it will still require SSL. If we want to disable that, we need to modify the SecurityConfig where we currently require that all calls to /api are thorough a secure channel

vidakovic:
@ptuomola yes, you are right... I would make that part (require HTTPS for REST calls) also configurable. I have something similar running in a modified version of Fineract that runs behind a proxy (NGinx); the proxy handles all the SSL related stuff... I think this is a quite common setup... and in my case I just wanted to save some CPU cycles (NGinx is - I think - faster and more flexible).
I'll attack this PR again when I have the improved database configuration ready.
Thanks again for checking Petri.

SubhamPramanik:
Thanks for the insight. Much appreciated!

Something to add, if you can make the HTTPS requirement configurable that would be great. I will send in a PR for Fineract to work with GitPod properly. Currently, the APIs don't really work in GitPod as HTTPS traffic/tunnelling is not supported in GitPod.

galovics (approved):
Looks good to me.

ptuomola:
@vidakovic looks like some checkstyle errors... can you fix and we can then merge?

vidakovic:
@ptuomola yep, finally have some time to take care of this

vidakovic:
... still investigating why one test is failing. FYI

vidakovic:
I think I know what's going on... the SSL settings are not properly applied  in Tomcat/WAR use case (i. e. integration testing)... ok, will check.

vidakovic:
I just verified locally the failing tests... and they run without a hitch...

vidakovic:
@ptuomola @galovics ... hosanna... it worked! In the end Cargo (well, Tomcat) didn't like the base path property in application.properties... so I just removed it, defaults are fine. Having said that: this PR is needed, because Tomcat deployments will very likely fail without this fix. FYI

francisguchie:
> @ptuomola @galovics ... hosanna... it worked! In the end Cargo (well, Tomcat) didn't like the base path property in application.properties... so I just removed it, defaults are fine. Having said that: this PR is needed, because Tomcat deployments will very likely fail without this fix. FYI

@vidakovic  thank you for thinking about Tomcat users - i am heavily one of them

vidakovic:
@francisguchie no problem at all... and as long as the community wants/needs a WAR file there will be one... in between releases 1.5 and 1.6 there was just a short moment when the WAR build was not properly working, but those issues are fixed. Things will get even easier once we have full Testcontainer support (working on it) which will replace the Cargo plugin based integration testing that we currently have. 

Here's another idea: in those Testcontainer unit tests we could also do - kind of - integration tests that cover WAR deployments in multiple application containers like Jetty, Tomcat, Undertow, JBoss... at least to show that the WAR is properly started in those environments... have to think about this... wondering if the community would like to use something else than Tomcat?

vorburger:
FYI I think Fineract (Spring Boot, realy) actually does work without HTTP IFF it's behind a Proxy. I had set up https://www.fineract.dev like that on GCP. I forgot the details, it's been two years ago, I would need to look up. There is some standard HTTP headers, if I remember correctly. Please create a new JIRA about that if anyone wants to follow-up and assign it to me and I'll try to add some documentation about this.

vorburger:
Oh wait now I know why this felt so familiar... I already wrote that doc, back then!

See https://github.com/apache/fineract/blob/develop/fineract-doc/src/docs/en/deployment.adoc#https

#2084 raised to link what was added here to that.

PR #2024, opened 2022-01-17
FINERACT-1492: Hikari configuration based on application properties

vidakovic:
## Description

Hikari configuration based entirely on application.properties. Tuning can be done via environment variables without recompiling. I got also rid unnecessary configuration classes and environment variables. No additional environment variables for the tenant databases; everything necessary is discovered via the Hikari configuration.


vidakovic:
@awasum could you have a look at this please and approve? Simplifies configuration. Thanks in advance :-)

ptuomola:
Hi - I was looking at this and fully agree it's a great improvement - makes
things a lot cleaner and simpler.

The only concern I have is: it will break everyone's current configuration.

Do you think it would make sense to make this change backward compatible?
Ie use the same environment variables if they have been defined.

francisguchie:
@ptuomola  @vidakovic - if you read  [FINERACT-1115](https://issues.apache.org/jira/browse/FINERACT-1115), you can see the dilemma that you just solved. In my simple understanding, I agree to removal of recompilation.

vidakovic:
@ptuomola ... valid point... how about this: I'll add a easily removable compatibility - java - configuration that is activated only if these old (deprecated) configs are detected (environment variables etc.) and spins up things with a fat warning message on the console telling people to migrate their configuration?

ptuomola:
Sounds like a great plan! That allows us to move to the new standard while keeping backward compatibility for at least a few releases….

vidakovic:
@ptuomola ... finally I found a quiet moment to wrap this up. While I was at it I introduced a mapping class for Fineract related properties (everything prefixed `fineract.*` in `application.properties`); those properties are mapped now to `org.apache.fineract.infrastructure.core.config.FineractProperties` for type safe access and avoiding all these `@Value(..)` annotations. Mapping the properties to a class also has advantages when doing configuration changes in a Kubernetes environment. The criteria I use to activate the backwards compatible configuration is a check on environment variable `fineract_tenants_driver` (if it's empty then the new configuration is assumed, if not empty then `org.apache.fineract.infrastructure.core.config.CompatibilityConfig` is activated. As we discussed: if the deprecated configuration is used then a warning is printed to the console.

I've added some additional hints to the README and to docker-compose.yml. The Docker Compose file is configured to use the new stuff (with more options to tweak the Hikari pool configuration).

I've tried both configs, work as advertised. Would be great if you could approve to get this stuff out of the way for the next release. Thanks again.

galovics (approved):

PR #2029, opened 2022-02-07
FINERACT-1497: Fixing Swagger Code generation on Windows

galovics:
## Description

## Checklist

### DIFF
--- a/fineract-client/build.gradle
+++ b/fineract-client/build.gradle
@@ -31,7 +31,7 @@ openApiMeta {
 }
 
 openApiValidate {
-    inputSpec = "file://$swaggerFile"
+    inputSpec = "file:///$swaggerFile"
     recommend = true
 }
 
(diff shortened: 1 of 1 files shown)


galovics:
Note that this is aimed to be included in 1.7. Not necessary for 1.6.

ptuomola:
I recall that file:// was added because otherwise the generator tried to use classloader to load the file, and this did not work in some scenarios. But let's try - maybe there have been other changes that mean this now works...

ptuomola:
Btw @vidakovic any idea why the integration tests are now super-fast? All of them completed in 10 minutes in this build - they used to take around 1 hour... whatever you've changed has really made a huge difference! :-)

ptuomola:
Now looking into this, I thought this was changed because without this, the incremental build did not work correctly. But looks like the incremental build is now broken for fineract-client anyway and a full rebuild is always triggered. What's even worse looks like the rebuild now fails in :fineract-client:licenseGenerated when the first build has worked. So something is quite broken with the fineract-client build. Need to investigate when it was broken - at the very least, clean / build or just build should work, and ideally even work as an incremental build running only the steps that are required.

galovics:
@ptuomola Thanks for taking the time and giving feedback. I appreciate.
Your point is probably valid and in some corner cases this doesn't work and I'm not sure if this is going to work in all environments.

The reason why on Windows this didn't work is because the file URI schema is not followed. 
The `file://` prefix is missing an extra slash since it's pointing to a local file. According to the RFC, this is not a valid URI scheme for files but some systems accept it. [RFC here](https://datatracker.ietf.org/doc/html/rfc8089#appendix-B)

To make sure this still works the way it should, instead of removing the file URI prefix, I added the extra missing slash to make sure it's compliant with the RFC and that's something Windows also accepts. The end result is `file:///`

See the updated PR.

ptuomola (approved):
LGTM

PR #2041, opened 2022-02-21
FINERACT-1509: Gradle wrapper missing in source distribution

vidakovic:
## Description

Looks like we forgot to add the Gradle wrapper files to the source distribution; without it the "gradlew" command doesn't work.

### DIFF
--- a/fineract-war/build.gradle
+++ b/fineract-war/build.gradle
@@ -93,7 +93,7 @@ distributions {
 
         contents {
             from "$rootDir/"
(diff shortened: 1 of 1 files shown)


vidakovic:
@ptuomola ... would be great if you could approve this one as soon as possible. I'd like to start voting on the release 1.6.0 as early as possible tomorrow. The distribution works otherwise; looks like we didn't have the Gradle wrapper JAR included in previous releases either...

galovics (approved):

PR #2053, opened 2022-02-24
Merge release 1.6.0

vidakovic:


vidakovic:
@ptuomola ... would be great if you could approve this one... as you can see there's only a small diff... only thing this does: apply automatically proper version based on Git tag/hash (in release we have a different pattern). Note: for the next release (1.7.x) I'll try to automate this more with JReleaser (for most parts) and maybe a little custom plugin to handle the Apache specifics.

ptuomola:
Hi @vidakovic do you still need this? The only delta shown is a commented out line, so I'm assuming this needs to be at least rebased before it makes sense...

vidakovic:
@ptuomola Hi Petri... this is already released (see: https://github.com/apache/fineract/blob/1.6.0/build.gradle#L89). It's just to keep this in sync. The only reason to have those two lines is to have the right version pattern when doing a release. This is at the moment done by manually enabling/disabling those lines (on a release branch). I have another improvement lined up to make the release less of a manual experience (including the version pattern)... I think we can reduce the manual steps a lot by using JReleaser + a little custom Gradle plugin to take care of any Fineract/Apache specific steps.

vidakovic:
FYI: conflict resolved.

ptuomola:
LGTM

PR #2058, opened 2022-03-05
Create initial dependabot.yml (FINERACT-1453)

vorburger:
see FINERACT-1453


vorburger:
I hope none of the active committers minds if as the rare "exception to the general rule" I'll go ahead and "self merge" this PR? (I generally think we should almost never do that, because peer review even for small changes almost always does have value, IMHO. This kind of "repo infrastructure" kind of thing may be the rare exception, because you can't really "test" this type of change, without just merging it.)

ptuomola:
Agreed @vorburger - happy to merge this for you, or for you to self merge

PR #2144, opened 2022-03-11
FINERACT-1483: fix Sonar bugs with Blocker and Critical severity

taskain7:
## Description

Bugs reported by Sonar with `Blocker` and `Critical` severities have been fixed

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/ClientRepositoryWrapper.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/ClientRepositoryWrapper.java
@@ -44,6 +44,7 @@ public class ClientRepositoryWrapper {
         this.context = context;
     }
 
+    @Transactional(readOnly = true)
     public Client findOneWithNotFoundDetection(final Long id) {
         return this.findOneWithNotFoundDetection(id, false);
     }
@@ -77,6 +78,7 @@ public class ClientRepositoryWrapper {
         this.repository.flush();
     }
 
+    @Transactional(readOnly = true)
     public Client getActiveClientInUserScope(Long clientId) {
         final Client client = this.findOneWithNotFoundDetection(clientId);
         if (client.isNotActive()) {
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanRepositoryWrapper.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanRepositoryWrapper.java
@@ -44,6 +44,7 @@ public class LoanRepositoryWrapper {
         this.repository = repository;
     }
 
+    @Transactional(readOnly = true)
     public Loan findOneWithNotFoundDetection(final Long id) {
         return this.findOneWithNotFoundDetection(id, false);
     }
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/LoanApplicationTerms.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/LoanApplicationTerms.java
@@ -1773,7 +1773,7 @@ public final class LoanApplicationTerms {
 
     }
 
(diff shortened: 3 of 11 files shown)


fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/ClientRepositoryWrapper.java
```diff
@@ -44,6 +44,7 @@ public ClientRepositoryWrapper(final ClientRepository repository, final Platform
         this.context = context;
     }
 
+    @Transactional
```
vidakovic: Are the "@Transactional" annotations really needed here? This is technically not solving the Sonar issues... I'd remove them for now... and if there's a reason why to put them then please add them to another PR

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/LoanApplicationTerms.java
```diff
@@ -1773,7 +1773,7 @@ private void updateRecurringMoratoriumOnPrincipalPeriods(Integer periodNumber) {
 
     }
 
-    public void setTotalPrincipalAccounted(Money totalPrincipalAccounted) {
+    public void setTotalPrincipalAccountedForInterestCalculation(Money totalPrincipalAccounted) {
```
vidakovic, line 1776: Why is the name of the method a problem?

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/ClientRepositoryWrapper.java
taskain7: Self-invocation cannot make @Transactional to take effect. @Transactional tells spring to create a proxy around the object. The proxy intercepts calls to the object from other objects, but it does not intercept calls within the object. And nested @Transactional is not a problem, because the default option is REQUIRED, so if `findOneWithNotFoundDetection(final Long id)` method has already opened a transaction, then `findOneWithNotFoundDetection(final Long clientId, final boolean loadLazyCollections)` won't be open a new one.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/LoanApplicationTerms.java
taskain7: There is also a `totalPrincipalAccounted` field in this object, so the original name of the setter is misleading, and does not do what says.

awasum:
@taskain7 Thanks for the contributions. Looks like there is a checkstyle issue somewhere in the PR: https://github.com/apache/fineract/runs/5512932230?check_suite_focus=true

Check the logs of the build failures above.

taskain7:
Thank you @awasum , I fixed, now it should pass the build

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/LoanApplicationTerms.java
vidakovic: ... makes sense

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanRepositoryWrapper.java
```diff
@@ -44,6 +44,7 @@ public LoanRepositoryWrapper(final LoanRepository repository) {
         this.repository = repository;
     }
 
+    @Transactional
```
vidakovic: Missing: @Transactional(readOnly = true)

vidakovic: @taskain7 ... could please fix this one remaining read-only transaction? Then we can tick this one of and merge. Thanks again for this contribution.

vidakovic (approved):
LGTM

PR #2171, opened 2022-03-22
FINERACT-1532: Create an example for pluggability

vidakovic:
Proof of concept to demonstrate modularity with a minimum of additional configuration. Documentation started.


fineract-provider/src/main/resources/META-INF/spring.factories
```diff
@@ -0,0 +1,2 @@
+org.springframework.boot.autoconfigure.EnableAutoConfiguration=\
```
galovics: Since this is part of the intermediate solution - while the modules are properly extracted - does it make sense to have it configured like this? The component scan would pick this up after all in fineract-provider. Thoughts?

galovics (commented):
Looks really good.
One question though @vidakovic, wouldn't it make sense to extract the NoteReadPlatformService/NoteWritePlatformService interfaces along with the NodeData DTO class into a separate "note" module as well? That could kick things off and we have a very real example.

fineract-provider/src/main/resources/META-INF/spring.factories
vidakovic: ... yeah... was thinking the same when I put it in place... I guess technically we don't need it in this setup... probably my thinking was to put stuff in place and later move to separate modules; I guess there's an argument to keep only the bare minimum.
What I'll do: add a section in the documentation describing the anatomy of a clean module;  anyone who wants to port an existing or creating a new one has then a detailed todo list. Alright, working on it.

galovics (approved):

PR #2186, opened 2022-03-25
FINERACT-1547: Cleanup credit bureau dependency in infrastructure core

vidakovic:
In the infrastructure core packages we should never use any cross dependency to any other package. "CommandProcessingResult" contains an attribute with a type "CreditBureauReportData". We should apply the same strategy as with attribute "changes": convert to "Map<String, Object>".


galovics (approved):
Looks good for now but it seriously hits me in the face why a generic CommandProcessingResult should contain any specifics to the creditBureauReport.

Let's merge it but later on we might need to come back and further improve.

PR #2188, opened 2022-03-26
FINERACT-1550: Fix repetitive JdbcTemplate/RoutingDataSource injection

vidakovic:
We are creating a lot of JdbcTemplate instances just to ensure that we use RoutingDataSource with it. It's way easier to do this once in a Spring Java config class and then just inject the JdbcTemplate. I'm not entirely sure what impact this change will have on memory usage, doing 150x `new JdbcTemplate(...)` must have an effect.

In a couple of places the data source object wsa stored in service classes, but never used. I cleaned this up.

Another nice side effect of this cleanup action: we are **NOT** exposing any internal details (RoutingDataSource)  in our ser

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/JdbcConfig.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/JdbcConfig.java
@@ -1,0 +1,39 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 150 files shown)


fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/JdbcConfig.java
```diff
@@ -0,0 +1,42 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Primary;
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+
+@Configuration
+public class JdbcConfig {
+
+    @Bean
+    @Primary
```
galovics: Question, is the @Primary really needed or it's just for safety?

galovics (commented):
Looks good. One comment though.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/JdbcConfig.java
vidakovic: Nothing escapes your eyes ;-) ... yes, just for safety, but I'll actually try to remove before final approval, I think it should work... the slimmer the better.

vidakovic:
Alright... the cancelled build yesterday was only a hiccup on Github Actions.

IOhacker:
LGTM

PR #2278, opened 2022-04-12
FINERACT-1573: Credit Balance Refund loan transaction

John-Woodlock:
## Description
Changes as per JIRA ticket but some notes:
This PR will result in some extra loan related code as it involves the addition of a new loan transaction.
I did reuse/share some existing functionality but there are places where the code of one (or more) of other loan transactions is quite similar  but not necessarily similar enough to justify the local optimisation.  There is definitely value in looking at DRYing transactions as a whole but not for this JIRA issue.

## Checklist


fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
```diff
@@ -3615,10 +3615,11 @@ public ChangedTransactionDetail adjustExistingTransaction(final LoanTransaction
         validateActivityNotBeforeClientOrGroupTransferDate(LoanEvent.LOAN_REPAYMENT_OR_WAIVER,
                 transactionForAdjustment.getTransactionDate());
 
-        if (transactionForAdjustment.isNotRepayment() && transactionForAdjustment.isNotWaiver()) {
-            final String errorMessage = "Only transactions of type repayment or waiver can be adjusted.";
-            throw new InvalidLoanTransactionTypeException("transaction", "adjustment.is.only.allowed.to.repayment.or.waiver.transaction",
-                    errorMessage);
+        if (transactionForAdjustment.isNotRepayment() && transactionForAdjustment.isNotWaiver()
```
galovics, line 3618: This is rather a comment for the changes on the whole file and not specific to this line.

John, do you think it would be possible to extract all these changes to a separate class instead of putting everything into the Loan class?

I brought this up on the mailing list as well that unless we have a really really good reason, we must start extracting new functionality outside of this class to decrease the maintenance cost.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanAccountDomainServiceJpa.java
```diff
@@ -623,6 +623,41 @@ private AppUser getAppUserIfPresent() {
         return user;
     }
 
+    @Override
+    public CommandProcessingResultBuilder creditBalanceRefund(Long loanId, LocalDate transactionDate, BigDecimal transactionAmount,
```
galovics, line 627: Same here just like with the Loan class. This is also getting too big.

I could imagine this LoanAccountDomainServiceJpa class to be kind of a facade and have a specific CreditBalanceRefundService class which has this implementation only and the LoanAccountDomainServiceJpa class just delegates the call.
Thoughts?

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/handler/CreditBalanceRefundCommandHandler.java
```diff
@@ -0,0 +1,78 @@
...
+import org.springframework.dao.DataIntegrityViolationException;
+import org.springframework.orm.jpa.JpaSystemException;
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+
+@Service
+@CommandType(entity = "LOAN", action = "CREDITBALANCEREFUND")
+public class CreditBalanceRefundCommandHandler implements NewCommandSourceHandler {
+
+    private static final Logger LOG = LoggerFactory.getLogger(CreditBalanceRefundCommandHandler.class);
```
galovics: You can use Lombok now with the @SLF4J annotation.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/handler/CreditBalanceRefundCommandHandler.java
```diff
@@ -0,0 +1,78 @@
...
+    private final LoanWritePlatformService writePlatformService;
+
+    @Autowired
+    public CreditBalanceRefundCommandHandler(final LoanWritePlatformService writePlatformService) {
```
galovics: Just go with @RequiredArgsConstructor

galovics (changes requested):
Looks good in terms of functionality but please take a look at the suggestions around the Loan and related classes.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanAccountDomainServiceJpa.java
John-Woodlock: Thanks Arnold,

I have updated the Lombok related changes.
I accept the need to reduce/refactor Java code.  I'll like an amnesty on this issue because, as said in the notes, it is a new loan transaction (very rare) so there will be some extra (but not major) additional code to loan.java - but I think I can do a better job of reducing the code soon as I'm looking at the core loan area and expect it would be more natural there to reduce/refactor... so in short, I'm fully onboard the care for the code drive but will be more productive soon.

galovics (approved):

IOhacker:
@John-Woodlock I think that this PR is missing the "FINERACT-1573" as prefix. https://github.com/apache/fineract/#pull-requests

galovics (approved):

PR #2281, opened 2022-04-14
FINERACT-1571: indroducing enviroment variables for read/write/batch …

taskain7:
## Description

Three enviroment variables have been introduced to be able to set the instance type of Fineract. (read/write/batch)

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractModeValidationConfig.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractModeValidationConfig.java
@@ -1,0 +1,50 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.infrastructure.core.config;
+
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.InitializingBean;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.context.annotation.Conditional;
+import org.springframework.core.Ordered;
+import org.springframework.core.annotation.Order;
+import org.springframework.stereotype.Component;
+
+@Slf4j
+@Component
+@Order(Ordered.HIGHEST_PRECEDENCE)
+@RequiredArgsConstructor
+@Conditional(FineractModeValidationCondition.class)
+public class FineractModeValidationConfig implements InitializingBean {
+
+    private final ApplicationContext applicationContext;
+
+    @Override
+    public void afterPropertiesSet() throws Exception {
+        terminateApplication();
+    }
+
+    private void terminateApplication() {
+        log.error(
+                "The Fineract instance type is not configured properly. At least one of these environment variables should be true: FINERACT_MODE_READ_ENABLED, FINERACT_MODE_WRITE_ENABLED, FINERACT_MODE_BATCH_ENABLED");
+        ((ConfigurableApplicationContext) this.applicationContext).close();
+    }
+}
--- a/fineract-provider/src/main/resources/application.properties
+++ b/fineract-provider/src/main/resources/application.properties
@@ -33,6 +33,10 @@ fineract.tenant.identifier=${FINERACT_DEFAULT_TENANTDB_IDENTIFIER:default}
 fineract.tenant.name=${FINERACT_DEFAULT_TENANTDB_NAME:fineract_default}
 fineract.tenant.description=${FINERACT_DEFAULT_TENANTDB_DESCRIPTION:Default Demo Tenant}
 
+fineract.mode.read-enabled=${FINERACT_MODE_READ_ENABLED:true}
+fineract.mode.write-enabled=${FINERACT_MODE_WRITE_ENABLED:true}
+fineract.mode.batch-enabled=${FINERACT_MODE_BATCH_ENABLED:true}
+
 management.health.jms.enabled=false
 
 # FINERACT 1296
--- a/fineract-provider/src/main/java/org/apache/fineract/ServerApplication.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/ServerApplication.java
@@ -23,7 +23,6 @@ import org.apache.fineract.infrastructure.core.boot.AbstractApplicationConfigura
 import org.springframework.boot.autoconfigure.jdbc.DataSourceProperties;
 import org.springframework.boot.builder.SpringApplicationBuilder;
 import org.springframework.boot.web.servlet.support.SpringBootServletInitializer;
-import org.springframework.context.ConfigurableApplicationContext;
 
 /**
  * Fineract main() application which launches Fineract in an embedded Tomcat HTTP (using Spring Boot).
@@ -52,7 +51,7 @@ public class ServerApplication extends SpringBootServletInitializer {
     }
 
     public static void main(String[] args) throws IOException {
-        ConfigurableApplicationContext ctx = configureApplication(new SpringApplicationBuilder(ServerApplication.class)).run(args);
+        configureApplication(new SpringApplicationBuilder(ServerApplication.class)).run(args);
         // ApplicationExitUtil.waitForKeyPressToCleanlyExit(ctx);
     }
 }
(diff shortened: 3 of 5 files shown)


fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/boot/TerminateService.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+package org.apache.fineract.infrastructure.core.boot;
+
+import javax.annotation.PreDestroy;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Service;
+
+@Slf4j
+@Service
+public class TerminateService {
```
vidakovic: Do we really need this? There's already a lot of feedback from Spring and Spring Boot themselves. If we can avoid code I'd vote to remove.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/configuration/data/InstanceTypeData.java
```diff
@@ -0,0 +1,33 @@
...
+package org.apache.fineract.infrastructure.configuration.data;
+
+import lombok.Data;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+@Data
+@Configuration
+@ConfigurationProperties(prefix = "fineract")
+public class InstanceTypeData {
```
vidakovic: That's going to clash a bit with "FineractProperties.java", because it has the same prefix. What I would do here instead is to define an inline class in FineractProperties... in the end it's just a utility class that's only used for configuration.
That would look like this (sketching):

```
@Data
@Configuration
@ConfigurationProperties(prefix = "fineract")
public class FineractProperties {
    // here all properties from Fineract
    private FineractModeProperties mode;

    @Data
    public static class FineractModeProperties {
        private boolean readEnabled;
        private boolean writeEnabled;
        private boolean batchEnabled;
    }
}
```

... which in turn translates then into the following properties:

```
fineract.mode.read-enabled=${FINERACT_MODE_READ_ENABLED:true}
fineract.mode.write-enabled=${FINERACT_MODE_WRITE_ENABLED:true}
fineract.mode.batch-enabled=${FINERACT_MODE_BATCH_ENABLED:true}
```

Advantages of this:
- you keep the config/application.properties related classes together, namely FineractProperties and FineractModeProperties
- consistent naming (prefix "Fineract", suffix "Properties"); then everyone knows what they are used for
- I think we could even make FineractModeProperties package private; then no one can be tempted to use it in any other context than configuration
- related concepts (I called it here "Mode", but doesn't really matter, could be anything else) are grouped with a kind of "namespace" (actually it's a prefix, here "fineract.mode.*"); once we have even more properties this will make life a lot easier and the FineractProperties class doesn't grow endlessly

fineract-provider/src/main/java/org/apache/fineract/infrastructure/configuration/service/InstanceTypeService.java
```diff
@@ -0,0 +1,32 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.infrastructure.configuration.service;
+
+public interface InstanceTypeService {
```
vidakovic: Do we really need this service? Doesn't seem to do anything else than returning the properties. We could just inject FineractProperties (see my earlier comment) and use the getters. Would save us 1 Java interface, 1 implementation class.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/configuration/service/InstanceTypeServiceImpl.java
```diff
@@ -0,0 +1,63 @@
...
+        return instanceTypeData.isBatchModeEnabled();
+    }
+
+    @Override
+    public boolean isReadOnlyMode() {
+        return instanceTypeData.isReadModeEnabled() && !instanceTypeData.isWriteModeEnabled() && !instanceTypeData.isBatchModeEnabled();
+    }
+
+    @Override
+    public boolean isInstanceTypeConfigured() {
```
vidakovic: We could do this check immediately on boot in a Java configuration class (similar to the way that we check that an authentication scheme is configured) and fail early. Injectable service for that not really necessary.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/boot/AbstractApplicationConfiguration.java
```diff
@@ -43,7 +44,7 @@
         LiquibaseAutoConfiguration.class })
 @EnableTransactionManagement
 @EnableWebSecurity
-@EnableConfigurationProperties({ FineractProperties.class, LiquibaseProperties.class })
+@EnableConfigurationProperties({ FineractProperties.class, LiquibaseProperties.class, InstanceTypeData.class })
```
vidakovic: See my earlier comments. Not needed if we would rearrange things a bit.

fineract-provider/src/main/resources/application.properties
```diff
@@ -33,6 +33,10 @@ fineract.tenant.identifier=${FINERACT_DEFAULT_TENANTDB_IDENTIFIER:default}
 fineract.tenant.name=${FINERACT_DEFAULT_TENANTDB_NAME:fineract_default}
 fineract.tenant.description=${FINERACT_DEFAULT_TENANTDB_DESCRIPTION:Default Demo Tenant}
 
+fineract.read-mode-enabled=${FINERACT_READ_MODE_ENABLED:true}
```
vidakovic: See earlier comments on properties mapping.

fineract-provider/src/test/java/org/apache/fineract/infrastructure/instancetype/DefaultInstanceTypeTest.java
```diff
@@ -0,0 +1,76 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+import io.cucumber.java8.En;
+import org.apache.fineract.infrastructure.configuration.data.InstanceTypeData;
+import org.apache.fineract.infrastructure.configuration.service.InstanceTypeService;
+import org.springframework.beans.factory.annotation.Autowired;
+
+public class DefaultInstanceTypeTest implements En {
```
vidakovic: I might be missing something here, but aren't the "*.feature" files missing? Please also see comments on properties mapping (probably less tests needed).

vidakovic (approved):
LGTM

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractModeValidationConfig.java
```diff
@@ -0,0 +1,47 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
+import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.stereotype.Component;
+
+@Slf4j
+@Component
+@ConditionalOnExpression("#{ systemEnvironment['FINERACT_MODE_READ_ENABLED'] == false and systemEnvironment['FINERACT_MODE_WRITE_ENABLED'] == false and systemEnvironment['FINERACT_MODE_BATCH_ENABLED'] == false}")
+public class FineractModeValidationConfig implements ApplicationContextAware {
```
galovics: I think we need an `@Order` annotation with the highest order to make sure Fineract is not going to do anything if it's none of the instances, for example no Liquibase migration runs.
@taskain7 @vidakovic thoughts?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractModeValidationConfig.java
```diff
@@ -0,0 +1,47 @@
...
+import org.springframework.beans.BeansException;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
+import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.stereotype.Component;
+
+@Slf4j
+@Component
+@ConditionalOnExpression("#{ systemEnvironment['FINERACT_MODE_READ_ENABLED'] == false and systemEnvironment['FINERACT_MODE_WRITE_ENABLED'] == false and systemEnvironment['FINERACT_MODE_BATCH_ENABLED'] == false}")
```
galovics: Hmm, I'm thinking that with this we are forcing people to use the new environment variables despite the fact that the new properties are configured with default values in the application.properties

I think we should move to validating the properties instead of the env vars directly.

galovics (changes requested):

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractModeValidationConfig.java
vidakovic: @taskain7 what you could do here is just use the properties instead... that way you make no assumption about variables in any shape or form... so even if you start things e. g. from you IntelliJ environment for development things will just fall properly in place.

taskain7: I can create a custom implementation for `Condition` interface, that uses the properties, not the env variables.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractModeValidationConfig.java
vidakovic: @galovics ... you are right... the order could be a concern here... of course we wouldn't want Liquibase accidentally kicking in even though the instance was intended as read-only. Makes sense to use `@Order` here @taskain7

galovics (approved):

PR #2288, opened 2022-04-21
FINERACT-1571: Allow Liquibase data migration for write instances only

taskain7:
## Description

Allowing Liquibase migration is based on the FINERACT_MODE_WRITE_ENABLED environment variable, so the migration is executed only on write instances. Also log message warns if the Liquibase migration is disabled.

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/migration/TenantDatabaseUpgradeService.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/migration/TenantDatabaseUpgradeService.java
@@ -68,8 +68,11 @@ public class TenantDatabaseUpgradeService implements InitializingBean {
 
     @Override
     public void afterPropertiesSet() throws Exception {
-        if (databaseStateVerifier.isLiquibaseDisabled()) {
+        if (databaseStateVerifier.isLiquibaseDisabled() || !fineractProperties.getMode().isWriteEnabled()) {
             LOG.warn("Liquibase is disabled. Not upgrading any database.");
+            if (!fineractProperties.getMode().isWriteEnabled()) {
+                LOG.warn("Liquibase is disabled because the current instance is configured as a non-write Fineract instance");
+            }
             return;
         }
         try {
--- a/fineract-provider/src/test/java/org/apache/fineract/infrastructure/core/LiquibaseStepDefinitions.java
+++ b/fineract-provider/src/test/java/org/apache/fineract/infrastructure/core/LiquibaseStepDefinitions.java
@@ -40,12 +40,14 @@ import org.apache.fineract.infrastructure.core.service.migration.TenantDataSourc
 import org.apache.fineract.infrastructure.core.service.migration.TenantDatabaseStateVerifier;
 import org.apache.fineract.infrastructure.core.service.migration.TenantDatabaseUpgradeService;
 import org.apache.fineract.infrastructure.security.service.TenantDetailsService;
+import org.springframework.beans.factory.annotation.Autowired;
 
 public class LiquibaseStepDefinitions implements En {
 
     private TenantDataSourceFactory tenantDataSourceFactory;
     private TenantDetailsService tenantDetailsService;
     private TenantDatabaseStateVerifier databaseStateVerifier;
+    @Autowired
     private FineractProperties fineractProperties;
     private ExtendedSpringLiquibaseFactory liquibaseFactory;
     private ExtendedSpringLiquibase initialTenantStoreLiquibase;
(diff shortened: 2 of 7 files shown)


fineract-provider/src/main/resources/application.properties
```diff
@@ -99,7 +99,7 @@ spring.datasource.hikari.dataSourceProperties['dumpQueriesOnException']=${FINERA
 spring.jpa.open-in-view=false
 
 # Liquibase configuration
-spring.liquibase.enabled=true
+spring.liquibase.enabled=${FINERACT_MODE_WRITE_ENABLED:true}
```
galovics: I don't think we need this since the check is already present in the code.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/migration/TenantDatabaseUpgradeService.java
```diff
@@ -70,6 +70,9 @@ public TenantDatabaseUpgradeService(final TenantDetailsService detailsService,
     public void afterPropertiesSet() throws Exception {
         if (databaseStateVerifier.isLiquibaseDisabled()) {
             LOG.warn("Liquibase is disabled. Not upgrading any database.");
+            if (!this.fineractProperties.getMode().isWriteEnabled()) {
+                LOG.warn("Liquibase is disabled, because the `FINERACT_MODE_WRITE_ENABLED` environment variable is set to false.");
```
galovics: Just styling, I think we can leave out the `this`
Plus, probably it makes more sense to show an environment variable name indepdenent message to the users in case it changes in the future; like "Liquibase is disabled because the current instance is configured as a non-write Fineract instance".
What do you think?

taskain7: You're right, I'm gonna change it.

fineract-provider/src/main/resources/application.properties
```diff
@@ -99,7 +99,7 @@ spring.datasource.hikari.dataSourceProperties['dumpQueriesOnException']=${FINERA
 spring.jpa.open-in-view=false
 
 # Liquibase configuration
-spring.liquibase.enabled=true
+spring.liquibase.enabled=${fineract.mode.write-enabled}
```
galovics: Thinking.. Do we need this change at all?

galovics (changes requested):
I'd love to see 3 new test cases cases verifying that:
1. In case of a read instance, migration doesn't start
2. In case of a batch instance, the migration doesn't start
3. In case of a write instance, the migration starts (but I guess this case is already implemented, just the property setup is not made explicit in the tests)

fineract-provider/src/main/resources/application.properties
taskain7: At some point we have to tell to Spring that Liquibase migration should we skipped, and as far as I am aware this way we can be sure Spring does not run any Liquibase related code. If you have any suggestion where should we put this condition then feel free to share, you are way more familiar with the Liquibase configuration in Fineract than me.

galovics: Oh okay, I get it now. The default Liquibase migration you're used to is actually disabled because the AutoConfiguration is excluded from the Spring context (see `AbstractApplicationConfiguration`). Essentially the only Liquibase migration done is within the `TenantDatabaseUpgradeService` where you have the check exactly if the instance is a write instance.

The reason the property is there in the application.properties is because the `TenantDatabaseUpgradeService` checks this same property whether the migration should be enabled/disabled.

fineract-provider/src/test/java/org/apache/fineract/infrastructure/core/LiquibaseStepDefinitions.java
```diff
@@ -65,6 +65,13 @@ public LiquibaseStepDefinitions() {
         });
         Given("Liquibase is enabled with a default tenant", () -> {
             initializeLiquibase(true);
+            initializeWriteInstance();
```
galovics: Let's create a new StepDefinitions class for the isntance types, like InstanceTypeStepDefinitions that only stores the related functionalities.

Technically what you implemented works but it's definitely not at the right place.

fineract-provider/src/test/java/org/apache/fineract/infrastructure/core/LiquibaseStepDefinitions.java
```diff
@@ -99,9 +106,18 @@ public LiquibaseStepDefinitions() {
...
             verify(databaseStateVerifier).isLiquibaseDisabled();
+            verify(fineractProperties).getMode();
             verifyNoMoreInteractions(databaseStateVerifier);
-            verifyNoInteractions(tenantDetailsService, tenantStoreDataSource, fineractProperties, liquibaseFactory,
-                    tenantDataSourceFactory);
+            verifyNoMoreInteractions(fineractProperties);
+            verifyNoInteractions(tenantDetailsService, tenantStoreDataSource, liquibaseFactory, tenantDataSourceFactory);
+        });
+
+        Then("The database migration did not do anything, because it is not a write instance", () -> {
```
galovics, line 108: Just a note: since this is related to liquibase, you can leave it here instead of moving this into the other class I mentioned above.

galovics (changes requested):

fineract-provider/src/test/java/org/apache/fineract/infrastructure/core/LiquibaseStepDefinitions.java
```diff
@@ -158,7 +164,8 @@ private void initializeLiquibase(boolean liquibaseEnabled) {
         defaultTenantDataSource = mock(DataSource.class);
 
         given(databaseStateVerifier.isLiquibaseDisabled()).willReturn(!liquibaseEnabled);
-        given(fineractProperties.getTenant()).willReturn(new FineractProperties.FineractTenantProperties());
+        // given(fineractProperties.getTenant()).willReturn(new FineractProperties.FineractTenantProperties());
```
galovics: comments.

fineract-provider/src/test/resources/application-test.properties
```diff
@@ -33,6 +33,10 @@ fineract.tenant.identifier=default
 fineract.tenant.name=fineract_default
 fineract.tenant.description=Default Demo Tenant
 
+fineract.mode.read-enabled=${FINERACT_MODE_READ_ENABLED:true}
```
galovics: You can skip using the env vars for the tests. Nobody is using them here.

galovics (approved):
LGTM

PR #2298, opened 2022-04-26
FINERACT-1594: Create setenv.sh file for Tomcat installations

vidakovic:
## Description

Script to give Tomcat users the same convenience as Docker Compose users.

### DIFF
--- a/fineract-war/setenv.sh
+++ b/fineract-war/setenv.sh
@@ -1,0 +1,55 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements. See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership. The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License. You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


vidakovic:
Note: I am adding more documentation on how to use this with FINERACT-1470

galovics:
@vidakovic it's missing the license header. It looks good but I'm not really sure I like the idea of providing an example setenv.shand here's why. 
I feel like people use what's available and I honestly think people will try to use the setenv.sh file as it is. And the thing is, it contains hardcoded values for the docker-compose MySQL setup (hostname, password, etc).
What I'd really like instead of it is having a documentation section for Tomcat deployments where we provide all the environment variables described with example values, so instead of people copy pasting stuff, they have to read through it and configure it for their environment.
Maybe I'm overthinking this, not sure.

vidakovic:
To stay with your argument: you could say the same of the docker compose file in the root folder. That said: I am not too religious about where to put it. But even as documentation people will copy and paste and you'll end up in the same place. And: if they have nothing in place they end up with the default the default values anyway... which are exactly the same as in this setenv.sh.
Something though needs to be done, because there are more WAR users out there that don't know how to translate the Docker env vars for their deployments.
I would leave it in place, add it to documentation and emphasize that this is an example config.

galovics:
@vidakovic alright.

galovics (approved):

galovics (approved):

PR #2308, opened 2022-05-02
Filter for instances read, write, batch

josehernandezfintecheandomx:
## Description

Use of `FineractProperties` to enable or disable Fineract Instance types to Read, Write and Batch jobs execution

The env variables are:

```
FINERACT_MODE_READ_ENABLED
FINERACT_MODE_WRITE_ENABLED
FINERACT_MODE_BATCH_ENABLED
```

The values are boolean, so the values allowed are: `true` or `false` 
If there is not any of these env variables, Fineract will works as default

There is a custom error code: `error.msg.invalid.instance.type` 

For instance in a Read Only Fineract instance If you send a POST method you will see the next error:

Screen Shot 2022-05-03 at 17 51 50

## Checklist


fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/SecurityConfig.java
```diff
@@ -80,6 +87,11 @@ protected void configure(HttpSecurity http) throws Exception {
                 .addFilterAfter(tenantAwareBasicAuthenticationFilter(), SecurityContextPersistenceFilter.class) //
                 .addFilterAfter(twoFactorAuthenticationFilter, BasicAuthenticationFilter.class); //
 
+        // Apply the HTTP method Filter only If the instance is Read or Write
+        if (fineractProperties.getMode().isReadOnlyMode() || fineractProperties.getMode().isWriteEnabled()) {
```
galovics: Nah, let's apply the filter everytime. No need to exclude it when it's not running in in a certain mode.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/filters/HttpMethodFilter.java
```diff
@@ -0,0 +1,81 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Scope;
+import org.springframework.stereotype.Component;
+import org.springframework.web.filter.GenericFilterBean;
+
+@Provider
+@Component
+@Scope("singleton")
+public class HttpMethodFilter extends GenericFilterBean {
```
galovics: Can we rename the class to represent what it does instead of how it's implemented?

Instead of HttpMethodFilter, we could do let's say InstanceTypeApiFilter or something.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/filters/HttpMethodFilter.java
```diff
@@ -0,0 +1,81 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Scope;
+import org.springframework.stereotype.Component;
+import org.springframework.web.filter.GenericFilterBean;
+
+@Provider
+@Component
+@Scope("singleton")
```
galovics: Why the Scope annotation?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/filters/HttpMethodFilter.java
```diff
@@ -0,0 +1,81 @@
...
+public class HttpMethodFilter extends GenericFilterBean {
+
+    private static final Logger LOG = LoggerFactory.getLogger(HttpMethodFilter.class);
```
galovics: Lombok `@Slf4j` annotation please.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/filters/HttpMethodFilter.java
```diff
@@ -0,0 +1,81 @@
...
+
+    private final String DEFAULT_HTTP_METHODS_READ = "GET";
```
galovics: Yeah, this is not gonna fly with me, hardcoding these. There's an HttpMethod enum for this reason both in Spring and in the Apache Commons lib.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/filters/HttpMethodFilter.java
```diff
@@ -0,0 +1,81 @@
...
+    private final String DEFAULT_HTTP_METHODS_WRITE = "POST,PUT,DELETE,OPTIONS";
```
galovics: Please don't. If we want to represent a Set structure, let's use a set.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/filters/HttpMethodFilter.java
```diff
@@ -0,0 +1,81 @@
...
+    private final String DEFAULT_HTTP_METHODS_ALL = "GET,POST,PUT,DELETE,OPTIONS";
+
+    private final String fineractInstanceType;
```
galovics: Why is this needed as a String? I don't get it.
You could just get a reference to the FineractProperties (which exactly tells you which instance type you're running) and use it.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/filters/HttpMethodFilter.java
```diff
@@ -0,0 +1,81 @@
...
+    private final List<String> httpMethodsAllowedList;
+
+    @Autowired
+    public HttpMethodFilter(final FineractProperties fineractProperties) {
```
galovics: Yeah, this is not the way, especially with the String checks and stuff.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/filters/HttpMethodFilter.java
```diff
@@ -0,0 +1,81 @@
...
+            this.httpMethodsAllowedList = Arrays.asList(DEFAULT_HTTP_METHODS_WRITE.toUpperCase().split(","));
+        } else {
+            this.fineractInstanceType = "Read/Write";
+            this.httpMethodsAllowedList = Arrays.asList(DEFAULT_HTTP_METHODS_ALL.toUpperCase().split(","));
+        }
+        LOG.info("Fineract Instance type: {} {} methods", this.fineractInstanceType, this.httpMethodsAllowedList.size());
+    }
+
+    @Override
+    public void doFilter(final ServletRequest req, final ServletResponse res, final FilterChain chain)
```
galovics: A pseudo code for this doFilter would look the following:
```
doFilter {
  if (isGetMethod() && isReadInstance()) {
    proceed()
  } else if (isAnythingElse() && isWriteInstance()) {
    proceed()
  } else {
    error()
  }
}
```
And later on if we need an exception list we could have a list of API paths like:
```
Set<String> EXCEPTION_API_PATHS = Sets.of(
    "/loans", ...) // just as an example of course
```
Then the doFilter's beginning gets extended with
```
if (EXCEPTION_API_PATHS.contains(currentApiPath)) {
   proceed()
}
```

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/filters/HttpMethodFilter.java
galovics: Also, any reason you didn't use a OncePerRequestFilter but a GenericFilterBean?

galovics (changes requested):
Tests are completely missing.

galovics:
There's one thing I'm also missing. Batch job APIs should be only callable from batch instances.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/filters/HttpMethodFilter.java
josehernandezfintecheandomx: Done! FineractInstanceModeApiFilter

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/filters/HttpMethodFilter.java
josehernandezfintecheandomx: Removed the Scope annotation

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/filters/HttpMethodFilter.java
josehernandezfintecheandomx: Done!

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/filters/HttpMethodFilter.java
josehernandezfintecheandomx: Done!

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/filters/HttpMethodFilter.java
josehernandezfintecheandomx: It was based in similar approach as the Tenant Identifier header

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/filters/HttpMethodFilter.java
josehernandezfintecheandomx: Done!

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/filters/HttpMethodFilter.java
josehernandezfintecheandomx: Done! some methods were included in FineractProperties

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/filters/HttpMethodFilter.java
josehernandezfintecheandomx: Done!

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/filters/HttpMethodFilter.java
josehernandezfintecheandomx: Done!

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java
```diff
@@ -60,5 +60,25 @@ public static class FineractModeProperties {
         public boolean isReadOnlyMode() {
             return readEnabled && !writeEnabled && !batchEnabled;
         }
+
+        public boolean isWriteOnlyMode() {
```
galovics, line 64: Why do we need this?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java
```diff
@@ -60,5 +60,25 @@ public static class FineractModeProperties {
...
+            return !readEnabled && writeEnabled && !batchEnabled;
+        }
+
+        public boolean isBatchOnlyMode() {
```
galovics, line 68: Why do we need this?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java
```diff
@@ -60,5 +60,25 @@ public static class FineractModeProperties {
...
+            return !readEnabled && !writeEnabled && batchEnabled;
+        }
+
+        public boolean isReadInstance() {
```
galovics, line 72: What about `isReadEnabled()`? It's already available on the class.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java
```diff
@@ -60,5 +60,25 @@ public static class FineractModeProperties {
...
+            return isReadOnlyMode() || readEnabled;
+        }
+
+        public boolean isWriteInstance() {
```
galovics, line 76: `isWriteEnabled()`?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java
```diff
@@ -60,5 +60,25 @@ public static class FineractModeProperties {
...
+            return isWriteOnlyMode() || writeEnabled;
+        }
+
+        public boolean isBatchInstance() {
```
galovics, line 80: `isBatchEnabled()`?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/data/ApiGlobalErrorResponse.java
```diff
@@ -72,7 +78,17 @@ public static ApiGlobalErrorResponse invalidTenantIdentifier() {
...
         globalErrorResponse.setUserMessageGlobalisationCode("error.msg.invalid.tenant.identifier");
-        globalErrorResponse.setDefaultUserMessage("Invalide tenant identifier provided with request.");
+        globalErrorResponse.setDefaultUserMessage("Invalid tenant identifier provided with request.");
+
+        return globalErrorResponse;
+    }
+
+    public static ApiGlobalErrorResponse invalidInstanceTypeMethod(final String method) {
+        final ApiGlobalErrorResponse globalErrorResponse = new ApiGlobalErrorResponse();
+        globalErrorResponse.setHttpStatusCode("405");
```
galovics: HttpStatus enum pls.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exceptionmapper/InvalidTenantIdentifierExceptionMapper.java
```diff
@@ -42,7 +42,7 @@ public class InvalidTenantIdentifierExceptionMapper implements ExceptionMapper<I
 
     @Override
     public Response toResponse(@SuppressWarnings("unused") final InvalidTenantIdentiferException e) {
-        return Response.status(Status.UNAUTHORIZED).entity(ApiGlobalErrorResponse.invalidTenantIdentifier())
+        return Response.status(Status.METHOD_NOT_ALLOWED).entity(ApiGlobalErrorResponse.invalidTenantIdentifier())
```
galovics: Why the change?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/filter/FineractInstanceModeApiFilter.java
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.fineract.infrastructure.security.data.PlatformRequestLog;
+import org.apache.http.HttpStatus;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import org.springframework.web.filter.GenericFilterBean;
+
+@Provider
+@Component
+@Slf4j
+public class FineractInstanceModeApiFilter extends GenericFilterBean {
```
galovics: `OncePerRequestFilter`?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/filter/FineractInstanceModeApiFilter.java
```diff
@@ -0,0 +1,86 @@
...
+    public void doFilter(final ServletRequest req, final ServletResponse res, final FilterChain chain)
+            throws IOException, ServletException {
+
+        final HttpServletRequest request = (HttpServletRequest) req;
+        final HttpServletResponse response = (HttpServletResponse) res;
+
+        log.debug("{} {} {} {}", request.getMethod(), fineractProperties.getMode().isReadEnabled(),
+                fineractProperties.getMode().isWriteEnabled(), fineractProperties.getMode().isBatchEnabled());
+
+        final StopWatch task = new StopWatch();
```
galovics: Why do we need this?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/filter/FineractInstanceModeApiFilter.java
```diff
@@ -0,0 +1,86 @@
...
+        task.start();
+
+        if (fineractProperties.getMode().isReadInstance() && isReadMethod(request)) {
+            chain.doFilter(request, response);
+        } else if (fineractProperties.getMode().isWriteInstance() && isWriteMethod(request)) {
+            chain.doFilter(request, response);
+        } else {
+            task.stop();
+            final PlatformRequestLog platformRequestLog = PlatformRequestLog.from(task, request);
```
galovics: Why do we need this `PlatformRequestLog`?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/filter/FineractInstanceModeApiFilter.java
```diff
@@ -0,0 +1,86 @@
...
+            log.debug("{}", this.toApiJsonSerializer.serialize(platformRequestLog));
```
galovics: This is definitely something we don't want. I mean why do we want to turn an object into a JSON within the logs? It's gonna be unreadable anyway.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/filter/FineractInstanceModeApiFilter.java
```diff
@@ -0,0 +1,86 @@
...
+            response.getWriter().write(errorResponse.toJson());
+        }
+    }
+
+    private boolean isReadMethod(HttpServletRequest request) {
+        return HttpMethod.GET.equals(request.getMethod());
+    }
+
+    private boolean isWriteMethod(HttpServletRequest request) {
+        return !HttpMethod.GET.equals(request.getMethod());
```
galovics: Maybe negating the isReadMethod is more readable, like `!isReadMethod()`

galovics (commented):
I'm still missing the implementation for limiting the batch job APIs to only batch instances.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/filter/FineractInstanceModeApiFilter.java
josehernandezfintecheandomx: Removed

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exceptionmapper/InvalidTenantIdentifierExceptionMapper.java
josehernandezfintecheandomx: updated

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java
josehernandezfintecheandomx: Due mixing the setup, you could have Read only or Read/Write or all and in all cases the instance is a Read instance

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java
josehernandezfintecheandomx: Due mixing the setup, you could have Write only or Read/Write or all and in all cases the instance is a Write instance

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java
josehernandezfintecheandomx: Due mixing the setup, you could have Write only or Read/Write or all but just in the first case you will have Write Only, then block the Read and Batch Jobs

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java
josehernandezfintecheandomx: Due mixing the setup, you could have Batch only or Read/Write with Batch or all but just in the first case you will have Batch Only, then block the Read and Write

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java
josehernandezfintecheandomx: Due mixing the setup, you could have Batch only or Read/Write with Batch or all and in all cases the instance is a Batch instance

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/filter/FineractInstanceModeApiFilter.java
josehernandezfintecheandomx: Removed

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/filter/FineractInstanceModeApiFilter.java
josehernandezfintecheandomx: Removed

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/data/ApiGlobalErrorResponse.java
josehernandezfintecheandomx: Done

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exceptionmapper/InvalidTenantIdentifierExceptionMapper.java
josehernandezfintecheandomx: Done

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/filter/FineractInstanceModeApiFilter.java
josehernandezfintecheandomx: Done

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/filter/FineractInstanceModeApiFilter.java
josehernandezfintecheandomx: Done

fineract-provider/src/main/java/org/apache/fineract/infrastructure/jobs/api/SchedulerJobApiResource.java
```diff
@@ -74,20 +77,23 @@ public class SchedulerJobApiResource {
     private final ToApiJsonSerializer<JobDetailHistoryData> jobHistoryToApiJsonSerializer;
     private final PortfolioCommandSourceWritePlatformService commandsSourceWritePlatformService;
     private final PlatformSecurityContext context;
+    private final FineractProperties fineractProperties;
 
     @Autowired
```
galovics: Let's use Lombok to generate the constructor (RequiredArgsConstructor).

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/filter/FineractInstanceModeApiFilter.java
```diff
@@ -0,0 +1,65 @@
...
+
+@Provider
+@Component
+public class FineractInstanceModeApiFilter extends OncePerRequestFilter {
+
+    @Autowired
+    private FineractProperties fineractProperties;
+
+    @Override
+    public void destroy() {
```
galovics: We don't need the destroy method override, do we?

galovics (changes requested):
Please add test cases for covering the read/write/batch APIs.

galovics:
Also, please do a rebase from latest develop.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java
taskain7: ((readEnabled && !writeEnabled && !batchEnabled) || readEnabled) is equivalent with (readEnabled)

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java
taskain7: same as isReadInstance()

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java
taskain7: same as isReadInstance

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/filter/FineractInstanceModeApiFilter.java
```diff
@@ -0,0 +1,65 @@
...
+
+    }
+
+    @Override
+    protected void doFilterInternal(final HttpServletRequest request, final HttpServletResponse response, final FilterChain filterChain)
+            throws ServletException, IOException {
+
+        if (fineractProperties.getMode().isReadInstance() && isReadMethod(request)) {
```
taskain7, line 46: As far as I'm aware read APIs should be allowed for write instances as well.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/filter/FineractInstanceModeApiFilter.java
```diff
@@ -0,0 +1,65 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import org.springframework.web.filter.OncePerRequestFilter;
+
+@Provider
+@Component
+public class FineractInstanceModeApiFilter extends OncePerRequestFilter {
+
+    @Autowired
+    private FineractProperties fineractProperties;
```
taskain7: constructor injection is more preffered than property injection

fineract-provider/src/main/java/org/apache/fineract/infrastructure/jobs/api/SchedulerJobApiResource.java
josehernandezfintecheandomx: Done!

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/filter/FineractInstanceModeApiFilter.java
josehernandezfintecheandomx: Done!

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/filter/FineractInstanceModeApiFilter.java
josehernandezfintecheandomx: Done! use @RequiredArgsConstructor lombok annotaton

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/filter/FineractInstanceModeApiFilter.java
josehernandezfintecheandomx: Until I understand there will be a separation for each case, I mean Read is just Read, Write is just Write and Batch is for the Quartz Jobs. You can mixed, Read/Write or Read/Batch or Write/Batch or all (default)

galovics:
@josehernandezfintecheandomx Please add test cases for covering the read/write/batch APIs.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/SecurityConfig.java
```diff
@@ -78,6 +85,7 @@ protected void configure(HttpSecurity http) throws Exception {
                 .sessionCreationPolicy(SessionCreationPolicy.STATELESS) //
                 .and() //
                 .addFilterAfter(tenantAwareBasicAuthenticationFilter(), SecurityContextPersistenceFilter.class) //
+                .addFilterAfter(fineractInstanceModeApiFilter, TenantAwareBasicAuthenticationFilter.class) //
```
adamsaghy: Please reorder the filter. Right now the "instance mode filter" is between the basic auth and twofactor auth which logically is incorrect. Either it is before all authentication or after...

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java
```diff
@@ -60,5 +60,25 @@ public static class FineractModeProperties {
...
+        public boolean isWriteOnlyMode() {
+            return !readEnabled && writeEnabled && !batchEnabled;
+        }
+
+        public boolean isBatchOnlyMode() {
+            return !readEnabled && !writeEnabled && batchEnabled;
+        }
+
+        public boolean isReadInstance() {
+            return isReadOnlyMode() || readEnabled;
```
adamsaghy, line 73: This logic is faulty. Just because  the "readEnabled" the instance might not be "read instance"...

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/data/ApiGlobalErrorResponse.java
```diff
@@ -18,15 +18,22 @@
...
 import java.util.List;
+import javax.ws.rs.core.Response.Status;
 
 /**
  *
  */
 public class ApiGlobalErrorResponse {
 
+    @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "dd-MM-yyyy hh:mm:ss")
+    private Date timestamp;
```
adamsaghy: It should be not Java util Date, rather  java.time.DateTime

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/data/ApiGlobalErrorResponse.java
```diff
@@ -187,7 +204,7 @@ public static ApiGlobalErrorResponse serviceUnavailable(final String globalisati
     }
 
     protected ApiGlobalErrorResponse() {
-        //
+        timestamp = new Date();
```
adamsaghy: This is wrong. You must provide proper timezone (in this case tenant timezone) or it will use the system timezone which is not the wanted behaviour.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/filter/FineractInstanceModeApiFilter.java
adamsaghy: This filter logic is faulty. If the node is neither read instance, neither write instance, so it is batch instance, should still be able to accept API request, see: Run Job API

josehernandezfintecheandomx:
> 

@galovics We have review that and with the current test implementation with Gradle/Cargo is not possible to change the Fineract setup. That means we can not change the Fineract Properties from the Integration Tests (as Fineract client)

galovics (commented):
A few test cases I'm still missing:

- Write only instance, read API works
- Write only instance, batch API doesn't work
- Write and batch instance, read API doesn't work

Please add them and then we can merge.

josehernandezfintecheandomx:
> A few test cases I'm still missing:
> 
> * Write only instance, read API works
> * Write only instance, batch API doesn't work
> * Write and batch instance, read API doesn't work
> 
> Please add them and then we can merge.

Due the Database initial state, we just can add the instance mode as Write Only, and with that we have the next three tests as showed in the next image:

Screen Shot 2022-05-30 at 21 29 59

galovics:
@josehernandezfintecheandomx just by looking at the code, it seems that a batch only instance will not be able to process any of the API calls since the Filter will reject those. Any chance you've tested this use-case?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/filter/FineractInstanceModeApiFilter.java
adamsaghy: @josehernandezfintecheandomx The above use case is still not covered...

josehernandezfintecheandomx:
> @josehernandezfintecheandomx just by looking at the code, it seems that a batch only instance will not be able to process any of the API calls since the Filter will reject those. Any chance you've tested this use-case?

It has been added with the line and the use of the request PathInfo:
`if (fineractProperties.getMode().isBatchInstance() && request.getPathInfo().startsWith("/jobs"))`

adamsaghy (approved):

galovics (approved):

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/SecurityConfig.java
josehernandezfintecheandomx: Done!

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java
josehernandezfintecheandomx: The logic is to know If the instance might read, so there are two options:
1. ReadOnly, that means not write and not batch
2. Read enabled, that means Write and Read enabled, Batch and Read enabled or all (default) Write, Read, Batch
In both cases is a ReadInstance

Read Instance not means ReadOnly instance

galovics (approved):

PR #2314, opened 2022-05-05
Fix for datatable serialization genericResultSet in false

josehernandezfintecheandomx:
## Description

Fix for datatable serialization `genericResultSet` in `false` in the GET datatable data endpoint, due date format exception for `datetime` values

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/GenericDataServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/GenericDataServiceImpl.java
@@ -25,6 +25,7 @@ import java.util.ArrayList;
 import java.util.List;
 import javax.sql.DataSource;
 import org.apache.fineract.infrastructure.core.exception.PlatformDataIntegrityException;
+import org.apache.fineract.infrastructure.core.service.DateUtils;
 import org.apache.fineract.infrastructure.core.service.RoutingDataSource;
 import org.apache.fineract.infrastructure.core.service.database.DatabaseIndependentQueryService;
 import org.apache.fineract.infrastructure.dataqueries.data.GenericResultsetData;
@@ -32,8 +33,6 @@ import org.apache.fineract.infrastructure.dataqueries.data.ResultsetColumnHeader
 import org.apache.fineract.infrastructure.dataqueries.data.ResultsetColumnValueData;
 import org.apache.fineract.infrastructure.dataqueries.data.ResultsetRowData;
 import org.apache.fineract.infrastructure.dataqueries.exception.DatatableNotFoundException;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.dao.DataAccessException;
 import org.springframework.jdbc.core.JdbcTemplate;
@@ -47,7 +46,6 @@ public class GenericDataServiceImpl implements GenericDataService {
     private final JdbcTemplate jdbcTemplate;
     private final DataSource dataSource;
     private final DatabaseIndependentQueryService databaseIndependentQueryService;
-    private static final Logger LOG = LoggerFactory.getLogger(GenericDataServiceImpl.class);
 
     @Autowired
     public GenericDataServiceImpl(final RoutingDataSource dataSource, final JdbcTemplate jdbcTemplate,
@@ -169,9 +167,10 @@ public class GenericDataServiceImpl implements GenericDataService {
                             writer.append(
                                     "[" + localDate.getYear() + ", " + localDate.getMonthValue() + ", " + localDate.getDayOfMonth() + "]");
                         } else if (currColType.equals("DATETIME")) {
-                            final LocalDateTime localDateTime = LocalDateTime.parse(currVal);
+                            final LocalDateTime localDateTime = LocalDateTime.parse(formatDateTimeValue(currVal),
+                                    DateUtils.DEFAULT_DATETIME_FORMATER);
                             writer.append("[" + localDateTime.getYear() + ", " + localDateTime.getMonthValue() + ", "
-                                    + localDateTime.getDayOfMonth() + " " + localDateTime.getHour() + ", " + localDateTime.getMinute()
+                                    + localDateTime.getDayOfMonth() + ", " + localDateTime.getHour() + ", " + localDateTime.getMinute()
                                     + ", " + localDateTime.getSecond() + ", " + localDateTime.get(ChronoField.MILLI_OF_SECOND) + "]");
                         } else {
                             writer.append(doubleQuote + replace(currVal, doubleQuote, slashDoubleQuote) + doubleQuote);
@@ -261,5 +260,15 @@ public class GenericDataServiceImpl implements GenericDataService {
         } catch (IllegalArgumentException e) {
             throw new DatatableNotFoundException(datatable);
         }
+    }
+
+    private String formatDateTimeValue(String dateTimeValue) {
+        if (dateTimeValue.length() > 19) {
+            dateTimeValue = dateTimeValue.substring(0, 19);
+        }
+        if (dateTimeValue.contains("T")) {
+            dateTimeValue = dateTimeValue.replace("T", " ");
+        }
+        return dateTimeValue;
     }
 }
--- a/integration-tests/src/test/java/org/apache/fineract/integrationtests/common/organisation/EntityDatatableChecksIntegrationTest.java
+++ b/integration-tests/src/test/java/org/apache/fineract/integrationtests/common/organisation/EntityDatatableChecksIntegrationTest.java
@@ -82,6 +82,26 @@ public class EntityDatatableChecksIntegrationTest {
     }
 
     @Test
+    public void validateCreateDeleteDatatableCheck() {
+        // creating datatable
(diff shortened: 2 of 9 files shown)


fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/GenericDataServiceImpl.java
```diff
@@ -169,7 +170,8 @@ public String generateJsonFromGenericResultsetData(final GenericResultsetData gr
                             writer.append(
                                     "[" + localDate.getYear() + ", " + localDate.getMonthValue() + ", " + localDate.getDayOfMonth() + "]");
                         } else if (currColType.equals("DATETIME")) {
-                            final LocalDateTime localDateTime = LocalDateTime.parse(currVal);
+                            final DateTimeFormatter simpleDateFormat = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss");
```
galovics: Let's extract this into a constant and name the variable what it represent. We don't need to waste an object creation every time we reach this execution path.

galovics (changes requested):

galovics:
Please also add an integration test-case for this.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/GenericDataServiceImpl.java
josehernandezfintecheandomx: Done!

galovics:
@josehernandezfintecheandomx could you please add a corresponding test case?

josehernandezfintecheandomx:
> @josehernandezfintecheandomx could you please add a corresponding test case?

Done test case added @galovics

integration-tests/src/test/java/org/apache/fineract/integrationtests/common/organisation/EntityDatatableChecksIntegrationTest.java
```diff
@@ -81,6 +81,26 @@ public void setup() {
         this.datatableHelper = new DatatableHelper(this.requestSpec, this.responseSpec);
     }
 
+    @Test
+    public void validateCreateReadDeleteDatatableCheck() {
```
galovics: I don't think this test-case covers the changed functionality that the datetimes are converted properly.
Could you please explain if I'm wrong there and point out how it's covered?
Thanks!

integration-tests/src/test/java/org/apache/fineract/integrationtests/common/system/DatatableHelper.java
```diff
@@ -48,6 +48,19 @@ public String createDatatable(final String apptableName, final boolean multiRow)
                 getTestDatatableAsJSON(apptableName, multiRow), "resourceIdentifier");
     }
 
+    public Integer createDatatableEntry(final String apptableName, final String datatableName, final Integer apptableId,
```
galovics, line 54: This method is not called anywhere in your test.

integration-tests/src/test/java/org/apache/fineract/integrationtests/common/system/DatatableHelper.java
```diff
@@ -48,6 +48,19 @@ public String createDatatable(final String apptableName, final boolean multiRow)
...
+            final boolean genericResultSet, final String dateFormat) {
+        return Utils.performServerPost(
+                this.requestSpec, this.responseSpec, DATATABLE_URL + "/" + datatableName + "/" + apptableId + "?genericResultSet="
+                        + Boolean.toString(genericResultSet) + "&" + Utils.TENANT_IDENTIFIER,
+                getTestDatatableEntryAsJSON(dateFormat), "resourceId");
+    }
+
+    public String readDatatableEntry(final String datatableName, final Integer resourceId, final boolean genericResultset) {
```
galovics, line 62: This method is not called anywhere in your test.

galovics (changes requested):

galovics:
The tests are failing: `DatatableIntegrationTest.validateCreateReadDeleteDatatableCheck(DatatableIntegrationTest.java:72)`

integration-tests/src/test/java/org/apache/fineract/integrationtests/common/system/DatatableIntegrationTest.java
```diff
@@ -0,0 +1,83 @@
...
+        final Integer clientID = ClientHelper.createClientAsPerson(requestSpec, responseSpec);
+
+        // creating new entity datatable check
+        final boolean genericResultSet = true;
+        Integer datatableResourceID = this.datatableHelper.createDatatableEntry(CLIENT_APP_TABLE_NAME, datatableName, clientID,
+                genericResultSet, DATE_TIME_FORMAT);
+        assertNotNull(datatableResourceID, "ERROR IN CREATING THE ENTITY DATATABLE RECORD");
+
+        // Read the Datatable entry generated with genericResultSet in true (default)
+        this.datatableHelper.readDatatableEntry(datatableName, clientID, genericResultSet);
```
galovics: Please make sure that if you read the datatable entry you assert on the result.

integration-tests/src/test/java/org/apache/fineract/integrationtests/common/system/DatatableIntegrationTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+        // Read the Datatable entry generated with genericResultSet in false
+        this.datatableHelper.readDatatableEntry(datatableName, clientID, !genericResultSet);
```
galovics: Same as above.

galovics (changes requested):

integration-tests/src/test/java/org/apache/fineract/integrationtests/common/organisation/EntityDatatableChecksIntegrationTest.java
josehernandezfintecheandomx: We have created a new test case special for the Datatable operations

integration-tests/src/test/java/org/apache/fineract/integrationtests/common/system/DatatableIntegrationTest.java
josehernandezfintecheandomx: As part of the Datatable action were included in the Test case, Datatable creation including a datetime field, add new record to this datatable, read in the two cases (genericResultSet in True and False) then finally delete de datatable created

integration-tests/src/test/java/org/apache/fineract/integrationtests/common/system/DatatableIntegrationTest.java
josehernandezfintecheandomx: The previous explanation applies here too

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/GenericDataServiceImpl.java
```diff
@@ -262,4 +261,14 @@ private SqlRowSet getDatatableMetaData(final String datatable) {
             throw new DatatableNotFoundException(datatable);
         }
     }
+
+    private String formatDateTimeValue(String dateTimeValue) {
```
galovics, line 265: Even though this solves the problem for now, this logic shouldn't be here.
On one note, we should definitely not decide based on the string representation of the datetime value, neither based on the length and whether it contains a `T`.

I looked at it a tiny bit deeper and we're completely missing the standardization on the datatable reading side.

If you look at `ReadWriteNonCoreDataServiceImpl#fillDatatableResultSetDataRows`, we're always reading the column value as string even though it's not necessarily a string.

In case of Date values, we could simply read them as Date type and then we don't have to deal with the non-sense DateTime parsing here.
It'd simplify a lot of things and set up proper boundaries on reading and processing the data.

integration-tests/src/test/java/org/apache/fineract/integrationtests/common/system/DatatableIntegrationTest.java
galovics: I understand the test case. My point is that the result of the `datatableHelper.readDatatableEntry()` method call is never checked.
For example we should check the amount of rows in the result along with their value.

galovics (changes requested):

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/GenericDataServiceImpl.java
josehernandezfintecheandomx: There are three cases due three different databases, two are same length the third one has more data then is more longer, and in between the two similars the difference is the T
Maybe you are right in terms to have a better solution reading the dates as dates, numbers as numbers instead of just strings, but It can to impact no just the API, can to impact the front end too

galovics: Yeah so my idea is to read dates as dates from the database and everything else as strings (for now at least) and use the existing date format on the API level so we don't break anything.

How I'd imagine, this is where we are right now:
Database -> string reading -> preparing the generic resultset -> format the strings -> send back on the API
Instead, let's do
Database -> string or date reading -> preparing the generic resultset -> format the date -> send back on the API

Hope that explains.

PR #2320, opened 2022-05-09
FINERACT-1471: AsciiDoc extension for read/write/batch separation

taskain7:
## Description

The AsciiDoc has been extended for read/write/batch separation.

### DIFF
--- a/fineract-doc/src/docs/en/chapters/deployment/index.adoc
+++ b/fineract-doc/src/docs/en/chapters/deployment/index.adoc
@@ -8,6 +8,8 @@ include::local.adoc[leveloffset=+1]
 
 include::appserver.adoc[leveloffset=+1]
 
+include::instance-type.adoc[leveloffset=+1]
+
 include::kubernetes.adoc[leveloffset=+1]
 
 include::aws.adoc[leveloffset=+1]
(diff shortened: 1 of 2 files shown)


fineract-doc/src/docs/en/chapters/development/instance-type.adoc
```diff
@@ -0,0 +1,178 @@
+= Fineract Instance types
+
+In cases where Fineract has to deal with high load, it can cause a performance problem for a single Fineract instance.
+To overcome this problem, Fineract instances can be started in different instance types for better scalability and performance in a multi-instance environment:
+
+Read instance
```
galovics: Let's make this a proper list: https://docs.asciidoctor.org/asciidoc/latest/lists/unordered/

fineract-doc/src/docs/en/chapters/development/index.adoc
```diff
@@ -6,6 +6,8 @@ include::git.adoc[leveloffset=+1]
 
 include::gpg.adoc[leveloffset=+1]
 
+include::instance-type.adoc[leveloffset=+1]
```
galovics: I think this section should rather be in the architecture one (or maybe deployment). Thoughts?

galovics (changes requested):
Other than those 2 comments, it's a good starting point. Let's fix 'em and we're good.

fineract-doc/src/docs/en/chapters/development/index.adoc
taskain7: Yeah, it could be any of them, but I would vote for deployment rather than architecture.

galovics (approved):

PR #2339, opened 2022-05-28
Fix NullPointerException at Base64$Encoder in ImagesApiResource (FINERACT-1224)

temi-ro:
## Description

Should fix the NullPointerException at Base64$Encoder.encode() in ImagesApiResource from https://issues.apache.org/jira/browse/FINERACT-1224 .

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/documentmanagement/api/ImagesApiResource.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/documentmanagement/api/ImagesApiResource.java
@@ -61,8 +61,9 @@ import org.springframework.stereotype.Component;
 @Component
 @Scope("singleton")
 @Path("{entity}/{entityId}/images")
-
 public class ImagesApiResource {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ImagesApiResource.class);
 
     private final PlatformSecurityContext context;
     private final ImageReadPlatformService imageReadPlatformService;
@@ -165,12 +166,12 @@ public class ImagesApiResource {
 
         try {
             byte[] resizedImageBytes = resizedImage.getByteSource().read();
-
-            if(resizedImageBytes != null){
+            if (resizedImageBytes != null) {
                 final String clientImageAsBase64Text = imageDataURISuffix + Base64.getMimeEncoder().encodeToString(resizedImageBytes);
                 return Response.ok(clientImageAsBase64Text, MediaType.TEXT_PLAIN_TYPE).build();
-            }else{
-                return Response.noContent().build();
+            } else {
+                LOG.error("resizedImageBytes is null for entityName={}, entityId={}, maxWidth={}, maxHeight={}", entityName, entityId, maxWidth, maxHeight);
+                return Response.serverError().build();
             }
         } catch (IOException e) {
             throw new ContentManagementException(imageData.name(), e.getMessage(), e);
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/documentmanagement/api/ImagesApiResource.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/documentmanagement/api/ImagesApiResource.java
@@ -54,6 +54,8 @@ import org.apache.fineract.portfolio.client.data.ClientData;
 import org.glassfish.jersey.media.multipart.FormDataBodyPart;
 import org.glassfish.jersey.media.multipart.FormDataContentDisposition;
 import org.glassfish.jersey.media.multipart.FormDataParam;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Scope;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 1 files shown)


vorburger:
@temi-ro can you please fix the code format and add the missing spaces so that it's pretty? 😄 

You can either "manually" fix it, or use the command that it shows that does it for you.

Please don't hesitate to let me know if you need any help.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/documentmanagement/api/ImagesApiResource.java
```diff
@@ -165,8 +165,13 @@ public Response retrieveImage(@PathParam("entity") final String entityName, @Pat
 
         try {
             byte[] resizedImageBytes = resizedImage.getByteSource().read();
-            final String clientImageAsBase64Text = imageDataURISuffix + Base64.getMimeEncoder().encodeToString(resizedImageBytes);
-            return Response.ok(clientImageAsBase64Text, MediaType.TEXT_PLAIN_TYPE).build();
+
```
vorburger: could you also remove this empty line?

vorburger (changes requested):

fineract-provider/src/main/java/org/apache/fineract/infrastructure/documentmanagement/api/ImagesApiResource.java
```diff
@@ -165,8 +165,13 @@ public Response retrieveImage(@PathParam("entity") final String entityName, @Pat
...
+            if(resizedImageBytes != null){
```
galovics: Not sure if this is right to handle this way.

I assume the resizedImageBytes is only null when there wasn't any image to resize. So I guess if we handle that use-case this way, the octet-stream branch might still fail.

I'd rather ask the question, is it reasonable to deal with these type of images?

Maybe at the beginning of this API, we shall simply put a check for this case and return an HTTP 400 as retrieving an image that doesn't exist (to some sense) is not necessarily something that makes sense for me, at least now.

Do you mind explaining the exact use-case for this extra logic?

galovics (changes requested):
Whatever will be the implementation at the end, please add test cases to cover the changed functionality. Thanks.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/documentmanagement/api/ImagesApiResource.java
```diff
@@ -165,8 +165,13 @@ public Response retrieveImage(@PathParam("entity") final String entityName, @Pat
...
+                final String clientImageAsBase64Text = imageDataURISuffix + Base64.getMimeEncoder().encodeToString(resizedImageBytes);
+                return Response.ok(clientImageAsBase64Text, MediaType.TEXT_PLAIN_TYPE).build();
+            }else{
+                return Response.noContent().build();
```
vorburger: `serverError()` seems more appropriate than `noContent()` in this particular case, I think

Also, let's log the fact that this happenend...

temi-ro: Done!

vorburger (approved):
LGTM as a first improvement for the NPE in FINERACT-1224. (I'll reply to @galovics next.)

fineract-provider/src/main/java/org/apache/fineract/infrastructure/documentmanagement/api/ImagesApiResource.java
vorburger: @galovics Hey! Thanks for jumping on this review - on a Saturday no less. So, the context here is that @temi-ro is a student who I'm teaching how to FLOSS and GitHub... we picked FINERACT-1224 as an example, and this is his very first contribution. We haven't gotten to actually running Fineract, and uploading images, and fully reproducing the issue. This would take more time than we have together today. The question is if this PR as-is (once it passed the build, of course) could be acceptable to you as a first step to at least avoid the NPE and clearer diagnostic log, as a first step? Still better than the NPE, IMHO. (I may follow-up with @temi-ro to further improve this case when I show him how to actually run everything etc.)

galovics: Sure thing. Thanks for the context @vorburger.

galovics (approved):

vorburger:
@temi-ro The build is still failing, probably because it's missing the import for the Logger. Do you still want to fix that? Then I can merge this, as your very first (real, useful, actually fixing a small bug in a real world open source project) Pull Request!

galovics:
> @temi-ro The build is still failing, probably because it's missing the import for the Logger. Do you still want to fix that? Then I can merge this, as your very first (real, useful, actually fixing a small bug in a real world open source project) Pull Request!

@vorburger @temi-ro you guys could even use Lombok there with the `@Slf4j` annotation, then no need to explicitly create the logger.

temi-ro:
@galovics at the moment I tried to use the Logger, since I have never seen Lombok and I am short in time today. Moreover, I am not sure if the checks will pass since on my laptop (Windows), there are some errors in relation with the line separators...

vorburger:
Build still fails, but now it's just due to what looks like a "flaky" test (@temi-ro that means a test failure NOT caused by changed you made here). I searched JIRA for it, and found FINERACT-1557 about it, which I've taken the liberty to re-open (hope that's OK). I suggest that we do not merge this PR until someone (else) has a change to fix that and we have a clean build again which is green for this and future PRs? (Flaky test failures can be very confusing for newcomers to projects.)

temi-ro:
Ok! I see.

vorburger:
I've now gone ahead and merged this - after I've convinced myself that this change does not actually break that `ClientLoanIntegrationTest`, by rebasing this change on today's `develop`, and seen it build green in #2346. (So the build failure seen on this PR was either a temporary flaky test, or was meanwhile fixed.)

@temi-ro note how a project typically may not merge your 3 Git commits from this PR exactly as-is; what I did is called a "Squash and Merge" - combining your 3 separate commits into 1 new single commit, with slightly updating wording. You are still "attributed" as the author though. So congratulations on your very very first contribution to an open source project! 🥳 🎉

PR #2340, opened 2022-05-28
Enable/Disable Messaging in Read Only instance mode

josehernandezfintecheandomx:
## Description

When a Fineract instance is running as read-only, all event receiving/sending shall be disabled

In case of a read-only instance, the best way to prevent events from being sent and received is to simply disable the messaging configuration (JMS/etc), that way there’s not even a connection established for the brokers.

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/EnableFineractEventsCondition.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/EnableFineractEventsCondition.java
@@ -1,0 +1,38 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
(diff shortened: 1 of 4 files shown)


fineract-provider/src/main/java/org/apache/fineract/notification/config/MessagingConfiguration.java
```diff
@@ -35,6 +36,7 @@
 
 @Configuration
 @Profile("activeMqEnabled")
+@ConditionalOnExpression("${fineract.mode.write-enabled:true} || ${fineract.mode.batch-enabled:true}")
```
galovics: I wish we could use programmatic conditions here with the FineractModeProperties POJO.

Before merging, please look at this: https://stackoverflow.com/a/69901997/2863995
And let's try to make it work.

@taskain7 already created a programmatic Condition here: `FineractModeValidationCondition`

josehernandezfintecheandomx: Done! Condition created: `EnableFineractEventsCondition`

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/EnableFineractEventsCondition.java
```diff
@@ -0,0 +1,40 @@
...
+package org.apache.fineract.infrastructure.core.config;
+
+import java.util.Optional;
+import lombok.RequiredArgsConstructor;
+import org.springframework.context.annotation.Condition;
+import org.springframework.context.annotation.ConditionContext;
+import org.springframework.core.type.AnnotatedTypeMetadata;
+
+@RequiredArgsConstructor
+public class EnableFineractEventsCondition implements Condition {
```
galovics, line 26: Nice job. I don't think we need the `@RequiredArgsConstructor` because we don't have any fields.

Also, since this is now a programmatic condition, we can introduce *unit* tests to cover the behavior. Please add it and then we're good.

josehernandezfintecheandomx: The annotation has been removed, which kind of UT we can create? Remember If we don't have write mode in the UT we have not database

galovics (approved):

PR #2344, opened 2022-05-31
Fineract with ReadOnly tenant database connection

josehernandezfintecheandomx:
## Description

With read separation, there’s a possibility to use read-only database connections for read-only instances.

Screen Shot 2022-06-02 at 23 07 12

## Checklist

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -539,6 +539,18 @@ configure(project.fineractJavaProjects) {
                 excludeTestsMatching project.property('excludeTests')
             }
         }
+
+        // Database type used in the Test
+        systemProperty 'dbType', 'mariadb'
+        if (project.hasProperty('dbType')) {
+            if ('postgresql'.equalsIgnoreCase(dbType)) {
+                systemProperty 'dbType', 'postgresql'
+            } else if ('mysql'.equalsIgnoreCase(dbType)) {
+                systemProperty 'dbType', 'mysql'
+            } else {
+                throw new GradleException('Provided dbType is not supported')
+            }
+        }
     }
 
     testlogger {
--- a/fineract-provider/build.gradle
+++ b/fineract-provider/build.gradle
@@ -223,6 +223,7 @@ bootRun {
     dependencies {
         implementation 'org.mariadb.jdbc:mariadb-java-client:2.7.5'
         implementation 'org.postgresql:postgresql:42.3.5'
+        implementation 'mysql:mysql-connector-java:8.0.29'
     }
 }
 
@@ -288,6 +289,7 @@ jib {
     dependencies {
         implementation 'org.mariadb.jdbc:mariadb-java-client:2.7.5'
         implementation 'org.postgresql:postgresql:42.3.5'
+        implementation 'mysql:mysql-connector-java:8.0.29'
     }
 
     pluginExtensions {
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/TomcatJdbcDataSourcePerTenantService.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/TomcatJdbcDataSourcePerTenantService.java
@@ -18,11 +18,6 @@
  */
 package org.apache.fineract.infrastructure.core.service;
 
-import static org.apache.fineract.infrastructure.core.domain.FineractPlatformTenantConnection.toJdbcUrl;
-import static org.apache.fineract.infrastructure.core.domain.FineractPlatformTenantConnection.toProtocol;
-
-import com.zaxxer.hikari.HikariConfig;
-import com.zaxxer.hikari.HikariDataSource;
 import java.util.HashMap;
 import java.util.Map;
 import javax.sql.DataSource;
@@ -44,12 +39,13 @@ public class TomcatJdbcDataSourcePerTenantService implements RoutingDataSourceSe
     private final Map<Long, DataSource> tenantToDataSourceMap = new HashMap<>(1);
     private final DataSource tenantDataSource;
 
-    @Autowired
-    private HikariConfig hikariConfig;
+    private final DataSourcePerTenantServiceFactory dataSourcePerTenantServiceFactory;
 
     @Autowired
-    public TomcatJdbcDataSourcePerTenantService(final @Qualifier("hikariTenantDataSource") DataSource tenantDataSource) {
+    public TomcatJdbcDataSourcePerTenantService(final @Qualifier("hikariTenantDataSource") DataSource tenantDataSource,
+            final DataSourcePerTenantServiceFactory dataSourcePerTenantServiceFactory) {
         this.tenantDataSource = tenantDataSource;
+        this.dataSourcePerTenantServiceFactory = dataSourcePerTenantServiceFactory;
     }
 
     @Override
@@ -68,42 +64,12 @@ public class TomcatJdbcDataSourcePerTenantService implements RoutingDataSourceSe
                 if (possibleDS != null) {
                     tenantDataSource = possibleDS;
                 } else {
-                    tenantDataSource = createNewDataSourceFor(tenantConnection);
+                    tenantDataSource = dataSourcePerTenantServiceFactory.createNewDataSourceFor(this.tenantDataSource, tenantConnection);
                     this.tenantToDataSourceMap.put(tenantConnection.getConnectionId(), tenantDataSource);
                 }
             }
         }
 
         return tenantDataSource;
-    }
-
-    // creates the tenant data source for the oltp and report database
-    private DataSource createNewDataSourceFor(final FineractPlatformTenantConnection tenantConnectionObj) {
-        String protocol = toProtocol(this.tenantDataSource);
-        String jdbcUrl = toJdbcUrl(protocol, tenantConnectionObj.getSchemaServer(), tenantConnectionObj.getSchemaServerPort(),
-                tenantConnectionObj.getSchemaName(), tenantConnectionObj.getSchemaConnectionParameters());
-
-        HikariConfig config = new HikariConfig();
-        config.setDriverClassName(hikariConfig.getDriverClassName());
-        config.setPoolName(tenantConnectionObj.getSchemaName() + "_pool");
-        config.setJdbcUrl(jdbcUrl);
-        config.setUsername(tenantConnectionObj.getSchemaUsername());
-        config.setPassword(tenantConnectionObj.getSchemaPassword());
-        config.setMinimumIdle(tenantConnectionObj.getInitialSize());
-        config.setMaximumPoolSize(tenantConnectionObj.getMaxActive());
-        config.setConnectionTestQuery(hikariConfig.getConnectionTestQuery());
-        config.setValidationTimeout(tenantConnectionObj.getValidationInterval());
-        config.setAutoCommit(hikariConfig.isAutoCommit());
-
-        // https://github.com/brettwooldridge/HikariCP/wiki/MBean-(JMX)-Monitoring-and-Management
-        config.setRegisterMbeans(true);
-
-        // https://github.com/brettwooldridge/HikariCP/wiki/MySQL-Configuration
-        // These are the properties for each Tenant DB; the same configuration
-        // is also in src/main/resources/META-INF/spring/hikariDataSource.xml
-        // for the all Tenants DB -->
-        config.setDataSourceProperties(hikariConfig.getDataSourceProperties());
-
-        return new HikariDataSource(config);
     }
 }
--- a/fineract-provider/src/main/resources/db/changelog/tenant-store/parts/0004_readonly_database_connection.xml
+++ b/fineract-provider/src/main/resources/db/changelog/tenant-store/parts/0004_readonly_database_connection.xml
@@ -22,6 +22,24 @@
 <databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
                    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.1.xsd">
-     <include file="parts/0003_reset_postgresql_sequences.xml" relativeToChangelogFile="true"/>
-     <include file="parts/0004_readonly_database_connection.xml" relativeToChangelogFile="true"/>
+    <changeSet author="fineract" id="add-readonly-database-connection-properties" context="tenant_store_db">
+        <addColumn tableName="tenant_server_connections">
+            <column name="readonly_schema_server" type="VARCHAR(100)"/>
+        </addColumn>
+        <addColumn tableName="tenant_server_connections">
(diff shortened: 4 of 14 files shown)


fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/TomcatJdbcDataSourcePerTenantService.java
```diff
@@ -80,15 +86,28 @@ public DataSource retrieveDataSource() {
...
+        String schemaPassword = tenantConnectionObj.getSchemaPassword();
+        // Properties to ReadOnly case
+        if (this.fineractProperties.getMode().isReadOnlyMode()) {
+            schemaServer = tenantConnectionObj.getReadOnlySchemaServer();
+            schemaPort = tenantConnectionObj.getReadOnlySchemaServerPort();
+            schemaName = tenantConnectionObj.getReadOnlySchemaName();
+            schemaUsername = tenantConnectionObj.getReadOnlySchemaUsername();
+            schemaPassword = tenantConnectionObj.getReadOnlySchemaPassword();
+        }
+        String jdbcUrl = toJdbcUrl(protocol, schemaServer, schemaPort, schemaName, tenantConnectionObj.getSchemaConnectionParameters());
```
galovics: I think we should be able to provide custom connection parameters just for the readonly DB, dont you think?

galovics: I think we should be able to provide custom connection parameters just for the readonly DB, dont you think?

fineract-provider/src/main/resources/db/changelog/tenant-store/parts/0004_readonly_database_connection.xml
```diff
@@ -0,0 +1,47 @@
...
+        <addColumn tableName="tenant_server_connections">
+            <column name="readonly_schema_username" type="VARCHAR(10)"/>
+        </addColumn>
+        <addColumn tableName="tenant_server_connections">
+            <column name="readonly_schema_password" type="VARCHAR(100)"/>
+        </addColumn>
+    </changeSet>
+    <changeSet author="fineract" id="set-readonly-database-connection-properties" context="tenant_store_db">
+        <sql>
+            UPDATE tenant_server_connections SET readonly_schema_server=schema_server, readonly_schema_name=schema_name, readonly_schema_server_port=schema_server_port, readonly_schema_username=schema_username, readonly_schema_password=schema_password
```
galovics: Well, I'm not sure if we want to "lie" here. I'd say the read only columns could stay null and when the app starts up in a read-only mode, it smartly recognizes that  if not all of the readonly parameters are filled, it simply falls back to the normal parameters.
Thoughts?

galovics: Well, I'm not sure if we want to "lie" here. I'd say the read only columns could stay null and when the app starts up in a read-only mode, it smartly recognizes that  if not all of the readonly parameters are filled, it simply falls back to the normal parameters.
Thoughts?

fineract-provider/src/main/resources/db/changelog/tenant-store/parts/0004_readonly_database_connection.xml
```diff
@@ -0,0 +1,47 @@
...
+            <column name="readonly_schema_server" type="VARCHAR(100)"/>
+        </addColumn>
+        <addColumn tableName="tenant_server_connections">
+            <column name="readonly_schema_name" type="VARCHAR(100)"/>
+        </addColumn>
+        <addColumn tableName="tenant_server_connections">
+            <column name="readonly_schema_server_port" type="VARCHAR(10)"/>
+        </addColumn>
+        <addColumn tableName="tenant_server_connections">
+            <column name="readonly_schema_username" type="VARCHAR(10)"/>
```
galovics: I think VARCHAR(10) is too small here. Should be VARCHAR(100) at least.

galovics: I think VARCHAR(10) is too small here. Should be VARCHAR(100) at least.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/TomcatJdbcDataSourcePerTenantService.java
```diff
@@ -80,15 +86,28 @@ public DataSource retrieveDataSource() {
...
-        String jdbcUrl = toJdbcUrl(protocol, tenantConnectionObj.getSchemaServer(), tenantConnectionObj.getSchemaServerPort(),
-                tenantConnectionObj.getSchemaName(), tenantConnectionObj.getSchemaConnectionParameters());
+        // Default properties for Writing
+        String schemaServer = tenantConnectionObj.getSchemaServer();
+        String schemaPort = tenantConnectionObj.getSchemaServerPort();
+        String schemaName = tenantConnectionObj.getSchemaName();
+        String schemaUsername = tenantConnectionObj.getSchemaUsername();
+        String schemaPassword = tenantConnectionObj.getSchemaPassword();
+        // Properties to ReadOnly case
+        if (this.fineractProperties.getMode().isReadOnlyMode()) {
```
galovics: I know I say this all the time but this should be covered by unit tests.

galovics: I know I say this all the time but this should be covered by unit tests.

galovics (changes requested):

fineract-provider/src/main/resources/db/changelog/tenant-store/parts/0004_readonly_database_connection.xml
adamsaghy: Agree with you here. We should set either the proper values or null everywhere (if read only table functionality not to be used).
However i would go even further and do it this way:
- Read the values from variables (same as we did for the default tenant)
- Add runOnChange=true 
- Add a precondition which removes the actual entry
- Add onFail="WARN" to the precondition as if it is the first time when the changeset is executed there is no row to be deleted
- Add the new entry (read only db settings)

What we earn:
- This changeset will not fail when the configuration changes
- Our tenant connection settings will be updated on changes automatically

If we are happy with this approach, i would suggest to apply the same changes to the original tenant connection changeset.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/TomcatJdbcDataSourcePerTenantService.java
adamsaghy: Yeah...true. that one was missing from the POC as well...
However can it happen the connection parameters to be different from the "original"? I mean "similar" DB, same data.

fineract-provider/src/main/resources/db/changelog/tenant-store/parts/0004_readonly_database_connection.xml
josehernandezfintecheandomx: Done!

josehernandezfintecheandomx: Done!

fineract-provider/src/main/resources/db/changelog/tenant-store/parts/0004_readonly_database_connection.xml
josehernandezfintecheandomx: The default values will be read like this:
`.append(" COALESCE(ts.readonly_schema_connection_parameters, ts.schema_connection_parameters) as readOnlySchemaConnectionParameters ")`

It was added the Connection Parameters field `readonly_schema_connection_parameters` for ReadOnly datasource

josehernandezfintecheandomx: The default values will be read like this:
`.append(" COALESCE(ts.readonly_schema_connection_parameters, ts.schema_connection_parameters) as readOnlySchemaConnectionParameters ")`

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/TomcatJdbcDataSourcePerTenantService.java
josehernandezfintecheandomx: How we can run the ReadOnly instance mode type, remember that If the instance mode is ReadOnly, Liquibase is not executed in the startup time and then we'll have not an initial datatabase schema.

josehernandezfintecheandomx: How we can run the ReadOnly instance mode type, remember that If the instance mode is ReadOnly, Liquibase is not executed in the startup time and then we'll have not an initial database schema.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/TomcatJdbcDataSourcePerTenantService.java
josehernandezfintecheandomx: Done! the `readOnlySchemaConnectionParameters` have been added in to the table and the class

fineract-provider/src/main/resources/db/changelog/tenant-store/parts/0004_readonly_database_connection.xml
adamsaghy: @josehernandezfintecheandomx I dont see any changes here... have you committed?

fineract-provider/src/main/resources/db/changelog/tenant/parts/0014_remove_unused_jobs.xml
```diff
@@ -21,6 +21,9 @@
 -->
 <databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.1.xsd">
     <changeSet author="fineract" id="1">
+        <sql>
```
adamsaghy, line 26: Blocker change!
Please revert this.

@josehernandezfintecheandomx I believe we should never remove history entries!

Also, please never modify an already existing and committed changeset!
Reason: it will fail for everyone who already pulled that changeset

josehernandezfintecheandomx: @adamsaghy 
OK! I can revert without problem, but It'll fails for everyone that wants to run this with run_job_history data, that was the reason that I updated this, the next screen shows the error

Screen Shot 2022-06-02 at 21 34 39

galovics: @taskain7 can you please take a look at the above?

galovics (changes requested):

fineract-provider/src/main/resources/db/changelog/tenant/parts/0014_remove_unused_jobs.xml
saghytech: @josehernandezfintecheandomx in this case, you are right. This is a must. However pls add a "validChecksum" as well, that way it will not fail for others (who might already executed this changelog)

adamsaghy: @josehernandezfintecheandomx in this case, you are right. This is a must. However pls add a "validChecksum" as well, that way it will not fail for others (who might already executed this changelog)

taskain7: nice catch @josehernandezfintecheandomx , I didn't find this issue when I made this change.
IMHO we can keep this delete in this PR, just make sure that the migration can run for anybody (who already ran my changes and aalso who didn't)

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/TomcatJdbcDataSourcePerTenantService.java
josehernandezfintecheandomx: Yes! And It's already considered in the code too. The logic is to take the current properties as default, but If the instance is ReadOnly we'll get the readonly values from the next three places:

1. Database table field values,
2. System env property, If the previous is null
3. Reuse the Database table field values (non readonly) If the previous are null

Screen Shot 2022-06-03 at 18 40 31

fineract-provider/src/main/resources/db/changelog/tenant/parts/0014_remove_unused_jobs.xml
josehernandezfintecheandomx: Resolved using validChecksum as suggested @adamsaghy 

Screen Shot 2022-06-03 at 19 12 49

FYI @galovics / @taskain7

adamsaghy: @josehernandezfintecheandomx In the meantime this problem was solved by an another strategy. The FK was removed from the job_run_history table, so we dont need to delete the history entry of the "deprecated" jobs.


Would you be so kind to remove this whole changeset from this PR?

adamsaghy: @josehernandezfintecheandomx In the meantime this problem was solved by an another strategy. The FK was removed from the job_run_history table, so we dont need to delete the history entry of the "deprecated" jobs.


Would you be so kind to remove this whole changeset from this PR?

build.gradle
```diff
@@ -539,6 +539,18 @@ configure(project.fineractJavaProjects) {
...
+
+        // Database type used in the Test
+        systemProperty 'dbType', 'mariadb'
+        if (project.hasProperty('dbType')) {
+            if ('postgresql'.equalsIgnoreCase(dbType)) {
+                systemProperty 'dbType', 'postgresql'
+            } else if ('mysql'.equalsIgnoreCase(dbType)) {
+                systemProperty 'dbType', 'mysql'
+            } else {
+                throw new GradleException('Provided dbType is not supported')
```
adamsaghy, line 551: What if the customer set 'mariadb' as the dbType? I am afraid it will fail now.

fineract-provider/build.gradle
```diff
@@ -223,6 +223,7 @@ bootRun {
     dependencies {
         implementation 'org.mariadb.jdbc:mariadb-java-client:2.7.5'
         implementation 'org.postgresql:postgresql:42.3.5'
+        implementation 'mysql:mysql-connector-java:8.0.29'
```
adamsaghy, line 226: Isnt this violating the License?

galovics:
Closing. The content of this PR is tackled in #2359

PR #2367, opened 2022-06-17
Delete fineract.dev.yaml (for now, may be restored again later)

vorburger:
Due to https://github.com/vorburger/www.fineract.dev/issues/15, which I still haven't sorted out yet, this action, despite been green on https://github.com/apache/fineract/actions/workflows/fineract.dev.yaml, currently actually does NOT update https://www.fineract.dev anymore. The only thing this has been doing for the past 6 months is it keeps sending me GCP notification about GCB failures... ;-) Let's therefore remove this? I'll propose reverting this removal when / if I get around to fixing https://github.com/vorburger/www.fineract.dev/issues/15.


galovics (approved):
Sounds reasonable to me. We don't want you @vorburger to get spammed by failure notifications. :))

PR #2369, opened 2022-06-19
Implement integration tests for Accounting rules

josehernandezfintecheandomx:
## Description

Implement integration tests for Accounting rules

Screen Shot 2022-06-19 at 10 23 59

## Checklist

### DIFF
--- a/integration-tests/src/test/java/org/apache/fineract/integrationtests/accounting/AccountingRuleIntegrationTest.java
+++ b/integration-tests/src/test/java/org/apache/fineract/integrationtests/accounting/AccountingRuleIntegrationTest.java
@@ -1,0 +1,79 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.integrationtests.accounting;
+
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import io.restassured.builder.RequestSpecBuilder;
+import io.restassured.builder.ResponseSpecBuilder;
+import io.restassured.http.ContentType;
+import io.restassured.specification.RequestSpecification;
+import io.restassured.specification.ResponseSpecification;
+import java.util.ArrayList;
+import org.apache.fineract.client.models.GetAccountRulesResponse;
+import org.apache.fineract.client.models.GetOfficesResponse;
+import org.apache.fineract.client.models.PostAccountingRulesResponse;
+import org.apache.fineract.integrationtests.common.OfficeHelper;
+import org.apache.fineract.integrationtests.common.Utils;
+import org.apache.fineract.integrationtests.common.accounting.Account;
+import org.apache.fineract.integrationtests.common.accounting.AccountHelper;
+import org.apache.fineract.integrationtests.common.accounting.AccountRuleHelper;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class AccountingRuleIntegrationTest {
+
+    private ResponseSpecification responseSpec;
+    private RequestSpecification requestSpec;
+
+    private AccountHelper accountHelper;
+    private AccountRuleHelper accountRuleHelper;
+
+    @BeforeEach
+    public void setup() {
+        Utils.initializeRESTAssured();
+
+        requestSpec = new RequestSpecBuilder().setContentType(ContentType.JSON).build();
+        requestSpec.header("Authorization", "Basic " + Utils.loginIntoServerAndGetBase64EncodedAuthenticationKey());
+        responseSpec = new ResponseSpecBuilder().expectStatusCode(200).build();
+
+        accountRuleHelper = new AccountRuleHelper(requestSpec, responseSpec);
+        accountHelper = new AccountHelper(requestSpec, responseSpec);
+    }
+
+    @Test
+    public void testAccountingRuleCreation() {
+        // given
+        final Account accountToCredit = accountHelper.createIncomeAccount();
+        final Account accountToDebit = accountHelper.createExpenseAccount();
+        final GetOfficesResponse headOffice = OfficeHelper.getHeadOffice(requestSpec, responseSpec);
+
+        // when
+        final PostAccountingRulesResponse accountingRule = accountRuleHelper.createAccountRule(headOffice.getId(), accountToCredit,
+                accountToDebit);
+        final ArrayList<GetAccountRulesResponse> accountingRules = accountRuleHelper.getAccountingRules();
+
+        // then
+        assertNotNull(accountingRule);
+        assertNotNull(accountingRule.getResourceId());
+        assertNotNull(accountingRules);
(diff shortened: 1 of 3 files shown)


integration-tests/src/test/java/org/apache/fineract/integrationtests/common/accounting/AccountRuleHelper.java
```diff
@@ -0,0 +1,50 @@
...
+    public AccountRuleHelper(final RequestSpecification requestSpec, final ResponseSpecification responseSpec) {
+        this.requestSpec = requestSpec;
+        this.responseSpec = responseSpec;
+    }
+
+    public ArrayList<HashMap> getAccountingRules() {
+        return Utils.performServerGet(this.requestSpec, this.responseSpec, ACCOUNTINGRULES_URL, "$");
+    }
+
+    public Integer createAccountRule(final Long officeId, final Account accountToCredit, final Account accountToDebit) {
```
galovics: Since this is a new helper and a new domain to create tests for, do you mind switching the old "hashmap" style implementation to using the typed fineract-client generated code? 
For example if you look at `org.apache.fineract.integrationtests.common.ClientHelper#createClientAddress`, I've been using the generated typed code. You could do the same for the AccountingRules.
Thoughts?

josehernandezfintecheandomx: Done! Using the `GetAccountRulesResponse` class

integration-tests/src/test/java/org/apache/fineract/integrationtests/accounting/AccountingRuleIntegrationTest.java
```diff
@@ -0,0 +1,80 @@
...
+        responseSpec = new ResponseSpecBuilder().expectStatusCode(200).build();
+
+        accountRuleHelper = new AccountRuleHelper(requestSpec, responseSpec);
+        accountHelper = new AccountHelper(requestSpec, responseSpec);
+    }
+
+    @Test
+    public void testAccountingRuleCreation() {
+        // given
+        accountToCredit = accountHelper.createIncomeAccount();
```
galovics: Let's keep this as local variables. It's important to not store state on an object/class level to prevent potential concurrency issues in the future (if we decide to run the tests in parallel mode)

integration-tests/src/test/java/org/apache/fineract/integrationtests/common/accounting/AccountRuleHelper.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    public ArrayList<GetAccountRulesResponse> getAccountingRules() {
+        final String response = Utils.performServerGet(this.requestSpec, this.responseSpec, ACCOUNTINGRULES_URL);
+        Type accountRuleListType = new TypeToken<ArrayList<GetAccountRulesResponse>>() {}.getType();
+        return GSON.fromJson(response, accountRuleListType);
+    }
+
+    public Integer createAccountRule(final Long officeId, final Account accountToCredit, final Account accountToDebit) {
+        final String assetAccountJSON = new AccountingRuleBuilder()
```
galovics, line 53: Why don't we use the generated models here as well?

galovics (changes requested):

integration-tests/src/test/java/org/apache/fineract/integrationtests/accounting/AccountingRuleIntegrationTest.java
josehernandezfintecheandomx: Done!

integration-tests/src/test/java/org/apache/fineract/integrationtests/common/accounting/AccountRuleHelper.java
josehernandezfintecheandomx: Done!

integration-tests/src/test/java/org/apache/fineract/integrationtests/accounting/AccountingRuleIntegrationTest.java
```diff
@@ -0,0 +1,79 @@
...
+    public void testAccountingRuleCreation() {
+        // given
+        final Account accountToCredit = accountHelper.createIncomeAccount();
+        final Account accountToDebit = accountHelper.createExpenseAccount();
+        final GetOfficesResponse headOffice = OfficeHelper.getHeadOffice(requestSpec, responseSpec);
+
+        // when
+        final PostAccountingRulesResponse accountingRule = accountRuleHelper.createAccountRule(headOffice.getId(), accountToCredit,
+                accountToDebit);
+        assertNotNull(accountingRule);
```
adamsaghy: Usually the checks are going under the "then" case not the "when"... but its minor

integration-tests/src/test/java/org/apache/fineract/integrationtests/accounting/AccountingRuleIntegrationTest.java
```diff
@@ -0,0 +1,79 @@
...
+        assertNotNull(accountingRule.getResourceId());
+
+        // then
+        final ArrayList<GetAccountRulesResponse> accountingRules = accountRuleHelper.getAccountingRules();
+        assertNotNull(accountingRules);
+        assertEquals(1, accountingRules.size());
```
adamsaghy: This is not recommended way of checking existence of a created entry. You cannot be sure whether it was not there already earlier, also you cannot be sure whether any other integration test will not created an account rules as well and if that one is executed earlier, this one will fail.

My recommendation is to check the actual id and some of the most important part of the account rules details.

integration-tests/src/test/java/org/apache/fineract/integrationtests/accounting/AccountingRuleIntegrationTest.java
josehernandezfintecheandomx: But the check is for the Get Accounting rules, so the Accounting Rule creation is before of the check

integration-tests/src/test/java/org/apache/fineract/integrationtests/accounting/AccountingRuleIntegrationTest.java
josehernandezfintecheandomx: In this case, there were not any Integration Test for Accounting rules, so It is brand new, to cover the test for the Get Accounting Rules API that was failing with the SQL using 0 and 1 instead of boolean values as It was fixed in the PR https://github.com/apache/fineract/pull/2335

Screen Shot 2022-06-28 at 16 32 09

integration-tests/src/test/java/org/apache/fineract/integrationtests/accounting/AccountingRuleIntegrationTest.java
adamsaghy: Like i said, it is nothing major... but usually:
GIVEN
           everything to prepare the context before executing the targeted actions
 WHEN
          execute all the targeted actions
 THEN
          check the outcome of the targeted actions

integration-tests/src/test/java/org/apache/fineract/integrationtests/accounting/AccountingRuleIntegrationTest.java
adamsaghy: I understand there is no any integration tests for accounting rules...yet. 
1. But what if i created one tomorrow and as the execution order of the tests are not specified it could happen that the new test is executed earlier and if that integration test is creating an accounting rule, your test will fail.
2. You are not testing whether the accounting rule were created properly as there are missing checks for:
- officeId
- accountCreditId
- accountDebitId
- name
- description

My recommendation is to avoid checks like:
- assertEquals(1, accountingRules.size());

And preferrably find the created accounting rule (from the list or retrieve it by ID) entry and check its details one by one

integration-tests/src/test/java/org/apache/fineract/integrationtests/accounting/AccountingRuleIntegrationTest.java
josehernandezfintecheandomx: Done!

integration-tests/src/test/java/org/apache/fineract/integrationtests/accounting/AccountingRuleIntegrationTest.java
josehernandezfintecheandomx: Done!

galovics (approved):

PR #2380, opened 2022-06-24
FINERACT-1430:ClientSocialStatus

rrpawar96:
JIRA: https://issues.apache.org/jira/browse/FINERACT-1430


galovics (changes requested):
Please update the PR with the latest develop changes.
Also, please add test cases to verify this functionality.
Thanks.

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/service/ClientWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -459,13 +466,14 @@ public CommandProcessingResult updateClient(final Long clientId, final JsonComma
                 clientForUpdate.updateSavingsProduct(savingsProductId);
             }
 
-            if (changes.containsKey(ClientApiConstants.genderIdParamName)) {
```
adamsaghy, line 462: @rrpawar96 Why did you override the update logic of gender?

PR #2402, opened 2022-07-03
FINERACT-1684: Business date part 3

adamsaghy:
## Description

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/data/ResultsetColumnHeaderData.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/data/ResultsetColumnHeaderData.java
@@ -179,7 +179,8 @@ public final class ResultsetColumnHeaderData implements Serializable {
     }
 
     private boolean isChar() {
-        return "char".equalsIgnoreCase(this.columnType) || "CHARACTER VARYING".equalsIgnoreCase(this.columnType);
+        return "char".equalsIgnoreCase(this.columnType) || "CHARACTER VARYING".equalsIgnoreCase(this.columnType)
+                || "bpchar".equalsIgnoreCase(this.columnType);
     }
 
     private boolean isVarchar() {
@@ -201,7 +202,8 @@ public final class ResultsetColumnHeaderData implements Serializable {
     }
 
     private boolean isInteger() {
-        return "integer".equalsIgnoreCase(this.columnType) || "int4".equalsIgnoreCase(this.columnType);
+        return "integer".equalsIgnoreCase(this.columnType) || "int2".equalsIgnoreCase(this.columnType)
+                || "int4".equalsIgnoreCase(this.columnType);
     }
 
     private boolean isSmallInt() {
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/GenericDataServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/GenericDataServiceImpl.java
@@ -24,6 +24,7 @@ import java.time.temporal.ChronoField;
 import java.util.ArrayList;
 import java.util.List;
 import javax.sql.DataSource;
+import lombok.extern.slf4j.Slf4j;
 import org.apache.fineract.infrastructure.core.exception.PlatformDataIntegrityException;
 import org.apache.fineract.infrastructure.core.service.DateUtils;
 import org.apache.fineract.infrastructure.core.service.RoutingDataSource;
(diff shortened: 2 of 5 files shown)


fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/database/DatabaseSpecificSQLGenerator.java
```diff
@@ -168,4 +168,14 @@ public String currentSchema() {
             throw new IllegalStateException("Database type is not supported for current schema " + databaseTypeResolver.databaseType());
         }
     }
+
+    public String getBoolean(boolean condition) {
+        if (databaseTypeResolver.isMySQL()) {
```
galovics: Do we need this? MySQL supports `true` and `false` values.

fineract-provider/src/main/java/org/apache/fineract/notification/eventandlistener/SpringNotificationEventListener.java
```diff
@@ -38,6 +39,7 @@ public class SpringNotificationEventListener implements ApplicationListener<Noti
     public void onApplicationEvent(NotificationEvent event) {
         log.debug("Processing Spring notification event {}", event);
         ThreadLocalContextUtil.init(event.getContext());
+        ThreadLocalContextUtil.setActionContext(ActionContext.COB);
```
galovics: Do you mind explaining it why we need to have COB context here? Thanks.

adamsaghy: :o 
It's a mistake! Nice catch!

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/database/DatabaseSpecificSQLGenerator.java
adamsaghy: Mysql is not supporting true or false :) It supports only the tinyint (1/0)

galovics: https://dev.mysql.com/doc/refman/5.7/en/boolean-literals.html @adamsaghy

galovics: We have true/false values in native queries all over the codebase and they work flawlessly.

adamsaghy: i have the feeling, i might learn something new today :)

adamsaghy: Well...the truth lies between.
select * from stretchy_report where self_service_user_report = '0'; -> Valid
select * from stretchy_report where self_service_user_report = '1'; -> Valid
select * from stretchy_report where self_service_user_report = 0; -> Valid
select * from stretchy_report where self_service_user_report = 1; -> Valid
select * from stretchy_report where self_service_user_report = false; -> Valid
select * from stretchy_report where self_service_user_report = true; -> Valid
select * from stretchy_report where self_service_user_report = 'false'; -> Valid, but by sheer luck (no error, but evaluate as false)
select * from stretchy_report where self_service_user_report = 'true'; -> Invalid, no error and evaluate it as false

Unfortunately, the stretchy report SQLs are using sometimes as a string this value.

galovics: Right but the thing is, if they search for the String version of true/false/1/0 that's already indicating a very bad pattern to me which should be fixed, don't you think?

adamsaghy: I can accept that, but... it is not "worse" pattern like providing a boolean to the tinyInt (which is number).
Let me fix that 5 occasion then :)

galovics: >  it is not "worse" pattern like providing a boolean to the tinyInt 

I don't agree with that. If the database engine provides the true/false literals and it'll translate them into 1/0 values, that's much better if you're really representing boolean values instead of using the numbers themselves. It's just semantically better.

Same for the booleans in Java. The underlying JVM represents them as 0/1 bits and I bet you don't want to use numbers there too. ;-)

PR #2415, opened 2022-07-11
Fineract-1646

apurbraj:
FINERACT-1646 (System is posting multiple Overdraft Interest for same day)


galovics (changes requested):
Add a test-case to cover this bug please so it never reoccurs. Thanks.

PR #2424, opened 2022-07-12
FINERACT-1658:Fix Group Filter in Center

rrpawar96:
https://issues.apache.org/jira/browse/FINERACT-1658


ikimbra:
@rrpawar96 Thanks for this fix. @francisguchie kindly test this as I also test to see if it works. 

I will update the team if it is fixed then close the issue.

galovics (changes requested):
Can we add a test case for this bugfix? Thanks.

rrpawar96:
> Can we add a test case for this bugfix? Thanks.

Thanks, @galovics for the review, Test case has been added.

galovics (approved):
LGTM

PR #2436, opened 2022-07-19
FINERACT-1330: Pay Charge & withdrawals is violating Overdraft conditions on current and past Transactions

rrpawar96:
https://issues.apache.org/jira/browse/FINERACT-1330

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccount.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccount.java
@@ -1521,6 +1521,7 @@ public class SavingsAccount extends AbstractPersistableCustom {
             final List<DepositAccountOnHoldTransaction> depositAccountOnHoldTransactions, final boolean backdatedTxnsAllowedTill) {
 
         List<SavingsAccountTransaction> transactionsSortedByDate = null;
+        BigDecimal transactionAmount = null;
         if (backdatedTxnsAllowedTill) {
             transactionsSortedByDate = retrieveSortedTransactions();
         } else {
@@ -1535,6 +1536,9 @@ public class SavingsAccount extends AbstractPersistableCustom {
         Money minRequiredBalance = minRequiredBalanceDerived(getCurrency());
         LocalDate lastSavingsDate = null;
         for (final SavingsAccountTransaction transaction : transactionsSortedByDate) {
+
+            transactionAmount = transaction.getAmount();
+
             if (transaction.isNotReversed() && transaction.isCredit()) {
                 runningBalance = runningBalance.plus(transaction.getAmount(this.currency));
             } else if (transaction.isNotReversed() && transaction.isDebit()) {
@@ -1573,17 +1577,26 @@ public class SavingsAccount extends AbstractPersistableCustom {
                         throw new PlatformApiDataValidationException(dataValidationErrors);
                     }
                 }
-
             }
             lastSavingsDate = transaction.transactionLocalDate();
-
         }
 
         BigDecimal withdrawalFee = null;
-        BigDecimal transactionAmount = null;
         if (isOverdraft()) {
             if (runningBalance.minus(minRequiredBalance).isLessThanZero()) {
                 throw new InsufficientAccountBalanceException("transactionAmount", getAccountBalance(), withdrawalFee, transactionAmount);
+            }
+        }
+    }
+
+    public void validateAccountBalanceDoesNotViolateOverdraft(final List<SavingsAccountTransaction> savingsAccountTransaction,
+            final BigDecimal amountPaid) {
+        if (savingsAccountTransaction != null) {
+            SavingsAccountTransaction savingsAccountTransactionFirst = savingsAccountTransaction.get(0);
+            if (!this.allowOverdraft) {
+                if (savingsAccountTransactionFirst.getRunningBalance(this.currency).minus(amountPaid).isLessThanZero()) {
+                    throw new InsufficientAccountBalanceException("transactionAmount", getAccountBalance(), null, amountPaid);
+                }
             }
         }
     }
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccountTransactionRepository.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccountTransactionRepository.java
@@ -21,6 +21,7 @@ package org.apache.fineract.portfolio.savings.domain;
 import java.time.LocalDate;
 import java.util.List;
 import javax.persistence.LockModeType;
+import org.springframework.data.domain.Pageable;
 import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.data.jpa.repository.JpaSpecificationExecutor;
 import org.springframework.data.jpa.repository.Lock;
@@ -47,6 +48,9 @@ public interface SavingsAccountTransactionRepository
     @Lock(LockModeType.PESSIMISTIC_WRITE)
     List<SavingsAccountTransaction> findBySavingsAccount(@Param("savingsAccount") SavingsAccount savingsAccount);
 
-    @Query("select sat from SavingsAccountTransaction sat where sat.refNo = :refNo")
-    List<SavingsAccountTransaction> findAllTransactionByRefNo(@Param("refNo") String refNo);
+    List<SavingsAccountTransaction> findByRefNo(@Param("refNo") String refNo);
+
+    @Query("select sat from SavingsAccountTransaction sat where sat.savingsAccount.id = :savingsId and sat.dateOf <= :transactionDate and sat.reversed=false")
+    List<SavingsAccountTransaction> findBySavingsAccountIdAndLessThanDateOfAndReversedIsFalse(@Param("savingsId") Long savingsId,
+            @Param("transactionDate") LocalDate transactionDate, Pageable pageable);
 }
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountWritePlatformServiceJpaRepositoryImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountWritePlatformServiceJpaRepositoryImpl.java
@@ -130,6 +130,9 @@ import org.apache.fineract.useradministration.domain.AppUserRepositoryWrapper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.data.domain.PageRequest;
+import org.springframework.data.domain.Pageable;
+import org.springframework.data.domain.Sort;
 import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Propagation;
@@ -718,7 +721,7 @@ public class SavingsAccountWritePlatformServiceJpaRepositoryImpl implements Savi
         List<SavingsAccountTransaction> savingsAccountTransactions = null;
         if (isBulk) {
             String transactionRefNo = savingsAccountTransaction.getRefNo();
-            savingsAccountTransactions = this.savingsAccountTransactionRepository.findAllTransactionByRefNo(transactionRefNo);
+            savingsAccountTransactions = this.savingsAccountTransactionRepository.findByRefNo(transactionRefNo);
             reversal = this.savingsAccountDomainService.handleReversal(account, savingsAccountTransactions, backdatedTxnsAllowedTill);
         } else {
             reversal = this.savingsAccountDomainService.handleReversal(account, Collections.singletonList(savingsAccountTransaction),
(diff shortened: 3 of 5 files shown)


francisguchie:
@rrpawar96 
Please check the failed build

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccountTransactionRepository.java
```diff
@@ -44,4 +44,8 @@ List<SavingsAccountTransaction> findTransactionsAfterPivotDate(@Param("savingsAc
 
     @Query("select sat from SavingsAccountTransaction sat where sat.refNo = :refNo")
     List<SavingsAccountTransaction> findAllTransactionByRefNo(@Param("refNo") String refNo);
+
+    @Query("select sat from SavingsAccountTransaction sat where sat.savingsAccount.id = :savingsId and sat.dateOf <= :transactionDate and sat.reversed=0 ORDER BY sat.dateOf DESC")
```
galovics: `sat.reversed=0` should be `sat.reversed=false`

galovics (changes requested):

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccountTransactionRepository.java
rrpawar96: > `sat.reversed=0` should be `sat.reversed=false`

@galovics Comment Addressed, Please review.

galovics (changes requested):
Can you please add new tests to cover the bugfix? Thanks.

rrpawar96:
> Can you please add new tests to cover the bugfix? Thanks.

@galovics Test case has been added, Please re-review.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccount.java
```diff
@@ -1574,21 +1579,36 @@ public void validateAccountBalanceDoesNotBecomeNegative(final String transaction
...
+                    throw new InsufficientAccountBalanceException("chargeAmount", getAccountBalance(), null, amountPaid);
+                }
+            }
+        }
+    }
+
+    private void overdraftLimitExceededException(final String paramName, final BigDecimal accountBalance, final BigDecimal withdrawalFee,
+            final BigDecimal transactionAmount) {
+        String showMessageForLimitExaustion = "Overdraft Limit is Exceeded";
+        throw new PlatformDataIntegrityException(showMessageForLimitExaustion, showMessageForLimitExaustion);
```
adamsaghy: Is there a reason to use PlatformDataIntegrityException instead of InsufficientAccountBalanceException?
Also all of the passed parameters are ignored: paramName, accountBalance, withdrawalFee, transactionAmount. Can you please write them into log with WARN at least?

francisguchie: @adamsaghy thank you for this review, @rrpawar96 if you need my input as we address this review, please call me

rrpawar96: Thanks for the review, @adamsaghy Comment addressed.

francisguchie:
@rrpawar96 thank you.  @adamsaghy kindly review

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccount.java
```diff
@@ -1573,21 +1576,36 @@ public void validateAccountBalanceDoesNotBecomeNegative(final String transaction
...
-                throw new InsufficientAccountBalanceException("transactionAmount", getAccountBalance(), withdrawalFee, transactionAmount);
+                overdraftLimitExceededException("transactionAmount", getAccountBalance(), transactionAmount);
+            }
+        }
+    }
+
+    public void validateAccountBalanceDoesNotViolateOverdraft(final List<SavingsAccountTransaction> savingsAccountTransaction,
+            final BigDecimal amountPaid) {
+        if (savingsAccountTransaction != null) {
+            SavingsAccountTransaction savingsAccountTransactionFirst = savingsAccountTransaction.get(0);
```
adamsaghy, line 1595: @rrpawar96 I was wondering whether the actual accountBalance of the Savings account would not be better candidate here instead of the running balance of the last transaction?

I dont think we should fetch all of the transactions to only use one of them...

francisguchie: @adamsaghy , I do agree with this as this uses lots of memory space. I have seen challenges with accounts that have lots of transactions. However, the account balance is prone to inaccuracy. This means that if we are using it at this point, there must be another process class the accurately updates the account balance. 

@rrpawar96  if you remember the [issue](https://issues.apache.org/jira/browse/FINERACT-1147)

adamsaghy: @francisguchie thanks for highlighting this issue. I think if the account balance is not accurate  at particular scenarios that is a major issue and should be addressed with priority.
However if the fix is not straightforward, i can imagine a workaround where not all of the transactions are fetched, but only the very last (if the transaction running balance is more reliable than the account balance). That way the performance and memory impact would be less.

what do you think?

rrpawar96: Yes, the JPA doesn't support the 'Limit 1' keyword in the Query to fetch a single record, So, the alternative that I will be using now is Pageable Param, is there any other alternative you think would fix this part? or Pageable would be fine? @adamsaghy

adamsaghy: @rrpawar96 You can use Pageable, or something like this:
`Optional<SavingsAccount> findFirstBySavingsAccountAndReversedFalseAndDateOfEqualsOrderByIdDesc();
`
However I have changed a little bit the query... 
Provided transaction date is equals or less... do we really need the "less" part?
Order by date is not enough as multiple transaction could happen on that day. Use Id as the secondary order by param to get the last transaction.

However i was wondering what is the purpose of this query. Shall it return the very last transaction of the savings account which is not reversed? In that case we dont need the transaction date and just order by id Desc.

Please help me to understand better this query...

adamsaghy: @rrpawar96 and @francisguchie Using the account balance is really that problematic? I thought the account balance is automatically recalculated for every payment activity (transaction).

francisguchie: @adamsaghy , I think its not problematic to use the account balance

adamsaghy: @francisguchie In that case i would rather rely on the account balance instead of trying to figure out which transaction was the very last and what is the running balance of it. I mean both could work and i believe both of them should be always the same (hopefully).

rrpawar96: @adamsaghy, The account balance history on the basis of the past dates is saved in only one table i.e. "m_savings_account_transaction (SavingsAccountTransaction)" in the "running account balance" field and there is no field as "account balance" in SavingsAccountTransaction (m_savings_account_transaction table). Although the "account balance" field is in another table "m_savings_account" that keeps the update of the latest balance of the account and not of the past date (history)

adamsaghy: @rrpawar96 Yeah... i should have read more carefully the story about this change :)
Seems you're right, you need to find the very last transaction of that particular day (pay charge transaction date) and based on the running balance of that transaction decide whether it is violating the overdraft rule or not.

To do that, I would recommend this:
`Optional<SavingsAccount> findFirstBySavingsAccountAndReversedFalseAndDateOfLessThanEqualOrderByDateOfDescIdDesc();`

But this one is pretty complex, so maybe Pagination would be just fine as well:
```
@Query("select sat from SavingsAccountTransaction sat where sat.savingsAccount.id = :savingsId and sat.dateOf <= :transactionDate and sat.reversed=false ORDER BY sat.dateOf DESC, sat.id DESC")
     List<SavingsAccountTransaction> retrieveRunningBalanceBySavingsAccountIdAndDate(@Param("savingsId") Long savingsId,
             @Param("transactionDate") LocalDate transactionDate, PageRequest.of(0, 1));
```

It is still a List, but now only 1 item will be there

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccountTransactionRepository.java
```diff
@@ -49,4 +50,8 @@ List<SavingsAccountTransaction> findTransactionRunningBalanceBeforePivotDate(@Pa
 
     @Query("select sat from SavingsAccountTransaction sat where sat.refNo = :refNo")
```
vidakovic: @rrpawar96 ... why is that additional @Query annotation needed here? Should just work the same way if you rename the function to "findByRefNo"... and is shorter... and you just have to "maintain" one thing (the function name). For short queries (i. e. not too many parameters) I think using the repository function magic is nicer.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccountTransactionRepository.java
```diff
@@ -49,4 +50,8 @@ List<SavingsAccountTransaction> findTransactionRunningBalanceBeforePivotDate(@Pa
...
     List<SavingsAccountTransaction> findAllTransactionByRefNo(@Param("refNo") String refNo);
+
+    @Query("select sat from SavingsAccountTransaction sat where sat.savingsAccount.id = :savingsId and sat.dateOf <= :transactionDate and sat.reversed=false ORDER BY sat.dateOf DESC, sat.id DESC")
```
vidakovic: To be consistent with other (query) methods rename this function to: "findBySavingsAccountIdAndLessThanDateOfAndReversedIsFalse". Just to be clear: in this case you need the @Query annotation (because of the nested "sat.savingsAccount.id") and there is (I think) no way to express this query with Spring Data's "function magic". But another thing I'd do here is to do sorting via a function parameter... this is usually the first thing that changes if someone wants to reuse this query... and if sorting is defined in the annotation then it's set in  stone. So instead I'd define a function (all together):

```
@Query("select sat from SavingsAccountTransaction sat where sat.savingsAccount.id = :savingsId and sat.dateOf <= :transactionDate and sat.reversed=false")
findBySavingsAccountIdAndLessThanDateOfAndReversedIsFalse(@Param("savingsId") Long savingsId, @Param("transactionDate") LocalDate transactionDate, Pageable pageable, Sort sort)
```

Like this is more likely that this function gets reused instead of someone else needing in another situation just a insignificant change (like sort order) and creates then a similar function (if history is an indicator with yet another naming scheme)... very hard for everyone else to wrap their heads around. Of course this is not rocket science if we just look at this isolated case... but if people read larger portions of code then the meandering naming strategy we have is a bit of a problem.

Admittedly the function name is a bit long, but it's consistent with the rest of the framework. I know that we have authors who introduced "retrieveXXX" and similar prefixes, but I think we should go for consistency in naming.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -1471,6 +1472,11 @@ private SavingsAccountTransaction payCharge(final SavingsAccountCharge savingsAc
         this.savingAccountAssembler.assignSavingAccountHelpers(account);
         final Set<Long> existingTransactionIds = new HashSet<>();
         final Set<Long> existingReversedTransactionIds = new HashSet<>();
+        List<SavingsAccountTransaction> savingsAccountTransaction = this.savingsAccountTransactionRepository
```
vidakovic, line 1479: "findBy" instead of "retrieve", see previous comments.

vidakovic:
@rrpawar96 ... might seem like a minor thing, but let's be consistent with the PR titles... here the title should be:

```
FINERACT-1330: Pay Charge & withdrawals is violating Overdraft conditions on current and past Transactions
```

This ensures that things are properly matched in Jira (aka that this PR is linked to the Jira ticket).... and it's also nicer to read. The camel case thing here is hard to grok if you read the git logs... and using spaces (after the colon) is not forbidden; please keep in mind that someone needs to read this.

rrpawar96:
@vidakovic, All Comments have been addressed.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccount.java
francisguchie: @adamsaghy thanks

galovics (approved):
LGTM

PR #2481, opened 2022-08-03
[FINERACT-1678] configuration validation for COBs

taskain7:
## Description

If any COB has partition size, chunk size or thread count below 1, or the partiotion size is less than chunk size * thread count, then the startup fails on the validation.

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractPartitionJobConfigValidationCondition.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractPartitionJobConfigValidationCondition.java
@@ -1,0 +1,59 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.infrastructure.core.config;
+
+import java.util.List;
+import java.util.function.Predicate;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.context.annotation.Condition;
+import org.springframework.context.annotation.ConditionContext;
+import org.springframework.core.type.AnnotatedTypeMetadata;
+
+@Slf4j
+public class FineractPartitionJobConfigValidationCondition implements Condition {
+
+    @Override
+    public boolean matches(ConditionContext context, AnnotatedTypeMetadata metadata) {
+        FineractProperties.FineractPartitionedJob partitionedJobProperties = ExplicitConfigurationPropertiesFactory.getProperty(context,
+                "fineract.partitioned-job", FineractProperties.FineractPartitionedJob.class);
+        if (partitionedJobProperties != null) {
+            List<FineractProperties.PartitionedJobProperty> invalidConfigs = partitionedJobProperties.getPartitionedJobProperties().stream()
+                    .filter(isAnyConfigBelowOne().or(FineractPartitionJobConfigValidationCondition::isPartitionRatioInvalid)).toList();
+            if (!invalidConfigs.isEmpty()) {
+                for (FineractProperties.PartitionedJobProperty invalidConfig : invalidConfigs) {
+                    log.error(
+                            "{} partitioned job is not configured properly. The partition size, chunk size and thread count must be more than 0, and partition size must be less then chunk size * thread count",
+                            invalidConfig.getJobName());
+                }
+            }
+            return !invalidConfigs.isEmpty();
+        } else {
+            return false;
+        }
+    }
+
+    private static Predicate<FineractProperties.PartitionedJobProperty> isAnyConfigBelowOne() {
+        return partitionedJobProperty -> !(partitionedJobProperty.getPartitionSize() > 0 && partitionedJobProperty.getChunkSize() > 0
+                && partitionedJobProperty.getThreadCount() > 0);
+    }
+
+    private static boolean isPartitionRatioInvalid(FineractProperties.PartitionedJobProperty partitionedJobProperty) {
+        return partitionedJobProperty.getPartitionSize() < partitionedJobProperty.getThreadCount() * partitionedJobProperty.getChunkSize();
+    }
+}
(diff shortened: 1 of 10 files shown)


fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractPartitionJobConfigValidationCondition.java
```diff
@@ -0,0 +1,49 @@
...
+import org.springframework.core.type.AnnotatedTypeMetadata;
+
+@Slf4j
+public class FineractPartitionJobConfigValidationCondition implements Condition {
+
+    @Override
+    public boolean matches(ConditionContext context, AnnotatedTypeMetadata metadata) {
+        FineractProperties.FineractPartitionedJobProperties partitionedJobProperties = Binder.get(context.getEnvironment())
+                .bind("fineract.partitioned-job", FineractProperties.FineractPartitionedJobProperties.class).orElse(null);
+        List<FineractProperties.JobProperty> invalidConfigs = partitionedJobProperties.getJobProperties().stream()
```
galovics: null-check for `partitionedJobProperties` is required based on the `orElse(null)` call

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractPartitionJobConfigValidationCondition.java
```diff
@@ -0,0 +1,49 @@
...
+                .filter(jobProperty -> !(jobProperty.getPartitionSize() > 0 && jobProperty.getChunkSize() > 0
```
galovics: Can we extract the filter condition to a method so it's more easier to read? thanks.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractPartitionJobConfigValidationCondition.java
```diff
@@ -0,0 +1,49 @@
...
+                        && jobProperty.getThreadCount() > 0
+                        && jobProperty.getPartitionSize() >= jobProperty.getThreadCount() * jobProperty.getChunkSize()))
```
galovics: Actually 2 filter methods would be cool. One for the greater than 0 check and one for the actual logic.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractPartitionJobConfigValidationCondition.java
```diff
@@ -0,0 +1,49 @@
...
+import org.springframework.boot.context.properties.bind.Binder;
+import org.springframework.context.annotation.Condition;
+import org.springframework.context.annotation.ConditionContext;
+import org.springframework.core.type.AnnotatedTypeMetadata;
+
+@Slf4j
+public class FineractPartitionJobConfigValidationCondition implements Condition {
+
+    @Override
+    public boolean matches(ConditionContext context, AnnotatedTypeMetadata metadata) {
```
galovics, line 32: Please add a test case to verify the logic. Thanks.

galovics (dismissed):

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractPartitionJobConfigValidationCondition.java
taskain7: There is an OR relationship between the 2 conditions, so I'm afraid we have to use 1 filter method.

adamsaghy (approved):
LGTM

fineract-provider/src/main/resources/application.properties
```diff
@@ -43,6 +43,11 @@ fineract.mode.batch-manager-enabled=${FINERACT_MODE_BATCH_MANAGER_ENABLED:true}
 fineract.correlation.enabled=${FINERACT_LOGGING_HTTP_CORRELATION_ID_ENABLED:false}
 fineract.correlation.header-name=${FINERACT_LOGGING_HTTP_CORRELATION_ID_HEADER_NAME:X-Correlation-ID}
 
+fineract.partitioned-job.job-properties[0].job-name=LOAN_COB
+fineract.partitioned-job.job-properties[0].chunk-size=${LOAN_COB_CHUNK_SIZE:100}
```
vidakovic: Two things:

- can't we shorten the property prefix from `fineract.partitioned-job.job-properties` to just `fineract.partitioned-jobs`; seems a bit redundant to nest the list as the only attribute in yet another nested class (see my comment above)
- we should prefix the environment variables properly to indicate they are introduced by Fineract. Instead of `LOAN_COB_PARTITION_SIZE` I'd suggest `FINERACT_PARTIONED_JOB_PARTITION_SIZE` which is closer to the property name; we had already a cleanup in the past to remove inconsistent variable names, would be great if we could stick with this.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java
```diff
@@ -72,4 +75,21 @@ public static class FineractCorrelationProperties {
         private boolean enabled;
         private String headerName;
     }
+
+    @Getter
+    @Setter
+    public static class FineractPartitionedJobProperties {
```
vidakovic: Not sure why this nested class is needed with only one attribute... why not just take that list and move one level up? Saves space and easier to read and follow.

fineract-provider/src/main/resources/application.properties
taskain7: for the first thing: you're right, I'm gonna fix that
for the second thing: we're gonna have different partition sizes for different jobs, because there could be a scenario where someone has a huge jobs that take a long time, and they want them to run on 100 workers, and also they have very tiny jobs, that they want to run on 5 workers. That's why I named it after the specific job.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java
```diff
@@ -36,6 +37,8 @@ public class FineractProperties {
 
     private FineractCorrelationProperties correlation;
 
+    private FineractPartitionedJobProperties fineractPartitionedJobProperties;
```
vidakovic: I think the "fineract" prefix is not necessary here; this nested class is already in FineractProperties... so kind of redundant. Same for the suffix "Properties"; that kind of is self-evident.
I'd suggest to name this attribute just `partionedJob` if you insist using the class `FineractPartitionedJobProperties`; but as this class only contains 1 attribute I'd just move `List<JobProperty> jobProperties` here and name the variable `partitionedJobProperties`

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java
```diff
@@ -72,4 +75,21 @@ public static class FineractCorrelationProperties {
...
+    @Getter
+    @Setter
+    public static class FineractPartitionedJobProperties {
+
+        private List<JobProperty> jobProperties;
+    }
+
+    @Getter
+    @Setter
+    public static class JobProperty {
```
vidakovic: It would be more consistent to name this class `PartitionedJobProperty` see also my comments above.

If you take all these changes together then the configuration becomes significantly shorter and less repetitive (in the naming).

fineract-provider/src/test/resources/application-test.properties
```diff
@@ -37,6 +37,11 @@ fineract.mode.read-enabled=true
 fineract.mode.write-enabled=true
 fineract.mode.batch-enabled=true
 
+fineract.partitioned-job.job-properties[0].job-name=LOAN_COB
```
vidakovic: Same here, see my comments above.

vidakovic (approved):
LGTM

PR #2526, opened 2022-08-22
[FINERACT-1678] Spring Batch JMS support

taskain7:
## Description

JMS support for Spring Batch message handling.
JMS support is configurable via application properties and environment variables.

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/cob/messagehandler/COBJmsManagerConfig.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/cob/messagehandler/COBJmsManagerConfig.java
@@ -1,0 +1,59 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.cob.messagehandler;
+
+import javax.jms.ConnectionFactory;
+import org.apache.fineract.cob.COBOutputChannelInterceptor;
+import org.apache.fineract.cob.messagehandler.conditions.JmsManagerCondition;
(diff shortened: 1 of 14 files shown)


fineract-provider/src/main/java/org/apache/fineract/cob/messagehandler/COBJmsManagerConfig.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @Bean
+    public DirectChannel inboundStaging() {
+        return new DirectChannel();
+    }
+
+    @Bean
+    public IntegrationFlow inboundStagingFlow() {
+        return IntegrationFlows.from(Jms.messageDrivenChannelAdapter(connectionFactory) //
+                .configureListenerContainer(c -> c.subscriptionDurable(false)) //
```
galovics: I think we'll need a durable subscription (true instead of the false value) to make sure while a subscriber is down, when it comes back online it'll receive the messages properly. Thoughts?

fineract-provider/src/main/java/org/apache/fineract/cob/messagehandler/COBJmsWorkerConfig.java
```diff
@@ -0,0 +1,71 @@
...
+    private ConnectionFactory connectionFactory;
+    @Autowired
+    private QueueChannel inboundRequests;
+    @Autowired
+    private FineractProperties fineractProperties;
+
+    @Bean
+    public IntegrationFlow inboundFlow() {
+        return IntegrationFlows.from(Jms.messageDrivenChannelAdapter(connectionFactory) //
+                .configureListenerContainer(c -> c.subscriptionDurable(false)) //
```
galovics: Durable subscription.

fineract-provider/src/main/java/org/apache/fineract/cob/messagehandler/COBJmsManagerConfig.java
```diff
@@ -0,0 +1,74 @@
...
+                .get();
+    }
+
+    @Bean
+    public DirectChannel inboundStaging() {
+        return new DirectChannel();
+    }
+
+    @Bean
+    public IntegrationFlow inboundStagingFlow() {
```
galovics: Giving it another thought, do we even need an inbound flow here? I don't think so since the manager will only send messages.

fineract-provider/src/main/java/org/apache/fineract/cob/messagehandler/COBJmsWorkerConfig.java
```diff
@@ -0,0 +1,71 @@
...
+                .get();
+    }
+
+    @Bean
+    public DirectChannel outboundStaging() {
+        return new DirectChannel();
+    }
+
+    @Bean
+    public IntegrationFlow outboundStagingFlow() {
```
galovics: I don't think we'll need an outbound flow, do we?

galovics (changes requested):

galovics (approved):

PR #2547, opened 2022-08-26
FINERACT-1708 migration fails when upgrading from 1.6.0 to 1.7.0

daviestobialex:
the unique constraint name in `fineract_default` database `m_savings_account_transaction` is `transaction_ref_no` instead of `ref_no` which cause migration to fail.

## Description

## Checklist

### DIFF
--- a/fineract-provider/src/main/resources/db/changelog/tenant/parts/0016_changed_unique_constraint_of_ref_no.xml
+++ b/fineract-provider/src/main/resources/db/changelog/tenant/parts/0016_changed_unique_constraint_of_ref_no.xml
@@ -26,6 +26,6 @@
         <dropUniqueConstraint tableName="m_savings_account_transaction" constraintName="m_savings_account_transaction_ref_no_key"/>
     </changeSet>
     <changeSet author="fineract" id="1" context="mysql">
-        <dropUniqueConstraint tableName="m_savings_account_transaction" constraintName="ref_no"/>
+        <dropUniqueConstraint tableName="m_savings_account_transaction" constraintName="transaction_ref_no"/>
     </changeSet>
 </databaseChangeLog>


fineract-provider/src/main/resources/db/changelog/tenant/parts/0016_changed_unique_constraint_of_ref_no.xml
```diff
@@ -26,6 +26,6 @@
         <dropUniqueConstraint tableName="m_savings_account_transaction" constraintName="m_savings_account_transaction_ref_no_key"/>
     </changeSet>
     <changeSet author="fineract" id="1" context="mysql">
-        <dropUniqueConstraint tableName="m_savings_account_transaction" constraintName="ref_no"/>
+        <dropUniqueConstraint tableName="m_savings_account_transaction" constraintName="transaction_ref_no"/>
```
galovics, line 29: Have you validated this with an empty 1.8 database? I think it's gonna fail due to the fact that the checksum is different.
You should use validateChecksum in case you're touching older changesets

galovics (changes requested):

fineract-provider/src/main/resources/db/changelog/tenant/parts/0016_changed_unique_constraint_of_ref_no.xml
adamsaghy: I am a little bit confused here. This changeset is 2-3 month old. We have integration tests which executed on all 3 db engine, and there were no any issues so far.

adamsaghy: if the problem is the liquibase is generating a unique constraint name which is differs from the priorly generated one (with flyway or manually), i would recomment to add conditions into the changesets here which precheck which unique constraint is existing and just ignore the drop unique constraint changeset if it does not exist

daviestobialex: exactly so on my part I just changed it to the unique constraint name that was generated at the time I installed it but it seemed like the new changes had the unique constraint in another name.

daviestobialex: > Have you validated this with an empty 1.8 database? I think it's gonna fail due to the fact that the checksum is different. You should use validateChecksum in case you're touching older changesets

This works on an empty 1.8 as the unique constraint name generated is the same.

PR #2560, opened 2022-09-01
FINERACT-1694: Integration of external event storage

galovics:
## Description

## Checklist


fineract-provider/src/main/java/org/apache/fineract/infrastructure/event/external/repository/domain/ExternalEvent.java
```diff
@@ -0,0 +1,75 @@
...
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import org.apache.fineract.infrastructure.core.domain.AbstractPersistableCustom;
+import org.apache.fineract.infrastructure.core.service.DateUtils;
+
+@Entity
+@Table(name = "m_external_event")
+@Getter
+@NoArgsConstructor
+public class ExternalEvent extends AbstractPersistableCustom {
```
adamsaghy, line 40: Please use the AbstractAuditableWithUTCDateTimeCustom and auditing will be handled automatically!

fineract-provider/src/main/java/org/apache/fineract/infrastructure/event/external/repository/domain/ExternalEvent.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @Column(name = "created_at", nullable = false)
+    private OffsetDateTime createdAt;
+
+    @Enumerated(EnumType.STRING)
+    @Column(name = "status", nullable = false)
+    @Setter
+    private ExternalEventStatus status;
+
+    @Column(name = "sent_at", nullable = false)
```
adamsaghy: It should be nullable, we will send it asynchronously at a later point in time.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/event/external/repository/domain/ExternalEventStatus.java
```diff
@@ -0,0 +1,23 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.infrastructure.event.external.repository.domain;
+
+public enum ExternalEventStatus {
+    TO_BE_SENT, SENT
```
adamsaghy, line 22: Should be there an error status as well?

adamsaghy (commented):
I was wondering whether we should have tests to execute the serializers with "real" data to validate their proper behaviour.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/event/external/repository/domain/ExternalEvent.java
galovics: Intentionally didn't use that one since I didn't want auditing, only a creation date.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/event/external/repository/domain/ExternalEvent.java
galovics: Yep, nice catch. I already changed the Liquibase script to be nullable. ;-)

fineract-provider/src/main/java/org/apache/fineract/infrastructure/event/external/repository/domain/ExternalEventStatus.java
galovics: Well, not really because if for some reason we won't be able to send the message, for example due to an outage on the message channel, we'll just retry it next time.

galovics:
> I was wondering whether we should have tests to execute the serializers with "real" data to validate their proper behaviour.

The external events feature is enabled during the integration tests and all the serializers are triggered. Note, if any of the serialization logic fails, the test case will also fail since the serializers are executed within the same transaction as the business operation.

adamsaghy (approved):
LGTM

build.gradle
```diff
@@ -279,6 +279,8 @@ configure(project.fineractJavaProjects) {
     sourceSets.main.output.resourcesDir = sourceSets.main.java.outputDir
     sourceSets.test.output.resourcesDir = sourceSets.test.java.outputDir
 
+    check.dependsOn('cucumber')
```
vidakovic, line 282: Not sure if I follow here... doesn't this mean that the cucumber tests kick in when we do linting with Spotless etc.? Linting and testing being two different build lifecycle stages?

galovics: Good catch Aleks. The reason I moved the cucumber execution here is because originally if I wanted to execute simple unit tests in fineract-provider, it was first running ALL the cucumber tests. And considering they are Spring tests, it's quite slow. For a single unit test it was taking 30-60 seconds to run which is way too long.

Moving the cucumber execution to depend on the check task means that the unit tests can be easily executed without the cucumber tests being run beforehand.

Also, this change is in line with the official Gradle recommendation:

> check: Aggregate task that performs verification tasks, such as running the tests. Some plugins add their own verification tasks to check. You should also attach any custom Test tasks to this lifecycle task if you want them to execute for a full build. This task is added by the Base Plugin.

Source: https://docs.gradle.org/current/userguide/java_plugin.html

vidakovic: alright... interesting.

vidakovic (approved):
LGTM

PR #2576, opened 2022-09-07
FINERACT-1678: Lock under processing loan accounts

adamsaghy:
## Description

With the database structure in place for loan account locks, we need to utilize it. 

In this story, we need to create a new Spring Batch step (not business steps) and add it into the Loan COB job. The new step will be locking the loan accounts. Also, at the end of the execution of the business steps, we need to unlock the loan accounts within the chunk (it’s important to do it in the same transaction as the business steps).

**Acceptance criteria**

A new lock loan accounts Spring Batch step is introduced

The lock loan accounts step locks the loan accounts which are in the chunk meaning that rows need to be inserted into the lock table

A new unlock loan accounts mechanism is introduced 

The unlock loan accounts mechanism unlocks the loan accounts which are in the chunk meaning that rows need to be deleted from the lock table

The lock loan accounts step is in the beginning of the execution, before executing the business steps

The unlock loan accounts step is at the end of the business step execution in the same transaction

If there was an error during processing a loan account, the exception shall be written in the corresponding lock row for later analysis

The lock_owner column is filled properly with LOAN_COB_CHUNK_PROCESSING

**Note**

Skipping for the Spring Batch step shall be enabled

A SkipListener could be useful

## Checklist


fineract-provider/src/main/java/org/apache/fineract/cob/common/InitialisationTasklet.java
```diff
@@ -0,0 +1,62 @@
...
+import org.springframework.security.authentication.UsernamePasswordAuthenticationToken;
+import org.springframework.security.core.authority.mapping.NullAuthoritiesMapper;
+import org.springframework.security.core.context.SecurityContextHolder;
+
+@Slf4j
+@RequiredArgsConstructor
+public class InitialisationTasklet implements Tasklet, StepExecutionListener {
+
+    private final AppUserRepositoryWrapper userRepository;
+    private StepExecution stepExecution;
```
galovics: Don't see any usage for this. Was it left here accidentally?

fineract-provider/src/main/java/org/apache/fineract/cob/common/InitialisationTasklet.java
```diff
@@ -0,0 +1,62 @@
...
+    public RepeatStatus execute(@NotNull StepContribution contribution, @NotNull ChunkContext chunkContext) throws Exception {
+        AppUser user = userRepository.fetchSystemUser();
+        UsernamePasswordAuthenticationToken auth = new UsernamePasswordAuthenticationToken(user, user.getPassword(),
+                new NullAuthoritiesMapper().mapAuthorities(user.getAuthorities()));
+        SecurityContextHolder.getContext().setAuthentication(auth);
+        return RepeatStatus.FINISHED;
+    }
+
+    @Override
+    public void beforeStep(@NotNull StepExecution stepExecution) {
```
galovics: Any reason for the before and afterStep methods?

fineract-provider/src/main/java/org/apache/fineract/cob/domain/LoanAccountLock.java
```diff
@@ -18,19 +18,22 @@
...
+import javax.persistence.UniqueConstraint;
 import javax.persistence.Version;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
+import org.apache.fineract.infrastructure.core.service.DateUtils;
 
 @Entity
-@Table(name = "m_loan_account_locks")
+@Table(name = "m_loan_account_locks", uniqueConstraints = {
+        @UniqueConstraint(columnNames = { "loan_id", "lock_owner" }, name = "UNIQUE_ACCOUNT_LOCK") })
```
galovics: As we discussed, the loan_id is already unique in the table, any reason to do a composite unique constraint?

fineract-provider/src/main/java/org/apache/fineract/cob/listener/LoanItemListener.java
```diff
@@ -0,0 +1,127 @@
...
+                loanAccountLock.ifPresent(accountLock -> accountLock.setError(String.format("Loan (id: %d) reading is failed", ee.getId()),
+                        ThrowableSerialization.serialize(e)));
+            }
+        });
+
+    }
+
+    @OnSkipInRead
+    public void onSkipInRead(@NotNull Exception e) {
+        log.warn("Skipping was triggered during read! Due to: ");
```
galovics: Missing the exception from the param list here.

fineract-provider/src/main/java/org/apache/fineract/cob/listener/LoanItemListener.java
```diff
@@ -0,0 +1,127 @@
...
+    @OnReadError
+    public void onReadError(Exception e) {
+        LoanReadException ee = (LoanReadException) e;
+        log.warn("Error was triggered during reading of Loan (id={}) due to: {}", ee.getId(), ThrowableSerialization.serialize(e));
+
+        transactionTemplate.setPropagationBehavior(PROPAGATION_REQUIRES_NEW);
+        transactionTemplate.execute(new TransactionCallbackWithoutResult() {
+
+            @Override
+            protected void doInTransactionWithoutResult(@NotNull TransactionStatus status) {
```
galovics, line 58: Isn't this a duplicated logic in the below methods? Any chance we could extract this similar piece of code into a method and reuse it?

fineract-provider/src/main/java/org/apache/fineract/cob/loan/ApplyLoanLockTasklet.java
```diff
@@ -0,0 +1,93 @@
...
+@Slf4j
+@RequiredArgsConstructor
+public class ApplyLoanLockTasklet implements Tasklet, StepExecutionListener {
+
+    private final LoanAccountLockRepository accountLockRepository;
+    private StepExecution stepExecution;
+
+    @Override
+    public RepeatStatus execute(@NotNull StepContribution contribution, @NotNull ChunkContext chunkContext) throws Exception {
+        ExecutionContext executionContext = stepExecution.getExecutionContext();
```
galovics: Question, can't we get the stepExecution from the StepContribution? I recall you might have told me already it wasn't working for some reason, but I'm not sure if I remember correctly.

fineract-provider/src/main/java/org/apache/fineract/cob/loan/ApplyLoanLockTasklet.java
```diff
@@ -0,0 +1,93 @@
...
+                .filter(e -> LockOwner.LOAN_COB_PARTITIONING.equals(e.getLockOwner()))
+                .collect(Collectors.toMap(LoanAccountLock::getLoanId, Function.identity()));
+
+        remainingLoanIds.removeAll(alreadyHardLockedAccountIds);
+        remainingLoanIds.removeAll(alreadyUnderProcessingAccountIds);
+
+        for (Long loanId : remainingLoanIds) {
+            LoanAccountLock loanAccountLock;
+            if (alreadySoftLockedAccountsMap.containsKey(loanId)) {
+                loanAccountLock = alreadySoftLockedAccountsMap.get(loanId);
```
galovics: I guess we could make this a tiny bit more descriptive for the future. 
Maybe extract this whole if branch into a method and name it like (random idea) `upgradeLock`.

fineract-provider/src/main/java/org/apache/fineract/cob/loan/ApplyLoanLockTasklet.java
```diff
@@ -0,0 +1,93 @@
...
+                loanAccountLock.setNewLockOwner(LockOwner.LOAN_COB_CHUNK_PROCESSING);
+            } else {
+                loanAccountLock = new LoanAccountLock(loanId, LockOwner.LOAN_COB_CHUNK_PROCESSING);
```
galovics: Can this even occur? If something wasn't soft locked before (considering it wasn't inline COB locked either), it means that it was already assigned to a worker in the chunk processing but the worker dropped it, probably due to a crash. In that case, the lock stays on the loanId and we don't need to create a new lock row in the table.

What do you think? Am I missing something, maybe this is a case I haven't thought about?

fineract-provider/src/main/java/org/apache/fineract/cob/loan/LoanCOBManagerConfiguration.java
```diff
@@ -63,13 +63,13 @@ public class LoanCOBManagerConfiguration {
...
     public LoanCOBPartitioner partitioner() {
-        return new LoanCOBPartitioner(cobPropertyService, cobBusinessStepService, jobOperator, jobExplorer, retrieveLoanIdService);
+        return new LoanCOBPartitioner(propertyService, cobBusinessStepService, jobOperator, jobExplorer, retrieveLoanIdService);
     }
 
     @Bean
     public Step loanCOBStep() {
-        return stepBuilderFactory.get(JobName.LOAN_COB.name()).partitioner("Loan COB worker", partitioner()).outputChannel(outboundRequests)
-                .build();
+        return stepBuilderFactory.get(JobName.LOAN_COB.name()).partitioner("loanCOBWorkerStep", partitioner())
```
galovics: Do you think we could extract the string `loanCOBWorkerStep` into a constant so it can be referenced from the manager and worker as well (to avoid future maintenance issues with potential renamings)

fineract-provider/src/main/java/org/apache/fineract/cob/loan/LoanItemReader.java
```diff
@@ -0,0 +1,77 @@
...
+    @Override
+    public Loan read() throws Exception {
+        try {
+            if (remainingData.size() > 0) {
+                loanId = remainingData.remove(0);
+                if (alreadyLockedAccounts != null && alreadyLockedAccounts.remove(loanId)) {
+                    throw new LoanAccountWasAlreadyLocked(loanId);
+                }
+
+                return loanRepository.findById(loanId).orElse(null);
```
galovics: Maybe instead of returning a null in case the loanId doesn't correspond to an actual loan, we could throw an exception cause that indicates an inconsistency in data.

galovics (changes requested):
Couple of comments but overall really really nice job. In fact I quite like the tiny reorganizations & renamings you did.
Plus the unit tests, splendid work. :-)

fineract-provider/src/main/java/org/apache/fineract/cob/loan/ApplyLoanLockTasklet.java
adamsaghy: This only happens till the soft lock logic is not in place or it will be handled differently and for (for now) unknown reasons there will be no soft lock applied. We still need to put a lock if there was none (by mistake or by design).

I think this way is more resilient. What do you think?

fineract-provider/src/main/java/org/apache/fineract/cob/loan/ApplyLoanLockTasklet.java
adamsaghy: You are right! We can gather the stepExecution from the stepContribution.

fineract-provider/src/main/java/org/apache/fineract/cob/listener/LoanItemListener.java
adamsaghy: Removed... not necessary. Already logged as part of the error listener

fineract-provider/src/main/java/org/apache/fineract/cob/common/InitialisationTasklet.java
adamsaghy: yes

galovics (approved):

PR #2623, opened 2022-09-28
FINERACT-1694 Categorization for Events

ruchiD:
## Description
Changes for introducing event categorization for business events.
Events are categorized based on domain objects on which events occur.

Why:
There might be downstream consumers who are interested in all - for example - Loan related events but they don’t want to specify it down to fine-grained level events like Loan Approval, Loan Disbursement.

Changes made:
Introduced category column for events table.
For each event type category must be provided.
External Events are saved with category information.

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/event/business/domain/BulkBusinessEvent.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/event/business/domain/BulkBusinessEvent.java
@@ -22,12 +22,20 @@ import java.util.List;
 
 public class BulkBusinessEvent extends AbstractBusinessEvent<List<BusinessEvent<?>>> {
 
+    private static final String CATEGORY = "Bulk";
+    private static final String TYPE = "BulkBusinessEvent";
+
     public BulkBusinessEvent(List<BusinessEvent<?>> value) {
         super(value);
     }
 
     @Override
     public String getType() {
-        return "BulkBusinessEvent";
+        return TYPE;
+    }
+
+    @Override
+    public String getCategory() {
+        return CATEGORY;
     }
 }
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/event/business/domain/BusinessEvent.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/event/business/domain/BusinessEvent.java
@@ -23,4 +23,6 @@ public interface BusinessEvent<T> {
     T get();
 
     String getType();
+
+    String getCategory();
 }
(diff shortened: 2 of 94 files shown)


adamsaghy (approved):
LGTM

fineract-provider/src/main/java/org/apache/fineract/infrastructure/event/business/domain/BulkBusinessEvent.java
```diff
@@ -30,4 +30,9 @@ public BulkBusinessEvent(List<BusinessEvent<?>> value) {
     public String getType() {
         return "BulkBusinessEvent";
```
vidakovic: I know, a bit picky, but could we add somewhere `static final String` variables with these kind of values (if in doubt then put it in class `BulkBusinessEvent.java` as a private variable. Like it is now the string will always be created new when the function `getType()` is called. Having a variable is also a good practice in terms of avoiding "magic strings".

Another suggestion: why not use `BulkBusinessEvent.class.getSimpleName()` (="BulkBusinessEvent")? In that case this whole thing would be immune against refactoring errors...

If I were you I would do this:

```
public class BulkBusinessEvent imiplements BusinessEvent {
    private static final String TYPE = getClass().getSimpleName();

...

    @Override
    public String getType() {
        return TYPE;
    }

...

}
```

And finally: as the function `getType()` is required by all `BusinessEvent` classes you could create an abstract class and move above code there. Benefits:
- only have to write this code once
- completely type-safe
- avoiding all magic strings (aka refactoring works like a charm)
- ... and memory usage is a tiny bit better (no "new strings" on every call to `getType()`)

This could look like this:

```
public class AbstractBusinessEvent imiplements BusinessEvent {
    protected static final String TYPE = getClass().getSimpleName();

...

    @Override
    public String getType() {
        return TYPE;
    }

...

}

public class BulkBusinessEvent extends AbstractBusinessEvent {

...

}
```

fineract-provider/src/main/java/org/apache/fineract/infrastructure/event/business/domain/BulkBusinessEvent.java
```diff
@@ -30,4 +30,9 @@ public BulkBusinessEvent(List<BusinessEvent<?>> value) {
...
     }
+
+    @Override
+    public String getCategory() {
+        return "Bulk";
```
vidakovic: Similar situation here as with `getType()` in my last comment. To avoid the "magic string" and avoid creating a new string on every call to that function I'd rearrange the code like this:

```
public class BulkBusinessEvent imiplements BusinessEvent {
    private static final String CATEGORY = "Bulk";

...

    @Override
    public String getCategory() {
        return CATEGORY;
    }

...

}
```

Rinse repeat for similar cases...

vidakovic (approved):
LGTM

PR #2659, opened 2022-10-12
FINERACT-1747 add datatable query api

ghost:
## Description

See details in https://issues.apache.org/jira/browse/FINERACT-1747

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/api/DatatablesApiResource.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/api/DatatablesApiResource.java
@@ -18,6 +18,7 @@
  */
 package org.apache.fineract.infrastructure.dataqueries.api;
 
+import com.google.gson.JsonObject;
 import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.Parameter;
 import io.swagger.v3.oas.annotations.enums.ParameterIn;
@@ -216,6 +217,25 @@ public class DatatablesApiResource {
 
         final boolean prettyPrint = ApiParameterHelper.prettyPrint(uriInfo.getQueryParameters());
         return this.toApiJsonSerializer.serializePretty(prettyPrint, result);
+    }
+
+    @GET
+    @Path("{datatable}/query")
+    @Produces({ MediaType.APPLICATION_JSON })
+    @Operation(summary = "Query Data Table values", description = "Query values from a registered data table.")
+    @ApiResponses({
+            @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = List.class))) })
+    public String queryValues(@PathParam("datatable") @Parameter(description = "datatable") final String datatable,
+            @QueryParam("columnFilter") @Parameter(description = "columnFilter") final String columnFilter,
+            @QueryParam("valueFilter") @Parameter(description = "valueFilter") final String valueFilter,
+            @QueryParam("resultColumns") @Parameter(description = "resultColumns") final String resultColumns,
+            @Context final UriInfo uriInfo) {
+        this.context.authenticatedUser().validateHasDatatableReadPermission(datatable);
+
+        final List<JsonObject> result = this.readWriteNonCoreDataService.queryDataTable(datatable, columnFilter, valueFilter,
+                resultColumns);
+
+        return this.toApiJsonSerializer.serializePretty(ApiParameterHelper.prettyPrint(uriInfo.getQueryParameters()), result);
     }
 
     @GET
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
@@ -18,6 +18,9 @@
  */
 package org.apache.fineract.infrastructure.dataqueries.service;
 
+import static java.util.Arrays.asList;
+import static org.apache.fineract.infrastructure.core.data.ApiParameterError.parameterErrorWithValue;
+
 import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableMap;
 import com.google.gson.JsonArray;
@@ -30,12 +33,18 @@ import java.sql.Date;
 import java.sql.SQLException;
 import java.sql.Statement;
 import java.sql.Timestamp;
+import java.sql.Types;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Stream;
 import javax.persistence.PersistenceException;
@@ -74,6 +83,7 @@ import org.apache.fineract.infrastructure.security.service.SqlInjectionPreventer
 import org.apache.fineract.infrastructure.security.utils.ColumnValidator;
 import org.apache.fineract.infrastructure.security.utils.SQLInjectionValidator;
 import org.apache.fineract.useradministration.domain.AppUser;
+import org.jetbrains.annotations.NotNull;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.dao.DataAccessException;
@@ -104,11 +114,24 @@ public class ReadWriteNonCoreDataServiceImpl implements ReadWriteNonCoreDataServ
             .put("number", "INT").put("boolean", "BIT").put("decimal", "DECIMAL").put("date", "DATE").put("datetime", "DATETIME")
             .put("text", "TEXT").put("dropdown", "INT").build();
     private static final ImmutableMap<String, String> apiTypeToPostgreSQL = ImmutableMap.<String, String>builder().put("string", "VARCHAR")
-            .put("number", "INT").put("boolean", "BIT").put("decimal", "DECIMAL").put("date", "DATE").put("datetime", "TIMESTAMP")
+            .put("number", "INT").put("boolean", "boolean").put("decimal", "DECIMAL").put("date", "DATE").put("datetime", "TIMESTAMP")
             .put("text", "TEXT").put("dropdown", "INT").build();
 
     private static final List<String> stringDataTypes = Arrays.asList("char", "varchar", "blob", "text", "tinyblob", "tinytext",
             "mediumblob", "mediumtext", "longblob", "longtext");
+    private static final DateTimeFormatter DATA_TABLE_DATE_FORMAT = DateTimeFormatter.ofPattern("yyyy-MM-dd");
+    private static final DateTimeFormatter DATA_TABLE_DATETIME_FORMAT = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss");
+    public static final String PG_BOOLEAN_TYPE = "boolean";
+    public static final String PG_BOOL_TYPE = "bool";
+    public static final String PG_INT_TYPE = "integer";
+    public static final String PG_BIGINT_TYPE = "bigint";
+    public static final String PG_DATE_TYPE = "date";
+    public static final String PG_NUMERIC_TYPE = "numeric";
+    public static final String PG_TEXT_TYPE = "text";
+    public static final String PG_VARCHAR_TYPE = "character varying";
+    public static final String BIT_TYPE = "bit";
+    public static final String MYSQL_DATETIME_TYPE = "datetime";
+    public static final String MYSQL_DATE_TYPE = "date";
 
     private final JdbcTemplate jdbcTemplate;
     private final DatabaseTypeResolver databaseTypeResolver;
@@ -160,9 +183,9 @@ public class ReadWriteNonCoreDataServiceImpl implements ReadWriteNonCoreDataServ
 
         // PERMITTED datatables
         SQLInjectionValidator.validateSQLInput(datatable);
-        final String sql = "select application_table_name, registered_table_name, entity_subtype" + " from x_registered_table "
-                + " where exists" + " (select 'f'" + " from m_appuser_role ur " + " join m_role r on r.id = ur.role_id"
-                + " left join m_role_permission rp on rp.role_id = r.id" + " left join m_permission p on p.id = rp.permission_id"
+        final String sql = "select application_table_name, registered_table_name, entity_subtype from x_registered_table "
+                + " where exists (select 'f' from m_appuser_role ur join m_role r on r.id = ur.role_id"
+                + " left join m_role_permission rp on rp.role_id = r.id left join m_permission p on p.id = rp.permission_id"
                 + " where ur.appuser_id = ? and registered_table_name=? and (p.code in ('ALL_FUNCTIONS', "
                 + "'ALL_FUNCTIONS_READ') or p.code = concat('READ_', registered_table_name))) "
                 + " order by application_table_name, registered_table_name";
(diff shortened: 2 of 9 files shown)


build.gradle
```diff
@@ -171,6 +171,7 @@ allprojects  {
         reportDir = file("$buildDir/reports/rat")
         excludes = [
             '**/src/main/templates/**/*.mustache',
+            '**/src/main/generated/**',
```
galovics: The generated Java files should also have licenses. Please remove this.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/api/DatatablesApiResource.java
```diff
@@ -218,6 +226,55 @@ public String getDatatable(@PathParam("datatable") @Parameter(description = "dat
...
+        }
+
+        if (valueFilter == null || valueFilter.isEmpty()) {
+            paramErrors.add(parameterErrorWithValue("400", "Value filter is empty!", "valueFilter", valueFilter));
+        }
+
+        if (resultColumns == null || resultColumns.isEmpty()) {
+            paramErrors.add(parameterErrorWithValue("400", "Result columns filter is empty!", "resultColumns", resultColumns));
+        } else {
+            List<String> resultColumnNames = Stream.of(resultColumns.split(",")).toList();
```
galovics: Why create a stream from it? Simple Arrays.asList?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/api/DatatablesApiResource.java
```diff
@@ -218,6 +226,55 @@ public String getDatatable(@PathParam("datatable") @Parameter(description = "dat
...
+    @Operation(summary = "Query Data Table values", description = "Query values from a registered data table.")
+    @ApiResponses({
+            @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = List.class))) })
+    public String queryValues(@PathParam("datatable") @Parameter(description = "datatable") final String datatable,
+            @QueryParam("columnFilter") @Parameter(description = "columnFilter") final String columnFilter,
+            @QueryParam("valueFilter") @Parameter(description = "valueFilter") final String valueFilter,
+            @QueryParam("resultColumns") @Parameter(description = "resultColumns") final String resultColumns,
+            @Context final UriInfo uriInfo) {
+        this.context.authenticatedUser().validateHasDatatableReadPermission(datatable);
+        List<String> dataTableColumnNames = genericDataService.fillResultsetColumnHeaders(datatable).stream()
```
galovics: I think we should move this logic one layer deeper and not have it on the API level, at least according to the current practices in fineract.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
```diff
@@ -183,6 +192,116 @@ public DatatableData retrieveDatatable(final String datatable) {
         return datatableData;
     }
 
+    @Override
+    public List<JsonObject> queryDataTable(String datatable, String columnFilter, String valueFilter, String resultColumns) {
+        Arrays.asList(datatable, columnFilter, valueFilter, resultColumns).forEach(SQLInjectionValidator::validateDynamicQuery);
+        List<String> resultColumnNames = Stream.of(resultColumns.split(",")).toList();
```
galovics: Arrays.asList?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
```diff
@@ -183,6 +192,116 @@ public DatatableData retrieveDatatable(final String datatable) {
...
+                .findFirst().map(ResultsetColumnHeaderData::getColumnType).orElse(columnFilter + " does not exist in datatable");
+        if (databaseTypeResolver.isPostgreSQL()) {
+            int[] argType = new int[1];
+            if ("bit".equalsIgnoreCase(filterColumnType)) {
+                finalValueFilter = BooleanUtils.toString(BooleanUtils.toBooleanObject(valueFilter), "1", "0", "null");
+                argType[0] = Types.BIT;
+            } else if ("boolean".equalsIgnoreCase(filterColumnType) || "bool".equalsIgnoreCase(filterColumnType)) {
+                finalValueFilter = BooleanUtils.toString(BooleanUtils.toBooleanObject(valueFilter), "true", "false", "null");
+                argType[0] = Types.BOOLEAN;
+            } else if ("integer".equalsIgnoreCase(filterColumnType)) {
```
galovics: Not a fan of using these magic strings in the conditions, would've been nicer if you extracted and reused them.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
```diff
@@ -183,6 +192,116 @@ public DatatableData retrieveDatatable(final String datatable) {
...
+                argType[0] = Types.DATE;
+            } else if (filterColumnType.toLowerCase().contains("timestamp")) {
+                argType[0] = Types.TIMESTAMP;
+            } else if ("numeric".equalsIgnoreCase(filterColumnType)) {
+                argType[0] = Types.DECIMAL;
+            } else if ("text".equalsIgnoreCase(filterColumnType) || "character varying".equalsIgnoreCase(filterColumnType)) {
+                argType[0] = Types.VARCHAR;
+            } else {
+                List<ApiParameterError> paramErrors = new ArrayList<>();
+                paramErrors.add(parameterErrorWithValue("400",
```
galovics: Okay so let's decide where to put the validation logic. On the API, I already mentioned that the validations should be one layer deeper and this confirms it.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
```diff
@@ -183,6 +192,116 @@ public DatatableData retrieveDatatable(final String datatable) {
...
+                        "Unsupported input type for datatable query! Column filter: " + filterColumnType, "valueFilter", valueFilter));
+                throw new PlatformApiDataValidationException(paramErrors);
+            }
+            rowSet = jdbcTemplate.queryForRowSet(sql, new Object[] { finalValueFilter }, argType);
```
galovics: I don't think we need the `new Object[]`

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
```diff
@@ -183,6 +192,116 @@ public DatatableData retrieveDatatable(final String datatable) {
...
+        } else if (databaseTypeResolver.isMySQL()) {
+            if ("bit".equalsIgnoreCase(filterColumnType)) {
+                int[] argType = new int[1];
+                argType[0] = Types.BIT;
```
galovics: Isn't this duplicated logic from the postgresql branch?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
```diff
@@ -183,6 +192,116 @@ public DatatableData retrieveDatatable(final String datatable) {
...
+                finalValueFilter = BooleanUtils.toString(BooleanUtils.toBooleanObject(valueFilter), "1", "0", "null");
+                rowSet = jdbcTemplate.queryForRowSet(sql, new Object[] { finalValueFilter }, argType);
+            } else if ("date".equalsIgnoreCase(filterColumnType)) {
+                int[] argType = new int[1];
+                argType[0] = Types.DATE;
+                try {
+                    rowSet = jdbcTemplate.queryForRowSet(sql, new Object[] { LocalDate.parse(valueFilter, DATA_TABLE_DATE_FORMAT_MYSQL) },
```
galovics: Don't we need a similar logic for PG?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
```diff
@@ -183,6 +192,116 @@ public DatatableData retrieveDatatable(final String datatable) {
...
+                Object rowValue = rowSet.getObject(rcn);
+                if (rowValue != null) {
+                    if (rowValue instanceof Character) {
+                        json.addProperty(rcn, (Character) rowValue);
+                    } else if (rowValue instanceof Number) {
+                        json.addProperty(rcn, new BigDecimal(rowValue.toString()));
+                    } else if (rowValue instanceof Boolean) {
+                        json.addProperty(rcn, (Boolean) rowValue);
+                    } else if (rowValue instanceof LocalDateTime) {
+                        json.addProperty(rcn, DATA_TABLE_DATETIME_FORMAT_MYSQL.format((LocalDateTime) rowValue));
```
galovics: Very misleading to use the MySQL format here. Let's dupe this member and name it properly like DATA_TABLE_RESULT_DATETIME_FORMAT, or something similar

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/SQLInjectionValidator.java
```diff
@@ -35,7 +35,7 @@ private SQLInjectionValidator() {
 
     private static final String[] COMMENTS = { "--", "({", "/*", "#" };
 
-    private static final String SQL_PATTERN = "[a-zA-Z_=,\\-'!><.?\"`% ()0-9*\n\r]*";
+    private static final String SQL_PATTERN = "[a-zA-Z_=,\\-:'!><.?\"`% ()0-9*\n\r]*";
```
galovics, line 38: One thing here. If you really wanna touch this class then cover these with tests. We literally have zero coverage on this and we could easily break it.

galovics (changes requested):

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
ghost: you made a change at 2022.02.22 putting a TODO comment in that service class:
// TODO: Extract these types out of here
`.put("text", "TEXT").put("dropdown", "INT").build();`
Those strings could be also reused or extracted.
Also there are countless occurrences of various column type names scattered in the code.
If the scope of this ticket is to refactor the codebase from database specific column type names, I will gladly do it but I don't think now is the time, it does not bring any real value. Nobody will search the entire codebase in a approx ~423.000 line project for reusable texts for words like "boolean" and "integer".

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
ghost: It is the same but at postgres every type is manually casted, however at mysql only a few type casting is needed and the rest is automatic,

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
ghost: The code runs on all 3 database backends if you checked the integ results and the they tests for successful http request and the result column values also. See other type casting comment.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
ghost: any other idea? Stream.of(....)? List.of(....)? what is the question?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/api/DatatablesApiResource.java
ghost: I don't agree why are we validating input params on database service layer instead at API level where it should occour but if the project is doing that way, I am moving it there.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/api/DatatablesApiResource.java
ghost: done

build.gradle
ghost: done

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
ghost: I used the same syntax as before. There are numerous cases around the project.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/SQLInjectionValidator.java
ghost: I don't want to touch it, the function needs it. If you want to search for a date like this: 2100-12-12 20:20:20. The current implementation will treat it as SQL injection however it is not. I will write some testcases for it.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
ghost: done

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/api/DatatablesApiResource.java
galovics: It's a matter of design overall in the codebase. It's one level deeper and as I said in another comment, right now the validation logic is scattered between this level and the service layer.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
galovics: No question. I'm simply saying to replace Stream.of(..).toList() with an Arrays.asList since you don't use the Stream API anyhow and it's just a waste of resources.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
galovics: I hope I don't need to expain the Boy Scout Rule: “Always leave the campground cleaner than you found it”

In fact, you don't need to extract magic Strings all over the codebase, obviously. I'm just stating that you duplicated/triplicated the magic Strings so why don't you locally extract it at least?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
galovics: Got it, but still this is not needed. As a matter of fact we started to get rid of those slowly but surely.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
galovics: Okay but then why don't we form the branch logic in a way that only differences are captured instead of doing everything all over again for both branches?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/SQLInjectionValidator.java
galovics: Yeah I got it you didn't want to touch it but the thing is you did hence the testing is on you. Please add proper test cases to cover this functionality (not only for your date case).

galovics (changes requested):
Still some things to do.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
ghost: Sorry my bad, github showed this comment on another line. I already fixed this.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
ghost: done

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
ghost: the problem with this is that the queryForRowSet method has 2 different signatures:
public SqlRowSet queryForRowSet(String sql, Object[] args, int[] argTypes) 
public SqlRowSet queryForRowSet(String sql, @Nullable Object... args)
if I omit the object array, it will resolve to the latter and it won't take the argTypes correctly

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
ghost: done

galovics (approved):

PR #2669, opened 2022-10-13
FINERACT-1758: Improve custom module example

vidakovic:
Adding custom code/modules was still not sufficiently explained. We have currently 2 different examples concerning modules and how to add your own code to Fineract while ensuring at the same time that downstream syncs work without a hassle. In the current upstream code there are still too many files that need to be changed when a custom module is added (e. g. settings.gradle etc.).

The goal of this PR is to make adding extensions/customizations to Fineract as easy as possible:

- no file edits on upstream related files
- avoid clashes with Fineract internals
- avoid clashes with other companies' extensions
- provide a custom Fineract Docker image that automatically includes all custom modules
- give examples and best practices (replace note service, add a COB business step)


galovics (dismissed):
Looks really good overall.
@vidakovic one thing I'm missing is an extension point for injecting new liquibase changesets. That's gonna be needed for example for the new custom Business Step for the COB you implemented in this PR. The class is one thing, holding the implementation but to be able to configure, a new row needs to be inserted into the DB (`m_batch_business_steps` table).

adamsaghy:
> 

I was thinking we can add an extra `<includeAll path="custom-changelog/"/>` which would looking for the `custom-changelog` directory on the classpath (also we can use `searchPath` of liquibase as well to define multiple search path) and execute any changelog be found by alphabetical order and if the path is missing it will ignore by default.

@vidakovic @galovics What do you think?

galovics (approved):

PR #2689, opened 2022-10-20
FINERACT-1758: Improve custom module example (fix)

vidakovic:
Small fix related to custom migration scripts (were accidentally executed twice) and making sure the migrations are applied to the right context ("tenant_db"). Updated documentation.


adamsaghy (approved):
LGTM

fineract-doc/src/docs/en/chapters/architecture/modules.adoc
```diff
@@ -63,6 +63,15 @@ If database migrations are needed as part of your customizations then you can ad
...
 ----
 
+And here an example migration script:
+
+[source,xml]
+----
+include::{rootdir}/custom/acme/note/starter/src/main/resources/db/custom-changelog/0001_acme_note_initial.xml[lines=22..]
+----
+
+IMPORTANT: Please make sure your `changeSet` tag has the attribute `context` set to `tenant_db`. That makes sure your changes will be applied to the tenant database (read: main database and not the tenant store database). In theory you could also target the tenant configuration database, but it's not recommended to do that.
```
galovics: What if we just disallow touching the tenant configuration DB? I think the tenant config should be tied to the shipped Fineract implementation. No custom module should touch it.

Having said that, the context=tenant_db could go one level upper to only include changeSets when its the tenant_db

Thoughts?

vidakovic: Well, from my perspective that tenant configuration DB should disappear anyway, but that's a topic for another day. In general I agree (they shouldn't touch that one), but then in the end: they own the system they are running on... so there's not much that we can do here if they really wanted to; which makes this pointless I guess.
As for "level up": I guess you mean in fineract-provider? Sure, we can set it there.

galovics: I mean here:

```
<includeAll path="db/custom-changelog" errorIfMissingOrEmpty="false"  />
```
We could also set the context that the custom changelog is only picked up for `tenant_db`. That way nobody can screw up the tenant config DB.

vidakovic: did it

galovics (approved):
LGTM

PR #2693, opened 2022-10-23
FINERACT-1744 add idempotency key storing for http requests

ghost:
## Description

Check https://issues.apache.org/jira/browse/FINERACT-1744

## Checklist

### DIFF
--- a/fineract-provider/src/main/resources/application.properties
+++ b/fineract-provider/src/main/resources/application.properties
@@ -59,6 +59,7 @@ fineract.events.external.producer.jms.enabled=${FINERACT_EXTERNAL_EVENTS_PRODUCE
 fineract.events.external.producer.jms.event-queue-name=${FINERACT_EXTERNAL_EVENTS_PRODUCER_JMS_QUEUE_NAME:JMS-event-queue}
 fineract.events.external.producer.jms.broker-url=${FINERACT_EXTERNAL_EVENTS_PRODUCER_JMS_BROKER_URL:tcp://127.0.0.1:61616}
 
+fineract.idempotency-key-header-name=${FINERACT_IDEMPOTENCY_KEY_HEADER_NAME:Idempotency-Key}
 
 # Logging pattern for the console
 logging.pattern.console=${CONSOLE_LOG_PATTERN:%clr(%d{yyyy-MM-dd HH:mm:ss.SSS}){faint} %clr(${LOG_LEVEL_PATTERN:-%5p}) %clr(${PID:- }){magenta} %clr(%replace([%X{correlationId}]){'\\[\\]', ''}) %clr(---){faint} %clr([%15.15t]){faint} %clr(%-40.40logger{39}){cyan} %clr(:){faint} %m%n${LOG_EXCEPTION_CONVERSION_WORD:%wEx}}
--- a/fineract-provider/src/main/java/org/apache/fineract/commands/service/SynchronousCommandProcessingService.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/commands/service/SynchronousCommandProcessingService.java
@@ -38,6 +38,7 @@ import org.apache.fineract.commands.handler.NewCommandSourceHandler;
 import org.apache.fineract.commands.provider.CommandHandlerProvider;
 import org.apache.fineract.infrastructure.configuration.domain.ConfigurationDomainService;
 import org.apache.fineract.infrastructure.core.api.JsonCommand;
+import org.apache.fineract.infrastructure.core.config.FineractProperties;
 import org.apache.fineract.infrastructure.core.data.CommandProcessingResult;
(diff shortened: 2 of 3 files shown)


fineract-provider/src/main/resources/application.properties
```diff
@@ -59,6 +59,7 @@ fineract.events.external.producer.jms.enabled=${FINERACT_EXTERNAL_EVENTS_PRODUCE
 fineract.events.external.producer.jms.event-queue-name=${FINERACT_EXTERNAL_EVENTS_PRODUCER_JMS_QUEUE_NAME:JMS-event-queue}
 fineract.events.external.producer.jms.broker-url=${FINERACT_EXTERNAL_EVENTS_PRODUCER_JMS_BROKER_URL:tcp://127.0.0.1:61616}
 
+fineract.idempotency-key-header-name=Idempotency-Key
```
galovics: Please make this configurable via an environment variable and have the default value as `Idempotency-Key`

fineract-provider/src/main/java/org/apache/fineract/infrastructure/idempotency/IdempotencyKeyFilter.java
```diff
@@ -0,0 +1,63 @@
...
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.context.SecurityContext;
+import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.stereotype.Component;
+import org.springframework.web.filter.GenericFilterBean;
+
+@Component
+public class IdempotencyKeyFilter extends GenericFilterBean {
+
+    @Value("${fineract.idempotency-key-header-name:Idempotency-Key}")
```
galovics: Not the way to go. Please look at `FineractProperties`.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/idempotency/IdempotencyKeyFilter.java
```diff
@@ -0,0 +1,63 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.http.HttpMethod;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.context.SecurityContext;
+import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.stereotype.Component;
+import org.springframework.web.filter.GenericFilterBean;
+
+@Component
+public class IdempotencyKeyFilter extends GenericFilterBean {
```
galovics: I'm sorry, why do we need this filter and storing the idempotency key in a ThreadLocal?

Wouldn't it be easier to just use the RequestContextHolder from Spring to get access to the HttpServletRequest from anywhere (in the command executor) and read the idempotency key there instead of polluting the already not so great ThreadLocal pattern?

galovics (changes requested):

fineract-provider/src/main/java/org/apache/fineract/infrastructure/idempotency/IdempotencyKeyFilter.java
ghost: done

fineract-provider/src/main/resources/application.properties
ghost: done

fineract-provider/src/main/java/org/apache/fineract/infrastructure/idempotency/IdempotencyKeyFilter.java
ghost: There is no performance or any other problem with this approach (under the hood RequestContextHolder also uses ThreadLocal), however I added a remove command for safety after the pass of the value. Beside me 4 other developers used the ThreadLocalContextUtil for their job and the reason why I also implemented this way is because to see every Fineract specific headers at 1 place for more readable and understandable code instead of just "hiding" the logic deep in the PortfolioCommandSourceWritePlatformServiceImpl class without any trace.

The 2 solution is basically equivalent, the cost of +1 filter calling is just almost equal to zero however RequestContextHolder is not used anywhere in the codebase yet.

If you insist to implement it with your solution, I will change it gladly.

galovics: The ThreadLocalContextUtil class is exposed to everywhere and accessible. The idempotencyKey as a notion is tied to the command execution framework and I'd like to keep it in that bounded context within Fineract. That's why I'm suggesting the alternative.

ghost: suggestion seems logical

galovics (approved):

galovics (approved):

PR #2701, opened 2022-10-25
FINERACT-1783: Customizeable LoanRepaymentScheduleTransactionProcessor

vidakovic:
Currently Fineract supports 6 loan transaction processing strategies:

- standard
- heavens family
- early payment
- creocore
- Indian RBI
- principal/interest/penalty/fees order (default)

The strategy is loaded/instantiated via a factory class "LoanRepaymentScheduleTransactionProcessor" and it's only method "determineProcessor" by passing a "LoanTransactionProcessingStrategy" object. "LoanTransactionProcessingStrategy" is effectively treated like an enum (only its "code" attribute is used). The other attributes "LoanTransactionProcessingStrategy" seem to be only for display purposes in a UI.

Proposed improvements:

- remove hardcoded instantiations of processors
- remove "LoanTransactionProcessingStrategy" and avoid storing this information in the database (table "ref_loan_transaction_processing_strategy") in the first place. This would also remove some unnecessary constraints/joins to this table.
- use dependency injection to make loan transaction processing customizable
- provide possibility to list all available processors for display in a UI


fineract-doc/src/docs/en/chapters/architecture/modules.adoc
```diff
@@ -1,91 +1,3 @@
...
-
-The Docker image with included custom modules is called `fineract-custom`.
-
-NOTE: We'll provide soon a way to customize the Docker image parameters (image name, JVM implementation, JVM args, ports etc.).
-
-
-=== Outlook
-
-If this *proof of concept* is accepted we could prepare Fineract's services to be replaceable. This approach works already very well even if we don't have proper JAR libraries published on Maven Central. It's an important goal to separate customized code from Fineract's internals to have soon real modules.
+TBD
```
adamsaghy, line 3: I think it is fine to keep this description...better than TBD

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/data/LoanSummaryData.java
```diff
@@ -96,7 +96,7 @@ public LoanSummaryData(final CurrencyData currency, final BigDecimal principalDi
         this.interestOverdue = interestOverdue;
         this.feeChargesCharged = feeChargesCharged;
         this.feeChargesDueAtDisbursementCharged = feeChargesDueAtDisbursementCharged;
-        this.feeChargesPaid = feeChargesPaid;
+        this.feeChargesPaid = feeChargesPaid != null ? feeChargesPaid : BigDecimal.ZERO;
```
adamsaghy: For generic behaviour, i would not set only the feeChargesPaid to Zero.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
```diff
@@ -226,9 +225,11 @@ public class Loan extends AbstractAuditableWithUTCDateTimeCustom {
...
     private CodeValue loanPurpose;
 
-    @ManyToOne(fetch = FetchType.EAGER)
-    @JoinColumn(name = "loan_transaction_strategy_id", nullable = true)
-    private LoanTransactionProcessingStrategy transactionProcessingStrategy;
+    @Column(name = "loan_transaction_strategy_code", nullable = false)
+    private String transactionProcessingStrategyCode;
+
+    @Column(name = "loan_transaction_strategy_name")
+    private String transactionProcessingStrategyName;
```
adamsaghy, line 232: Do we need to store the name of the strategy? Wouldnt an enum be better?

vidakovic: Strictly speaking it's not necessary (and we do have an "enum"), but it's there for backward compatibility (read: REST API) with the user interfaces (they used the descriptive field in the old lookup table).
The point here is to avoid any lookups/joins for simple stuff like that and that adding custom transaction strategies is just a matter of creating a custom module with the necessary Spring/Boot configuration (no additional Liquibase migrations etc.).
This column is only descriptive and not used for any lookups of strategies, should be fine as is and is the best way I think to provide descriptive texts for all strategies (even those that we don't know of yet).

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/data/LoanSummaryData.java
vidakovic: Well spotted. I checked if the reason for this line was outside of this constructor, but it's used only in one place. The Git history doesn't show any use of "ZERO", will revert that line.

adamsaghy (approved):
LGTM

PR #2705, opened 2022-10-26
[FINERACT-1678] Reject write API calls for hard-locked loan accounts

taskain7:
## Description

Reject write loan API calls if loan is hard-locked via filter.


fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -2564,6 +2578,12 @@ public CommandProcessingResult bulkLoanReassignment(final JsonCommand command) {
                 .build();
     }
 
+    private boolean isLoanLocked(Long loanId) {
```
galovics: Do you think we could put this into a separate component to be reused from other places in the future?
Maybe it'd make sense to create a service just for the locking domain.

galovics (approved):

PR #2718, opened 2022-11-01
FINEARCT-1744 add command execution 2 step transaction handling

ghost:
## Description

Check FINERACT-1744

## Checklist


fineract-provider/src/main/java/org/apache/fineract/commands/data/AuditSearchData.java
```diff
@@ -34,5 +34,5 @@ public final class AuditSearchData {
     private final Collection<AppUserData> appUsers;
     private final List<String> actionNames;
     private final List<String> entityNames;
-    private final Collection<ProcessingResultLookup> processingResults;
+    private final Collection<ProcessingResultLookup> status;
```
galovics: status**es**, plural.

fineract-provider/src/main/java/org/apache/fineract/commands/service/SynchronousCommandProcessingService.java
```diff
@@ -69,45 +76,43 @@ public class SynchronousCommandProcessingService implements CommandProcessingSer
...
+        String idempotencyKey = getIdempotencyKey(wrapper);
+        CommandSource existingCommand = commandSourceRepository.findByActionNameAndEntityNameAndIdempotencyKey(wrapper.actionName(),
+                wrapper.entityName(), idempotencyKey);
+        final CommandProcessingResult result;
+        if (existingCommand != null) {
+            if (UNDER_PROCESSING.getValue().equals(existingCommand.getStatus())) {
+                throw new DuplicateCommandException(wrapper.actionName(), wrapper.entityName(), wrapper.getIdempotencyKey(),
+                        wrapper.getJson(), UNDER_PROCESSING.getValue());
+            } else if (ERROR.getValue().equals(existingCommand.getStatus())) {
+                throw new DuplicateCommandException(wrapper.actionName(), wrapper.entityName(), wrapper.getIdempotencyKey(),
```
galovics: I don't think this is right. In an error case, we want to return the original error result from the command execution.

fineract-provider/src/main/java/org/apache/fineract/commands/service/SynchronousCommandProcessingService.java
```diff
@@ -69,45 +76,43 @@ public class SynchronousCommandProcessingService implements CommandProcessingSer
...
+                        existingCommand.getResult(), ERROR.getValue());
+            } else if (PROCESSED.getValue().equals(existingCommand.getStatus())) {
+                throw new DuplicateCommandException(wrapper.actionName(), wrapper.entityName(), wrapper.getIdempotencyKey(),
```
galovics: Not right. In case it's been already processed, we should return the original successful result.

fineract-provider/src/main/java/org/apache/fineract/commands/service/SynchronousCommandProcessingService.java
```diff
@@ -69,45 +76,43 @@ public class SynchronousCommandProcessingService implements CommandProcessingSer
...
+                        existingCommand.getResult(), PROCESSED.getValue());
+            }
+        }
 
-        final NewCommandSourceHandler handler = findCommandHandler(wrapper);
+        final AppUser maker = context.authenticatedUser(wrapper);
+        CommandSource commandSourceResult = getInitialCommandSource(wrapper, command, maker, idempotencyKey);
+        initialCommandSaver.save(commandSourceResult);
```
galovics: Couldn't we make this separation better? Just for creating the InitialCommandSaver, to work with the declarative transaction, it's fishy.

Instead why don't we move the whole initial command creation logic into there (including the getInitialCommandSource method).

That way the implementation would be more robust by just calling a "initialCommandCreator.create" method which would create and save the new command. 

Thoughts?

Or... you could create a separate repository method which always saves the entity in a new transaction (using the Custom Spring repository way for example) and then you don't need to have this InitialCommandSaver.

fineract-provider/src/main/java/org/apache/fineract/commands/service/SynchronousCommandProcessingService.java
```diff
@@ -69,45 +76,43 @@ public class SynchronousCommandProcessingService implements CommandProcessingSer
...
-                            .getHeader(fineractProperties.getIdempotencyKeyHeaderName());
-                }
-            }
-
-            commandSourceResult = CommandSource.fullEntryFrom(wrapper, command, maker,
-                    wrapper.getIdempotencyKey() == null
-                            ? (requestIdempotencyKey == null ? idempotencyKeyGenerator.create() : requestIdempotencyKey)
-                            : wrapper.getIdempotencyKey());
-        }
+        commandSourceResult.setResult(toApiJsonSerializer.serializeResult(result)); // TODO
```
galovics: Explain the TODO please.

fineract-provider/src/main/java/org/apache/fineract/commands/service/SynchronousCommandProcessingService.java
```diff
@@ -147,6 +151,34 @@ public CommandProcessingResult executeCommand(final CommandWrapper wrapper, fina
         return result;
     }
 
+    private String getIdempotencyKey(CommandWrapper wrapper) {
```
galovics: Don't we want to extract this method to a separate class just for the sake of making this class easier to test?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exception/DuplicateCommandException.java
```diff
@@ -0,0 +1,54 @@
...
+        super(null, null);
+        this.action = action;
+        this.entity = entity;
+        this.idempotencyKey = idempotencyKey;
+        this.response = response;
+        this.processingResult = processingResult;
+    }
+
+    @Override
+    public String toString() {
```
galovics: Replace with Lombok ToString pls.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exception/DuplicateCommandException.java
```diff
@@ -0,0 +1,54 @@
...
+        return "DuplicateCommandException{" + "action='" + action + '\'' + ", entity='" + entity + '\'' + ", idempotencyKey='"
+                + idempotencyKey + '\'' + ", processingResult=" + processingResult + '}';
+    }
+
+    public int getProcessingResult() {
```
galovics: Lombok Getter pls.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exception/DuplicateCommandException.java
```diff
@@ -0,0 +1,54 @@
...
+        return processingResult;
+    }
+
+    public String getResponse() {
```
galovics: Same as above.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exceptionmapper/DuplicateCommandExceptionMapper.java
```diff
@@ -0,0 +1,52 @@
...
+
+@Provider
+@Component
+@Slf4j
+public class DuplicateCommandExceptionMapper implements ExceptionMapper<DuplicateCommandException> {
+
+    @Override
+    public Response toResponse(final DuplicateCommandException exception) {
+        log.debug("Duplicate request: {}", exception.toString());
+        if (PROCESSED.getValue() == exception.getProcessingResult() || ERROR.getValue() == exception.getProcessingResult()) {
```
galovics: Please don't do this. :-)
Rather create different exceptions for the different use-cases.

galovics (changes requested):
The build is failing and this change is now calling for unit tests. Please work on it a little more.

fineract-provider/src/main/java/org/apache/fineract/commands/service/SynchronousCommandProcessingService.java
ghost: this solution with the custom error mapper completely satisfy the specification, tested with multiple threads, works

fineract-provider/src/main/java/org/apache/fineract/commands/service/SynchronousCommandProcessingService.java
ghost: this solution with the custom error mapper completely satisfy the specification, tested with multiple threads, works, but it is an open question what and how to save to the result column in case of a generic error, waiting for answer...

fineract-provider/src/main/java/org/apache/fineract/commands/service/SynchronousCommandProcessingService.java
ghost: just leftover

fineract-provider/src/main/java/org/apache/fineract/commands/service/SynchronousCommandProcessingService.java
ghost: basically a few getters and setters but I will move it, the class is already a bloated dependency chaos but a 11. injected field will do no harm

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exception/DuplicateCommandException.java
ghost: will do

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exception/DuplicateCommandException.java
ghost: will do

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exception/DuplicateCommandException.java
ghost: will do

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exceptionmapper/DuplicateCommandExceptionMapper.java
ghost: rather create 2 more unnecessary classes than simply pass a primitive int parameter?

fineract-provider/src/main/java/org/apache/fineract/commands/service/SynchronousCommandProcessingService.java
ghost: will do but YAGNI, it is just a simple extracted method for cleaner understanding, the initial command saver can not be avoided because transaction forcing inside transactions only works with public method calls

ghost:
> The build is failing and this change is now calling for unit tests. Please work on it a little more.

will do

fineract-provider/src/main/java/org/apache/fineract/commands/data/AuditSearchData.java
ghost: will do

fineract-provider/src/main/java/org/apache/fineract/commands/service/CommandSaver.java
```diff
@@ -0,0 +1,75 @@
...
+import org.springframework.transaction.annotation.Propagation;
+import org.springframework.transaction.annotation.Transactional;
+
+/**
+ * Two phase transactional command processing: save initial...work...finish/failed to handle idempotent requests.
+ */
+@Component
+public class CommandSaver {
+
+    @Autowired
```
galovics: Please us constructor injection in combination with @RequiredArgsConstructor

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/jpa/JpaExceptionHandler.java
```diff
@@ -0,0 +1,88 @@
...
+import org.springframework.beans.BeansException;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
+import org.springframework.stereotype.Component;
+import org.springframework.web.context.request.RequestAttributes;
+import org.springframework.web.context.request.RequestContextHolder;
+import org.springframework.web.context.request.ServletRequestAttributes;
+
+@Component
+public class JpaExceptionHandler implements ExceptionHandler, ApplicationContextAware {
```
galovics, line 35: I don't get why you need this class especially with mixing Spring's and JPA's lifecycle.
Cant you just put a try catch around the command execution and unwrap the proper exception there?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exceptionmapper/DuplicateCommandExceptionMapper.java
galovics: Nope. Rather create specific exceptions for specific use-cases. It's always fishy to use an exception and pass a parameter within that to decide later on some logic.

galovics (changes requested):

fineract-provider/src/main/java/org/apache/fineract/commands/service/SynchronousCommandProcessingService.java
galovics: > because transaction forcing inside transactions only works with public method calls

That's only true for declarative tx management. If you use a TransactionTemplate, you're pretty much good to go.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/jpa/JpaExceptionHandler.java
ghost: - in case 2 threads are calling with the same IDK and trying to save the same request initially then a java.sql.SQLIntegrityConstraintViolationException is thrown down the bottom, wrapped in a DatabaseException which is handled by jpa and can not be catched outside -> simply a 500 error will be generated and a dummy result will be sent to the caller, we can not execute custom logic after/before/anywhere of it to save FAILED status or to creat customresult for the caller
- spring is mixed in because Idempotency header's name is configurable therefore we need to access the environment

fineract-provider/src/main/java/org/apache/fineract/commands/service/SynchronousCommandProcessingService.java
ghost: The codebase is already using declarative tx mangement at numerous places. (very appropriately instead of mixing techniques)

fineract-provider/src/main/java/org/apache/fineract/commands/service/CommandSaver.java
ghost: done

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exceptionmapper/DuplicateCommandExceptionMapper.java
ghost: done

ghost:
failed command:
image

success command:
image

fineract-provider/src/main/java/org/apache/fineract/commands/service/SynchronousCommandProcessingService.java
```diff
@@ -64,69 +66,58 @@ public class SynchronousCommandProcessingService implements CommandProcessingSer
...
         } catch (Throwable t) {
-            publishHookErrorEvent(wrapper, command, t);
+            ErrorInfo ex;
+            if (t instanceof final RuntimeException e) {
+                ex = ErrorHandler.handler(e);
+            } else {
+                ex = new ErrorInfo(500, 9999, "{\"Exception\": " + t.toString() + "}");
+            }
+            publishHookEvent(wrapper.entityName(), wrapper.actionName(), command, ex);
+            commandSourceService.saveFailed(ex.getMessage(), commandSourceService.findCommandSource(wrapper, idempotencyKey));
```
galovics, line 97: So why do you need to call "findCommandSource" once more here? Why don't you reuse the result of the saveInitial call from above?

fineract-provider/src/main/java/org/apache/fineract/commands/service/SynchronousCommandProcessingService.java
```diff
@@ -64,69 +66,58 @@ public class SynchronousCommandProcessingService implements CommandProcessingSer
...
     @Retry(name = "executeCommand", fallbackMethod = "fallbackExecuteCommand")
     public CommandProcessingResult executeCommand(final CommandWrapper wrapper, final JsonCommand command,
             final boolean isApprovedByChecker) {
 
         final boolean rollbackTransaction = configurationDomainService.isMakerCheckerEnabledForTask(wrapper.taskPermissionName());
+        String idempotencyKey = idempotencyKeyResolver.resolve(wrapper);
+        checkExistingCommand(wrapper, idempotencyKey);
 
-        final NewCommandSourceHandler handler = findCommandHandler(wrapper);
+        commandSourceService.saveInitial(wrapper, command, context.authenticatedUser(wrapper), idempotencyKey);
```
galovics, line 84: Why don't you move the context.authenticatedUser logic into the saveInitial call? That' mean less parameters here. Same for the idempotency key.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/jpa/JpaExceptionHandler.java
galovics: Can you pls share the other implementation where you tried to catch the exception? I'm having doubts to understand why this exception cannot be caught simply on a higher level and handled properly.

On the other hand, if this class is really needed, it still shouldn't be a Bean as well as instantiated by EclipseLink. Mixing objects like that is a smell.
If you need access to the AppContext from a static context, rather create a separate class that's ONLY a Spring Bean and provides access statically.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/exceptionmapper/DuplicateCommandExceptionMapper.java
```diff
@@ -0,0 +1,70 @@
...
+
+@Provider
+@Component
+@Slf4j
+public class DuplicateCommandExceptionMapper implements ExceptionMapper<DuplicateCommandException> {
+
+    @Override
+    public Response toResponse(final DuplicateCommandException exception) {
+        log.debug("Duplicate request: {}", exception.getMessage());
+        if (exception instanceof CommandProcessedException) {
```
galovics, line 45: I'm sorry but I don't understand it. Why the if statements here?
Why not just create separate exception mappers for each type of exception.

galovics (changes requested):
I don't see integration tests for verifying the idempotency functionality. Both positive and negative cases.

PR #2727, opened 2022-11-06
FINERACT-1795: Improve resilience of command processing service

vidakovic:
Improve the resilience of the command processing service and make retry (and other parameters) configurable (provide reasonable defaults).

The current retry mechanics in place are causing various congestion issues in high traffic deployments. Besides that, we always have to lookup the tenant to retrieve the retry configuration; Spring Boot's application.properties are the proper place for that.

Retries are used currently with 4 service function calls and the code used to implement retries is very repetitive and not very adjustable.

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/domain/FineractPlatformTenantConnection.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/domain/FineractPlatformTenantConnection.java
@@ -54,8 +54,6 @@ public class FineractPlatformTenantConnection implements Serializable {
     private final int suspectTimeout;
     private final int timeBetweenEvictionRunsMillis;
     private final int minEvictableIdleTimeMillis;
-    private final int maxRetriesOnDeadlock;
-    private final int maxIntervalBetweenRetries;
     private final boolean testOnBorrow;
 
     public FineractPlatformTenantConnection(final Long connectionId, final String schemaName, String schemaServer,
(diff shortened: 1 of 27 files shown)


vidakovic:
I've investigated that failure a bit and can't make any sense of it. The server side exception mapping seems to be OK as far as I can tell (investigated all execution paths). Nevertheless, on the integration test we get a HTTP 204 instead of 404. As far as I can see this is not related to my PR (made no changes in the area where things are breaking).

vidakovic:
I think I found it... tricky. Needed to re-throw exception in `logCommand` fallback method; it was swallowing the exceptions and that resulted utlimately in the HTTP 204 code.

vidakovic:
Adjusting the tests took more time than should be necessary... the whole tenant based database setup is impossible to mock for testing... there are just to many factories inside of factories calling static functions in some obscure utility classes.
Found a solution for now and left a comment.

vidakovic:
Alright... I think that's it. Removed some overly complex error handling for some retry scenarios and fixed a hidden issue in one of the retry fallback functions (missing return type leads to wrong method signature).

vidakovic:
Added documentation and more cleanups.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/domain/FineractPlatformTenantConnection.java
```diff
@@ -171,10 +171,12 @@ public int getMinEvictableIdleTimeMillis() {
         return this.minEvictableIdleTimeMillis;
     }
 
+    @Deprecated
```
galovics: I think let's just cut this out from here and from the DB as well. We'll mention it in the changelog.

vidakovic: Very well... on the way

vidakovic: @galovics done... dropped also the columns from the tenant configuration DB.

galovics (approved):

PR #2731, opened 2022-11-07
[FINERACT-1678] Introducing synchronous COB execution for loan account changing APIs

taskain7:
## Description

When a loan is soft locked by COB partitioning during write operation, then the Inline COB runs on the loan instead of regular COB and after that the write operation goes through.

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/cob/service/InlineLoanCOBExecutorServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/cob/service/InlineLoanCOBExecutorServiceImpl.java
@@ -66,7 +66,7 @@ import org.springframework.transaction.support.TransactionTemplate;
 @Service
 @Slf4j
 @RequiredArgsConstructor
-public class InlineLoanCOBExecutorServiceImpl implements InlineExecutorService, InitializingBean {
+public class InlineLoanCOBExecutorServiceImpl implements InlineExecutorService<Long>, InitializingBean {
 
     private static final String JOB_EXECUTION_FAILED_MESSAGE = "Job execution failed for job with name: ";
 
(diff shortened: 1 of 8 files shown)


fineract-provider/src/main/java/org/apache/fineract/cob/service/LoanAccountLockServiceImpl.java
```diff
@@ -51,4 +51,11 @@ public boolean isLoanHardLocked(Long loanId) {
                         || LockOwner.LOAN_INLINE_COB_PROCESSING.equals(loanAccountLock.getLockOwner())) //
                 .filter(loanAccountLock -> StringUtils.isBlank(loanAccountLock.getError())).isPresent();
     }
+
+    @Override
+    public boolean isLoanSoftLocked(Long loanId) {
+        Optional<LoanAccountLock> loanAccountLockOptional = loanAccountLockRepository.findById(loanId);
```
galovics: I think it's not needed to fetch the actual entity. Let's write a custom repository method for this.
Something like `existsByIdAndLockOwner`

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/SecurityConfig.java
```diff
@@ -90,7 +89,7 @@ protected void configure(HttpSecurity http) throws Exception {
                 .addFilterAfter(fineractInstanceModeApiFilter, SecurityContextPersistenceFilter.class) //
                 .addFilterAfter(tenantAwareBasicAuthenticationFilter(), FineractInstanceModeApiFilter.class) //
                 .addFilterAfter(loanCOBApiFilter, TenantAwareBasicAuthenticationFilter.class) //
-                .addFilterAfter(twoFactorAuthenticationFilter, BasicAuthenticationFilter.class); //
+                .addFilterAfter(twoFactorAuthenticationFilter, LoanCOBApiFilter.class); //
```
galovics: So where do the BasicAuthenticationFilter go?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/jobs/filter/LoanCOBApiFilter.java
```diff
@@ -67,35 +70,51 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
...
             } else {
                 proceed(filterChain, request, response);
             }
         }
     }
 
-    private boolean isBypassUser() {
-        return context.getAuthenticatedUserIfPresent().isBypassUser();
+    private void executeInlineCob(List<Long> loanIds) {
+        inlineLoanCOBExecutorService.lockLoanAccounts(loanIds);
```
galovics: Hmm, I'm thinking about the responsibilities. I think it'd be better if the service itself locks the loan accounts, don't you think?
Like that's a unit in itself. You instruct the service "let's run the inline COB for these loan accounts" and you should expect that everything is taken care of.

fineract-provider/src/main/java/org/apache/fineract/cob/service/InlineLoanCOBExecutorServiceImpl.java
```diff
@@ -121,7 +121,7 @@ private List<LoanAccountLock> getLoanAccountLocks(List<Long> loanIds) {
         return loanAccountLocks;
     }
 
-    private void execute(List<Long> loanIds, String jobName) {
+    public void execute(List<Long> loanIds, String jobName) {
```
galovics: Can't we add this method to the interface with a generic T type which refers to the entity ID type?

galovics (changes requested):

fineract-provider/src/main/java/org/apache/fineract/cob/service/InlineLoanCOBExecutorServiceImpl.java
```diff
@@ -66,7 +66,7 @@
 @Service
 @Slf4j
 @RequiredArgsConstructor
-public class InlineLoanCOBExecutorServiceImpl implements InlineExecutorService, InitializingBean {
+public class InlineLoanCOBExecutorServiceImpl implements InlineExecutorService<List<Long>>, InitializingBean {
```
galovics: What if we  simply don't specify the List part of the generic parameter? Just the Long. That way all the interface implementors will have the same List input type for the method.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/jobs/service/InlineExecutorService.java
```diff
@@ -21,7 +21,9 @@
 import org.apache.fineract.infrastructure.core.api.JsonCommand;
 import org.apache.fineract.infrastructure.core.data.CommandProcessingResult;
 
-public interface InlineExecutorService {
+public interface InlineExecutorService<T> {
 
     CommandProcessingResult executeInlineJob(JsonCommand command, String jobName);
+
+    void execute(T id, String jobName);
```
galovics: So rather this should look like:
```
void execute(List<T> elements, String jobName);
```

Plus, it'd be good to provide a way to do the same thing for a single item, so like we could add this:
```
default void execute(T element, String jobName) {
    execute(singletonList(element), jobName)
}
```
Thoughts?

galovics (changes requested):

fineract-provider/src/main/java/org/apache/fineract/infrastructure/jobs/service/InlineExecutorService.java
taskain7: makes sense, I've chenged it.

galovics (approved):

PR #2739, opened 2022-11-11
FINERACT-1706: Fix Loan Delinquency Classification after Repayment

josehernandezfintecheandomx:
## Description

The delinquency calculation must to happen as of the latest unpaid chargeback transaction date in order to reflect the right delinquency days and amount

[FINERACT-1706](https://issues.apache.org/jira/browse/FINERACT-1706).

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyReadPlatformServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyReadPlatformServiceImpl.java
@@ -18,15 +18,10 @@
  */
 package org.apache.fineract.portfolio.delinquency.service;
 
-import java.math.BigDecimal;
-import java.time.LocalDate;
 import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
-import java.util.Set;
 import lombok.RequiredArgsConstructor;
-import org.apache.fineract.infrastructure.core.service.DateUtils;
-import org.apache.fineract.organisation.monetary.domain.MonetaryCurrency;
 import org.apache.fineract.portfolio.delinquency.data.DelinquencyBucketData;
 import org.apache.fineract.portfolio.delinquency.data.DelinquencyRangeData;
 import org.apache.fineract.portfolio.delinquency.data.LoanDelinquencyTagHistoryData;
@@ -41,11 +36,8 @@ import org.apache.fineract.portfolio.delinquency.mapper.DelinquencyRangeMapper;
 import org.apache.fineract.portfolio.delinquency.mapper.LoanDelinquencyTagMapper;
 import org.apache.fineract.portfolio.loanaccount.data.CollectionData;
 import org.apache.fineract.portfolio.loanaccount.domain.Loan;
-import org.apache.fineract.portfolio.loanaccount.domain.LoanRepaymentScheduleInstallment;
 import org.apache.fineract.portfolio.loanaccount.domain.LoanRepository;
 import org.apache.fineract.portfolio.loanaccount.domain.LoanTransaction;
-import org.apache.fineract.portfolio.loanaccount.domain.LoanTransactionToRepaymentScheduleMapping;
-import org.apache.fineract.portfolio.loanaccount.domain.LoanTransactionType;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 
@@ -61,6 +53,7 @@ public class DelinquencyReadPlatformServiceImpl implements DelinquencyReadPlatfo
     private final DelinquencyBucketMapper mapperBucket;
     private final LoanDelinquencyTagMapper mapperLoanDelinquencyTagHistory;
     private final LoanRepository loanRepository;
+    private final LoanDelinquencyDomainService loanDelinquencyDomainService;
 
     @Override
     public Collection<DelinquencyRangeData> retrieveAllDelinquencyRanges() {
@@ -109,89 +102,21 @@ public class DelinquencyReadPlatformServiceImpl implements DelinquencyReadPlatfo
 
     @Override
     public CollectionData calculateLoanCollectionData(final Long loanId) {
-        final LocalDate businessDate = DateUtils.getBusinessLocalDate();
         final Optional<Loan> optLoan = this.loanRepository.findById(loanId);
 
         CollectionData collectionData = CollectionData.template();
         if (optLoan.isPresent()) {
             final Loan loan = optLoan.get();
 
+            collectionData = loanDelinquencyDomainService.getOverdueCollectionData(loan);
             collectionData.setAvailableDisbursementAmount(loan.getApprovedPrincipal().subtract(loan.getDisbursedAmount()));
             collectionData.setNextPaymentDueDate(loan.possibleNextRepaymentDate());
-
-            final MonetaryCurrency currency = loan.getCurrency();
-            BigDecimal delinquentAmount = BigDecimal.ZERO;
-            // Overdue Days calculation
-            Long overdueDays = 0L;
-            LocalDate overdueSinceDate = null;
-            final List<LoanTransaction> chargebackTransactions = loan.retrieveListOfTransactionsByType(LoanTransactionType.CHARGEBACK);
-            for (LoanTransaction loanTransaction : chargebackTransactions) {
-                Set<LoanTransactionToRepaymentScheduleMapping> loanTransactionToRepaymentScheduleMappings = loanTransaction
-                        .getLoanTransactionToRepaymentScheduleMappings();
-                LoanTransactionToRepaymentScheduleMapping loanTransactionToRepaymentScheduleMapping = loanTransactionToRepaymentScheduleMappings
-                        .iterator().next();
-                if (!loanTransactionToRepaymentScheduleMapping.getLoanRepaymentScheduleInstallment().isPrincipalCompleted(currency)) {
-                    overdueSinceDate = loanTransaction.getTransactionDate();
-                    break;
-                }
-            }
-
-            final List<LoanRepaymentScheduleInstallment> installments = loan.getRepaymentScheduleInstallments();
-            if (overdueSinceDate == null) {
-                for (LoanRepaymentScheduleInstallment installment : installments) {
-                    if (installment.isNotFullyPaidOff()) {
-                        overdueSinceDate = installment.getDueDate();
-                        break;
-                    }
-                }
-            }
-
-            Integer graceDays = 0;
-            if (overdueSinceDate != null) {
-                if (loan.loanProduct().getLoanProductConfigurableAttributes().getGraceOnArrearsAgingBoolean()) {
-                    graceDays = loan.loanProduct().getLoanProductRelatedDetail().getGraceOnArrearsAgeing();
-                    if (graceDays == null) {
-                        graceDays = 0;
-                    }
-                    overdueSinceDate = overdueSinceDate.plusDays(graceDays);
-                }
-                overdueDays = DateUtils.getDifferenceInDays(overdueSinceDate, businessDate);
-                if (overdueDays < 0) {
-                    overdueDays = 0L;
-                }
-                collectionData.setDelinquentDate(overdueSinceDate);
-            }
-
-            collectionData.setPastDueDays(overdueDays);
-            if (overdueDays > 0) {
-                collectionData.setDelinquentDays(overdueDays - graceDays);
-            }
 
             final LoanTransaction lastRepaymenTransaction = loan.getLastRepaymentTransaction();
             if (lastRepaymenTransaction != null) {
                 collectionData.setLastPaymentDate(lastRepaymenTransaction.getTransactionDate());
                 collectionData.setLastPaymentAmount(lastRepaymenTransaction.getAmount());
             }
-
-            // Calculate Delinquency Amount
-            for (LoanRepaymentScheduleInstallment installment : installments) {
-                if (installment.getDueDate().isBefore(businessDate)) {
-                    delinquentAmount = delinquentAmount.add(installment.getTotalOutstanding(currency).getAmount());
-                }
-            }
-            for (LoanTransaction loanTransaction : chargebackTransactions) {
-                Set<LoanTransactionToRepaymentScheduleMapping> loanTransactionToRepaymentScheduleMappings = loanTransaction
-                        .getLoanTransactionToRepaymentScheduleMappings();
-                for (LoanTransactionToRepaymentScheduleMapping loanTransactionToRepaymentScheduleMapping : loanTransactionToRepaymentScheduleMappings) {
-                    LoanRepaymentScheduleInstallment loanRepaymentScheduleInstallment = loanTransactionToRepaymentScheduleMapping
-                            .getLoanRepaymentScheduleInstallment();
-                    if (!loanRepaymentScheduleInstallment.isPrincipalCompleted(currency)) {
-                        delinquentAmount = delinquentAmount.add(loanTransaction.getAmount());
-                    }
-                }
-            }
-
-            collectionData.setDelinquentAmount(delinquentAmount);
         }
 
         return collectionData;
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyWritePlatformServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyWritePlatformServiceImpl.java
@@ -35,7 +35,6 @@ import org.apache.fineract.infrastructure.core.exception.PlatformDataIntegrityEx
 import org.apache.fineract.infrastructure.core.service.DateUtils;
 import org.apache.fineract.infrastructure.event.business.domain.loan.LoanDelinquencyRangeChangeBusinessEvent;
 import org.apache.fineract.infrastructure.event.business.service.BusinessEventNotifierService;
-import org.apache.fineract.organisation.monetary.domain.MonetaryCurrency;
 import org.apache.fineract.portfolio.delinquency.api.DelinquencyApiConstants;
 import org.apache.fineract.portfolio.delinquency.data.DelinquencyBucketData;
 import org.apache.fineract.portfolio.delinquency.data.DelinquencyRangeData;
@@ -51,13 +50,10 @@ import org.apache.fineract.portfolio.delinquency.exception.DelinquencyBucketAges
 import org.apache.fineract.portfolio.delinquency.exception.DelinquencyRangeInvalidAgesException;
 import org.apache.fineract.portfolio.delinquency.validator.DelinquencyBucketParseAndValidator;
 import org.apache.fineract.portfolio.delinquency.validator.DelinquencyRangeParseAndValidator;
+import org.apache.fineract.portfolio.loanaccount.data.CollectionData;
 import org.apache.fineract.portfolio.loanaccount.data.LoanScheduleDelinquencyData;
 import org.apache.fineract.portfolio.loanaccount.domain.Loan;
-import org.apache.fineract.portfolio.loanaccount.domain.LoanRepaymentScheduleInstallment;
 import org.apache.fineract.portfolio.loanaccount.domain.LoanRepositoryWrapper;
-import org.apache.fineract.portfolio.loanaccount.domain.LoanTransaction;
-import org.apache.fineract.portfolio.loanaccount.domain.LoanTransactionToRepaymentScheduleMapping;
-import org.apache.fineract.portfolio.loanaccount.domain.LoanTransactionType;
 import org.apache.fineract.portfolio.loanproduct.domain.LoanProductRepository;
 import org.springframework.stereotype.Service;
 
@@ -76,6 +72,7 @@ public class DelinquencyWritePlatformServiceImpl implements DelinquencyWritePlat
     private final LoanRepositoryWrapper loanRepository;
     private final LoanProductRepository loanProductRepository;
     private final BusinessEventNotifierService businessEventNotifierService;
+    private final LoanDelinquencyDomainService loanDelinquencyDomainService;
 
     @Override
     public CommandProcessingResult createDelinquencyRange(JsonCommand command) {
@@ -147,66 +144,13 @@ public class DelinquencyWritePlatformServiceImpl implements DelinquencyWritePlat
 
     @Override
     public LoanScheduleDelinquencyData calculateDelinquencyData(LoanScheduleDelinquencyData loanScheduleDelinquencyData) {
-        final LocalDate businessDate = DateUtils.getBusinessLocalDate();
         Loan loan = loanScheduleDelinquencyData.getLoan();
         if (loan == null) {
             loan = this.loanRepository.findOneWithNotFoundDetection(loanScheduleDelinquencyData.getLoanId());
         }
-
-        final Integer graceOnArrearAgeing = loan.getLoanProduct().getLoanProductRelatedDetail().getGraceOnArrearsAgeing();
-        final LocalDate overdueSinceDate = getOverdueSinceDate(loan, businessDate, graceOnArrearAgeing);
-        final Long overdueDays = calculateOverdueDays(businessDate, overdueSinceDate);
-        return new LoanScheduleDelinquencyData(loan.getId(), overdueSinceDate, overdueDays, loan);
-    }
-
-    @Override
-    public LocalDate getOverdueSinceDate(final Loan loan, final LocalDate businessDate, final Integer graceOnArrearAgeing) {
-        LoanRepaymentScheduleInstallment loanRepaymentSchedule = null;
-        for (LoanRepaymentScheduleInstallment installment : loan.getRepaymentScheduleInstallments()) {
-            if (installment.getDueDate().isBefore(businessDate) && installment.isNotFullyPaidOff()) {
-                loanRepaymentSchedule = installment;
-                break;
-            }
-        }
-
-        LocalDate overdueSinceDate = null;
-        final MonetaryCurrency loanCurrency = loan.getCurrency();
-        final List<LoanTransaction> loanTransactions = loan.retrieveListOfTransactionsByType(LoanTransactionType.CHARGEBACK);
-        if (loanRepaymentSchedule != null) {
-            // Default Due date
-            overdueSinceDate = loanRepaymentSchedule.getDueDate();
-        }
-
-        // If there is some Loan Transaction Chargeback
-        for (LoanTransaction loanTransaction : loanTransactions) {
-            if (loanTransaction.getLoanTransactionToRepaymentScheduleMappings().iterator().hasNext()) {
-                final LoanTransactionToRepaymentScheduleMapping transactionMapping = loanTransaction
-                        .getLoanTransactionToRepaymentScheduleMappings().iterator().next();
-
-                if (transactionMapping != null
-                        && !transactionMapping.getLoanRepaymentScheduleInstallment().isPrincipalCompleted(loanCurrency)) {
-                    overdueSinceDate = loanTransaction.getTransactionDate();
-                    break;
-                }
-            }
-        }
-
-        // Include grace days If It's defined
-        if (overdueSinceDate != null && graceOnArrearAgeing != null) {
-            overdueSinceDate = overdueSinceDate.plusDays(graceOnArrearAgeing.longValue());
-        }
-        return overdueSinceDate;
-    }
-
-    private Long calculateOverdueDays(final LocalDate businessDate, LocalDate overdueSinceDate) {
-        Long overdueDays = 0L;
-        if (overdueSinceDate != null) {
-            overdueDays = DateUtils.getDifferenceInDays(overdueSinceDate, businessDate);
-            if (overdueDays < 0) {
-                overdueDays = 0L;
-            }
-        }
-        return overdueDays;
+        final CollectionData collectionData = loanDelinquencyDomainService.getOverdueCollectionData(loan);
+        log.debug("Delinquency {}", collectionData);
+        return new LoanScheduleDelinquencyData(loan.getId(), collectionData.getDelinquentDate(), collectionData.getDelinquentDays(), loan);
     }
 
     @Override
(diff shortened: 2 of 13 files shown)


fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyReadPlatformServiceImpl.java
```diff
@@ -109,52 +111,30 @@ public Collection<LoanDelinquencyTagHistoryData> retrieveDelinquencyRangeHistory
...
+            Integer graceDays = 0;
+            if (loan.loanProduct().getLoanProductConfigurableAttributes().getGraceOnArrearsAgingBoolean()) {
+                graceDays = loan.loanProduct().getLoanProductRelatedDetail().getGraceOnArrearsAgeing();
+                if (graceDays == null) {
+                    graceDays = 0;
+                }
+            }
 
-            collectionData.setAvailableDisbursementAmount(loan.getApprovedPrincipal().subtract(loan.getDisbursedAmount()));
+            collectionData.setAvailableDisbursementAmount(loan.getApprovedPrincipal().subtract(loan.getNetDisbursalAmount()));
```
adamsaghy: i dont think this is correct.

The netDisbursalAmount will be overriden always by the very latest disbursement amount.
**Example:**
- Approved amount is 1000
- I do a disbursement with amount of 400
- The netDisbursalAmount will be 400, so it will set the available disbursement amount is 600
- If I do a 2nd disbursement with amount of 600
- The netDisbursalAmount will be 600, so the available disbursement amount will be set to 400, but that is not correct. We should consider all of the disbursement details to get the correct amount.

@josehernandezfintecheandomx what do you think?

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyReadPlatformServiceImpl.java
```diff
@@ -109,52 +111,30 @@ public Collection<LoanDelinquencyTagHistoryData> retrieveDelinquencyRangeHistory
...
-                Set<LoanTransactionToRepaymentScheduleMapping> loanTransactionToRepaymentScheduleMappings = loanTransaction
-                        .getLoanTransactionToRepaymentScheduleMappings();
-                LoanTransactionToRepaymentScheduleMapping loanTransactionToRepaymentScheduleMapping = loanTransactionToRepaymentScheduleMappings
-                        .iterator().next();
-                if (!loanTransactionToRepaymentScheduleMapping.getLoanRepaymentScheduleInstallment().isPrincipalCompleted(currency)) {
-                    overdueSinceDate = loanTransaction.getTransactionDate();
-                    break;
-                }
-            }
+            final LocalDate overdueSinceDate = DelinquencyUtils.getOverdueSinceDate(loan, businessDate, graceDays);
```
adamsaghy: You can get the overDueSinceDate and the amount in the same calculation. You dont need to do twice the same:
- DelinquencyUtils.getOverdueSinceDate(loan, businessDate, graceDays) 
- From line 159-177, it is the same.

Please merge them together.

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyReadPlatformServiceImpl.java
```diff
@@ -174,20 +154,27 @@ public CollectionData calculateLoanCollectionData(final Long loanId) {
             }
 
             // Calculate Delinquency Amount
-            for (LoanRepaymentScheduleInstallment installment : installments) {
+            Map<Integer, BigDecimal> installmentBalances = new HashMap<>();
```
adamsaghy: This can be removed, the logic in DelinquencyUtils.getOverdueSinceDate can be extended to get the delinquent amount as well.

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyUtils.java
```diff
@@ -0,0 +1,85 @@
...
+            }
+        }
+
+        LocalDate overdueSinceDate = null;
+        if (loanRepaymentSchedule != null) {
+            // Default Due date
+            overdueSinceDate = loanRepaymentSchedule.getDueDate();
+        }
+
+        Map<Integer, BigDecimal> installmentBalances = new HashMap<>();
```
adamsaghy: This is not needed. You already have the first unpaid instalment (see line 42)

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyUtils.java
```diff
@@ -0,0 +1,85 @@
...
+        for (LoanRepaymentScheduleInstallment installment : loan.getRepaymentScheduleInstallments()) {
+            installmentBalances.put(installment.getInstallmentNumber(), installment.getPrincipalCompleted(loanCurrency).getAmount());
+        }
+
+        // If there is some Loan Transaction Chargeback
+        final List<LoanTransaction> loanTransactions = loan.retrieveListOfTransactionsByType(LoanTransactionType.CHARGEBACK);
+        for (LoanTransaction loanTransaction : loanTransactions) {
```
adamsaghy: This can be simplified:
- You already knows which is the first unpaid instalment
- Put the total paid principal for that instalment into a temporary variable
- If there are chargeback transactions for that instalment (fetch those loantransactions through the repayment mapping or simply iterate through the loan transactions and check the from- & due date of instalment against the transaction date), as iterating through those transaction, deduct the transaction amount from this temporary variable. And check after the deduction it went to negative. If yes, that transaction is not fully paid yet, so the delinquent amount can be collected from that moment and the overdue since date can be set with the transaction date of that transaction.
BTW question: only chargeback transaction should increase the delinquent amount?

adamsaghy (changes requested):
See my comments!

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyUtils.java
```diff
@@ -0,0 +1,117 @@
...
+                }
+            }
+        }
+
+        if (nextInstallment == null && loan.getRepaymentScheduleInstallments().size() > 0) {
+            nextInstallment = loan.getRepaymentScheduleInstallments().get(loan.getRepaymentScheduleInstallments().size() - 1);
+        }
+
+        // If exists a next no overdue installment
+        if (nextInstallment != null) {
```
adamsaghy: I think this logic is not considering properly the chargeback transactions.
**Example:**
- There are 2 installments
- First one is not fully paid and overdue and contains a chargeback transaction
- The second one is not due yet 

The logic here will not set the overdueSinceDate by the chargeback transaction as it was never checked.
The nextInstallment will be set to check the second installment, which is not even due and  might overwrite the overdueSinceDate which is wrong, the first unpaid chargeback transaction is in the first unpaid installment

adamsaghy (changes requested):
See my comments.

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyUtils.java
```diff
@@ -0,0 +1,120 @@
...
+
+                for (LoanTransactionToRepaymentScheduleMapping mappngInstallment : installment
+                        .getLoanTransactionToRepaymentScheduleMappings()) {
+                    final LoanTransaction loanTransaction = mappngInstallment.getLoanTransaction();
+                    if (loanTransaction.isChargeback() && loanTransaction.getTransactionDate().isBefore(businessDate)) {
+                        amountAvailable = amountAvailable.subtract(loanTransaction.getAmount());
+                        if (amountAvailable.compareTo(BigDecimal.ZERO) < 0 && !overdueSinceDateWasSet) {
+                            overdueSinceDate = loanTransaction.getTransactionDate();
+                            overdueSinceDateWasSet = true;
+                            break;
```
adamsaghy: We dont want to break the iteration. We need to process all the due chargeback transactions. Please remove the `break`.

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyReadPlatformServiceImpl.java
adamsaghy: @josehernandezfintecheandomx This is still not fixed. Please share your thoughts about the issue I listed above.

adamsaghy (changes requested):
Please see my comments!

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanTransactionChargebackTest.java
```diff
@@ -160,7 +159,7 @@ public void applyLoanTransactionChargebackInLongTermLoan() {
         }
 
         loanTransactionHelper.evaluateLoanSummaryAdjustments(getLoansLoanIdResponse, Double.valueOf(baseAmount));
-        DelinquencyBucketsHelper.evaluateLoanCollectionData(getLoansLoanIdResponse, 0, Double.valueOf("333.33"));
+        DelinquencyBucketsHelper.evaluateLoanCollectionData(getLoansLoanIdResponse, 0, Double.valueOf("0.00"));
```
adamsaghy, line 191: Would you please extend this test case to increase the business date by 1 and run the COB and check the collectiondata will be updated with the proper details?

josehernandezfintecheandomx: Done!

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyUtils.java
josehernandezfintecheandomx: Removed

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyUtils.java
josehernandezfintecheandomx: Done

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyUtils.java
josehernandezfintecheandomx: Removed

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyReadPlatformServiceImpl.java
josehernandezfintecheandomx: Removed

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyReadPlatformServiceImpl.java
josehernandezfintecheandomx: Updared

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
```diff
@@ -6781,15 +6781,19 @@ public Collection<LoanCharge> getLoanCharges() {
...
-        checkAndFetchLazyCollection(this.loanTransactions);
+        initializeLazyBasicCollections();
         checkAndFetchLazyCollection(this.disbursementDetails);
         checkAndFetchLazyCollection(this.loanTermVariations);
         checkAndFetchLazyCollection(this.collateral);
         checkAndFetchLazyCollection(this.loanOfficerHistory);
         checkAndFetchLazyCollection(this.loanCollateralManagements);
     }
 
+    public void initializeLazyBasicCollections() {
```
adamsaghy: Please rename this. "Basic collections" is really non-descriptive method name. Suggestion: `fetchRepaymentScheduleAndTransactions()`

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanRepaymentScheduleInstallment.java
```diff
@@ -140,6 +140,9 @@ public class LoanRepaymentScheduleInstallment extends AbstractAuditableWithUTCDa
     @OneToMany(cascade = CascadeType.ALL, orphanRemoval = true, fetch = FetchType.LAZY, mappedBy = "installment")
     private Set<LoanInstallmentCharge> installmentCharges = new HashSet<>();
 
+    @OneToMany(cascade = CascadeType.ALL, orphanRemoval = true, fetch = FetchType.EAGER, mappedBy = "installment")
+    private Set<LoanTransactionToRepaymentScheduleMapping> loanTransactionToRepaymentScheduleMappings = new HashSet<>();
```
adamsaghy, line 144: I am a little hesitant whether we need to fetch the mapping eagerly. Can you tell me why did you decide to do so? Eager loading only required if we would read an association when the entity is already detached. I dont think that would happen here (or should)...i guess we need to make sure this entity is used in transactional block

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyReadPlatformServiceImpl.java
```diff
@@ -109,89 +103,23 @@ public Collection<LoanDelinquencyTagHistoryData> retrieveDelinquencyRangeHistory
...
     @Override
     public CollectionData calculateLoanCollectionData(final Long loanId) {
-        final LocalDate businessDate = DateUtils.getBusinessLocalDate();
         final Optional<Loan> optLoan = this.loanRepository.findById(loanId);
 
         CollectionData collectionData = CollectionData.template();
         if (optLoan.isPresent()) {
+            final LocalDate businessDate = DateUtils.getBusinessLocalDate();
             final Loan loan = optLoan.get();
+            loan.initializeLazyBasicCollections();
```
adamsaghy: I would rather put a `@Transactional` annotation on the method then manually try to fetch the lazy associations.

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyUtils.java
```diff
@@ -0,0 +1,132 @@
...
+                                break;
+                            }
+                        }
+                    }
+                }
+            } else if (!firstNotYetDueInstallment) {
+                log.debug("Loan Id: {} with installment {} due date {}", loan.getId(), installment.getInstallmentNumber(),
+                        installment.getDueDate());
+                firstNotYetDueInstallment = true;
+                amountAvailable = installment.getTotalPaid(loanCurrency).getAmount().subtract(installment.getTotalPaidInAdvance());
```
adamsaghy: Are you sure we need to deduct the total paid in advance?

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyReadPlatformServiceImpl.java
galovics: 100% agreed. TBH the initializeLazyXY method calls should be eliminated completely. They are mostly used to overcome the fact that the transaction boundaries are not properly defined.

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyUtils.java
```diff
@@ -0,0 +1,132 @@
...
+import org.apache.fineract.infrastructure.core.service.DateUtils;
+import org.apache.fineract.organisation.monetary.domain.MonetaryCurrency;
+import org.apache.fineract.portfolio.loanaccount.data.CollectionData;
+import org.apache.fineract.portfolio.loanaccount.domain.Loan;
+import org.apache.fineract.portfolio.loanaccount.domain.LoanRepaymentScheduleInstallment;
+import org.apache.fineract.portfolio.loanaccount.domain.LoanTransaction;
+import org.apache.fineract.portfolio.loanaccount.domain.LoanTransactionToRepaymentScheduleMapping;
+
+@Slf4j
+public final class DelinquencyUtils {
```
galovics: I don't like this. When I see the following names:
- Manager
- Utils
- Util
- etc

I'm always careful because usually it's something fishy. Not always but most of the times. I think this is rather a poor implementation for separation of concerns. On one hand, let's forbid ourselves from using static methods unless there's a good reason for it. 

In this case this class is handling business logic. That's not a Util and that shouldn't be handled in static methods.

Try to generalize the class/method around single responsibilities and try to split it that way into components.

galovics: Bonus, why don't I see unit tests for this class?

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
galovics: Let's get rid of this method rather. It's spreading a bad pattern across the system. Let's try to force ourselves to properly define our transaction boundaries and the methods like this will never be used.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanRepaymentScheduleInstallment.java
galovics: > we would read an association when the entity is already detached.

Not necessarily. Eager fetching comes handy when the parent entity - from a business point of view - can't live without the child entity. 
I'm not sure if this is the case here though so I'm also waiting for some explanation here.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanRepaymentScheduleInstallment.java
```diff
@@ -718,6 +721,9 @@ public void addToCredits(final BigDecimal amount) {
     }
 
     public BigDecimal getTotalPaidInAdvance() {
+        if (this.totalPaidInAdvance == null) {
```
galovics: This is kinda fishy to me again. 
If the column can be null, that means there was no payment in advance, right? We introduce that logic and the users of this class will never be able to make that decision, right? Since we hide it here with this condition.

Also, if this logic has been introduced here, did you check the whole codebase for native SQLs/JPQLs/etc where the same DB table has been used and the same column has been retrieved? We'll need this same behavior there, right?

galovics (changes requested):

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
josehernandezfintecheandomx: Done, removed

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanRepaymentScheduleInstallment.java
josehernandezfintecheandomx: Done, moved to LAZY

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyReadPlatformServiceImpl.java
josehernandezfintecheandomx: Done, removed

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyUtils.java
josehernandezfintecheandomx: Removed

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanRepaymentScheduleInstallment.java
josehernandezfintecheandomx: Removed

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyUtils.java
josehernandezfintecheandomx: It is now in a service, I can't put in the current Read service because It generates circular dependency

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyReadPlatformServiceImpl.java
```diff
@@ -109,89 +104,22 @@ public Collection<LoanDelinquencyTagHistoryData> retrieveDelinquencyRangeHistory
...
     public CollectionData calculateLoanCollectionData(final Long loanId) {
-        final LocalDate businessDate = DateUtils.getBusinessLocalDate();
         final Optional<Loan> optLoan = this.loanRepository.findById(loanId);
 
         CollectionData collectionData = CollectionData.template();
         if (optLoan.isPresent()) {
+            final LocalDate businessDate = DateUtils.getBusinessLocalDate();
             final Loan loan = optLoan.get();
 
+            collectionData = this.loanCollectionDomainService.getOverdueCollectionData(loan, businessDate);
```
galovics: `this.` not needed.

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyWritePlatformServiceImpl.java
```diff
@@ -152,61 +149,9 @@ public LoanScheduleDelinquencyData calculateDelinquencyData(LoanScheduleDelinque
...
-    private Long calculateOverdueDays(final LocalDate businessDate, LocalDate overdueSinceDate) {
-        Long overdueDays = 0L;
-        if (overdueSinceDate != null) {
-            overdueDays = DateUtils.getDifferenceInDays(overdueSinceDate, businessDate);
-            if (overdueDays < 0) {
-                overdueDays = 0L;
-            }
-        }
-        return overdueDays;
+        final CollectionData collectionData = this.loanCollectionDomainService.getOverdueCollectionData(loan, businessDate);
```
galovics: `this.` not needed

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyWritePlatformServiceImpl.java
```diff
@@ -152,61 +149,9 @@ public LoanScheduleDelinquencyData calculateDelinquencyData(LoanScheduleDelinque
...
+        log.debug("Delinquency {}", collectionData.toString());
```
galovics: No need for the explicit toString call.

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyWritePlatformServiceImpl.java
```diff
@@ -217,12 +162,8 @@ public CommandProcessingResult applyDelinquencyTagToLoan(Long loanId, JsonComman
         final DelinquencyBucket delinquencyBucket = loan.getLoanProduct().getDelinquencyBucket();
         if (delinquencyBucket != null) {
             final LocalDate businessDate = DateUtils.getBusinessLocalDate();
-
-            final Integer graceOnArrearAgeing = loan.getLoanProduct().getLoanProductRelatedDetail().getGraceOnArrearsAgeing();
-            final LocalDate overdueSinceDate = getOverdueSinceDate(loan, businessDate, graceOnArrearAgeing);
-
-            final Long overdueDays = calculateOverdueDays(businessDate, overdueSinceDate);
-            changes = lookUpDelinquencyRange(loan, delinquencyBucket, overdueDays);
+            final CollectionData collectionData = this.loanCollectionDomainService.getOverdueCollectionData(loan, businessDate);
```
galovics: No need for `this.`

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyWritePlatformServiceImpl.java
```diff
@@ -233,15 +174,9 @@ public void applyDelinquencyTagToLoan(LoanScheduleDelinquencyData loanDelinquenc
...
-            final Integer graceOnArrearAgeing = loan.getLoanProduct().getLoanProductRelatedDetail().getGraceOnArrearsAgeing();
-
-            LocalDate overdueSinceDate = loanDelinquencyData.getOverdueSinceDate();
-            if (overdueSinceDate == null) {
-                getOverdueSinceDate(loan, businessDate, graceOnArrearAgeing);
-            }
-
-            final Long overdueDays = calculateOverdueDays(businessDate, overdueSinceDate);
-            lookUpDelinquencyRange(loan, delinquencyBucket, overdueDays);
+            final CollectionData collectionData = this.loanCollectionDomainService.getOverdueCollectionData(loan, businessDate);
```
galovics: As above.

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyWritePlatformServiceImpl.java
```diff
@@ -233,15 +174,9 @@ public void applyDelinquencyTagToLoan(LoanScheduleDelinquencyData loanDelinquenc
...
+            log.debug("Delinquency {}", collectionData.toString());
```
galovics: No need for toString

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/LoanCollectionDomainService.java
```diff
@@ -0,0 +1,36 @@
...
+public interface LoanCollectionDomainService {
+
+    /**
+     * This method is to calculate the Overdue date and other properties, If the loan is overdue or If there is some
+     * Charge back transaction
+     *
+     * @param loan
+     * @param businessDate
+     */
+    CollectionData getOverdueCollectionData(Loan loan, LocalDate businessDate);
```
galovics: If we're always passing the current business date, why is it a parameter? Can't we just resolve the current business date from within the implementation?

adamsaghy (dismissed):

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/serialization/LoanEventApiJsonValidator.java
```diff
@@ -237,7 +237,7 @@ public void validateChargebackTransaction(final String json) {
         final BigDecimal transactionAmount = this.fromApiJsonHelper
                 .extractBigDecimalWithLocaleNamed(LoanApiConstants.TRANSACTION_AMOUNT_PARAMNAME, element);
         baseDataValidator.reset().parameter(LoanApiConstants.TRANSACTION_AMOUNT_PARAMNAME).value(transactionAmount).notNull()
-                .zeroOrPositiveAmount();
+                .positiveAmount();
```
galovics: This is a separate bugfix. Please extract it to a different PR so it can be reviewed separately along with its tests.

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/LoanCollectionDomainServiceImpl.java
```diff
@@ -0,0 +1,135 @@
...
+import org.apache.fineract.portfolio.loanaccount.data.CollectionData;
+import org.apache.fineract.portfolio.loanaccount.domain.Loan;
+import org.apache.fineract.portfolio.loanaccount.domain.LoanRepaymentScheduleInstallment;
+import org.apache.fineract.portfolio.loanaccount.domain.LoanTransaction;
+import org.apache.fineract.portfolio.loanaccount.domain.LoanTransactionToRepaymentScheduleMapping;
+import org.springframework.stereotype.Service;
+
+@Slf4j
+@Service
+public class LoanCollectionDomainServiceImpl implements LoanCollectionDomainService {
```
galovics: Don't like the naming still, especially the terminology CollectionData because it doesn't reflect anything about the data it represents.
But I can live with it for now.

The other thing though which I can't live without is proper unit test coverage for this class.

galovics (changes requested):

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyReadPlatformServiceImpl.java
josehernandezfintecheandomx: Removed

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyWritePlatformServiceImpl.java
josehernandezfintecheandomx: Removed

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyWritePlatformServiceImpl.java
josehernandezfintecheandomx: Removed

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyWritePlatformServiceImpl.java
josehernandezfintecheandomx: Removed

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyWritePlatformServiceImpl.java
josehernandezfintecheandomx: Removed

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyWritePlatformServiceImpl.java
josehernandezfintecheandomx: Removed

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/serialization/LoanEventApiJsonValidator.java
josehernandezfintecheandomx: Removed

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/LoanCollectionDomainService.java
josehernandezfintecheandomx: Updated

fineract-provider/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyUtils.java
josehernandezfintecheandomx: Done in class `LoanDelinquencyDomainServiceTest`

galovics (approved):

adamsaghy (approved):
LGTM

PR #2749, opened 2022-11-18
FINERACT-1807: Extension point - data service

vidakovic:
Community request to be able to override "ReadWriteNonCoreDataService".

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/starter/DataQueriesAutoConfiguration.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/starter/DataQueriesAutoConfiguration.java
@@ -1,0 +1,56 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/starter/DataQueriesAutoConfiguration.java
```diff
@@ -0,0 +1,56 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+
+@Configuration
+public class DataQueriesAutoConfiguration {
+
+    @Bean
+    @ConditionalOnMissingBean
```
galovics, line 44: You can put this annotation directly onto the component and move back the @Component annotation as well. That way you don't need to deal with the constructor parameters.

galovics (changes requested):

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/starter/DataQueriesAutoConfiguration.java
vidakovic: Yes... but this is already in preparation of factoring those packages out as separate modules... then you have less Spring dependent annotations on the service. And it would follow the "template" we established with the note service. If you don't mind I'd leave it like that.

galovics (approved):

jdailey (commented):
LGTM but I'm not the expert

PR #2752, opened 2022-11-22
FINERACT-1794: Address improvement in file upload APIs

vidakovic:
More rigours file validation to be introduces when file is uploaded to fineract.


adamsaghy (approved):
LGTM

fineract-provider/src/main/resources/application.properties
```diff
@@ -70,6 +70,17 @@ fineract.loan.transactionprocessor.principal-interest-penalties-fees.enabled=${F
...
 fineract.loan.transactionprocessor.error-not-found-fail=${FINERACT_LOAN_TRANSACTIONPROCESSOR_ERROR_NOT_FOUND_FAIL:true}
 
+fineract.content.regex-whitelist-enabled=${FINERACT_CONTENT_REGEX_WHITELIST_ENABLED:true}
+fineract.content.regex-whitelist=${FINERACT_CONTENT_REGEX_WHITELIST:.*\\.pdf$,.*\\.doc,.*\\.docx,.*\\.xls,.*\\.xlsx,.*\\.jpg,.*\\.jpeg,.*\\.png}
+fineract.content.mime-whitelist-enabled=${FINERACT_CONTENT_MIME_WHITELIST_ENABLED:true}
+fineract.content.mime-whitelist=${FINERACT_CONTENT_MIME_WHITELIST:application/pdf,application/msword,application/vnd.openxmlformats-officedocument.wordprocessingml.document,application/vnd.ms-excel,application/vnd.openxmlformats-officedocument.spreadsheetml.sheet,image/jpeg,image/png}
+fineract.content.filesystem.enabled=${FINERACT_CONTENT_FILESYSTEM_ENABLED:true}
+fineract.content.filesystem.rootFolder=${FINERACT_CONTENT_FILESYSTEM_ROOT_FOLDER:${user.home}/.fineract}
+fineract.content.s3.enabled=${FINERACT_CONTENT_S3_ENABLED:false}
+fineract.content.s3.bucketName=${FINERACT_CONTENT_S3_BUCKET_NAME:}
```
galovics, line 80: Prob you haven't tested it with AWS S3 @vidakovic because the property name here is wrong.
The FineractProperties class defines them as `accessKey` and `secretKey` yet here it's called `accessName` and `secretName` (same for the env var naming).

fineract-provider/src/test/resources/application-test.properties
```diff
@@ -60,6 +60,16 @@ fineract.loan.transactionprocessor.principal-interest-penalties-fees.enabled=tru
...
 fineract.loan.transactionprocessor.error-not-found-fail=true
 
+fineract.content.regex-whitelist-enabled=true
+fineract.content.regex-whitelist=.*\\.pdf$,.*\\.doc,.*\\.docx,.*\\.xls,.*\\.xlsx,.*\\.jpg,.*\\.jpeg,.*\\.png
+fineract.content.mime-whitelist-enabled=true
+fineract.content.mime-whitelist=application/pdf,application/msword,application/vnd.openxmlformats-officedocument.wordprocessingml.document,application/vnd.ms-excel,application/vnd.openxmlformats-officedocument.spreadsheetml.sheet,image/jpeg,image/png
+fineract.content.filesystem.enabled=true
+fineract.content.filesystem.rootFolder=${user.home}/.fineract
+fineract.content.s3.enabled=false
+fineract.content.s3.bucketName=
```
galovics, line 70: Same as in the other properties

fineract-provider/src/main/java/org/apache/fineract/infrastructure/documentmanagement/contentrepository/FileSystemContentPathSanitizer.java
```diff
@@ -0,0 +1,138 @@
...
+        regexWhitelist = fineractProperties.getContent().getRegexWhitelist().stream().map(Pattern::compile).toList();
+    }
+
+    @Override
+    public String sanitize(String path) {
+        return sanitize(path, null);
+    }
+
+    @Override
+    public String sanitize(String path, BufferedInputStream is) {
```
galovics, line 62: Not sure about the entire picture here but I'd be careful with passing a BufferedInputStream (or any other InputStream for that matter) as a public method parameter since there could be cases where the input stream cannot be read twice.
I assume the logic here is to decide what mime type the data has and after sanitizing, the actual data is saved into file system or S3. But in fact we're reading the input stream once to decide the mime type and once more to save the data. That might not work for every input stream.

Is it possible to read the input stream once at the beginning, get the data as a byte array (IOUtils.copy) and inspect that for mime type checks and save it as a file.

Thoughts?

galovics (changes requested):

fineract-provider/src/main/resources/application.properties
vidakovic: Shoot :gun: ... on it.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/documentmanagement/contentrepository/FileSystemContentPathSanitizer.java
vidakovic: Good catch... not possible to sneak anything around you :wink: ... as you can see I didn't include the sanitizer in S3, because it's not really needed there (now). To get this improvement over the finish line I've arranged things as you see them here, including the buffered input stream... which - kind of - forces everyone who uses this function to pay attention. In too many places the use of InputStream is just not done the right way (streams are copied into byte arrays which are immediately wrapped in ByteArrayInputStreams... and multiple times with the same data). None of the places where I had to make changes for this improvement paid attention - at all - if streams are still open or closed. If the authors were in doubt then they just took the byte array again and created a new stream. Obviously that defeats they whole idea of streams on not holding all the data in memory.
I intend to generalize this part of the system even further to make contributions (like this one https://issues.apache.org/jira/browse/FINERACT-1814) easier and cleanup the whole implementation:

- why do we have a distinction between images and other files... 
- actually images can only be associated - for whatever reason with clients and staff... 
- why do we have to replicate something like a file path and not just use a flat structure with UUIDs for the physical file names... we can attach any kind of metadata in the database structure if really needed

Just to say: let's leave it like this for a moment. I just didn't want to introduce too many changes at once (I already moved the configuration from the database to application.properties... which is the better way to do it nowadays with reloadable configurations etc.).

I'll mark this one as resolved for now, but expect more next week.

galovics (approved):

PR #2759, opened 2022-11-24
FINERACT-1678: Introducing Last COB Business Date for loans

taskain7:
## Description

Loans store the last date when the Loan COB job was done on it.
Custom parameters can be added for job execution via API.

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/cob/loan/AbstractLoanItemProcessor.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/cob/loan/AbstractLoanItemProcessor.java
@@ -18,6 +18,9 @@
  */
 package org.apache.fineract.cob.loan;
 
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.util.Objects;
 import java.util.TreeMap;
 import lombok.AccessLevel;
 import lombok.RequiredArgsConstructor;
@@ -40,17 +43,27 @@ public abstract class AbstractLoanItemProcessor implements ItemProcessor<Loan, L
 
     @Setter(AccessLevel.PROTECTED)
     private ExecutionContext executionContext;
+    private LocalDate businessDate;
 
     @Override
     public Loan process(@NotNull Loan item) throws Exception {
         TreeMap<Long, String> businessStepMap = (TreeMap<Long, String>) executionContext.get(LoanCOBConstant.BUSINESS_STEP_MAP);
 
-        return cobBusinessStepService.run(businessStepMap, item);
+        Loan alreadyProcessedLoan = cobBusinessStepService.run(businessStepMap, item);
+        alreadyProcessedLoan.setLastClosedBusinessDate(businessDate);
+        return alreadyProcessedLoan;
     }
 
     @AfterStep
     public ExitStatus afterStep(@NotNull StepExecution stepExecution) {
         return ExitStatus.COMPLETED;
+    }
+
+    protected void setBusinessDate(StepExecution stepExecution) {
+        this.businessDate = LocalDate.parse(
+                Objects.requireNonNull(
+                        (String) stepExecution.getJobExecution().getExecutionContext().get(LoanCOBConstant.BUSINESS_DATE_PARAMETER_NAME)),
+                DateTimeFormatter.ISO_DATE);
     }
 
 }
--- a/fineract-provider/src/main/java/org/apache/fineract/cob/loan/ResolveLoanCOBCustomJobParametersTasklet.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/cob/loan/ResolveLoanCOBCustomJobParametersTasklet.java
@@ -16,37 +16,26 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.fineract.infrastructure.jobs.service.executealldirtyjobs;
+package org.apache.fineract.cob.loan;
 
-import java.util.List;
 import lombok.RequiredArgsConstructor;
-import org.apache.fineract.infrastructure.core.config.FineractProperties;
-import org.apache.fineract.infrastructure.jobs.domain.ScheduledJobDetail;
-import org.apache.fineract.infrastructure.jobs.domain.ScheduledJobDetailRepository;
-import org.apache.fineract.infrastructure.jobs.service.JobRegisterService;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.fineract.cob.common.CustomJobParameterResolver;
 import org.springframework.batch.core.StepContribution;
 import org.springframework.batch.core.scope.context.ChunkContext;
 import org.springframework.batch.core.step.tasklet.Tasklet;
 import org.springframework.batch.repeat.RepeatStatus;
-import org.springframework.stereotype.Component;
 
+@Slf4j
 @RequiredArgsConstructor
-@Component
-public class ExecuteAllDirtyJobsTasklet implements Tasklet {
+public class ResolveLoanCOBCustomJobParametersTasklet implements Tasklet {
 
-    private final JobRegisterService jobRegisterService;
-    private final ScheduledJobDetailRepository scheduledJobDetailsRepository;
-    private final FineractProperties fineractProperties;
+    private final CustomJobParameterResolver customJobParameterResolver;
 
     @Override
     public RepeatStatus execute(StepContribution contribution, ChunkContext chunkContext) throws Exception {
-        List<ScheduledJobDetail> jobDetails = scheduledJobDetailsRepository.findAllMismatchedJobs(true);
-
-        for (ScheduledJobDetail scheduledJobDetail : jobDetails) {
-            if (scheduledJobDetail.getNodeId().toString().equals(fineractProperties.getNodeId())) {
-                jobRegisterService.executeJobWithParameters(scheduledJobDetail.getId(), null);
-            }
-        }
+        customJobParameterResolver.resolve(contribution, chunkContext, LoanCOBConstant.BUSINESS_DATE_PARAMETER_NAME,
+                LoanCOBConstant.BUSINESS_DATE_PARAMETER_NAME);
         return RepeatStatus.FINISHED;
     }
 }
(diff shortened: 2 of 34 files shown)


fineract-provider/src/main/java/org/apache/fineract/cob/loan/AbstractLoanItemWriter.java
```diff
@@ -18,25 +18,41 @@
...
+        ExecutionContext jobExecutionContext = stepExecution.getJobExecution().getExecutionContext();
+        String businessDate = (String) jobExecutionContext.get(LoanCOBConstant.BUSINESS_DATE_PARAMETER_NAME);
+        this.businessDate = businessDate != null ? LocalDate.parse(businessDate)
+                : ThreadLocalContextUtil.getBusinessDateByType(BusinessDateType.COB_DATE);
+    }
 
     @Override
     public void write(@NotNull List<? extends Loan> items) throws Exception {
         super.write(items);
+        items.forEach(loan -> loan.setLastCOBBusinessDate(businessDate));
```
galovics: Not sure about this. Wouldn't it be better within the processor?

fineract-provider/src/main/java/org/apache/fineract/cob/loan/ResolveLoanCOBCustomJobParametersTasklet.java
```diff
@@ -0,0 +1,76 @@
...
+import org.apache.fineract.infrastructure.springbatch.SpringBatchJobConstants;
+import org.springframework.batch.core.StepContribution;
+import org.springframework.batch.core.scope.context.ChunkContext;
+import org.springframework.batch.core.step.tasklet.Tasklet;
+import org.springframework.batch.repeat.RepeatStatus;
+import org.springframework.beans.factory.InitializingBean;
+
+@Slf4j
+@RequiredArgsConstructor
+public class ResolveLoanCOBCustomJobParametersTasklet implements Tasklet, InitializingBean {
```
galovics: I feel like we should do some abstraction here. 

Obviously the majority of this code is generic:
- Creating the GSON serializer
- Reading the custom job parameter from the DB
- deserializing it

The rest is specific to the Loan COB, i.e. that the business date parameter is used. I think other jobs could benefit from making an abstract class from this an have specializations for jobs.

Thoughts?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/jobs/service/JobStarter.java
```diff
@@ -45,14 +50,19 @@ public class JobStarter {
...
     private final JobLauncher jobLauncher;
     private final JobParameterRepository jobParameterRepository;
+    private final List<JobParameterProvider> jobParameterProviders;
+    private final JobNameService jobNameService;
 
-    public void run(Job job, ScheduledJobDetail scheduledJobDetail, FineractContext fineractContext)
-            throws JobInstanceAlreadyCompleteException, JobExecutionAlreadyRunningException, JobParametersInvalidException,
-            JobRestartException {
+    public void run(Job job, ScheduledJobDetail scheduledJobDetail, FineractContext fineractContext,
+            List<JobParameterDTO> jobParameterDTOList) throws JobInstanceAlreadyCompleteException, JobExecutionAlreadyRunningException,
```
galovics: I think we can make this a Set instead of a List and make the JobParameterDTO to implement an equals and hashcode. What do you think?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/jobs/service/jobparameterprovider/JobParameterProvider.java
```diff
@@ -0,0 +1,35 @@
...
+import org.apache.fineract.infrastructure.jobs.data.JobParameterDTO;
+import org.springframework.batch.core.JobParameter;
+
+public interface JobParameterProvider {
+
+    Map<String, JobParameter> provide(List<JobParameterDTO> jobParameterDTOList);
+
+    String getJobName();
+
+    default boolean canProvideParametersForJob(String jobName) {
```
galovics: Well. :))))
Would be nicer to put this logic into an abstract class and use the template method pattern but works this way too. It's just abusing the default methods a little bit :)))

fineract-provider/src/main/java/org/apache/fineract/infrastructure/jobs/service/jobparameterprovider/LoanCOBJobParameterProviderImpl.java
```diff
@@ -0,0 +1,82 @@
...
+import org.apache.fineract.infrastructure.jobs.domain.CustomJobParameterRepository;
+import org.apache.fineract.infrastructure.jobs.service.JobName;
+import org.apache.fineract.infrastructure.springbatch.SpringBatchJobConstants;
+import org.springframework.batch.core.JobParameter;
+import org.springframework.beans.factory.InitializingBean;
+import org.springframework.stereotype.Component;
+
+@Component
+@RequiredArgsConstructor
+public class LoanCOBJobParameterProviderImpl implements JobParameterProvider, InitializingBean {
```
galovics: Let's drop the Impl postfix. Never liked it and doesn't add value at all.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/jobs/service/jobparameterprovider/LoanCOBJobParameterProviderImpl.java
```diff
@@ -0,0 +1,82 @@
...
+    public void afterPropertiesSet() throws Exception {
+        this.gson = gsonFactory.createSimpleGson();
+    }
+
+    @Override
+    public Map<String, JobParameter> provide(List<JobParameterDTO> jobParameterDTOList) {
+        Map<String, JobParameter> jobParameterMap = new HashMap<>();
+        CustomJobParameter customJobParameter = new CustomJobParameter();
+        customJobParameter.setParameterJson(gson.toJson(getJobParameterDTOListWithCorrectBusinessDate(jobParameterDTOList)));
+        CustomJobParameter savedCustomJobParameter = customJobParameterRepository.saveAndFlush(customJobParameter);
```
galovics: No need for saveAndFlush. A regular save should be enough I think and put the @Transactional annotation onto the method so the boundaries are properly defined.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
```diff
@@ -430,6 +430,9 @@ public class Loan extends AbstractAuditableWithUTCDateTimeCustom {
     @Column(name = "fixed_principal_percentage_per_installment", scale = 2, precision = 5, nullable = true)
     private BigDecimal fixedPrincipalPercentagePerInstallment;
 
+    @Column(name = "last_cob_business_date")
```
galovics: I think this could be called more generically like `last_closed_business_date`

galovics (changes requested):
Let's extend the loan APIs with the last_closed_business_date attribute as well plus the swagger. And then we can write some integration test(s) to cover at least basic scenarios, i.e. loan is created/etc and LoanCOB executed, that sets the initial last_closed_business_date, plus if the loan has already been closed once, and the business date changes, it gets updated after a COB execution. Also, Loans which are behind more than 1 day are not picked up anymore.
Thoughts?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/jobs/service/jobparameterprovider/LoanCOBJobParameterProviderImpl.java
taskain7: With regular save I don't get the saved ID, that's why I used the saveAndFlush

fineract-provider/src/main/java/org/apache/fineract/cob/loan/AbstractLoanItemProcessor.java
```diff
@@ -40,17 +42,25 @@ public abstract class AbstractLoanItemProcessor implements ItemProcessor<Loan, L
...
+        alreadyProcessedLoan.setLastClosedBusinessDate(businessDate);
+        return alreadyProcessedLoan;
     }
 
     @AfterStep
     public ExitStatus afterStep(@NotNull StepExecution stepExecution) {
         return ExitStatus.COMPLETED;
     }
 
+    protected void setBusinessDate(StepExecution stepExecution) {
```
galovics, line 62: You sure we need this? We already have the executioncontext within this class and using it in the process method. Why don't we do it similarly?

fineract-provider/src/main/java/org/apache/fineract/cob/common/AbstractCustomJobParameterResolver.java
```diff
@@ -0,0 +1,56 @@
...
+import org.apache.fineract.infrastructure.core.serialization.GoogleGsonSerializerHelper;
+import org.apache.fineract.infrastructure.jobs.data.JobParameterDTO;
+import org.apache.fineract.infrastructure.jobs.domain.CustomJobParameter;
+import org.apache.fineract.infrastructure.jobs.domain.CustomJobParameterRepository;
+import org.apache.fineract.infrastructure.springbatch.SpringBatchJobConstants;
+import org.springframework.batch.core.scope.context.ChunkContext;
+import org.springframework.beans.factory.InitializingBean;
+
+@RequiredArgsConstructor
+public class AbstractCustomJobParameterResolver implements InitializingBean {
```
galovics: I more thought of this as either an Abstract Tasklet that ppl can reuse or just a regular resolver that Tasklets can use.

Now this is something else between the 2. :-) 
Maybe just go with a simple resolver, that's the most generic approach.

galovics (changes requested):

fineract-provider/src/main/java/org/apache/fineract/cob/loan/AbstractLoanItemProcessor.java
taskain7: we need this when tha business date is in different ExecutionContext than the other parameters we use (i.e. LoanCOB)

galovics (approved):

PR #2797, opened 2022-12-09
FINERACT-1785: Allow to add charge (penalty) with same disbursement date

josehernandezfintecheandomx:
## Description

When trying to apply specified due date fees or penalties on the date of the loan’s disbursement date, it won’t be effective and It's not added as outstanding

[FINERACT-1785](https://issues.apache.org/jira/browse/FINERACT-1785)

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanRepaymentScheduleProcessingWrapper.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanRepaymentScheduleProcessingWrapper.java
@@ -119,8 +119,7 @@ public class LoanRepaymentScheduleProcessingWrapper {
                 } else if (loanCharge.isDueForCollectionFromAndUpToAndIncluding(periodStart, periodEnd)) {
                     cumulative = cumulative.plus(loanCharge.amount());
                     // Special case for Loan Charges (Due Date) added the same disbursement date
-                } else if (period.isFirstPeriod()
-                        && loanCharge.isDueForCollectionFromIncludingAndUpToAndIncluding(periodStart, periodEnd)) {
+                } else if (period.isFirstPeriod() && periodStart.equals(loanCharge.getDueDate())) {
                     cumulative = cumulative.plus(loanCharge.amount());
                 }
             }
@@ -212,6 +211,9 @@ public class LoanRepaymentScheduleProcessingWrapper {
                     BigDecimal loanChargeAmt = amount.multiply(loanCharge.getPercentage()).divide(BigDecimal.valueOf(100));
                     cumulative = cumulative.plus(loanChargeAmt);
                 } else if (loanCharge.isDueForCollectionFromAndUpToAndIncluding(periodStart, periodEnd)) {
+                    cumulative = cumulative.plus(loanCharge.amount());
+                    // Special case for Loan Charges (Due Date) added the same disbursement date
+                } else if (period.isFirstPeriod() && periodStart.equals(loanCharge.getDueDate())) {
                     cumulative = cumulative.plus(loanCharge.amount());
                 }
             }
--- a/integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanChargeSpecificDueDateTest.java
+++ b/integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanChargeSpecificDueDateTest.java
@@ -62,7 +62,7 @@ public class LoanChargeSpecificDueDateTest {
     }
 
     @Test
-    public void testApplyLoanSpecificDueDateChargeWithDisbursementDate() {
+    public void testApplyLoanSpecificDueDateFeeWithDisbursementDate() {
 
         final LocalDate todaysDate = Utils.getLocalDateOfTenant();
 
@@ -82,7 +82,7 @@ public class LoanChargeSpecificDueDateTest {
 
         // Get loan details
         GetLoansLoanIdResponse getLoansLoanIdResponse = loanTransactionHelper.getLoan(requestSpec, responseSpec, loanId);
-        validateLoanAccount(getLoansLoanIdResponse, Double.valueOf(principalAmount), Double.valueOf("0.00"));
+        validateLoanAccount(getLoansLoanIdResponse, Double.valueOf(principalAmount), Double.valueOf("0.00"), false);
 
         // Apply Loan Charge with specific due date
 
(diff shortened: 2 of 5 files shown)


integration-tests/src/test/java/org/apache/fineract/integrationtests/common/loans/LoanTransactionHelper.java
```diff
@@ -1563,6 +1563,14 @@ public void validateLoanFeesOustandingBalance(GetLoansLoanIdResponse getLoansLoa
         }
     }
 
+    public void validateLoanPenaltiesOustandingBalance(GetLoansLoanIdResponse getLoansLoanIdResponse, Double amountExpected) {
+        GetLoansLoanIdSummary getLoansLoanIdSummary = getLoansLoanIdResponse.getSummary();
+        if (getLoansLoanIdSummary != null) {
```
galovics: Having a null-check here and in general within tests is a really really bad pattern cause it could prevent assertions from being evaluation therefore making the tests pass.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanRepaymentScheduleProcessingWrapper.java
```diff
@@ -213,6 +213,10 @@ private Money cumulativePenaltyChargesDueWithin(final LocalDate periodStart, fin
                     cumulative = cumulative.plus(loanChargeAmt);
                 } else if (loanCharge.isDueForCollectionFromAndUpToAndIncluding(periodStart, periodEnd)) {
                     cumulative = cumulative.plus(loanCharge.amount());
+                    // Special case for Loan Charges (Due Date) added the same disbursement date
+                } else if (period.isFirstPeriod()
```
galovics: Why don't we check directly whether the charge is due on disbursement date? This condition is kinda confusing to me and probably to you too, that's why you put the comment there.

Let's make it crystal clear, loan charge on disbursement date.

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanChargeSpecificDueDateTest.java
```diff
@@ -110,7 +110,60 @@ public void testApplyLoanSpecificDueDateChargeWithDisbursementDate() {
...
+        GetLoansLoanIdResponse getLoansLoanIdResponse = loanTransactionHelper.getLoan(requestSpec, responseSpec, loanId);
+        validateLoanAccount(getLoansLoanIdResponse, Double.valueOf(principalAmount), Double.valueOf("0.00"), true);
+
+        // Apply Loan Charge with specific due date
+
+        final String feeAmount = "10.00";
+        String payloadJSON = ChargesHelper.getLoanSpecifiedDueDateJSON(ChargesHelper.CHARGE_CALCULATION_TYPE_FLAT, feeAmount, true);
+        final PostChargesResponse postChargesResponse = ChargesHelper.createLoanCharge(requestSpec, responseSpec, payloadJSON);
+        assertNotNull(postChargesResponse);
+        final Long loanChargeId = postChargesResponse.getResourceId();
```
galovics, line 146: Let's assert that the resourceId is not null.

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanChargeSpecificDueDateTest.java
```diff
@@ -110,7 +110,60 @@ public void testApplyLoanSpecificDueDateChargeWithDisbursementDate() {
...
+        // Make a full repayment to close the Loan
+        Float amount = Float.valueOf("1010.00");
+        PostLoansLoanIdTransactionsResponse loanIdTransactionsResponse = loanTransactionHelper.makeLoanRepayment(operationDate, amount,
+                loanId);
+        assertNotNull(loanIdTransactionsResponse);
+        log.info("Loan Transaction Id: {} {}", loanId, loanIdTransactionsResponse.getResourceId());
+
+        getLoansLoanIdResponse = loanTransactionHelper.getLoan(requestSpec, responseSpec, loanId);
+        assertNotNull(getLoansLoanIdResponse);
+        validateLoanAccount(getLoansLoanIdResponse, Double.valueOf("0.00"), Double.valueOf("0.00"), true);
```
galovics, line 167: Can we also validate the loan status as well? (CLOSED_OBLIGATIONS_MET)

galovics (changes requested):

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanChargeSpecificDueDateTest.java
josehernandezfintecheandomx: Done

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanRepaymentScheduleProcessingWrapper.java
josehernandezfintecheandomx: This is a special case or requirement for the first installment, the other installments must to be in the other if

integration-tests/src/test/java/org/apache/fineract/integrationtests/common/loans/LoanTransactionHelper.java
josehernandezfintecheandomx: This is not the main assertion pf this method, that is below, but anyway, It is updated

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanChargeSpecificDueDateTest.java
josehernandezfintecheandomx: Done

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanRepaymentScheduleProcessingWrapper.java
adamsaghy: i dont think this newly added branch will ever be executed.

the condition in line 214 will always be true before it could reach the new branch in line 217.
Also both of them do exactly the same thing. 
@josehernandezfintecheandomx could you advise on this?

adamsaghy: I would rather do this:
- If this is the first period
AND
- first period startdate == loanCharge due date

I dont think your approach is wrong, but its harder to understand. `loanCharge.isDueForCollectionFromIncludingAndUpToAndIncluding` is hard to understand what it is doing.

Also this special case is only to handle when the charge due date is equals with first instalment start date. We dont necessarily need to check from - to, inclusive / including stuffs... only whether they are equals or not...
@josehernandezfintecheandomx what do you think?

josehernandezfintecheandomx: Well, It can be implemented in different ways. This one was used for the Fees in the previous change and now is the same code in the Penalties method. Plus the logic for validating the dates was reusing the same functions.

Based on that, I've updated the code as you suggest and I've removed the code added in the previous change that is not more needed. @adamsaghy 

Screenshot 2022-12-12 at 22 20 00

galovics (approved):

PR #2897, opened 2023-01-18
[FINERACT-1859] Avro schema for LoanAccountsStayedLockedBusinessEvent

taskain7:
## Description

Avro schema for `LoanAccountsStayedLockedBusinessEvent`

### DIFF
--- a/fineract-avro-schemas/src/main/avro/loan/v1/LoanAccountStayedLockedDataV1.avsc
+++ b/fineract-avro-schemas/src/main/avro/loan/v1/LoanAccountStayedLockedDataV1.avsc
@@ -1,0 +1,27 @@
+{
+    "name": "LoanAccountStayedLockedDataV1",
+    "namespace": "org.apache.fineract.avro.loan.v1",
+    "type": "record",
+    "fields": [
+        {
+            "name": "id",
+            "type": "long"
+        },
+        {
+            "default": null,
+            "name": "externalId",
+            "type": [
+                "null",
+                "string"
+            ]
+        },
+        {
+            "default": null,
+            "name": "accountNo",
+            "type": [
+                "null",
+                "string"
+            ]
+        }
+    ]
+}
--- a/fineract-provider/src/main/java/org/apache/fineract/cob/domain/LoanAccountLock.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/cob/domain/LoanAccountLock.java
@@ -18,6 +18,7 @@
  */
 package org.apache.fineract.cob.domain;
 
+import java.time.LocalDate;
 import java.time.OffsetDateTime;
 import javax.persistence.Column;
 import javax.persistence.Entity;
@@ -28,7 +29,9 @@ import javax.persistence.Table;
 import javax.persistence.Version;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
+import org.apache.fineract.infrastructure.businessdate.domain.BusinessDateType;
 import org.apache.fineract.infrastructure.core.service.DateUtils;
+import org.apache.fineract.infrastructure.core.service.ThreadLocalContextUtil;
 
 @Entity
 @Table(name = "m_loan_account_locks")
@@ -57,10 +60,14 @@ public class LoanAccountLock {
     @Column(name = "stacktrace")
     private String stacktrace;
 
+    @Column(name = "lock_placed_on_cob_business_date")
+    private LocalDate lockPlacedOnCobBusinessDate;
+
     public LoanAccountLock(Long loanId, LockOwner lockOwner) {
         this.loanId = loanId;
         this.lockOwner = lockOwner;
         this.lockPlacedOn = DateUtils.getOffsetDateTimeOfTenant();
+        this.lockPlacedOnCobBusinessDate = ThreadLocalContextUtil.getBusinessDateByType(BusinessDateType.COB_DATE);
     }
 
     public void setError(String errorMessage, String stacktrace) {
@@ -71,5 +78,6 @@ public class LoanAccountLock {
     public void setNewLockOwner(LockOwner newLockOwner) {
         this.lockOwner = newLockOwner;
         this.lockPlacedOn = DateUtils.getOffsetDateTimeOfTenant();
+        this.lockPlacedOnCobBusinessDate = ThreadLocalContextUtil.getBusinessDateByType(BusinessDateType.COB_DATE);
     }
 }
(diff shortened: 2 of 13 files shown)


fineract-avro-schemas/src/main/avro/loan/v1/LoanAccountStayedLockedDataV1.avsc
```diff
@@ -0,0 +1,31 @@
+{
+    "name": "LoanAccountStayedLockedDataV1",
+    "namespace": "org.apache.fineract.avro.loan.v1",
+    "type": "record",
+    "fields": [
+        {
+            "default": null,
```
adamsaghy: if the loan id cannot be null, please reflect it in the avro schema as well!

fineract-provider/src/main/java/org/apache/fineract/cob/loan/StayedLockedLoansTasklet.java
```diff
@@ -35,14 +40,24 @@ public class StayedLockedLoansTasklet implements Tasklet {
...
         if (!loanAccountLocks.isEmpty()) {
             List<Long> loanIds = loanAccountLocks.stream().map(LoanAccountLock::getLoanId).toList();
-            businessEventNotifierService.notifyPostBusinessEvent(new LoanAccountsStayedLockedBusinessEvent(loanIds));
+            businessEventNotifierService.notifyPostBusinessEvent(new LoanAccountsStayedLockedBusinessEvent(buildLoanAccountData(loanIds)));
         }
         return RepeatStatus.FINISHED;
     }
+
+    private LoanAccountsStayedLockedBusinessEvent.Data buildLoanAccountData(List<Long> loanIds) {
+        List<LoanIdAndExternalIdAndAccountNo> stayedLockedLoanAccounts = loanRepository.findAllStayedLockedByLoanIds(loanIds);
```
adamsaghy: Please make sure the loanIds array is not too big / request. Postgres has some restrictions on the parameter size.

adamsaghy (changes requested):
Kindly check my comments!

fineract-provider/src/main/java/org/apache/fineract/cob/loan/StayedLockedLoansTasklet.java
galovics: Do we need to at all go to the database to grab this data? Can't we pass this through some context?

taskain7: I changed the query, we don't have to deal with the limitation anymore.

fineract-provider/src/main/java/org/apache/fineract/cob/domain/LoanAccountLock.java
```diff
@@ -57,10 +60,14 @@ public class LoanAccountLock {
     @Column(name = "stacktrace")
     private String stacktrace;
 
+    @Column(name = "lock_placed_on_cob_business_date", nullable = false)
+    private LocalDate lockPlacedOnCobBusinessDate;
```
galovics, line 64: Can't we just call it lockPlacedOnBusinessDate and use it as DateUtils.getBusinessDate()?

fineract-provider/src/main/java/org/apache/fineract/cob/loan/StayedLockedLoansTasklet.java
```diff
@@ -33,16 +39,25 @@
...
+    private final LoanRepository loanRepository;
 
     @Override
     public RepeatStatus execute(StepContribution contribution, ChunkContext chunkContext) throws Exception {
-        List<LoanAccountLock> loanAccountLocks = loanAccountLockRepository.findAll();
-        if (!loanAccountLocks.isEmpty()) {
-            List<Long> loanIds = loanAccountLocks.stream().map(LoanAccountLock::getLoanId).toList();
-            businessEventNotifierService.notifyPostBusinessEvent(new LoanAccountsStayedLockedBusinessEvent(loanIds));
+        LoanAccountsStayedLockedBusinessEvent.Data lockedLoanAccounts = buildLoanAccountData();
+        if (!lockedLoanAccounts.getLoanAccounts().getLoanAccounts().isEmpty()) {
```
galovics: why the double getLoanAccounts()?

fineract-provider/src/main/resources/db/changelog/tenant/parts/0086_add_cob_business_date_to_loan_account_locks.xml
```diff
@@ -0,0 +1,32 @@
...
+    under the License.
+
+-->
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd">
+    <changeSet author="fineract" id="1" context="mysql">
+        <addColumn tableName="m_loan_account_locks">
+            <column name="lock_placed_on_cob_business_date" type="DATE">
+                <constraints nullable="false"/>
```
galovics: Let's give it a default value in this case.

galovics (changes requested):

fineract-provider/src/main/java/org/apache/fineract/cob/domain/LoanAccountLock.java
taskain7: we could, however, since COB places the lock on the account, the COB busines date made more sense to me.

fineract-provider/src/main/java/org/apache/fineract/cob/loan/StayedLockedLoansTasklet.java
taskain7: used unneccessary wrapper class, removed it.

fineract-provider/src/main/java/org/apache/fineract/cob/domain/LoanAccountLock.java
adamsaghy: Just to give a 3rd opinion: "lock_placed_on" makes more sense to me. We should not put business logic into db column naming (i know, i know...many cases its already there but...)

taskain7: we already have a "lock_placed_on", but that's the actual datetime, and we need the business date as well to be able to connect the lock and the COB execution

adamsaghy: i see. thanks.

fineract-provider/src/main/resources/db/changelog/tenant/parts/0086_add_cob_business_date_to_loan_account_locks.xml
taskain7: removed the constraint, we are more backward compatible with the nullable column

galovics (approved):

PR #2908, opened 2023-01-22
FINERACT-1840: Fee income is not realised if the loan got fully repaid

josehernandezfintecheandomx:
## Description

A Loan with a NSF fee is added (specific due date fee) and the loan got fully repaid on the same day the Fee income is not realised, the Accrual transaction is not added

[FINERACT-1840](https://issues.apache.org/jira/browse/FINERACT-1840)

- Repayment Schedule
Screenshot 2023-01-22 at 16 46 26

- Transactions
Screenshot 2023-01-22 at 16 46 17

- Another example
Screenshot 2023-01-26 at 0 31 48

Screenshot 2023-01-26 at 0 31 04

Screenshot 2023-01-26 at 0 30 50

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/accounting/journalentry/api/JournalEntriesApiResource.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/accounting/journalentry/api/JournalEntriesApiResource.java
@@ -104,7 +104,7 @@ public class JournalEntriesApiResource {
             + "\n" + "journalentries?orderBy=transactionId&sortOrder=DESC\n" + "\n" + "journalentries?runningBalance=true\n" + "\n"
             + "journalentries?transactionDetails=true\n" + "\n" + "journalentries?loanId=12\n" + "\n" + "journalentries?savingsId=24")
     @ApiResponses({
-            @ApiResponse(responseCode = "200", description = "OK", content = @Content(array = @ArraySchema(schema = @Schema(implementation = JournalEntryData.class)))) })
+            @ApiResponse(responseCode = "200", description = "OK", content = @Content(array = @ArraySchema(schema = @Schema(implementation = JournalEntriesApiResourceSwagger.GetJournalEntriesTransactionIdResponse.class)))) })
     public String retrieveAll(@Context final UriInfo uriInfo,
             @QueryParam("officeId") @Parameter(description = "officeId") final Long officeId,
             @QueryParam("glAccountId") @Parameter(description = "glAccountId") final Long glAccountId,
@@ -166,7 +166,7 @@ public class JournalEntriesApiResource {
             + "journalentries/1?fields=officeName,glAccountId,entryType,amount\n" + "\n" + "journalentries/1?runningBalance=true\n" + "\n"
             + "journalentries/1?transactionDetails=true")
     @ApiResponses({
-            @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = JournalEntryData.class))) })
+            @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = JournalEntriesApiResourceSwagger.JournalEntryTransactionItem.class))) })
     public String retrieveJournalEntryById(
             @PathParam("journalEntryId") @Parameter(description = "journalEntryId") final Long journalEntryId,
             @Context final UriInfo uriInfo,
--- a/fineract-provider/src/main/java/org/apache/fineract/accounting/journalentry/api/JournalEntriesApiResourceSwagger.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/accounting/journalentry/api/JournalEntriesApiResourceSwagger.java
@@ -19,6 +19,12 @@
 package org.apache.fineract.accounting.journalentry.api;
 
 import io.swagger.v3.oas.annotations.media.Schema;
+import java.math.BigDecimal;
+import java.time.LocalDate;
+import java.util.List;
+import org.apache.fineract.organisation.monetary.api.CurrenciesApiResourceSwagger.CurrencyItem;
+import org.apache.fineract.portfolio.note.api.NotesApiResourceSwagger.GetResourceTypeResourceIdNotesResponse;
+import org.apache.fineract.portfolio.paymenttype.api.PaymentTypeApiResourceSwagger.GetPaymentTypesResponse;
 
 /**
  * Created by sanyam on 25/7/17.
@@ -62,4 +68,112 @@ final class JournalEntriesApiResourceSwagger {
         @Schema(description = "1")
         public Long officeId;
     }
+
+    static final class EnumOptionType {
+
+        private EnumOptionType() {}
+
+        @Schema(example = "2")
+        public Long id;
+        @Schema(example = "accountType.asset")
+        public String code;
+        @Schema(example = "ASSET")
+        public String value;
+    }
+
+    static final class JournalEntryTransactionItem {
+
+        private JournalEntryTransactionItem() {}
+
+        static final class PaymentDetailData {
+
+            private PaymentDetailData() {}
+
+            @Schema(example = "62")
+            public Long id;
+            public GetPaymentTypesResponse paymentType;
+            @Schema(example = "acc123")
+            public String accountNumber;
+            @Schema(example = "che123")
+            public String checkNumber;
+            @Schema(example = "rou123")
+            public String routingCode;
+            @Schema(example = "rec123")
+            public String receiptNumber;
+            @Schema(example = "ban123")
+            public String bankNumber;
+        }
+
+        static final class TransactionDetails {
+
+            private TransactionDetails() {}
+
+            @Schema(example = "2")
+            public Long transactionId;
+            public EnumOptionType transactionType;
+            public GetResourceTypeResourceIdNotesResponse noteData;
+            public PaymentDetailData paymentDetails;
+        }
+
+        @Schema(example = "1")
+        public Long id;
+        @Schema(example = "L12")
+        public String transactionId;
+        @Schema(example = "1")
+        public Long entityId;
+        @Schema(example = "1")
+        public Long officeId;
+        @Schema(example = "Head Office")
+        public String officeName;
+        @Schema(example = "10")
+        public Long glAccountId;
+        @Schema(example = "Cash Account")
+        public String glAccountName;
+        @Schema(example = "0123-4567")
+        public String glAccountCode;
+        @Schema(example = "[2022, 07, 01]")
+        public LocalDate transactionDate;
+        @Schema(example = "[2022, 07, 01]")
+        public LocalDate submittedOnDate;
+
+        @Schema(example = "100.000000")
+        public Double amount;
+        @Schema(example = "false")
+        public boolean reversed;
+        @Schema(example = "false")
+        public boolean manualEntry;
+        @Schema(example = "Manual entry")
+        public String comments;
+        @Schema(example = "QWERTY")
+        public String referenceNumber;
+        @Schema(example = "1234.56")
+        public BigDecimal officeRunningBalance;
+        @Schema(example = "1234.56")
+        public BigDecimal organizationRunningBalance;
+        @Schema(example = "false")
+        public boolean runningBalanceComputed;
+        @Schema(example = "1")
+        public Long createdByUserId;
+        @Schema(example = "mifos")
+        public String createdByUserName;
+        @Schema(example = "[2022, 07, 01]")
+        public LocalDate createdDate;
+
+        public CurrencyItem currency;
+        public EnumOptionType glAccountType;
+        public EnumOptionType entryType;
+        public EnumOptionType entityType;
+        public TransactionDetails transactionDetails;
+    }
+
+    @Schema(description = "GetJournalEntriesTransactionIdResponse")
+    public static final class GetJournalEntriesTransactionIdResponse {
+
+        private GetJournalEntriesTransactionIdResponse() {}
+
+        @Schema(example = "2")
+        public Long totalFilteredRecords;
+        public List<JournalEntryTransactionItem> pageItems;
+    }
+
 }
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
@@ -3447,10 +3447,10 @@ public class Loan extends AbstractAuditableWithUTCDateTimeCustom {
             }
         }
         if (isAllChargesPaid) {
-            loanLifecycleStateMachine.transition(LoanEvent.REPAID_IN_FULL, this);
-
             this.closedOnDate = transactionDate;
             this.actualMaturityDate = transactionDate;
+            loanLifecycleStateMachine.transition(LoanEvent.REPAID_IN_FULL, this);
+
         } else if (LoanStatus.fromInt(this.loanStatus).isOverpaid()) {
             if (this.totalOverpaid == null || BigDecimal.ZERO.compareTo(this.totalOverpaid) == 0) {
                 this.overpaidOnDate = null;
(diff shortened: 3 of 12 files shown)


adamsaghy (changes requested):
We should not rework an existing business logic without a good reason and proper test coverage. 
Please discuss these changes further!

fineract-provider/src/main/java/org/apache/fineract/notification/service/NotificationDomainServiceImpl.java
```diff
@@ -345,4 +347,19 @@ private Set<Long> getNotifiableUserIds(Long officeId, String permission) {
         Collection<AppUser> usersWithPermission = users.stream().filter(aU -> aU.hasAnyPermission(permission, "ALL_FUNCTIONS")).toList();
         return usersWithPermission.stream().map(AppUser::getId).collect(toSet());
     }
+
+    private static class LoanStatusChangedListener implements BusinessEventListener<LoanStatusChangedBusinessEvent> {
```
adamsaghy: Please move this logic into LoanWritePlatformServiceJpaRepositoryImpl or create a new service which will take care of the logic. I reckon the NotificationDomainService is not the proper place!

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
```diff
@@ -3451,9 +3451,9 @@ private void handleLoanRepaymentInFull(final LocalDate transactionDate, final Lo
             }
         }
         if (isAllChargesPaid) {
+            this.closedOnDate = transactionDate;
```
adamsaghy, line 3450: Why did you move the date? If the reason was to make this date be part of the event, please move the actualMaturityDate as well to be consistent.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
```diff
@@ -3490,26 +3490,47 @@ private void processIncomeAccrualTransactionOnLoanClosure() {
...
 
+    public void applyIncomeAccrualTransaction(LocalDate closedDate) {
+        if (isPeriodicAccrualAccountingEnabledOnLoanProduct()) {
+            ExternalId externalId = ExternalId.empty();
+            final boolean isExternalIdAutoGenerationEnabled = TemporaryConfigurationServiceContainer.isExternalIdAutoGenerationEnabled();
+            if (isExternalIdAutoGenerationEnabled) {
+                externalId = ExternalId.generate();
+            }
+
+            HashMap<String, BigDecimal> cumulativeIncomeFromInstallments = new HashMap<>();
```
adamsaghy: The two logic is not matching!
Things that got extracted from 3493 to 3508 is not matching with the new logic of the "applyIncomeAccrualTransaction" method. Please double check and remove the unnecessary pieces!

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
```diff
@@ -3801,6 +3822,8 @@ public ChangedTransactionDetail closeAsWrittenOff(final JsonCommand command, fin
 
         validateAccountStatus(LoanEvent.WRITE_OFF_OUTSTANDING);
 
+        final LocalDate writtenOffOnLocalDate = command.localDateValueOfParameterNamed(TRANSACTION_DATE);
+        this.closedOnDate = writtenOffOnLocalDate;
```
adamsaghy, line 3834: Why to move these dates?

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
```diff
@@ -3969,14 +3990,13 @@ public ChangedTransactionDetail close(final JsonCommand command, final LoanLifec
             final Money totalLoanOverpayment = calculateTotalOverpayment();
             if (totalLoanOverpayment.isGreaterThanZero() && getInArrearsTolerance().isGreaterThanOrEqualTo(totalLoanOverpayment)) {
                 // TODO - KW - technically should set somewhere that this loan
-                // has
-                // 'overpaid' amount
+                // has 'overpaid' amount
+                this.closedOnDate = closureDate;
```
adamsaghy, line 4002: Why to move this date?

adamsaghy (changes requested):
Please see my comments and please dont forget to cover the new behaviour with proper tests!

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
josehernandezfintecheandomx: I needed to move the set of the `closedOnDate` value before the Life cycle machine method call because when the event listener was called It was in null at that moment

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
josehernandezfintecheandomx: I needed to move the set of the `closedOnDate` value before the Life cycle machine method call because when the event listener was called It was in null at that moment

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
adamsaghy: Please move the actualMaturityDate as well then...

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
josehernandezfintecheandomx: I needed to move the set of the `closedOnDate` value before the Life cycle machine method call because when the event listener was called It was in null at that moment

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
adamsaghy: What is weird...even the loan got changed later, since it is the same instance, the updated field should be reflected there as well.

fineract-provider/src/main/java/org/apache/fineract/notification/service/NotificationDomainServiceImpl.java
josehernandezfintecheandomx: Done!

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
josehernandezfintecheandomx: Done

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
josehernandezfintecheandomx: Done

fineract-provider/src/main/java/org/apache/fineract/notification/service/NotificationDomainServiceImpl.java
josehernandezfintecheandomx: Done It was moved

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
josehernandezfintecheandomx: The logic is there

fineract-provider/src/main/java/org/apache/fineract/notification/service/NotificationDomainServiceImpl.java
```diff
@@ -345,4 +345,5 @@ private Set<Long> getNotifiableUserIds(Long officeId, String permission) {
         Collection<AppUser> usersWithPermission = users.stream().filter(aU -> aU.hasAnyPermission(permission, "ALL_FUNCTIONS")).toList();
         return usersWithPermission.stream().map(AppUser::getId).collect(toSet());
     }
+
```
adamsaghy: Please remove this extra line

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
```diff
@@ -3451,9 +3451,9 @@ private void handleLoanRepaymentInFull(final LocalDate transactionDate, final Lo
             }
         }
         if (isAllChargesPaid) {
+            this.closedOnDate = transactionDate;
             loanLifecycleStateMachine.transition(LoanEvent.REPAID_IN_FULL, this);
 
-            this.closedOnDate = transactionDate;
             this.actualMaturityDate = transactionDate;
```
adamsaghy: Please move the actualMaturityDate as well...

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
```diff
@@ -3510,6 +3510,39 @@ private void processIncomeAccrualTransactionOnLoanClosure() {
...
+            BigDecimal fee = (BigDecimal) feeDetails.get(FEE);
+            if (fee == null) {
+                fee = BigDecimal.ZERO;
+            }
+            BigDecimal penalty = (BigDecimal) feeDetails.get(PENALTY);
+            if (penalty == null) {
+                penalty = BigDecimal.ZERO;
+            }
+            BigDecimal total = fee.add(penalty);
+            // TODO: calculate interest?
```
adamsaghy, line 3532: Please check whether we need the interest calculation as well, or do it later...

fineract-provider/src/main/java/org/apache/fineract/notification/service/NotificationDomainServiceImpl.java
josehernandezfintecheandomx: Done

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanChargeSpecificDueDateTest.java
```diff
@@ -233,6 +236,109 @@ public void testApplyAndWaiveLoanSpecificDueDatePenaltyWithDisbursementDate() {
...
+        log.info("Operation date {}", transactionDate);
+        payloadJSON = LoanTransactionHelper.getSpecifiedDueDateChargesForLoanAsJSON(chargeId.toString(), operationDate, feeAmount);
+        postLoansLoanIdChargesResponse = loanTransactionHelper.addChargeForLoan(loanId, payloadJSON, responseSpec);
+        assertNotNull(postLoansLoanIdChargesResponse);
+        final Long loanChargeId03 = postLoansLoanIdChargesResponse.getResourceId();
+        assertNotNull(loanChargeId03);
+
+        getLoansLoanIdResponse = loanTransactionHelper.getLoan(requestSpec, responseSpec, loanId);
+        validateLoanAccount(getLoansLoanIdResponse, Double.valueOf(principalAmount), Double.valueOf("40.00"), true);
+        loanTransactionHelper.evaluateLoanTransactionData(getLoansLoanIdResponse, "loanTransactionType.accrual", Double.valueOf("50.00"));
```
adamsaghy: Why is it 50? There was no periodic accrual execution yet...

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
josehernandezfintecheandomx: We have comment this will be later

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanChargeSpecificDueDateTest.java
josehernandezfintecheandomx: In my opinion and as I have seen in other cases the `runaccrual` call runs separately and when it is executed in the test, we already have passed the add the third Loan Charge

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
josehernandezfintecheandomx: Done!

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanChargeSpecificDueDateTest.java
adamsaghy: In that case this test is not sufficient, because the add periodic accrual job will calculate the Accrual transaction and not the event of loan got closed. We need a test which makes sure, the last accrual transaction was created by not a job, but by the system when the loan got closed.

adamsaghy: Would you also please check whether the very last accrual transaction amount, penalty portion, transaction date is correct and it created the journal entries for it as well?

josehernandezfintecheandomx: Fixed, there were some issues in this point, as we saw the IT was not using Accrual accounting and in the code the Fee Details was adding the PENALTIES and the read for the Accrual Transaction was reading PENALTY. That includes the old code too.

fineract-provider/src/main/java/org/apache/fineract/accounting/journalentry/api/JournalEntriesApiResource.java
```diff
@@ -166,7 +166,7 @@ public String retrieveAll(@Context final UriInfo uriInfo,
             + "journalentries/1?fields=officeName,glAccountId,entryType,amount\n" + "\n" + "journalentries/1?runningBalance=true\n" + "\n"
             + "journalentries/1?transactionDetails=true")
     @ApiResponses({
-            @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = JournalEntryData.class))) })
+            @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = JournalEntriesApiResourceSwagger.GetJournalEntriesTransactionIdResponse.class))) })
```
adamsaghy: This looks not right to me... This API is not returning the response with paging. I think the one you are looking for is the `retrieveAll` method.

I reckon for this method the right swagger class would be `JournalEntriesApiResourceSwagger.GetJournalEntriesTransactionIdResponse.JournalEntryTransactionPageItems.class`

However I would rename the `JournalEntryTransactionPageItems` to something more generic like `JournalEntryTransactionItem`

fineract-provider/src/main/java/org/apache/fineract/accounting/journalentry/api/JournalEntriesApiResourceSwagger.java
```diff
@@ -62,4 +64,77 @@ private PostJournalEntriesTransactionIdResponse() {
...
+            @Schema(example = "2")
+            public Long transactionId;
+            public EnumOptionType transactionType;
+        }
+
+        static final class JournalEntryTransactionPageItems {
+
+            private JournalEntryTransactionPageItems() {}
+
+            @Schema(example = "1")
```
adamsaghy: Please add all the relevant fields from the JournalEntryData

adamsaghy (changes requested):
Please kindly check my comments!

fineract-provider/src/main/java/org/apache/fineract/accounting/journalentry/api/JournalEntriesApiResource.java
josehernandezfintecheandomx: When you get the Journal entries using the Transaction Id, the current endpoint return an array (usually two items) with the gl journal entries but It returns with paging 

Screenshot 2023-02-02 at 12 56 42

Done, I've renamed the class

fineract-provider/src/main/java/org/apache/fineract/accounting/journalentry/api/JournalEntriesApiResourceSwagger.java
josehernandezfintecheandomx: Done, the other fields are used for the bulk import process

fineract-provider/src/main/java/org/apache/fineract/accounting/journalentry/api/JournalEntriesApiResource.java
adamsaghy: You are still annotating the wrong method. The "retrieveAll" is called not the "retrieveJournalEntryById"

josehernandezfintecheandomx: Done, updated both annotations

PR #2915, opened 2023-01-23
FINERACT-1707 - S3 export and dependent service

b0c1:
FINERACT-1707 - S3 export and dependent service
- [x] S3 export function
- [x] list all available export-type services
- [x] refactor export service
- [x] unit test for export name generator
- [x] s3 integration test
- [x] localstack integration for testing

## Description

## Checklist

### DIFF
--- a/.github/workflows/build-mariadb.yml
+++ b/.github/workflows/build-mariadb.yml
@@ -52,12 +52,39 @@ jobs:
         run: |
             ./gradlew --no-daemon -q createDB -PdbName=fineract_tenants
             ./gradlew --no-daemon -q createDB -PdbName=fineract_default
+      - name: Start LocalStack
+        env:
+          AWS_ENDPOINT_URL: http://localhost:4566
+          AWS_ACCESS_KEY_ID: localstack
+          AWS_SECRET_ACCESS_KEY: localstack
+          AWS_REGION: us-east-1
+        run: |
+          echo "Update python pyopenssl"
+          pip install --upgrade pyopenssl
+          echo "Install localstack"
+          pip install localstack awscli-local[ver1] # install LocalStack cli and awslocal
+          docker pull localstack/localstack         # Make sure to pull the latest version of the image
+          localstack start -d                       # Start LocalStack in the background
+
+          echo "Waiting for LocalStack startup..."  # Wait 30 seconds for the LocalStack container
+          localstack wait -t 30                     # to become ready before timing out
+          echo "Startup complete"
+          echo "Create fineract S3 bucket"
+          awslocal s3api create-bucket --bucket fineract-reports
+          echo "LocalStack initialization complete"
       - name: Install additional software
         run: |
             sudo apt-get update
             sudo apt-get install ghostscript graphviz -y
 
       - name: Build & Test
+        env:
+          AWS_ENDPOINT_URL: http://localhost:4566
+          AWS_ACCESS_KEY_ID: localstack
+          AWS_SECRET_ACCESS_KEY: localstack
+          AWS_REGION: us-east-1
+          FINERACT_REPORT_EXPORT_S3_ENABLED: true
+          FINERACT_REPORT_EXPORT_S3_BUCKET_NAME: fineract-reports
         run: |
             ./gradlew --no-daemon --console=plain build test --fail-fast -x doc -x :twofactor-tests:test -x :oauth2-test:test
             ./gradlew --no-daemon --console=plain :twofactor-tests:test --fail-fast
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/api/RunreportsApiResource.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/api/RunreportsApiResource.java
@@ -20,6 +20,7 @@ package org.apache.fineract.infrastructure.dataqueries.api;
 
 import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.Parameter;
+import io.swagger.v3.oas.annotations.media.ArraySchema;
 import io.swagger.v3.oas.annotations.media.Content;
 import io.swagger.v3.oas.annotations.media.Schema;
 import io.swagger.v3.oas.annotations.responses.ApiResponse;
@@ -39,6 +40,7 @@ import javax.ws.rs.core.Response;
 import javax.ws.rs.core.UriInfo;
 import org.apache.fineract.infrastructure.core.api.ApiParameterHelper;
 import org.apache.fineract.infrastructure.core.exception.PlatformServiceUnavailableException;
+import org.apache.fineract.infrastructure.dataqueries.data.ReportExportType;
 import org.apache.fineract.infrastructure.dataqueries.service.ReadReportingService;
 import org.apache.fineract.infrastructure.report.provider.ReportingProcessServiceProvider;
 import org.apache.fineract.infrastructure.report.service.ReportingProcessService;
@@ -68,6 +70,29 @@ public class RunreportsApiResource {
         this.context = context;
         this.readExtraDataAndReportingService = readExtraDataAndReportingService;
         this.reportingProcessServiceProvider = reportingProcessServiceProvider;
+    }
+
+    @GET
+    @Path("/availableExports/{reportName}")
+    @Consumes({ MediaType.APPLICATION_JSON })
+    @Produces({ MediaType.APPLICATION_JSON })
+    @Operation(summary = "Return all available export types for the specific report", description = "Returns the list of all available export types.")
+    @ApiResponses({
+            @ApiResponse(responseCode = "200", description = "", content = @Content(array = @ArraySchema(schema = @Schema(implementation = ReportExportType.class)))) })
+    public Response retrieveAllAvailableExports(@PathParam("reportName") @Parameter(description = "reportName") final String reportName,
+            @Context final UriInfo uriInfo,
+            @DefaultValue("false") @QueryParam(IS_SELF_SERVICE_USER_REPORT_PARAMETER) @Parameter(description = IS_SELF_SERVICE_USER_REPORT_PARAMETER) final boolean isSelfServiceUserReport) {
+        MultivaluedMap<String, String> queryParams = new MultivaluedStringMap();
+        queryParams.putAll(uriInfo.getQueryParameters());
+
+        final boolean parameterType = ApiParameterHelper.parameterType(queryParams);
+        String reportType = readExtraDataAndReportingService.getReportType(reportName, isSelfServiceUserReport, parameterType);
+        ReportingProcessService reportingProcessService = reportingProcessServiceProvider.findReportingProcessService(reportType);
+        if (reportingProcessService == null) {
+            throw new PlatformServiceUnavailableException("err.msg.report.service.implementation.missing",
+                    ReportingProcessServiceProvider.SERVICE_MISSING + reportType, reportType);
+        }
+        return Response.ok().entity(reportingProcessService.getAvailableExportTargets()).build();
     }
 
     @GET
@@ -108,8 +133,8 @@ public class RunreportsApiResource {
         // Pass through isSelfServiceUserReport so that ReportingProcessService implementations can use it
         queryParams.putSingle(IS_SELF_SERVICE_USER_REPORT_PARAMETER, Boolean.toString(isSelfServiceUserReport));
 
-        String reportType = this.readExtraDataAndReportingService.getReportType(reportName, isSelfServiceUserReport, parameterType);
-        ReportingProcessService reportingProcessService = this.reportingProcessServiceProvider.findReportingProcessService(reportType);
+        String reportType = readExtraDataAndReportingService.getReportType(reportName, isSelfServiceUserReport, parameterType);
+        ReportingProcessService reportingProcessService = reportingProcessServiceProvider.findReportingProcessService(reportType);
         if (reportingProcessService == null) {
             throw new PlatformServiceUnavailableException("err.msg.report.service.implementation.missing",
                     ReportingProcessServiceProvider.SERVICE_MISSING + reportType, reportType);
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/data/ReportExportType.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/data/ReportExportType.java
@@ -16,27 +16,16 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.fineract.infrastructure.dataqueries.api;
+package org.apache.fineract.infrastructure.dataqueries.data;
 
-import io.swagger.v3.oas.annotations.media.Schema;
-import java.util.List;
-import org.apache.fineract.infrastructure.dataqueries.data.ResultsetColumnHeaderData;
-import org.apache.fineract.infrastructure.dataqueries.data.ResultsetRowData;
+import java.io.Serializable;
+import lombok.Data;
+import lombok.RequiredArgsConstructor;
 
-/**
- * Created by sanyam on 5/8/17. Fixed ;) by Michael Vorburger.ch on 2020/11/21.
- */
-final class RunreportsApiResourceSwagger {
+@Data
+@RequiredArgsConstructor
+public class ReportExportType implements Serializable {
 
-    private RunreportsApiResourceSwagger() {}
-
-    @Schema
-    public static final class RunReportsResponse {
-
-        private RunReportsResponse() {}
-
-        public List<ResultsetColumnHeaderData> columnHeaders;
-        public List<ResultsetRowData> data;
-    }
-
+    private final String key;
+    private final String queryParameter;
 }
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/DatatableReportingProcessService.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/DatatableReportingProcessService.java
@@ -18,105 +18,71 @@
  */
 package org.apache.fineract.infrastructure.dataqueries.service;
 
-import java.io.File;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
-import javax.ws.rs.core.MediaType;
+import java.util.Optional;
 import javax.ws.rs.core.MultivaluedMap;
 import javax.ws.rs.core.Response;
-import javax.ws.rs.core.Response.ResponseBuilder;
-import javax.ws.rs.core.StreamingOutput;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.fineract.infrastructure.core.api.ApiParameterHelper;
-import org.apache.fineract.infrastructure.core.serialization.ToApiJsonSerializer;
+import org.apache.fineract.infrastructure.core.service.StreamUtil;
 import org.apache.fineract.infrastructure.dataqueries.api.RunreportsApiResource;
-import org.apache.fineract.infrastructure.dataqueries.data.GenericResultsetData;
-import org.apache.fineract.infrastructure.dataqueries.data.ReportData;
+import org.apache.fineract.infrastructure.dataqueries.data.ReportExportType;
+import org.apache.fineract.infrastructure.dataqueries.service.export.DatatableReportExportService;
+import org.apache.fineract.infrastructure.dataqueries.service.export.ResponseHolder;
 import org.apache.fineract.infrastructure.report.annotation.ReportService;
 import org.apache.fineract.infrastructure.report.service.ReportingProcessService;
-import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
 @Service
 @ReportService(type = { "Table", "Chart", "SMS" })
+@RequiredArgsConstructor
+@Slf4j
 public class DatatableReportingProcessService implements ReportingProcessService {
 
-    private final ReadReportingService readExtraDataAndReportingService;
-    private final ToApiJsonSerializer<ReportData> toApiJsonSerializer;
-    private final GenericDataService genericDataService;
-
-    @Autowired
-    public DatatableReportingProcessService(final ReadReportingService readExtraDataAndReportingService,
-            final GenericDataService genericDataService, final ToApiJsonSerializer<ReportData> toApiJsonSerializer) {
-        this.readExtraDataAndReportingService = readExtraDataAndReportingService;
-        this.toApiJsonSerializer = toApiJsonSerializer;
-        this.genericDataService = genericDataService;
-    }
+    private final List<DatatableReportExportService> exportServices;
 
     @Override
     public Response processRequest(String reportName, MultivaluedMap<String, String> queryParams) {
         boolean isSelfServiceUserReport = Boolean.parseBoolean(
                 queryParams.getOrDefault(RunreportsApiResource.IS_SELF_SERVICE_USER_REPORT_PARAMETER, List.of("false")).get(0));
 
-        DatatableExportTargetParameter exportMode = DatatableExportTargetParameter.checkTarget(queryParams);
+        DatatableExportTargetParameter exportMode = DatatableExportTargetParameter.resolverExportTarget(queryParams);
         final String parameterTypeValue = ApiParameterHelper.parameterType(queryParams) ? "parameter" : "report";
         final Map<String, String> reportParams = getReportParams(queryParams);
-        return switch (exportMode) {
-            case CSV -> exportCSV(reportName, queryParams, reportParams, isSelfServiceUserReport, parameterTypeValue);
-            case PDF -> exportPDF(reportName, queryParams, reportParams, isSelfServiceUserReport, parameterTypeValue);
-            case S3 -> exportS3(reportName, queryParams, reportParams, isSelfServiceUserReport, parameterTypeValue);
-            default -> exportJSON(reportName, queryParams, reportParams, isSelfServiceUserReport, parameterTypeValue,
-                    exportMode == DatatableExportTargetParameter.PRETTY_JSON);
-        };
-    }
-
-    private Response exportJSON(String reportName, MultivaluedMap<String, String> queryParams, Map<String, String> reportParams,
-            boolean isSelfServiceUserReport, String parameterTypeValue, boolean prettyPrint) {
-        final GenericResultsetData result = this.readExtraDataAndReportingService.retrieveGenericResultset(reportName, parameterTypeValue,
-                reportParams, isSelfServiceUserReport);
-
-        String json;
-        final boolean genericResultSetIsPassed = ApiParameterHelper.genericResultSetPassed(queryParams);
-        final boolean genericResultSet = ApiParameterHelper.genericResultSet(queryParams);
-        if (genericResultSetIsPassed) {
-            if (genericResultSet) {
-                json = this.toApiJsonSerializer.serializePretty(prettyPrint, result);
-            } else {
-                json = this.genericDataService.generateJsonFromGenericResultsetData(result);
-            }
-        } else {
-            json = this.toApiJsonSerializer.serializePretty(prettyPrint, result);
+        ResponseHolder response = findReportExportService(exportMode) //
+                .orElseThrow(() -> new IllegalArgumentException("Unsupported export target: " + exportMode)) //
+                .export(reportName, queryParams, reportParams, isSelfServiceUserReport, parameterTypeValue);
+        Response.ResponseBuilder builder = Response.status(response.status().getStatusCode());
+        if (StringUtils.isNotBlank(response.contentType())) {
+            builder = builder.type(response.contentType());
         }
-
-        return Response.ok().entity(json).type(MediaType.APPLICATION_JSON).build();
+        if (StringUtils.isNotBlank(response.fileName())) {
+            builder = builder.header("Content-Disposition", "attachment; filename=" + response.fileName());
+        }
+        if (response.entity() != null) {
+            builder = builder.entity(response.entity());
+        }
+        if (response.headers() != null && !response.headers().isEmpty()) {
+            builder = response.headers().stream().collect(StreamUtil.foldLeft(builder, (b, h) -> b.header(h.getKey(), h.getValue())));
+        }
+        return builder.build();
     }
 
-    private Response exportS3(String reportName, MultivaluedMap<String, String> queryParams, Map<String, String> reportParams,
-            boolean isSelfServiceUserReport, String parameterTypeValue) {
-        throw new UnsupportedOperationException("S3 export not supported for datatables");
+    @Override
+    public List<ReportExportType> getAvailableExportTargets() {
+        return Arrays //
+                .stream(DatatableExportTargetParameter.values()) //
+                .filter(target -> findReportExportService(target).isPresent()) //
+                .map(target -> new ReportExportType(target.name(), target.getValue())) //
+                .toList();
     }
 
-    private Response exportPDF(String reportName, MultivaluedMap<String, String> queryParams, Map<String, String> reportParams,
-            boolean isSelfServiceUserReport, String parameterTypeValue) {
-
-        final String pdfFileName = this.readExtraDataAndReportingService.retrieveReportPDF(reportName, parameterTypeValue, reportParams,
-                isSelfServiceUserReport);
-
-        final File file = new File(pdfFileName);
-
-        final ResponseBuilder response = Response.ok(file);
-        response.header("Content-Disposition", "attachment; filename=\"" + pdfFileName + "\"");
-        response.header("content-Type", "application/pdf");
-
-        return response.build();
+    private Optional<DatatableReportExportService> findReportExportService(DatatableExportTargetParameter target) {
+        return exportServices.stream().filter(service -> service.supports(target)).findFirst();
     }
 
-    private Response exportCSV(String reportName, MultivaluedMap<String, String> queryParams, Map<String, String> reportParams,
-            boolean isSelfServiceUserReport, String parameterTypeValue) {
-        final StreamingOutput result = this.readExtraDataAndReportingService.retrieveReportCSV(reportName, parameterTypeValue, reportParams,
-                isSelfServiceUserReport);
-
-        return Response.ok().entity(result).type("text/csv")
-                .header("Content-Disposition", "attachment;filename=" + reportName.replaceAll(" ", "") + ".csv").build();
-
-    }
 }
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/export/CsvDatatableReportExportServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/export/CsvDatatableReportExportServiceImpl.java
@@ -1,0 +1,51 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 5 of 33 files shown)


fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/data/ReportExportType.java
```diff
@@ -0,0 +1,34 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.infrastructure.dataqueries.data;
+
+import java.io.Serializable;
+import lombok.Data;
+
+@Data
+public class ReportExportType implements Serializable {
```
galovics, line 27: `@RequiredArgsConstructor`

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/api/RunreportsApiResource.java
```diff
@@ -70,6 +72,29 @@ public RunreportsApiResource(final PlatformSecurityContext context, final ReadRe
...
+    @ApiResponses({
+            @ApiResponse(responseCode = "200", description = "", content = @Content(array = @ArraySchema(schema = @Schema(implementation = ReportExportType.class)))) })
+    public Response retrieveAllAvailableExports(@PathParam("reportName") @Parameter(description = "reportName") final String reportName,
+            @Context final UriInfo uriInfo,
+            @DefaultValue("false") @QueryParam(IS_SELF_SERVICE_USER_REPORT_PARAMETER) @Parameter(description = IS_SELF_SERVICE_USER_REPORT_PARAMETER) final boolean isSelfServiceUserReport) {
+        MultivaluedMap<String, String> queryParams = new MultivaluedStringMap();
+        queryParams.putAll(uriInfo.getQueryParameters());
+
+        final boolean parameterType = ApiParameterHelper.parameterType(queryParams);
+        String reportType = this.readExtraDataAndReportingService.getReportType(reportName, isSelfServiceUserReport, parameterType);
```
galovics: `this.` not needed

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/api/RunreportsApiResource.java
```diff
@@ -70,6 +72,29 @@ public RunreportsApiResource(final PlatformSecurityContext context, final ReadRe
...
+        ReportingProcessService reportingProcessService = this.reportingProcessServiceProvider.findReportingProcessService(reportType);
+        if (reportingProcessService == null) {
```
galovics, line 91: Let's not put this logic here but in the provider.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/DatatableReportingProcessService.java
```diff
@@ -90,13 +112,64 @@ private Response exportJSON(String reportName, MultivaluedMap<String, String> qu
         return Response.ok().entity(json).type(MediaType.APPLICATION_JSON).build();
     }
 
-    private Response exportS3(String reportName, MultivaluedMap<String, String> queryParams, Map<String, String> reportParams,
-            boolean isSelfServiceUserReport, String parameterTypeValue) {
-        throw new UnsupportedOperationException("S3 export not supported for datatables");
+    private boolean isAwsCredentialsValid() {
```
galovics: This definitely doesn't belong here. Why can't we use the AWS SDK provided `AwsCredentialsProvider`?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/DatatableReportingProcessService.java
```diff
@@ -18,57 +18,79 @@
...
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.s3.S3Client;
 
 @Service
 @ReportService(type = { "Table", "Chart", "SMS" })
+@RequiredArgsConstructor
+@Slf4j
 public class DatatableReportingProcessService implements ReportingProcessService {
 
+    public static final String AWS_S_3_BUCKET_NAME = "AWS_S3_BUCKET_NAME";
```
galovics: This doesn't belong here, let's use the FineractProperties for this.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/DatatableReportingProcessService.java
```diff
@@ -90,13 +112,64 @@ private Response exportJSON(String reportName, MultivaluedMap<String, String> qu
...
+            String parameterTypeValue) {
+        if (!isAwsCredentialsValid()) {
+            return Response.status(Response.Status.BAD_REQUEST).entity("AWS credentials not set").build();
+        }
+        if (StringUtils.isBlank(System.getenv().get(AWS_S_3_BUCKET_NAME))) {
+            return Response.status(Response.Status.BAD_REQUEST).entity("AWS S3 bucket name not set").build();
+        }
+        try {
+            S3Client s3Client = S3Client.builder().region(Region.of(System.getenv(SdkSystemSetting.AWS_REGION.environmentVariable())))
+                    .credentialsProvider(EnvironmentVariableCredentialsProvider.create()).build();
```
galovics: Let's not hardcode the EnvironmentVariableCredentialsProvider, we need the `DefaultCredentialsProvider` so that it'll read from multiple places.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/DatatableReportingProcessService.java
```diff
@@ -90,13 +112,64 @@ private Response exportJSON(String reportName, MultivaluedMap<String, String> qu
...
+            StreamingOutput output = this.readExtraDataAndReportingService.retrieveReportCSV(reportName, parameterTypeValue, reportParams,
```
galovics: `this.` not needed

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/DatatableReportingProcessService.java
```diff
@@ -90,13 +112,64 @@ private Response exportJSON(String reportName, MultivaluedMap<String, String> qu
...
+    private Response exportS3(String reportName, Map<String, String> reportParams, boolean isSelfServiceUserReport,
+            String parameterTypeValue) {
+        if (!isAwsCredentialsValid()) {
+            return Response.status(Response.Status.BAD_REQUEST).entity("AWS credentials not set").build();
+        }
+        if (StringUtils.isBlank(System.getenv().get(AWS_S_3_BUCKET_NAME))) {
+            return Response.status(Response.Status.BAD_REQUEST).entity("AWS S3 bucket name not set").build();
+        }
+        try {
+            S3Client s3Client = S3Client.builder().region(Region.of(System.getenv(SdkSystemSetting.AWS_REGION.environmentVariable())))
```
galovics: Can't we create the S3Client upon startup?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/DatatableReportingProcessService.java
galovics: It's not your fault, just noting but I really really don't like this StreamingOutput interface here, it's a javax.ws.rs class which is not supposed to be on this level but on the API.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/DatatableReportingProcessService.java
```diff
@@ -90,13 +112,64 @@ private Response exportJSON(String reportName, MultivaluedMap<String, String> qu
...
+                    .credentialsProvider(EnvironmentVariableCredentialsProvider.create()).build();
+            StreamingOutput output = this.readExtraDataAndReportingService.retrieveReportCSV(reportName, parameterTypeValue, reportParams,
+                    isSelfServiceUserReport);
+            ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
+            output.write(byteArrayOutputStream);
+            byteArrayOutputStream.close();
+            String finalS3Path = getS3FolderName() + "" + exportFileNameGeneration(reportName, reportParams) + ".csv";
```
galovics: Just a very minor thing but can we construct this path value with String.format? Also, is it intentional to have an empty String in the middle?

Bonus: would be awesome if you'd extract this to a factory class so that you can separately write unit test for it.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/DatatableReportingProcessService.java
```diff
@@ -90,13 +112,64 @@ private Response exportJSON(String reportName, MultivaluedMap<String, String> qu
...
+        }
+        if (StringUtils.isBlank(System.getenv().get(AWS_S_3_BUCKET_NAME))) {
+            return Response.status(Response.Status.BAD_REQUEST).entity("AWS S3 bucket name not set").build();
+        }
+        try {
+            S3Client s3Client = S3Client.builder().region(Region.of(System.getenv(SdkSystemSetting.AWS_REGION.environmentVariable())))
+                    .credentialsProvider(EnvironmentVariableCredentialsProvider.create()).build();
+            StreamingOutput output = this.readExtraDataAndReportingService.retrieveReportCSV(reportName, parameterTypeValue, reportParams,
+                    isSelfServiceUserReport);
+            ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
```
galovics: try with resources?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/DatatableReportingProcessService.java
```diff
@@ -90,13 +112,64 @@ private Response exportJSON(String reportName, MultivaluedMap<String, String> qu
...
+            output.write(byteArrayOutputStream);
+            byteArrayOutputStream.close();
+            String finalS3Path = getS3FolderName() + "" + exportFileNameGeneration(reportName, reportParams) + ".csv";
+            s3Client.putObject(builder -> builder.bucket(System.getenv().get(AWS_S_3_BUCKET_NAME)).key(finalS3Path).build(),
+                    RequestBody.fromBytes(byteArrayOutputStream.toByteArray()));
+            return Response.noContent().build();
+        } catch (IOException e) {
+            log.error("Error while exporting to S3", e);
```
galovics: No need for the log message if you're rethrowing the exception below.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/DatatableReportingProcessService.java
```diff
@@ -90,13 +112,64 @@ private Response exportJSON(String reportName, MultivaluedMap<String, String> qu
...
+            throw new IllegalStateException("Error while exporting to S3!", e);
```
galovics: Exclamation mark not necessary, is it? :)

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/DatatableReportingProcessService.java
```diff
@@ -90,13 +112,64 @@ private Response exportJSON(String reportName, MultivaluedMap<String, String> qu
...
+        }
+    }
+
+    private String getS3FolderName() {
+        String folderName = configurationDomainService.retrieveReportExportS3FolderName();
+        if (StringUtils.isNotBlank(folderName)) {
+            folderName = folderName + "/";
+        }
+        return folderName.trim();
```
galovics: why don't we trim beforehand? the end will not be trimmed anyway because you append a `/` at the end.
I think I'm missing more validation here since this is a user editable value (the global config). I think the configurationDomainService should take care of validating whether the folder name doesn't contain a slash at the end and whitespaces anywhere so that from a contractual point of view, whoever uses the `retrieveReportExportS3FolderName` method, they don't need to do the same check over and over again.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/DatatableReportingProcessService.java
```diff
@@ -90,13 +112,64 @@ private Response exportJSON(String reportName, MultivaluedMap<String, String> qu
...
+    }
+
+    private String exportFileNameGeneration(String reportName, Map<String, String> reportParams) {
+        String timestamp = "_" + new SimpleDateFormat("yyyyMMddHHmmss").format(new Date());
```
galovics: Careful, this is the old Date API on one hand, it's better to use the new java.time package for this and bonus question, shall we have the business date here, the tenant datetime or the system datetime? (pointer: DateUtils class)

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/DatatableReportingProcessService.java
```diff
@@ -90,13 +112,64 @@ private Response exportJSON(String reportName, MultivaluedMap<String, String> qu
...
+        int reportMaximumFileName = 512 - timestamp.length() - 4;
```
galovics: Comment here please on where the 512 and 4 comes from.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/DatatableReportingProcessService.java
```diff
@@ -90,13 +112,64 @@ private Response exportJSON(String reportName, MultivaluedMap<String, String> qu
...
+        String fileName = reportName;
+        if (reportParams != null) {
```
galovics: This should really be extracted into another class & method to be able to properly unit test it.

galovics (changes requested):

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/api/RunreportsApiResource.java
b0c1: After discuss, currently not changed, because used by another method.

vidakovic:
@galovics @b0c1 ... is there any reason why we use the database (global configuration service) to configure these things? Why wouldn't we put this in application.properties?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/DatatableReportingProcessService.java
b0c1: It's the generation datetime, I thin it's generation timestamp

.github/workflows/build-mariadb.yml
```diff
@@ -52,12 +52,39 @@ jobs:
...
+          AWS_ENDPOINT_URL: http://localhost:4566
+          AWS_ACCESS_KEY_ID: localstack
+          AWS_SECRET_ACCESS_KEY: localstack
+          AWS_REGION: us-east-1
+        run: |
+          echo "Update python pyopenssl"
+          pip install --upgrade pyopenssl
+          echo "Install localstack"
+          pip install localstack awscli-local[ver1] # install LocalStack cli and awslocal
+          docker pull localstack/localstack         # Make sure to pull the latest version of the image
```
galovics, line 66: Aren't we mixing up 2 different approaches here? One is starting localstack via its own CLI tool and one via docker?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/export/CsvDatatableReportExportServiceImpl.java
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.stereotype.Service;
+
+@Service
+@RequiredArgsConstructor
+public class CsvDatatableReportExportServiceImpl implements DatatableReportExportService {
+
+    private final ReadReportingService readExtraDataAndReportingService;
+
+    @Override
+    public Response export(String reportName, MultivaluedMap<String, String> queryParams, Map<String, String> reportParams,
```
galovics: I'm not gonna push back on the PR because of this but just a note. This is a service layer class yet it's returning a Response which is a view layer class. We're mixing up the layers.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/export/CsvDatatableReportExportServiceImpl.java
```diff
@@ -0,0 +1,51 @@
...
+            boolean isSelfServiceUserReport, String parameterTypeValue) {
+        final StreamingOutput result = this.readExtraDataAndReportingService.retrieveReportCSV(reportName, parameterTypeValue, reportParams,
+                isSelfServiceUserReport);
+
+        return Response.ok().entity(result).type("text/csv")
+                .header("Content-Disposition", "attachment;filename=" + reportName.replaceAll(" ", "") + ".csv").build();
```
galovics: When it comes to the attachment's name, shouldn't we do the same thing as for DatatableExportUtil#normalizeFolderName?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/export/S3DatatableReportExportServiceImpl.java
```diff
@@ -0,0 +1,71 @@
...
+@RequiredArgsConstructor
+public class S3DatatableReportExportServiceImpl implements DatatableReportExportService {
+
+    public static final int AWS_S3_MAXIMUM_KEY_LENGTH = 1024;
+    private final ReadReportingService readExtraDataAndReportingService;
+
+    private final ConfigurationDomainService configurationDomainService;
+    private final S3Client s3Client;
+
+    @Value("${fineract.report.export.s3.bucket:reports}")
```
galovics: FineractProperties class?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/export/S3DatatableReportExportServiceImpl.java
```diff
@@ -0,0 +1,71 @@
...
+    @Override
+    public Response export(String reportName, MultivaluedMap<String, String> queryParams, Map<String, String> reportParams,
+            boolean isSelfServiceUserReport, String parameterTypeValue) {
+        try {
+            StreamingOutput output = this.readExtraDataAndReportingService.retrieveReportCSV(reportName, parameterTypeValue, reportParams,
+                    isSelfServiceUserReport);
+            try (ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream()) {
+                output.write(byteArrayOutputStream);
+                String folder = configurationDomainService.retrieveReportExportS3FolderName();
+                String filePath = DatatableExportUtil.generateDatatableExportFileName(AWS_S3_MAXIMUM_KEY_LENGTH, folder, ".csv", reportName,
```
galovics: Can't we change the extension parameter to not need the dot prefix?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/s3/AmazonS3Config.java
```diff
@@ -0,0 +1,75 @@
...
+    public DefaultCredentialsProvider awsCredentialsProvider() {
+        return DefaultCredentialsProvider.create();
+    }
+
+    @Bean
+    public AwsRegionProvider awsRegionProvider() {
+        return DefaultAwsRegionProviderChain.builder().build();
+    }
+
+    @Bean("s3Client")
```
galovics, line 50: No need to specify the bean name, it'll automatically get the method name.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/s3/AmazonS3Config.java
```diff
@@ -0,0 +1,75 @@
...
+    public S3Client s3Client(DefaultCredentialsProvider awsCredentialsProvider, AwsRegionProvider awsRegionProvider) {
+        return S3Client.builder().credentialsProvider(awsCredentialsProvider).region(awsRegionProvider.getRegion()).build();
+    }
+
+    @Bean("s3Client")
+    @Profile("test")
+    @Conditional(AmazonS3ConfigCondition.class)
+    public S3Client s3ClientTest(DefaultCredentialsProvider awsCredentialsProvider, AwsRegionProvider awsRegionProvider) {
+        S3ClientBuilder builder = S3Client.builder();
+        String env = System.getenv().getOrDefault("AWS_ENDPOINT_URL", "");
```
galovics: You can use the Environment class from Spring.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/s3/AmazonS3Config.java
```diff
@@ -0,0 +1,75 @@
...
+        if (StringUtil.isNotBlank(env)) {
+            builder.endpointOverride(URI.create(env)).forcePathStyle(true);
+        }
+        return builder.credentialsProvider(awsCredentialsProvider).region(awsRegionProvider.getRegion()).build();
+    }
+
+    @Bean
+    @ConditionalOnBean(S3Client.class)
+    public S3DatatableReportExportServiceImpl s3DatatableReportExportServiceImpl(ReadReportingService reportServiceImpl,
```
galovics, line 60: I don't think this bean definition should be here. Btw why don't we remove the definition completely and use the Component annotation on the class itelf?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/s3/AmazonS3Config.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @Bean("s3Client")
+    @Profile("!test")
+    @Conditional(AmazonS3ConfigCondition.class)
+    public S3Client s3Client(DefaultCredentialsProvider awsCredentialsProvider, AwsRegionProvider awsRegionProvider) {
+        return S3Client.builder().credentialsProvider(awsCredentialsProvider).region(awsRegionProvider.getRegion()).build();
+    }
+
+    @Bean("s3Client")
```
galovics: Btw I'd approach this differently a bit. I'm not saying it's not going to work but would be cleaner.

Let's assume we have this config:
```
    @Bean
    @Conditional(AmazonS3ConfigCondition.class)
    public S3Client s3Client(DefaultCredentialsProvider awsCredentialsProvider, AwsRegionProvider awsRegionProvider) {
        return S3Client.builder().credentialsProvider(awsCredentialsProvider).region(awsRegionProvider.getRegion()).build();
    }
```
Let's create a custom interface:
```
interface S3ClientCustomizer {
    void customize(S3ClientBuilder builder);
}
```
Then let's change the bean config:
```
    @Bean
    @Conditional(AmazonS3ConfigCondition.class)
    public S3Client s3Client(DefaultCredentialsProvider awsCredentialsProvider, AwsRegionProvider awsRegionProvider, List<S3ClientCustomizer> customizers) {
 
 
 
 S3ClientBuilder builder = S3Client.builder().credentialsProvider(awsCredentialsProvider).region(awsRegionProvider.getRegion());
        customizers.forEach(c -> c.customize(builder));
        return builder.build();
    }
```
And then we create a customizer for the tests:
```
@Component
@RequiredArgsConstructor
@Profile("test")
public class LocalstackS3ClientCustomizer implements S3ClientCustomizer {
    private final Environment environment;

    @Override
    public void customize(S3ClientBuilder builder) {
        String env = environment.....
        if (StringUtil.isNotBlank(env)) {
            builder.endpointOverride(URI.create(env)).forcePathStyle(true);
        }
    }
}
```

fineract-provider/src/main/java/org/apache/fineract/infrastructure/s3/AmazonS3ConfigCondition.java
```diff
@@ -0,0 +1,53 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Condition;
+import org.springframework.context.annotation.ConditionContext;
+import org.springframework.core.type.AnnotatedTypeMetadata;
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
+import software.amazon.awssdk.regions.providers.AwsRegionProvider;
+
+public class AmazonS3ConfigCondition implements Condition {
+
+    @Value("${fineract.report.export.s3.enabled}")
```
galovics: FineractProperties?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/s3/AmazonS3ConfigCondition.java
```diff
@@ -0,0 +1,53 @@
...
+
+import org.apache.commons.lang3.StringUtils;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Condition;
+import org.springframework.context.annotation.ConditionContext;
+import org.springframework.core.type.AnnotatedTypeMetadata;
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
+import software.amazon.awssdk.regions.providers.AwsRegionProvider;
+
+public class AmazonS3ConfigCondition implements Condition {
```
galovics: You can use PropertiesCondition.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/s3/AmazonS3ConfigCondition.java
```diff
@@ -0,0 +1,53 @@
...
+    @Override
+    public boolean matches(ConditionContext context, AnnotatedTypeMetadata metadata) {
+        return context.getEnvironment().getRequiredProperty("fineract.report.export.s3.enabled", Boolean.class)
+                && StringUtils.isNotBlank(context.getEnvironment().getProperty("fineract.report.export.s3.bucket", ""))
+                && isAwsCredentialValid(context);
+    }
+
+    private boolean isAwsCredentialValid(ConditionContext context) {
+        try {
+            context.getBeanFactory().getBean(AwsCredentialsProvider.class).resolveCredentials();
```
galovics: This is not necessarily the best idea. You're accessing the container before it even gets properly initialized. 
It's better if you explicitly instantiate the default credentials provider and region provider to test this.

https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/annotation/Condition.html

> Conditions must follow the same restrictions as [BeanFactoryPostProcessor](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/factory/config/BeanFactoryPostProcessor.html) and take care to never interact with bean instances.

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/ReportExportTest.java
```diff
@@ -0,0 +1,63 @@
...
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+import retrofit2.Response;
+
+/**
+ * Integration Test for /runreports/ API.
+ *
+ * @author Michael Vorburger.ch
+ */
+@Tag("github_only")
```
galovics: Do we need this tag?

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/ReportExportTest.java
```diff
@@ -0,0 +1,63 @@
...
+    void runClientListingTableReportCSV() throws IOException {
+        Response<ResponseBody> result = okR(
+                fineract().reportsRun.runReportGetFile("Client Listing", Map.of("R_officeId", "1", "exportCSV", "true"), false));
+        assertThat(result.body().contentType()).isEqualTo(MediaType.parse("text/csv"));
+        assertThat(result.body().string()).contains("Office/Branch");
+    }
+
+    @Test
+    @GithubOnly
+    @Disabled
```
galovics: Can't we put the `@Disabled` annotation into the GitHubOnly annotation?
This is kinda confusing to me that it's disabled but it's actually not. Also, if you move this to the GitHubOnly annotation, let's put a comment there and explain why.

integration-tests/src/test/java/org/apache/fineract/integrationtests/GithubOnly.java
```diff
@@ -0,0 +1,30 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariable;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.TYPE, ElementType.METHOD })
+@EnabledIfEnvironmentVariable(named = "GITHUB_ACTIONS", matches = "true")
+public @interface GithubOnly {}
```
galovics: Let's rename this to CIOnly.

galovics (changes requested):

.github/workflows/build-mariadb.yml
b0c1: It's come from localstack official documentation. It's using the docker but doesn't pull the image.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/s3/AmazonS3Config.java
b0c1: It should. Initialization moved to a different Configuration but `S3Client` must be available (or not) in the BeanFactory, because of this `@Component` will not work with `@ConditionOnBean` annotation

fineract-provider/src/main/java/org/apache/fineract/infrastructure/s3/AmazonS3ConfigCondition.java
b0c1: It provides a complex condition to enable S3 configuration

galovics:
> @galovics @b0c1 ... is there any reason why we use the database (global configuration service) to configure these things? Why wouldn't we put this in application.properties?

@vidakovic we wanna be able to change the configuration value runtime.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/s3/AmazonS3Config.java
```diff
@@ -0,0 +1,64 @@
...
+    public S3Client s3Client(DefaultCredentialsProvider awsCredentialsProvider, AwsRegionProvider awsRegionProvider,
+            List<S3ClientCustomizer> customizers) {
+        S3ClientBuilder builder = S3Client.builder().credentialsProvider(awsCredentialsProvider).region(awsRegionProvider.getRegion());
+        customizers.forEach(customizer -> customizer.customize(builder));
+        return builder.build();
+    }
+
+    @Bean
+    @ConditionalOnBean(S3Client.class)
+    public S3DatatableReportExportServiceImpl s3DatatableReportExportServiceImpl(ReadReportingService reportServiceImpl,
```
galovics, line 60: I think you forgot to remove this bean since it's defined in AmazonS3ExportConfiguration.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/s3/LocalstackS3ClientCustomizer.java
```diff
@@ -0,0 +1,43 @@
...
+import org.apache.poi.util.StringUtil;
+import org.springframework.context.annotation.Profile;
+import org.springframework.core.env.Environment;
+import org.springframework.stereotype.Component;
+import software.amazon.awssdk.services.s3.S3ClientBuilder;
+
+@Component
+@RequiredArgsConstructor
+@Profile("test")
+public class LocalstackS3ClientCustomizer implements S3ClientCustomizer {
```
galovics, line 32: Splendid ;)

galovics (changes requested):

fineract-provider/src/main/java/org/apache/fineract/infrastructure/s3/LocalstackS3ClientCustomizer.java
b0c1: ❤️

fineract-provider/src/main/java/org/apache/fineract/infrastructure/s3/AmazonS3ConfigCondition.java
galovics: The point of using PropertiesCondition is that it's gonna be a compilation error if anything changes while in case you rely on the property strings, you don't get any compilation-time checks.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/s3/AmazonS3ConfigCondition.java
```diff
@@ -0,0 +1,51 @@
...
+import software.amazon.awssdk.regions.providers.DefaultAwsRegionProviderChain;
+
+@Slf4j
+public class AmazonS3ConfigCondition implements Condition {
+
+    @Override
+    public boolean matches(ConditionContext context, AnnotatedTypeMetadata metadata) {
+        Boolean s3ReportEnabled = context.getEnvironment().getRequiredProperty("fineract.report.export.s3.enabled", Boolean.class);
+        boolean s3BucketIsSet = StringUtils.isNotBlank(context.getEnvironment().getProperty("fineract.report.export.s3.bucket", ""));
+        log.info("AWS S3 Enabled: {}", s3ReportEnabled);
```
galovics: Let's either remove this log message or also log out that the credentials are valid. I think this'll send the wrong message that S3 is enabled even in case the credentials are invalid.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/s3/AmazonS3ConfigCondition.java
```diff
@@ -0,0 +1,51 @@
...
+        return s3ReportEnabled && s3BucketIsSet && isAwsCredentialValid();
+    }
+
+    private boolean isAwsCredentialValid() {
+        try {
+            DefaultCredentialsProvider.create().resolveCredentials();
+            DefaultAwsRegionProviderChain.builder().build().getRegion();
+            return true;
+        } catch (Exception e) {
+            log.warn("AWS credential is not valid", e);
```
galovics: So I think this is not necessarily the best idea either since:
1. the logic is clear, in case its invalid, we'll return false
2. we're saying the credentials are invalid yet in case of incorrectly set region, we also return false and claim that the creds were invalid.

galovics (changes requested):

vidakovic:
> > @galovics @b0c1 ... is there any reason why we use the database (global configuration service) to configure these things? Why wouldn't we put this in application.properties?
> 
> @vidakovic we wanna be able to change the configuration value runtime.

... that is a solved problem in Spring/Boot (search: "reloadable configuration")... I'd say not a strong argument to bury it in the database...

galovics:
> > > @galovics @b0c1 ... is there any reason why we use the database (global configuration service) to configure these things? Why wouldn't we put this in application.properties?
> > 
> > 
> > @vidakovic we wanna be able to change the configuration value runtime.
> 
> ... that is a solved problem in Spring/Boot (search: "reloadable configuration")... I'd say not a strong argument to bury it in the database...

I'm not saying you're wrong but if I remember correctly reloading configuration is only supported with Spring Cloud and done by annotating classes with `@RefreshScope`. Now, upon refresh, you'll have the choice whether to just "refresh" the application context, reload or shutdown-reload it. 

I've used it with a cloud project but I think fineract in its current form is not suitable for this and could cause issues with context refreshes.

**TL;DR;** I think going with refreshable properties with Spring's approach is way too premature for Fineract.

galovics (approved):

PR #2932, opened 2023-01-26
FINERACT-1870: Run reports fix for 1.8.x

vidakovic:
Run report fix for 1.8.x

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/SQLInjectionValidator.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/SQLInjectionValidator.java
@@ -18,6 +18,7 @@
  */
 package org.apache.fineract.infrastructure.security.utils;
 
+import java.util.List;
 import java.util.StringTokenizer;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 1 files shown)


fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/SQLInjectionValidator.java
```diff
@@ -37,23 +38,32 @@ private SQLInjectionValidator() {
...
+    // https://larrysteinle.com/2011/02/20/use-regular-expressions-to-detect-sql-code-injection more examples
+    private static final List<String> INJECTION_PATTERNS = List.of("(?i).*[or|and]\s*[\"']?-1[\"']?\\s*(-*).*",
+            "(?i).*\\s+[\"']?(\\d+)[\"']?\\s*=\\s*[\"']?(\\1)[\"']?\\s*(-*).*");
+
     public static void validateSQLInput(final String sqlSearch) {
         if (StringUtils.isBlank(sqlSearch)) {
             return;
         }
+
+        // TODO: this should be replaced by INJECTION_PATTERNS
```
adamsaghy, line 51: Is the TODO for future?
Should we replace it now?

vidakovic: Well... maybe I should remove it here. This a maintenance release, so changes should be as minimal as possible. But for the upcoming upstream develop PR (not there yet) we could discuss this a bit more. There are definitely more cases to cover.

Here I am just trying to do the necessary to fix the issue. So, if you are ok with that then let's move the discussion to upstream develop and leave things as they are here. If you want I can remove the TODOs.

adamsaghy (approved):
LGTM

adamsaghy:
would you mind to write a test to cover the fixed behaviour?

PR #3011, opened 2023-02-27
FINERACT-1831 - Encrypted tenant passwords

b0c1:
FINERACT-1831 - Encrypted tenant passwords
- [x] Encrypt tenant passwords
- [x] Update integration tests
- [x] Check the application database using encrypted passwords
- [x] One-time migration for the existing password

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/database/DataSourcePerTenantServiceFactory.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/database/DataSourcePerTenantServiceFactory.java
@@ -24,9 +24,9 @@ import static org.apache.fineract.infrastructure.core.domain.FineractPlatformTen
 import com.zaxxer.hikari.HikariConfig;
 import javax.sql.DataSource;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.fineract.infrastructure.core.config.FineractProperties;
 import org.apache.fineract.infrastructure.core.domain.FineractPlatformTenantConnection;
-import org.apache.fineract.infrastructure.security.constants.TenantConstants;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.context.ApplicationContext;
 import org.springframework.stereotype.Component;
@@ -47,16 +47,24 @@ public class DataSourcePerTenantServiceFactory {
     private final DataSource tenantDataSource;
     private final HikariDataSourceFactory hikariDataSourceFactory;
 
+    private final DatabasePasswordEncryptor databasePasswordEncryptor;
+
     public DataSourcePerTenantServiceFactory(@Qualifier("hikariTenantDataSource") DataSource tenantDataSource, HikariConfig hikariConfig,
-            FineractProperties fineractProperties, ApplicationContext context, HikariDataSourceFactory hikariDataSourceFactory) {
+            FineractProperties fineractProperties, ApplicationContext context, HikariDataSourceFactory hikariDataSourceFactory,
+            DatabasePasswordEncryptor databasePasswordEncryptor) {
         this.hikariConfig = hikariConfig;
         this.fineractProperties = fineractProperties;
         this.context = context;
         this.tenantDataSource = tenantDataSource;
         this.hikariDataSourceFactory = hikariDataSourceFactory;
+        this.databasePasswordEncryptor = databasePasswordEncryptor;
     }
 
     public DataSource createNewDataSourceFor(final FineractPlatformTenantConnection tenantConnection) {
+        if (!databasePasswordEncryptor.isMasterPasswordHashValid(tenantConnection.getMasterPasswordHash())) {
+            throw new IllegalArgumentException(
+                    "Invalid master password on tenant connection %d.".formatted(tenantConnection.getConnectionId()));
+        }
         String protocol = toProtocol(tenantDataSource);
         // Default properties for Writing
         String schemaServer = tenantConnection.getSchemaServer();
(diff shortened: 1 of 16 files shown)


fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/database/DataSourcePerTenantServiceFactory.java
```diff
@@ -67,18 +68,13 @@ public DataSource createNewDataSourceFor(final FineractPlatformTenantConnection
...
-                    schemaServer);
-            schemaPort = getPropertyValue(tenantConnection.getReadOnlySchemaServerPort(), TenantConstants.PROPERTY_RO_SCHEMA_SERVER_PORT,
-                    schemaPort);
-            schemaName = getPropertyValue(tenantConnection.getReadOnlySchemaName(), TenantConstants.PROPERTY_RO_SCHEMA_SCHEMA_NAME,
-                    schemaName);
-            schemaUsername = getPropertyValue(tenantConnection.getReadOnlySchemaUsername(), TenantConstants.PROPERTY_RO_SCHEMA_USERNAME,
-                    schemaUsername);
-            schemaPassword = getPropertyValue(tenantConnection.getReadOnlySchemaPassword(), TenantConstants.PROPERTY_RO_SCHEMA_PASSWORD,
-                    schemaPassword);
+            schemaServer = getPropertyValue(tenantConnection.getReadOnlySchemaServer(), schemaServer);
```
galovics: I'd rather go with org.apache.commons.lang3.StringUtils#defaultIfBlank.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/migration/TenantPasswordEncryptionTask.java
```diff
@@ -0,0 +1,90 @@
...
+import liquibase.resource.ResourceAccessor;
+import org.apache.fineract.infrastructure.security.utils.EncryptionUtil;
+
+public class TenantPasswordEncryptionTask implements CustomTaskChange {
+
+    private final String encryptionType;
+    private final String masterKey;
+
+    public TenantPasswordEncryptionTask() {
+        this.encryptionType = Optional.ofNullable(System.getenv("FINERACT_DEFAULT_TENANTDB_ENCRYPTION"))
```
galovics: Unfortunately this is not enough. The properties can be overridden directly via program arguments for example so this is not going to be in an environment variable necessarily.
I know this object will be managed by Liquibase so no Spring DI at this point but you can do the ugly way to do it, look at TemporaryConfigurationServiceContainer

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/migration/TenantPasswordEncryptionTask.java
```diff
@@ -0,0 +1,90 @@
...
+                .orElse(EncryptionUtil.DEFAULT_CIPHER);
+        this.masterKey = Optional.ofNullable(System.getenv("FINERACT_DEFAULT_TENANTDB_KEY")).orElse(EncryptionUtil.DEFAULT_MASTER_PASSWORD);
+    }
+
+    @Override
+    public void execute(Database database) throws CustomChangeException {
+        var dbConn = (JdbcConnection) database.getConnection(); // autocommit is false
```
galovics: No var pls. Thanks ;-)

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/migration/TenantPasswordEncryptionTask.java
```diff
@@ -0,0 +1,90 @@
...
+            Statement selectStatement = dbConn.createStatement();
+            Statement updateStatement = dbConn.createStatement();
+
+            ResultSet rs = selectStatement.executeQuery("SELECT id, schema_password FROM tenant_server_connections");
+            while (rs.next()) {
+                String id = rs.getString("id");
+                String schemaPassword = rs.getString("schema_password");
+                String encryptedPassword = EncryptionUtil.encrypt(encryptionType, masterKey, schemaPassword);
+
+                var updateSql = String.format("update tenant_server_connections set schema_password = '%s' where id = %s",
```
galovics: Same here.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/migration/TenantPasswordEncryptionTask.java
```diff
@@ -0,0 +1,90 @@
...
+                        encryptedPassword, id);
+                updateStatement.execute(updateSql);
+            }
+
+            selectStatement.close();
```
galovics: Resource leaking potentially. You can use try with resources instead.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/EncryptionUtil.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.Base64;
+import java.util.Optional;
+import javax.crypto.Cipher;
+import javax.crypto.SecretKeyFactory;
+import javax.crypto.spec.IvParameterSpec;
+import javax.crypto.spec.PBEKeySpec;
+import javax.crypto.spec.SecretKeySpec;
+import org.jetbrains.annotations.NotNull;
+
+public final class EncryptionUtil {
```
galovics, line 34: Can we make this class a Spring bean instead so it can be easily mocked out?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/EncryptionUtil.java
```diff
@@ -0,0 +1,104 @@
...
+
+    public static final String DEFAULT_CIPHER = "AES/CBC/PKCS5Padding";
+
+    public static final String DEFAULT_MASTER_PASSWORD = "fineract";
+
+    private EncryptionUtil() {
+    }
+
+    public static String encrypt(String cipherType, String masterPassword, String data) {
```
galovics: Do we need this level of abstraction to be able to provide the cipherType and masterPassword? If it's a Spring bean you can easily take these from the environment and there'll be no need to pass them.

galovics (changes requested):

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/database/DataSourcePerTenantServiceFactory.java
```diff
@@ -47,16 +47,23 @@ public class DataSourcePerTenantServiceFactory {
...
         this.fineractProperties = fineractProperties;
         this.context = context;
         this.tenantDataSource = tenantDataSource;
         this.hikariDataSourceFactory = hikariDataSourceFactory;
+        this.databasePasswordEncryptor = databasePasswordEncryptor;
     }
 
     public DataSource createNewDataSourceFor(final FineractPlatformTenantConnection tenantConnection) {
+        if (!databasePasswordEncryptor.isPasswordValid(tenantConnection.getMasterPasswordHash())) {
+            throw new IllegalArgumentException("Invalid master password");
```
galovics: Can we include the tenant ID in the message? If somebody encounters this error, they won't know which tenant has the wrong password.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/database/DatabasePasswordEncryptor.java
```diff
@@ -0,0 +1,73 @@
...
+
+    public String getMasterPasswordHash() {
+        String masterPassword = Optional.ofNullable(fineractProperties) //
+                .map(FineractProperties::getTenant) //
+                .map(FineractProperties.FineractTenantProperties::getMasterPassword) //
+                .orElse(DEFAULT_MASTER_PASSWORD);
+        return BCrypt.hashpw(masterPassword.getBytes(StandardCharsets.UTF_8), BCrypt.gensalt());
+    }
+
+    public boolean isPasswordValid(String hashed) {
```
galovics: I'd call this isMasterPasswordHashValid

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/migration/TenantDatabaseUpgradeService.java
```diff
@@ -59,6 +60,8 @@ public class TenantDatabaseUpgradeService implements InitializingBean {
     private final TenantDataSourceFactory tenantDataSourceFactory;
     private final Environment environment;
 
+    private final List<CustomTaskChange> customTaskChangesForDependencyInjection;
```
galovics, line 64: Is this needed to ensure the ordering of bean creation? If so, can you make a comment here on that so nobody will delete it?

galovics (changes requested):

galovics (approved):

PR #3015, opened 2023-03-01
[FINERACT-1904] loanChargePaidByList fix

taskain7:
## Description

Fixing loanChargePaidByList in GET loan API and LoanTransactionMakeRepaymentPostBusinessEvent event

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanReadPlatformServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanReadPlatformServiceImpl.java
@@ -173,6 +173,7 @@ public class LoanReadPlatformServiceImpl implements LoanReadPlatformService {
     private final LoanTransactionRepository loanTransactionRepository;
     private final LoanTransactionRelationRepository loanTransactionRelationRepository;
     private final LoanTransactionRelationMapper loanTransactionRelationMapper;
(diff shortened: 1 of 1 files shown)


fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanAccountDomainServiceJpa.java
```diff
@@ -198,7 +199,8 @@ public LoanTransaction makeRepayment(final LoanTransactionType repaymentTransact
                 defaultLoanLifecycleStateMachine, existingTransactionIds, existingReversedTransactionIds, isRecoveryRepayment,
                 scheduleGeneratorDTO, isHolidayValidationDone);
 
-        saveLoanTransactionWithDataIntegrityViolationChecks(newRepaymentTransaction);
+        newRepaymentTransaction = saveLoanTransactionWithDataIntegrityViolationChecks(newRepaymentTransaction);
+        newRepaymentTransaction.setLoanChargesPaid(loanChargePaidByRepository.findByLoanTransaction(newRepaymentTransaction));
```
adamsaghy: I dont think we need this. The chargePaidBy associations are already assigned to the loanTransaction

taskain7: you're right. it's removed

galovics (approved):

PR #3018, opened 2023-03-01
[FINERACT-1678] Exception fix if every loan fail during COB

taskain7:
## Description

Fix for Exception when every loan fail durin Loan COB job.


adamsaghy (approved):
LGTM

PR #3026, opened 2023-03-07
FINERACT-1724 - COB Optimistic locking fix

b0c1:
FINERACT-1724 - Optimistic locking fix
- [x] ReloadService for lead entities
- [x] ReloaderService iterating over ReloadService and try to find and reload the entity
- [x] Fix COBBusinessStepServiceImpl to reload the entity before the business step

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/cob/service/ReloadService.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/cob/service/ReloadService.java
@@ -1,0 +1,28 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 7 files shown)


fineract-provider/src/main/java/org/apache/fineract/cob/service/ReloadService.java
```diff
@@ -0,0 +1,28 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.cob.service;
+
+import org.apache.fineract.infrastructure.core.domain.AbstractPersistableCustom;
+
+public interface ReloadService<S extends AbstractPersistableCustom> {
+
+    <X extends AbstractPersistableCustom> boolean canReload(X input);
```
galovics, line 25: Can't this be type of S instead of X?

galovics (approved):

PR #3043, opened 2023-03-13
[FINERACT-1724] use partitions for queries with big parameter list

taskain7:
## Description

Fix for the PostgreSQL parameter size limitation with list partitioning

### DIFF
--- a/fineract-provider/src/main/resources/application.properties
+++ b/fineract-provider/src/main/resources/application.properties
@@ -42,6 +42,8 @@ fineract.mode.write-enabled=${FINERACT_MODE_WRITE_ENABLED:true}
 fineract.mode.batch-worker-enabled=${FINERACT_MODE_BATCH_WORKER_ENABLED:true}
 fineract.mode.batch-manager-enabled=${FINERACT_MODE_BATCH_MANAGER_ENABLED:true}
 
+fineract.query.in-clause-parameter-size-limit=${FINERACT_QUERY_PARAMETER_SIZE:1000}
+
 fineract.correlation.enabled=${FINERACT_LOGGING_HTTP_CORRELATION_ID_ENABLED:false}
 fineract.correlation.header-name=${FINERACT_LOGGING_HTTP_CORRELATION_ID_HEADER_NAME:X-Correlation-ID}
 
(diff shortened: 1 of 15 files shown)


adamsaghy:
@taskain7 Kindly asking you to check the failing tests!

fineract-provider/src/main/resources/application.properties
```diff
@@ -42,6 +42,8 @@ fineract.mode.write-enabled=${FINERACT_MODE_WRITE_ENABLED:true}
 fineract.mode.batch-worker-enabled=${FINERACT_MODE_BATCH_WORKER_ENABLED:true}
 fineract.mode.batch-manager-enabled=${FINERACT_MODE_BATCH_MANAGER_ENABLED:true}
 
+fineract.query.parameter-size=${FINERACT_QUERY_PARAMETER_SIZE:65000}
```
galovics: 2 things, I'd name this specifically that this is the maximum amount of parameters within an IN clause, emphasize that please.
Second thing, I'd be way more careful with the default number here and go with like 10-20k but definitely not 65k.

The thing is, DBs are often really bad at executing queries with really big IN clauses.

galovics (dismissed):

fineract-provider/src/main/resources/application.properties
adamsaghy: As a default i would go with 1000 maximum.

taskain7: changed it

adamsaghy (dismissed):
LGTM

adamsaghy:
@taskain7 Would you mind to add one more test which explicitly checks whether the partitioning is working?

galovics (approved):

IOhacker (commented):
@taskain7 parameter limit come from the PG JDBC Driver https://github.com/pgjdbc/pgjdbc/blob/master/pgjdbc/src/main/java/org/postgresql/core/PGStream.java (method  sendInteger2(int val) where the upper limit is set to 65535) how this fix prevent not to reach/manual set this limit?

taskain7:
@IOhacker 
this setting won't partition the query to separate queries that each have 65K parameter, it just returns with `PSQLException: PreparedStatement can have at most 65,535 parameters`. So we have to manually partition the query parameters to smaller chunks to make the queries executable. And also, we should respect other DBs as well.

PR #3053, opened 2023-03-16
[FINERACT-1678] Inline loan COB api restriction

taskain7:
## Description

Limitate the size of the list in the request body on the Inline Loan COB API


fineract-provider/src/main/java/org/apache/fineract/cob/service/InlineLoanCOBExecutorServiceImpl.java
```diff
@@ -78,6 +79,7 @@
 public class InlineLoanCOBExecutorServiceImpl implements InlineExecutorService<Long> {
 
     private static final String JOB_EXECUTION_FAILED_MESSAGE = "Job execution failed for job with name: ";
+    private static final Integer INLINE_COB_REQUEST_ITEM_LIMIT = 1000;
```
adamsaghy, line 82: Should it be configurable in application.properties?
We might wanna use something similar like `fineract.query.in-clause-parameter-size-limit`?

adamsaghy: I think it would be better to be moved this into application.properties.

taskain7: fixed

adamsaghy (approved):
LGTM

PR #3054, opened 2023-03-16
FINERACT-1724 Read only tenant password encryption migration

b0c1:
FINERACT-1724 Read only tenant password encryption migration
- [x] ReadOnly tenant password encryption migration script
- [x] Documentation
- [x] Encryption password generator

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/database/DatabasePasswordEncryptor.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/database/DatabasePasswordEncryptor.java
@@ -19,6 +19,7 @@
 package org.apache.fineract.infrastructure.core.service.database;
 
 import java.nio.charset.StandardCharsets;
+import java.text.MessageFormat;
 import java.util.Optional;
 import lombok.RequiredArgsConstructor;
 import org.apache.fineract.infrastructure.core.config.FineractProperties;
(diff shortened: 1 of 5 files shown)


galovics (approved):

adamsaghy:
@b0c1 Please fix the checkstyle issues!

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/database/DatabasePasswordEncryptor.java
```diff
@@ -37,6 +38,20 @@ public class DatabasePasswordEncryptor implements PasswordEncryptor {
 
     private final FineractProperties fineractProperties;
 
+    //CHECKSTYLE:OFF
+    public static void main(String[] args) {
+        if (args.length < 2) {
+            System.out.println(
```
adamsaghy, line 44: Would it be better to use a logger here instead of System.out.println?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/database/DatabasePasswordEncryptor.java
```diff
@@ -37,6 +38,20 @@ public class DatabasePasswordEncryptor implements PasswordEncryptor {
 
     private final FineractProperties fineractProperties;
 
+    //CHECKSTYLE.OFF: IllegalCatch - Turn off checkstyle for main method
```
adamsaghy: @b0c1 You might wanna use this `@SuppressWarnings("checkstyle:regexpsinglelinejava")` ;)
and `SuppressionCommentFilter` is not needed

galovics (approved):

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/database/DatabasePasswordEncryptor.java
b0c1: Thanks.

PR #3056, opened 2023-03-17
[FINERACT-1678] Loan COB with 0 loans

taskain7:
## Description

Set all taskelts to `COMPLETED` status when there is no loan to be processed by the Loan COB


vidakovic (approved):
LGTM

galovics (approved):

PR #3098, opened 2023-04-03
Update fineract image source in docker-compose.yml to pull the correct `fineract` image artifact 

elijah0kello:
## Description

I updated the line in the docker-compose.yml that is supposed to reference the docker image for `fineract` from `fineract` to `apache/fineract`.

### DIFF
--- a/docker-compose-postgresql.yml
+++ b/docker-compose-postgresql.yml
@@ -39,7 +39,7 @@ services:
     ports:
       - "5432:5432"
   fineract-server:
-    image: fineract:latest
+    image: apache/fineract:latest
     volumes:
       - ./fineract-provider/build/data:/data
     healthcheck:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -35,7 +35,7 @@ services:
     ports:
       - "3306:3306"
   fineract-server:
-    image: fineract:latest
+    image: apache/fineract:latest
     volumes:
       - ./fineract-provider/build/data:/data
     healthcheck:


IOhacker:
@elijah0kello @vidakovic This PR is pointing to the right Docker Hub repository, but the latest push has not been updated over a year. https://hub.docker.com/r/apache/fineract/tags 

Github Actions must be pushing to the Apache Fineract Docker Hub repository before merging this PR to avoid any confusion about the "latest" tag.

elijah0kello:
> @elijah0kello @vidakovic This PR is pointing to the right Docker Hub repository, but the latest push has not been updated over a year. https://hub.docker.com/r/apache/fineract/tags
> 
> Github Actions must be pushing to the Apache Fineract Docker Hub repository before merging this PR to avoid any confusion about the "latest" tag.

Let me update the latest push in Github Actions to push to the right Apache Fineract Docker Hub repository. Hadn't considered that. Thank you

vidakovic:
@elijah0kello note that the Docker images for Fineract are published via Mifo's org on Docker Hub... this is happening in a downstream fork in a separate Github Workflow. Simple reason for not doing this here: we had issues with the credentials.

elijah0kello:
> @elijah0kello note that the Docker images for Fineract are published via Mifo's org on Docker Hub... this is happening in a downstream fork in a separate Github Workflow. Simple reason for not doing this here: we had issues with the credentials.

That explains why I can't find the point where the image is pushed to Docker Hub. So does that mean this PR can't be merged? If yes, is it okay to leave the things the way they are? Also kindly,  what do you advise that I should do. Note that I found the same issue in `docker-compose-postgresql.yml` and I have updated it.

PR #3134, opened 2023-04-19
[FINERACT-1922] Skip business date job when business date is not enabled

taskain7:
## Description

Increase Business Date and Increase COB date jobs will be skipped when business date is not enabled.


adamsaghy:
@taskain7 Please add some test cases as well to cover this new behaviour!

adamsaghy (approved):
LGTM

PR #3155, opened 2023-05-05
FINERACT-1724 Batch API HotFix

b0c1:
- [x] Fix Batch API internal error


mohitsinha (changes requested):
Can we have IT for this

galovics:
@mohitsinha I don't think this is testable with an IT test cause we're testing a scenario where something goes to an error, especially this is true if we want to simulate an HTTP 500.

I'm merging this but let's think about the ways to cover this.

galovics (approved):

PR #3158, opened 2023-05-09
FINERACT-1912 Savings Account Transactions query -Advanced

Mk9894:
## Note

**Please refer https://github.com/apache/fineract/pull/3192 for the changes. This PR will be closed**

## Description

## Checklist


fineract-provider/src/main/resources/static/legacy-docs/apiLive.htm
```diff
@@ -29369,6 +29369,242 @@ <h2>Retrieve Savings Account Transaction Template:</h2>
 	    </div>
 	</div>
 
+	<a id="savingsaccounts_transactions_list" name="savingsaccounts_transactions_list" class="old-syle-anchor">&nbsp;</a>
```
galovics, line 29372: This is not used anymore.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/SavingsAccountTransactionsApiResource.java
```diff
@@ -114,6 +121,53 @@ public String retrieveTemplate(@PathParam("savingsId") final Long savingsId,
...
+            @QueryParam("toDate") @Parameter(description = "toDate") final DateParam toDateParam,
+            @QueryParam("fromAmount") @Parameter(description = "fromAmount") @DecimalMin(value = "0", message = "must be greater than or equal to 0") final BigDecimal fromAmount,
+            @QueryParam("toAmount") @Parameter(description = "toAmount") @DecimalMin(value = "0", message = "must be greater than or equal to 0") final BigDecimal toAmount,
+            @QueryParam("locale") @Parameter(description = "locale") final String locale,
+            @QueryParam("transactionType") @Parameter(description = "transcationType") final String transactionType) {
+
+        this.context.authenticatedUser().validateHasReadPermission(SavingsApiConstants.SAVINGS_ACCOUNT_RESOURCE_NAME);
+        final SearchParameters searchParameters = SearchParameters.forPagination(offset, limit, orderBy, sortOrder);
+
+        LocalDate fromDate = null;
```
galovics, line 134: Why don't we use an argument resolver for this instead? That way we wouldn't have to duplicate this logic anywhere but just go with a plain LocalDate in the parameter list.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountReadPlatformService.java
```diff
@@ -45,6 +46,10 @@ public interface SavingsAccountReadPlatformService {
 
     Collection<SavingsAccountTransactionData> retrieveAllTransactions(Long savingsId, DepositAccountType depositAccountType);
 
+    Page<SavingsAccountTransactionData> retrieveAllTransactions(Long savingsId, DepositAccountType depositAccountType,
```
galovics, line 48: I don't like this signature. Why don't we create a parameter object which has all the filterable attributes instead of passing them one by one.
That way we could reduce and keep the amount of parameters at the number of 3. savingsId, accountType, searchParams

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountReadPlatformServiceImpl.java
```diff
@@ -1267,6 +1270,99 @@ public Collection<SavingsAccountTransactionData> retrieveAllTransactions(final L
         return this.jdbcTemplate.query(sql, this.transactionsMapper, new Object[] { savingsId, depositAccountType.getValue() }); // NOSONAR
     }
 
+    @Override
+    public Page<SavingsAccountTransactionData> retrieveAllTransactions(final Long savingsId, DepositAccountType depositAccountType,
+            SearchParameters searchParameters, LocalDate fromDate, LocalDate toDate,
+            @DecimalMin(value = "0", message = "must be greater than or equal to 0") BigDecimal fromAmount,
```
galovics: What's going to evaluate this validation annotation?

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountReadPlatformServiceImpl.java
```diff
@@ -1267,6 +1270,99 @@ public Collection<SavingsAccountTransactionData> retrieveAllTransactions(final L
...
+            @DecimalMin(value = "0", message = "must be greater than or equal to 0") BigDecimal toAmount, String transactionType) {
+
+        final StringBuilder sqlBuilder = new StringBuilder(200);
```
galovics: This is a super red flag, no native queries please unless you absolutely need them. Why don't you simply use a JPQL query with a projection for this?

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountReadPlatformServiceImpl.java
```diff
@@ -1267,6 +1270,99 @@ public Collection<SavingsAccountTransactionData> retrieveAllTransactions(final L
...
+        sqlBuilder.append(this.transactionsMapper.schema());
+        sqlBuilder.append(" where sa.id = ? and sa.deposit_type_enum = ? ");
+        final Object[] objectArray = new Object[12];
+        objectArray[0] = savingsId;
+        objectArray[1] = depositAccountType.getValue();
+        int arrayPos = 2;
+
+        if (StringUtils.isNotEmpty(transactionType)) {
+            Integer transcationTypeEnum = null;
+            if ("deposit".equals(transactionType.toLowerCase())) {
```
galovics: magic string

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountReadPlatformServiceImpl.java
```diff
@@ -1267,6 +1270,99 @@ public Collection<SavingsAccountTransactionData> retrieveAllTransactions(final L
...
+                transcationTypeEnum = SavingsAccountTransactionType.DEPOSIT.getValue();
+            } else if ("withdrawal".equals(transactionType.toLowerCase())) {
```
galovics: magic string

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountReadPlatformServiceImpl.java
```diff
@@ -1267,6 +1270,99 @@ public Collection<SavingsAccountTransactionData> retrieveAllTransactions(final L
...
+    public Page<SavingsAccountTransactionData> retrieveAllTransactions(final Long savingsId, DepositAccountType depositAccountType,
+            SearchParameters searchParameters, LocalDate fromDate, LocalDate toDate,
+            @DecimalMin(value = "0", message = "must be greater than or equal to 0") BigDecimal fromAmount,
+            @DecimalMin(value = "0", message = "must be greater than or equal to 0") BigDecimal toAmount, String transactionType) {
+
+        final StringBuilder sqlBuilder = new StringBuilder(200);
+        sqlBuilder.append("select " + sqlGenerator.calcFoundRows() + " ");
+        sqlBuilder.append(this.transactionsMapper.schema());
+        sqlBuilder.append(" where sa.id = ? and sa.deposit_type_enum = ? ");
+        final Object[] objectArray = new Object[12];
```
galovics: Nope, this won't fly.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountReadPlatformServiceImpl.java
```diff
@@ -48,6 +49,7 @@
 import org.apache.fineract.organisation.monetary.data.CurrencyData;
 import org.apache.fineract.organisation.staff.data.StaffData;
 import org.apache.fineract.organisation.staff.service.StaffReadPlatformService;
+import org.apache.fineract.organisation.teller.exception.InvalidDateInputException;
```
galovics: This is a teller exception, why did we import it in the savings module?

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountReadPlatformServiceImpl.java
```diff
@@ -1267,6 +1270,99 @@ public Collection<SavingsAccountTransactionData> retrieveAllTransactions(final L
...
+            }
+            sqlBuilder.append(" and ").append(" tr.transaction_type_enum = ? ");
+            objectArray[arrayPos] = transcationTypeEnum;
+            arrayPos = arrayPos + 1;
+        }
+
+        if (fromDate != null || toDate != null) {
+            if (fromDate != null && toDate != null) {
+                if (toDate.isBefore(fromDate)) {
+                    throw new InvalidDateInputException(fromDate.toString(), toDate.toString());
```
galovics: No message, nobody will understand whats the issue without looking at the code.

galovics (changes requested):
In addition to the comments, there are zero tests for this new API, please add proper coverage for it.

galovics:
Also, squash your commits pls.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/SavingsAccountTransactionsApiResourceSwagger.java
```diff
@@ -20,11 +20,172 @@
...
+            @Schema(example = "false")
+            public boolean reversed;
+            @Schema(example = "[2023, 05, 01]")
+            public LocalDate submittedOnDate;
+            @Schema(example = "false")
+            public boolean interestedPostedAsOn;
+            @Schema(example = "mifos")
+            public String submittedByUsername;
+            @Schema(example = "false")
+            public boolean isManualTransaction;
```
adamsaghy, line 171: use `Boolean` class instead

fineract-provider/src/main/resources/static/legacy-docs/apiLive.htm
Mk9894: @galovics Does it mean that we can remove this?

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountReadPlatformService.java
Mk9894: @galovics  Thanks for the feedback. I can update it accordingly.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/SavingsAccountTransactionsApiResourceSwagger.java
Mk9894: @adamsaghy  Thanks for the feedback. Will fix it.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountReadPlatformServiceImpl.java
Mk9894: Hello @galovics , I acknowledge your preference for using JPQL for implementation. However, it's important to note that there is a possibility of compatibility issues with the other APIs of the transaction endpoint. Specifically, the retrieveOne function of the transaction API returns a response of Type SavingsAccountTransactionData, which is a custom type. In contrast, when using JPQL, we need to return SavingsAccountTransaction as the data type, which is the entity itself. This could potentially lead to the return response being incompatible with other transaction endpoints. I would greatly appreciate hearing your thoughts on this. 

Thanks

adamsaghy: you might wanna consider using jpa projections:
https://www.baeldung.com/spring-data-jpa-projections

https://vladmihalcea.com/spring-jpa-dto-projection/

Mk9894: Hey @adamsaghy  Thanks for your response. I tried using JPA specifications. But when i try to build the code, i get license violation! Does this mean I wont be able to use JPA specifications?

Mk9894: @galovics Using JPQL seems like a potential option, however, I came across a challenge where I need to join an unrelated entity (AccountTransferTransaction) to retrieve account transfer information, which as far as I'm aware is not feasible with JPQL. Would you happen to have any suggestions or alternative solutions to this issue?

fineract-provider/src/main/resources/static/legacy-docs/apiLive.htm
jdailey: This is a judgement call I think.  You should see if what you do breaks the web UI from Mifos.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountReadPlatformServiceImpl.java
galovics: Please look at the EclipseLink docs:
https://wiki.eclipse.org/EclipseLink/UserGuide/JPA/Basic_JPA_Development/Querying/JPQL#EclipseLink_Extensions_.28EQL.29

> [Joins](https://wiki.eclipse.org/EclipseLink/UserGuide/JPA/Basic_JPA_Development/Querying/JPQL#ON) between independent entities (EL 2.4)

Another option is to do a theta join between the entities, that should be supported by JPA directly.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/SavingsAccountTransactionsApiResource.java
Mk9894: @galovics 

Thank you for sharing the information regarding the HandlerArgumentResolver. I found that HandlerArgumentResolver is specific to Spring MVC framework and  I understand that the HandlerArgumentResolver interface is not directly related to the Jersey framework, which is a separate framework for building RESTful web services.

Also it seems that **Fineract** is indeed using the Jersey framework for handling REST requests, rather than the MVC framework.

In a Spring Boot application, we have the flexibility to choose either the Spring MVC or Jersey framework for handling REST requests. When using Spring MVC, the HandlerArgumentResolver interface allows us to customize the resolution of method arguments in controller methods. However, in the case of Jersey, an equivalent mechanism called the ParamConverterProvider interface is used to handle method arguments.

In Jersey, if we want to achieve similar functionality to the HandlerArgumentResolver, we can implement the ParamConverterProvider interface. This interface enables us to define custom conversion logic for method parameters in Jersey-based REST endpoints. The ParamConverterProvider is responsible for converting the values from incoming requests into the desired method parameter types.

Regarding your specific concern, if we choose to implement the ParamConverter interface in our custom implementation, we might not have access to the "locale" and "dateFormat" Query parameters from the Request. This could limit our ability to convert the given date string to a LocalDate object based on the specified locale and dateFormat.

Considering this, it seems more appropriate to use the legacy approach as currently implemented in the PR. This approach allows us to retrieve the "**locale**" and "**dateFormat**" Query parameters and utilize them for converting the date string to a LocalDate object, satisfying the requirements.

Thanks. 
Please let me know if you have any further questions or concerns

galovics: That's right, it should be a jersey parameter converter.
It's possible to access other attributes like this:
- https://stackoverflow.com/questions/27584161/read-another-parameter-within-jerseys-paramconverter
- https://stackoverflow.com/questions/53657358/paramconverter-based-on-http-header-value

Anyhow, let's postpone this and keep the legacy version.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/SavingsAccountTransactionsApiResourceSwagger.java
Mk9894: @adamsaghy The respective response entity is having this field as primitive boolean.

adamsaghy: it is fine then :)

galovics (changes requested):
No tests, please cover the use-cases with integration/unit tests. Thanks.

Also, squash your commits.

adamsaghy:
@Mk9894 Please squash your commits!

PR #3170, opened 2023-05-11
[FINERACT-1906] Add disbursement charge to accrual transactions

taskain7:
## Description

Fineract creates an accrual transaction, when a disbursement fee is charged.

## Test
`ClientLoanIntegrationTest:loanWithChargesOfTypeAmountPercentageAndPeriodicAccrualAccountingEnabled()` existing test case covers the integration test

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
@@ -192,6 +192,7 @@ import org.apache.fineract.portfolio.savings.domain.SavingsAccount;
 import org.apache.fineract.portfolio.transfer.api.TransferApiConstants;
 import org.apache.fineract.useradministration.domain.AppUser;
 import org.springframework.dao.DataIntegrityViolationException;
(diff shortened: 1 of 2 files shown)


fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -484,6 +486,15 @@ public CommandProcessingResult disburseLoan(final Long loanId, final JsonCommand
                     && loanCharge.isChargePending()) {
                 disBuLoanCharges.put(loanCharge.getId(), loanCharge.amountOutstanding());
             }
+            if (loanCharge.isDisbursementCharge()) {
```
mohitsinha, line 489: Looks good.
But can this be handled using JPA rather than Raw queries?

galovics (approved):

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
adamsaghy: @taskain7 Agree on this. Why do it need to be native query?

PR #3181, opened 2023-05-15
Fineract modularization - stage 0

b0c1:
- [x] Create fineract-core module - used to common cross module classes and interfaces
- [x] Create fineract-investor module for build
- [x] Modify tenant liquibase loader to using module based liquibase processing
- [x] Sample liquibase into the fineract-investor module
- [x] Move NewCommandSourceHandler and dependencies to the fineract-core


galovics (approved):

galovics (approved):

vidakovic:
Just out of curiosity: was there an immediate need to create this core module? Personally, I would have waited a bit to get all that manual JSON parsing stuff (or at least as much as possible) out of the way first... was out of the loop lately, so might have missed the discussions... just curious.

galovics:
@vidakovic yeah there was the need. This is still an intermediate step. If we are finally getting rid of the json parsing, we can still do it. It adds no overhead.

PR #3192, opened 2023-05-22
FINERACT-1912 Support Pagination, sorting and filtering for Savings account transactions

Mk9894:
## Description

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/account/domain/AccountTransferTransaction.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/account/domain/AccountTransferTransaction.java
@@ -26,6 +26,7 @@ import jakarta.persistence.ManyToOne;
 import jakarta.persistence.Table;
 import java.math.BigDecimal;
 import java.time.LocalDate;
+import lombok.Getter;
 import org.apache.fineract.infrastructure.core.domain.AbstractPersistableCustom;
 import org.apache.fineract.organisation.monetary.domain.MonetaryCurrency;
 import org.apache.fineract.organisation.monetary.domain.Money;
@@ -34,6 +35,7 @@ import org.apache.fineract.portfolio.savings.domain.SavingsAccountTransaction;
 
 @Entity
 @Table(name = "m_account_transfer_transaction")
+@Getter
 public class AccountTransferTransaction extends AbstractPersistableCustom {
 
     @ManyToOne
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/SavingsAccountTransactionsApiResource.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/SavingsAccountTransactionsApiResource.java
@@ -19,6 +19,7 @@
 package org.apache.fineract.portfolio.savings.api;
 
 import io.swagger.v3.oas.annotations.Operation;
+import io.swagger.v3.oas.annotations.Parameter;
 import io.swagger.v3.oas.annotations.media.ArraySchema;
 import io.swagger.v3.oas.annotations.media.Content;
 import io.swagger.v3.oas.annotations.media.Schema;
@@ -48,13 +49,17 @@ import org.apache.fineract.infrastructure.core.exception.PlatformDataIntegrityEx
 import org.apache.fineract.infrastructure.core.exception.UnrecognizedQueryParamException;
 import org.apache.fineract.infrastructure.core.serialization.ApiRequestJsonSerializationSettings;
 import org.apache.fineract.infrastructure.core.serialization.DefaultToApiJsonSerializer;
+import org.apache.fineract.infrastructure.core.service.Page;
+import org.apache.fineract.infrastructure.core.service.PagedRequest;
 import org.apache.fineract.infrastructure.security.service.PlatformSecurityContext;
 import org.apache.fineract.portfolio.paymenttype.data.PaymentTypeData;
 import org.apache.fineract.portfolio.paymenttype.service.PaymentTypeReadPlatformService;
 import org.apache.fineract.portfolio.savings.DepositAccountType;
 import org.apache.fineract.portfolio.savings.SavingsApiConstants;
 import org.apache.fineract.portfolio.savings.data.SavingsAccountTransactionData;
+import org.apache.fineract.portfolio.savings.domain.search.SavingsTransactionSearch;
 import org.apache.fineract.portfolio.savings.service.SavingsAccountReadPlatformService;
+import org.apache.fineract.portfolio.savings.service.search.SavingsAccountTransactionsSearchServiceImpl;
 import org.springframework.dao.CannotAcquireLockException;
 import org.springframework.orm.ObjectOptimisticLockingFailureException;
 import org.springframework.stereotype.Component;
@@ -71,6 +76,7 @@ public class SavingsAccountTransactionsApiResource {
     private final ApiRequestParameterHelper apiRequestParameterHelper;
     private final SavingsAccountReadPlatformService savingsAccountReadPlatformService;
     private final PaymentTypeReadPlatformService paymentTypeReadPlatformService;
+    private final SavingsAccountTransactionsSearchServiceImpl transactionsSearchServiceImpl;
 
     private boolean is(final String commandParam, final String commandValue) {
         return StringUtils.isNotBlank(commandParam) && commandParam.trim().equalsIgnoreCase(commandValue);
@@ -116,6 +122,19 @@ public class SavingsAccountTransactionsApiResource {
 
         return this.toApiJsonSerializer.serialize(settings, transactionData,
                 SavingsApiSetConstants.SAVINGS_TRANSACTION_RESPONSE_DATA_PARAMETERS);
+    }
+
+    @POST
+    @Path("search")
+    @Consumes({ MediaType.APPLICATION_JSON })
+    @Produces({ MediaType.APPLICATION_JSON })
+    @Operation(summary = "Search Savings Account Transactions")
+    @ApiResponses({
+            @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = SavingsAccountTransactionsApiResourceSwagger.SavingsAccountTransactionsSearchResponse.class))) })
+    public String searchTransactions(@PathParam("savingsId") @Parameter(description = "savingsId") final Long savingsId,
+            @Parameter PagedRequest<SavingsTransactionSearch> searchRequest) {
+        Page<SavingsAccountTransactionData> transactionsData = transactionsSearchServiceImpl.searchTransactions(savingsId, searchRequest);
+        return toApiJsonSerializer.serialize(transactionsData);
     }
 
     @POST
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsTransactionSearchResult.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsTransactionSearchResult.java
@@ -1,0 +1,147 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.portfolio.savings.data;
+
+import java.math.BigDecimal;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.util.Objects;
+import java.util.Optional;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import org.apache.fineract.organisation.monetary.data.CurrencyData;
+import org.apache.fineract.organisation.monetary.domain.ApplicationCurrency;
+import org.apache.fineract.portfolio.account.data.AccountTransferData;
+import org.apache.fineract.portfolio.account.domain.AccountTransferTransaction;
+import org.apache.fineract.portfolio.paymentdetail.data.PaymentDetailData;
+import org.apache.fineract.portfolio.paymentdetail.domain.PaymentDetail;
+import org.apache.fineract.portfolio.paymenttype.data.PaymentTypeData;
+import org.apache.fineract.portfolio.savings.domain.SavingsAccount;
+import org.apache.fineract.portfolio.savings.service.SavingsEnumerations;
+import org.apache.fineract.useradministration.domain.AppUser;
+
+@Getter
+@AllArgsConstructor
+public class SavingsTransactionSearchResult {
+
+    private Long transactionId;
+    private Integer transactionType;
+    private LocalDate transactionDate;
+    private BigDecimal transactionAmount;
+    private Long releaseIdOfHoldAmountTransaction;
+    private String reasonForBlock;
+    private LocalDateTime createdDate;
+    private AppUser appUser;
+    private String note;
+    private BigDecimal runningBalance;
+    private boolean reversed;
+    private boolean reversalTransaction;
+    private Long originalTxnId;
+    private Boolean lienTransaction;
+    private boolean isManualTransaction;
+    private AccountTransferTransaction fromSavingsTransaction;
+    private AccountTransferTransaction toSavingsTransaction;
+    private SavingsAccount savingsAccount;
+    private PaymentDetail paymentDetail;
+    private ApplicationCurrency currency;
+
+    public static final SavingsAccountTransactionData toSavingsAccountTransactionData(SavingsTransactionSearchResult dto) {
+        final Long id = dto.getTransactionId();
+        final int transactionTypeInt = dto.getTransactionType();
+
+        final SavingsAccountTransactionEnumData transactionType = SavingsEnumerations.transactionType(transactionTypeInt);
+
+        final LocalDate date = dto.getTransactionDate();
+        final LocalDate submittedOnDate = Optional.ofNullable(dto.getCreatedDate()).map(LocalDateTime::toLocalDate).orElse(null);
+        final BigDecimal amount = Optional.ofNullable(dto.getTransactionAmount()).orElse(BigDecimal.ZERO);
+        final Long releaseTransactionId = Optional.ofNullable(dto.getReleaseIdOfHoldAmountTransaction()).orElse(null);
+        final String reasonForBlock = Optional.ofNullable(dto.getReasonForBlock()).orElse(null);
+        final BigDecimal outstandingChargeAmount = null;
+        final BigDecimal runningBalance = Optional.ofNullable(dto.getRunningBalance()).orElse(BigDecimal.ZERO);
+        final boolean reversed = dto.isReversed();
+        final boolean isReversal = dto.isReversalTransaction();
+        final Long originalTransactionId = Optional.ofNullable(dto.getOriginalTxnId()).orElse(null);
+        final Boolean lienTransaction = dto.getLienTransaction();
+
+        final Long savingsId = Optional.ofNullable(dto.getSavingsAccount()).map(savingsAccount -> savingsAccount.getId()).orElse(null);
+        final String accountNo = Optional.ofNullable(dto.getSavingsAccount()).map(savingsAccount -> savingsAccount.getAccountNumber())
+                .orElse(null);
+        final boolean postInterestAsOn = dto.isManualTransaction();
+
+        PaymentDetailData paymentDetailData = null;
+        if (Objects.nonNull(transactionType) && transactionType.isDepositOrWithdrawal()) {
+            final PaymentDetail paymentDetail = dto.getPaymentDetail();
+            if (Objects.nonNull(paymentDetail)) {
+                final Long paymentTypeId = Optional.ofNullable(paymentDetail.getPaymentType()).map(paymentType -> paymentType.getId())
+                        .orElse(null);
+                if (Objects.nonNull(paymentTypeId)) {
+                    final String typeName = Optional.ofNullable(paymentDetail.getPaymentType()).map(paymentType -> paymentType.getName())
+                            .orElse(null);
+                    final PaymentTypeData paymentType = PaymentTypeData.instance(paymentTypeId, typeName);
+                    final String accountNumber = paymentDetail.getAccountNumber();
+                    final String checkNumber = paymentDetail.getCheckNumber();
+                    final String routingCode = paymentDetail.getRoutingCode();
+                    final String receiptNumber = paymentDetail.getReceiptNumber();
+                    final String bankNumber = paymentDetail.getBankNumber();
+                    paymentDetailData = new PaymentDetailData(id, paymentType, accountNumber, checkNumber, routingCode, receiptNumber,
+                            bankNumber);
+                }
+            }
+        }
+
+        final String currencyCode = dto.getSavingsAccount().getCurrency().getCode();
+        final String currencyName = dto.getCurrency().getName();
+        final String currencyNameCode = dto.getCurrency().getNameCode();
+        final String currencyDisplaySymbol = dto.getCurrency().getDisplaySymbol();
+        final Integer currencyDigits = dto.getSavingsAccount().getCurrency().getDigitsAfterDecimal();
+        final Integer inMultiplesOf = dto.getSavingsAccount().getCurrency().getCurrencyInMultiplesOf();
+        final CurrencyData currency = new CurrencyData(currencyCode, currencyName, currencyDigits, inMultiplesOf, currencyDisplaySymbol,
+                currencyNameCode);
+
+        AccountTransferData transfer = null;
+        AccountTransferTransaction transferFrom = dto.getFromSavingsTransaction();
+        AccountTransferTransaction transferTo = dto.getToSavingsTransaction();
+        if (Objects.nonNull(transferFrom)) {
+            final Long fromTransferId = transferFrom.getId();
+            final LocalDate fromTransferDate = transferFrom.getDate();
+            final BigDecimal fromTransferAmount = Optional.ofNullable(transferFrom.getAmount()).orElse(BigDecimal.ZERO);
+            final boolean fromTransferReversed = transferFrom.isReversed();
+            final String fromTransferDescription = transferFrom.getDescription();
+
+            transfer = AccountTransferData.transferBasicDetails(fromTransferId, currency, fromTransferAmount, fromTransferDate,
+                    fromTransferDescription, fromTransferReversed);
+        } else if (Objects.nonNull(transferTo)) {
+            final Long toTransferId = transferTo.getId();
+            final LocalDate toTransferDate = transferTo.getDate();
+            final BigDecimal toTransferAmount = Optional.ofNullable(transferTo.getAmount()).orElse(BigDecimal.ZERO);
+            final boolean toTransferReversed = transferTo.isReversed();
+            final String toTransferDescription = transferTo.getDescription();
+
+            transfer = AccountTransferData.transferBasicDetails(toTransferId, currency, toTransferAmount, toTransferDate,
+                    toTransferDescription, toTransferReversed);
+        }
+        final String submittedByUsername = Optional.ofNullable(dto.getAppUser()).map(user -> user.getUsername()).orElse(null);
+        final String note = Optional.ofNullable(dto.getNote()).orElse(null);
+        return SavingsAccountTransactionData.create(id, transactionType, paymentDetailData, savingsId, accountNo, date, currency, amount,
+                outstandingChargeAmount, runningBalance, reversed, transfer, submittedOnDate, postInterestAsOn, submittedByUsername, note,
+                isReversal, originalTransactionId, lienTransaction, releaseTransactionId, reasonForBlock);
+
+    }
+
+}
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccountTransactionRepository.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccountTransactionRepository.java
@@ -21,6 +21,7 @@ package org.apache.fineract.portfolio.savings.domain;
 import jakarta.persistence.LockModeType;
 import java.time.LocalDate;
 import java.util.List;
+import org.apache.fineract.portfolio.savings.domain.search.SavingsTransactionsSearchRepository;
 import org.springframework.data.domain.Pageable;
 import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.data.jpa.repository.JpaSpecificationExecutor;
@@ -28,8 +29,8 @@ import org.springframework.data.jpa.repository.Lock;
 import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.query.Param;
 
-public interface SavingsAccountTransactionRepository
-        extends JpaRepository<SavingsAccountTransaction, Long>, JpaSpecificationExecutor<SavingsAccountTransaction> {
+public interface SavingsAccountTransactionRepository extends JpaRepository<SavingsAccountTransaction, Long>,
+        JpaSpecificationExecutor<SavingsAccountTransaction>, SavingsTransactionsSearchRepository {
 
     @Query("select sat from SavingsAccountTransaction sat where sat.id = :transactionId and sat.savingsAccount.id = :savingsId")
     SavingsAccountTransaction findOneByIdAndSavingsAccountId(@Param("transactionId") Long transactionId,
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/search/SavingsTransactionSearch.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/search/SavingsTransactionSearch.java
@@ -1,0 +1,51 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.portfolio.savings.domain.search;
+
+import java.math.BigDecimal;
+import java.time.LocalDate;
+import java.util.List;
+import lombok.Data;
+import org.apache.fineract.infrastructure.core.data.RangeOperator;
+import org.apache.fineract.portfolio.savings.SavingsAccountTransactionType;
+
+@Data
+public class SavingsTransactionSearch {
+
+    private Filters filters;
+
+    @Data
+    public static class Filters {
+
+        private List<RangeFilter<LocalDate>> transactionDate;
+
+        private List<RangeFilter<BigDecimal>> transactionAmount;
+
+        private List<SavingsAccountTransactionType> transactionType;
+    }
+
+    @Data
+    public static class RangeFilter<T> {
+
+        private RangeOperator operator;
+
+        private T value;
+    }
+
+}
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/search/SavingsTransactionsSearchRepositoryImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/search/SavingsTransactionsSearchRepositoryImpl.java
@@ -1,0 +1,186 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.portfolio.savings.domain.search;
+
+import jakarta.persistence.EntityManager;
+import jakarta.persistence.TypedQuery;
+import java.math.BigDecimal;
+import java.time.LocalDate;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.StringJoiner;
+import java.util.stream.Collectors;
+import lombok.RequiredArgsConstructor;
+import org.apache.commons.collections4.CollectionUtils;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.fineract.portfolio.savings.DepositAccountType;
+import org.apache.fineract.portfolio.savings.SavingsAccountTransactionType;
+import org.apache.fineract.portfolio.savings.data.SavingsTransactionSearchResult;
+import org.apache.fineract.portfolio.savings.domain.search.SavingsTransactionSearch.Filters;
+import org.apache.fineract.portfolio.savings.domain.search.SavingsTransactionSearch.RangeFilter;
+import org.springframework.data.domain.Page;
+import org.springframework.data.domain.Pageable;
+import org.springframework.data.domain.Sort;
+import org.springframework.data.domain.Sort.Order;
+import org.springframework.data.support.PageableExecutionUtils;
+import org.springframework.stereotype.Repository;
+
+@Repository
+@RequiredArgsConstructor
+public class SavingsTransactionsSearchRepositoryImpl implements SavingsTransactionsSearchRepository {
+
+    private static final String AMOUNT_FIELD_NAME = "amount";
+    private static final String ID_FIELD_NAME = "id";
+    private static final String CREATED_DATE_FIELD_NAME = "createdDate";
+    private static final String TRANSACTION_DATE_FIELD_NAME = "dateOf";
+    private final EntityManager entityManager;
+
+    @Override
+    public Page<SavingsTransactionSearchResult> searchTransactions(SavingsTransactionSearchParameters searchParameters) {
+        // Build base query with filters but without the selection
+        BaseQueryParametersMapResult baseQueryParameterMapResult = buildBaseQueryWithFilters(searchParameters.getSavingsId(),
+                searchParameters.getDepositAccountType(), searchParameters.getFilters());
+
+        // Attach the selection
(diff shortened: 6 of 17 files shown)


fineract-provider/src/main/java/org/apache/fineract/portfolio/account/domain/AccountTransferTransaction.java
```diff
@@ -132,6 +132,22 @@ public void reverse() {
         this.reversed = true;
     }
 
+    public LocalDate getDate() {
```
galovics: Why not Lombok @Getter?

fineract-provider/src/main/java/org/apache/fineract/portfolio/paymentdetail/domain/PaymentDetail.java
```diff
@@ -121,4 +121,16 @@ public String getReceiptNumber() {
     public String getRoutingCode() {
         return routingCode;
     }
+
+    public String getAccountNumber() {
```
galovics: Why not Lombok @Getter?

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/SavingsAccountTransactionType.java
```diff
@@ -124,6 +127,20 @@ public static SavingsAccountTransactionType fromInt(final Integer transactionTyp
         return savingsAccountTransactionType;
     }
 
+    public static SavingsAccountTransactionType fromString(String transactionType) {
+        for (SavingsAccountTransactionType savingsAccountTransactionType : SavingsAccountTransactionType.values()) {
```
galovics: I really really don't like this splitting logic. Why can't we solve it differently which doesn't involve string magic? this solution is super fragile.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountTransactionDTOV2.java
```diff
@@ -0,0 +1,147 @@
...
+import org.apache.fineract.portfolio.paymentdetail.data.PaymentDetailData;
+import org.apache.fineract.portfolio.paymentdetail.domain.PaymentDetail;
+import org.apache.fineract.portfolio.paymenttype.data.PaymentTypeData;
+import org.apache.fineract.portfolio.savings.domain.SavingsAccount;
+import org.apache.fineract.portfolio.savings.service.SavingsEnumerations;
+import org.apache.fineract.useradministration.domain.AppUser;
+
+@Getter
+@AllArgsConstructor
+public class SavingsAccountTransactionDTOV2 {
```
galovics: DTOV2? Where's DTOV1? Can't we get a better name for this that represents the use-case?

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountTransactionDTOV2.java
```diff
@@ -0,0 +1,147 @@
...
+            submittedByUsername = getOrDefault(dto.getAppUser().getUsername(), null);
+        }
+        final String note = dto.getNote();
+
+        return SavingsAccountTransactionData.create(id, transactionType, paymentDetailData, savingsId, accountNo, date, currency, amount,
+                outstandingChargeAmount, runningBalance, reversed, transfer, submittedOnDate, postInterestAsOn, submittedByUsername, note,
+                isReversal, originalTransactionId, lienTransaction, releaseTransactionId, reasonForBlock);
+    }
+
+    private static <T> T getOrDefault(T input, T defaultValue) {
```
galovics: Single responsibility. How come the SavingsAccountTransaction has to deal with getOrDefault logic? You can use Objects.requireNonNull(T object, T defaultObject) instead of this..

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountTransactionDTOV2.java
```diff
@@ -0,0 +1,147 @@
...
+    private final Long originalTxnId;
+    private final Boolean lienTransaction;
+    private final boolean isManualTransaction;
+    private final AccountTransferTransaction fromSavingsTransaction;
+    private final AccountTransferTransaction toSavingsTransaction;
+    private final SavingsAccount savingsAccount;
+    private final PaymentDetail paymentDetail;
+    private final ApplicationCurrency currency;
+
+    public static final SavingsAccountTransactionData tosavingsAccountTransactionData(SavingsAccountTransactionDTOV2 dto) {
```
galovics: Not conforming the Java naming convention (read: not camelCase)

galovics: Are you planning to create a unit test for this class? Seems like there's way too much logic here than it should be.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountTransactionDTOV2.java
```diff
@@ -0,0 +1,147 @@
...
+        final Long id = dto.getTransactionId();
+        final int transactionTypeInt = dto.getTransactionType();
+
+        final SavingsAccountTransactionEnumData transactionType = SavingsEnumerations.transactionType(transactionTypeInt);
+
+        final LocalDate date = dto.getTransactionDate();
+        final LocalDate submittedOnDate = Optional.ofNullable(dto.getCreatedDate().toLocalDate()).orElse(null);
+        final BigDecimal amount = dto.getTransactionAmount();
+        final Long releaseTransactionId = Optional.ofNullable(dto.getReleaseIdOfHoldAmountTransaction()).orElse(0L);
```
galovics: Is this right? If there's no release transaction, we simply return a 0 as Id? This is misleading in my opinion to clients unless you explicitly mark and document this in the swagger docs.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountTransactionDTOV2.java
```diff
@@ -0,0 +1,147 @@
...
+        final String reasonForBlock = dto.getReasonForBlock();
+        final BigDecimal outstandingChargeAmount = null;
+        final BigDecimal runningBalance = dto.getRunningBalance();
+        final boolean reversed = dto.isReversed();
+        final boolean isReversal = dto.isReversalTransaction();
+        final Long originalTransactionId = Optional.ofNullable(dto.getOriginalTxnId()).orElse(0L);
```
galovics: Same as above.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccountTransactionRepository.java
```diff
@@ -53,4 +56,28 @@ List<SavingsAccountTransaction> findTransactionRunningBalanceBeforePivotDate(@Pa
...
+                    LEFT JOIN tr.notes nt ON (nt.savingsTransaction = tr)
+                    WHERE tr.savingsAccount.id = :savingsId
+                    AND tr.savingsAccount.depositType = :depositType
+                    AND (:transactionType IS NULL OR tr.typeOf = :transactionType )
+                    AND (:#{#searchParameters.fromDate} IS NULL OR tr.dateOf >= :#{#searchParameters.fromDate} )
+                    AND (:#{#searchParameters.toDate} IS NULL OR tr.dateOf <= :#{#searchParameters.toDate} )
+                    AND (:#{#searchParameters.fromAmount} IS NULL OR tr.amount >= :#{#searchParameters.fromAmount} )
+                    AND (:#{#searchParameters.toAmount} IS NULL OR tr.amount <= :#{#searchParameters.toAmount} )
+            """)
+    Page<SavingsAccountTransactionDTOV2> findAll(@Param("savingsId") Long savingsId, @Param("depositType") Integer depositType,
```
galovics: findAll method naming is confusing cause it's not finding all and this is not an entity either. This use-case is for search, name it accordingly.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountReadPlatformService.java
```diff
@@ -45,6 +45,9 @@ public interface SavingsAccountReadPlatformService {
 
     Collection<SavingsAccountTransactionData> retrieveAllTransactions(Long savingsId, DepositAccountType depositAccountType);
 
+    Page<SavingsAccountTransactionData> retrieveAllTransactions(Long savingsId, DepositAccountType depositAccountType,
```
galovics: Same as I said for the repository, search use-case.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountReadPlatformServiceImpl.java
```diff
@@ -126,7 +141,7 @@ public SavingsAccountReadPlatformServiceImpl(final PlatformSecurityContext conte
             final ChargeReadPlatformService chargeReadPlatformService,
             final EntityDatatableChecksReadService entityDatatableChecksReadService, final ColumnValidator columnValidator,
             final SavingsAccountAssembler savingAccountAssembler, PaginationHelper paginationHelper,
-            DatabaseSpecificSQLGenerator sqlGenerator) {
+            DatabaseSpecificSQLGenerator sqlGenerator, SavingsAccountTransactionRepository savingsAccountTransactionsRepository) {
```
galovics: Could've replced this with a lombok annotation.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountReadPlatformServiceImpl.java
```diff
@@ -1267,6 +1283,136 @@ public Collection<SavingsAccountTransactionData> retrieveAllTransactions(final L
...
+        validateSearchParameters(searchParameters);
+
+        Integer page = 0;
+        Integer size = SavingsApiConstants.SAVINGS_ACCOUNT_TRANSACTIONS_DEFAULT_LIMIT;
+        if (searchParameters.isLimited() && searchParameters.isOffset()) {
+            Integer limit = searchParameters.getLimit();
+            page = searchParameters.getOffset() / limit;
+            size = limit;
+        }
+        Pageable pageable = PageRequest.of(page, size);
```
galovics: What's the point of creating this PageRequest object if you override it in the following if/else?

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountReadPlatformServiceImpl.java
```diff
@@ -1267,6 +1283,136 @@ public Collection<SavingsAccountTransactionData> retrieveAllTransactions(final L
...
+                    defaultEnglishMessage.toString(), SavingsApiConstants.toAmountParamName, searchParameters.getToAmount());
+            dataValidationErrors.add(error);
+        }
+
+        if (!dataValidationErrors.isEmpty()) {
+            throw new PlatformApiDataValidationException(dataValidationErrors);
+        }
+    }
+
+    private boolean isValidAmount(BigDecimal amount) {
```
galovics: I bet fineract already has code for this, please check.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountReadPlatformServiceImpl.java
```diff
@@ -1267,6 +1283,136 @@ public Collection<SavingsAccountTransactionData> retrieveAllTransactions(final L
...
+        BigDecimal zero = BigDecimal.ZERO;
+        BigDecimal maxValue = BigDecimal.valueOf(Double.MAX_VALUE);
+
+        return (amount.compareTo(zero) >= 0) && (amount.compareTo(maxValue) <= 0);
+    }
+
+    private boolean isValidAmount(BigDecimal fromAmount, BigDecimal toAmount) {
```
galovics: I bet fineract already has code for this, please check.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountReadPlatformServiceImpl.java
```diff
@@ -1267,6 +1283,136 @@ public Collection<SavingsAccountTransactionData> retrieveAllTransactions(final L
...
+        BigDecimal maxValue = BigDecimal.valueOf(Double.MAX_VALUE);
+
+        boolean isValidFromAmount = isValidAmount(fromAmount);
+        boolean isValidToAmount = isValidAmount(toAmount);
+        boolean isValidFromToAmount = fromAmount.compareTo(toAmount) <= 0;
+
+        return isValidFromAmount && isValidToAmount && isValidFromToAmount;
+    }
+
+    private boolean isValidTransactionType(String transactionType) {
```
galovics: I bet fineract already has code for this, please check.

integration-tests/src/test/java/org/apache/fineract/integrationtests/common/savings/SavingsAccountHelper.java
```diff
@@ -646,6 +649,27 @@ public HashMap getSavingsTransaction(final Integer savingsID, final Integer savi
...
+        final String response = Utils.performServerGet(requestSpec, responseSpec, URL, null);
+        return GSON.fromJson(response, GetSavingsAccountTransactionsResponse.class);
+    }
+
+    public GetSavingsAccountTransactionsResponse getSavingsTransactionsWithQueryParams(final Integer savingsID,
+            Map<String, String> queryParamMap) {
+        final String URL = SAVINGS_ACCOUNT_URL + "/" + savingsID + "/transactions" + "?" + Utils.TENANT_IDENTIFIER;
+        StringBuilder builder = new StringBuilder(URL);
+
+        for (Map.Entry<String, String> entry : queryParamMap.entrySet()) {
```
galovics: Why don't we use the generated fineract-client classes? That way you don't even need to do this string magic again on the query params.

integration-tests/src/test/java/org/apache/fineract/integrationtests/SavingsAccountTransactionsIntegrationTest.java
```diff
@@ -0,0 +1,379 @@
...
+        final String startDate = "01 May 2023";
+        final Integer clientID = ClientHelper.createClient(this.requestSpec, this.responseSpec, startDate);
+        Assertions.assertNotNull(clientID);
+
+        final Integer savingsId = createSavingsAccountDailyPosting(clientID, startDate);
+
+        this.savingsAccountHelper.depositToSavingsAccount(savingsId, "1000", startDate, CommonConstants.RESPONSE_RESOURCE_ID);
+
+        GetSavingsAccountTransactionsResponse transactionsResponse = this.savingsAccountHelper.getSavingsTransactionsV2(savingsId);
+        Assertions.assertNotNull(transactionsResponse);
```
galovics: Please import this statically to have a consistent structure.

integration-tests/src/test/java/org/apache/fineract/integrationtests/SavingsAccountTransactionsIntegrationTest.java
```diff
@@ -0,0 +1,379 @@
...
+        assertEquals(1, transactionsResponse.getTotalFilteredRecords());
```
galovics: The details of the savings transaction is not checked (amount, etc)

integration-tests/src/test/java/org/apache/fineract/integrationtests/SavingsAccountTransactionsIntegrationTest.java
```diff
@@ -0,0 +1,379 @@
...
+        this.savingsAccountHelper.depositToSavingsAccount(savingsId, "700", startDate3, CommonConstants.RESPONSE_RESOURCE_ID);
+
+        GetSavingsAccountTransactionsResponse transactionsResponse = this.savingsAccountHelper
+                .getSavingsTransactionsWithQueryParams(savingsId, queryParametersMap);
+        Assertions.assertNotNull(transactionsResponse);
+        assertEquals(3, transactionsResponse.getTotalFilteredRecords());
+        Assertions.assertNotNull(transactionsResponse.getPageItems());
+        assertEquals(3, transactionsResponse.getPageItems().size());
+        Set<GetSavingsAccountTransactionsPageItem> pageItems = transactionsResponse.getPageItems();
+        List<BigDecimal> actualList = new ArrayList<>();
```
galovics: I don't like this logic here. Why can't we simply check the 1., 2., 3. items in the list? If the ordering works, it should be in the proper order in the list as well.

integration-tests/src/test/java/org/apache/fineract/integrationtests/SavingsAccountTransactionsIntegrationTest.java
```diff
@@ -0,0 +1,379 @@
...
+        GetSavingsAccountTransactionsResponse transactionsResponse = this.savingsAccountHelper
+                .getSavingsTransactionsWithQueryParams(savingsId, queryParametersMap);
+        Assertions.assertNotNull(transactionsResponse);
+        assertEquals(3, transactionsResponse.getTotalFilteredRecords());
+        Assertions.assertNotNull(transactionsResponse.getPageItems());
+        assertEquals(2, transactionsResponse.getPageItems().size());
+    }
+
+    @Test
+    public void testSavingsTransactionsSortByAmountDesc() {
```
galovics: Don't see a test-case for Asc ordering.

integration-tests/src/test/java/org/apache/fineract/integrationtests/SavingsAccountTransactionsIntegrationTest.java
```diff
@@ -0,0 +1,379 @@
...
+        this.savingsAccountHelper.depositToSavingsAccount(savingsId, "700", startDate3, CommonConstants.RESPONSE_RESOURCE_ID);
+
+        GetSavingsAccountTransactionsResponse transactionsResponse = this.savingsAccountHelper
+                .getSavingsTransactionsWithQueryParams(savingsId, queryParametersMap);
+        Assertions.assertNotNull(transactionsResponse);
+        assertEquals(3, transactionsResponse.getTotalFilteredRecords());
+        Assertions.assertNotNull(transactionsResponse.getPageItems());
+        assertEquals(3, transactionsResponse.getPageItems().size());
+        Set<GetSavingsAccountTransactionsPageItem> pageItems = transactionsResponse.getPageItems();
+        List<BigDecimal> actualList = new ArrayList<>();
```
galovics: Same as above.

integration-tests/src/test/java/org/apache/fineract/integrationtests/SavingsAccountTransactionsIntegrationTest.java
```diff
@@ -0,0 +1,379 @@
...
+        this.savingsAccountHelper.depositToSavingsAccount(savingsId, "700", startDate3, CommonConstants.RESPONSE_RESOURCE_ID);
+
+        GetSavingsAccountTransactionsResponse transactionsResponse = this.savingsAccountHelper
+                .getSavingsTransactionsWithQueryParams(savingsId, queryParametersMap);
+        Assertions.assertNotNull(transactionsResponse);
+        assertEquals(3, transactionsResponse.getTotalFilteredRecords());
+        Assertions.assertNotNull(transactionsResponse.getPageItems());
+        assertEquals(3, transactionsResponse.getPageItems().size());
+        Set<GetSavingsAccountTransactionsPageItem> pageItems = transactionsResponse.getPageItems();
+        List<LocalDate> actualList = new ArrayList<>();
```
galovics: Same as above.

integration-tests/src/test/java/org/apache/fineract/integrationtests/SavingsAccountTransactionsIntegrationTest.java
```diff
@@ -0,0 +1,379 @@
...
+        pageItems.forEach(data -> {
+            actualList.add(data.getDate());
+        });
+        boolean isAscending = isDateListOrdered(actualList, "ASC");
+        assertEquals(true, isAscending);
+
+    }
+
+    @Test
+    public void testSavingsTransactionsFilterByTransactionType() {
```
galovics: This test is just half of the pie because you never tested whether without the filtering applied, the post interest transaction is also returned; i.e. you don't know if your filter parameter that you passed has been handled or it's because by default the API filters out non deposit transactions.

integration-tests/src/test/java/org/apache/fineract/integrationtests/SavingsAccountTransactionsIntegrationTest.java
```diff
@@ -0,0 +1,379 @@
...
+        GetSavingsAccountTransactionsResponse transactionsResponse = this.savingsAccountHelper
+                .getSavingsTransactionsWithQueryParams(savingsId, queryParametersMap);
+        Assertions.assertNotNull(transactionsResponse);
+        assertEquals(2, transactionsResponse.getTotalFilteredRecords());
+        Assertions.assertNotNull(transactionsResponse.getPageItems());
+        assertEquals(2, transactionsResponse.getPageItems().size());
+        Set<GetSavingsAccountTransactionsPageItem> pageItems = transactionsResponse.getPageItems();
+        for (GetSavingsAccountTransactionsPageItem item : pageItems) {
+            assertEquals(true, item.getTransactionType().getDeposit(), "Transaction Type is not as expected");
+            assertEquals(true, item.getAmount().compareTo(BigDecimal.valueOf(700)) <= 0);
```
galovics: if you're checking this, why don't you check for 400?

galovics (changes requested):

integration-tests/src/test/java/org/apache/fineract/integrationtests/SavingsAccountTransactionsIntegrationTest.java
Mk9894: @galovics There is a test case for Default sorting. The default sorting is ASC.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountReadPlatformServiceImpl.java
Mk9894: Yes. But I am afraid if it would break the existing implementation

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccountTransactionRepository.java
Mk9894: @galovics As per my current implementation findAll will also work. If you dont provide any query param to this endpoint, still you can get all the transactions without any filter applied to the result set! By adding query params you can get the filtered records.

If i misunderstood the requirement, then Is it okay to implement a POST endpoint for this ?
Eg: POST /savingsaccount/{id}/transactions/search

Please confirm.

integration-tests/src/test/java/org/apache/fineract/integrationtests/SavingsAccountTransactionsIntegrationTest.java
galovics: That's not enough. There are 3 use-cases for sorting:
- default
- asc
- desc

From which asc is not covered. Imagine the default changing from asc to desc sorting. Then you won't have any proof that asc sorting works.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountReadPlatformServiceImpl.java
galovics: This is a constructor. If it breaks the existing impl, you'll know it because the startup of the Spring context will simply fail.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccountTransactionRepository.java
galovics: Still I think this is a search use-case and should be named accordingly.

A POST API is also fine with me, but if you're planning to create a new POST API, then please take a look at ClientSearchV2ApiResource cause it could act as a base pattern for the implementation.

Mk9894: Sure. Thanks for the heads up!

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsTransactionSearchResult.java
```diff
@@ -0,0 +1,152 @@
...
+    public static final SavingsAccountTransactionSearchData toSavingsAccountTransactionSearchData(SavingsTransactionSearchResult dto) {
+        final Long id = dto.getTransactionId();
+        final int transactionTypeInt = dto.getTransactionType();
+
+        final SavingsAccountTransactionEnumData transactionType = SavingsEnumerations.transactionType(transactionTypeInt);
+
+        final LocalDate date = dto.getTransactionDate();
+        final LocalDate submittedOnDate = Optional.ofNullable(dto.getCreatedDate()).map(LocalDateTime::toLocalDate).orElse(null);
+        final BigDecimal amount = Optional.ofNullable(dto.getTransactionAmount()).orElse(BigDecimal.ZERO);
+        final Long releaseTransactionId = Optional.ofNullable(dto.getReleaseIdOfHoldAmountTransaction()).orElse(0L);
```
Mk9894: @galovics I can return a null value if getReleaseIdOfHoldAmountTransaction is null . But The existing implementation of SavingsAccountTransactions returns a 0 value.

chtompki:
I'm going to read in on this PR to see if I can help out with the lift.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/v2/search/SavingsAccountTransactionsSearchApiResource.java
```diff
@@ -0,0 +1,66 @@
...
+    private final DefaultToApiJsonSerializer<SavingsAccountTransactionSearchData> toApiJsonSerializer;
+
+    @POST
+    @Path("search")
+    @Consumes({ MediaType.APPLICATION_JSON })
+    @Produces({ MediaType.APPLICATION_JSON })
+    @Operation(summary = "Search Savings Account Transactions")
+    @ApiResponses({
+            @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = SavingsAccountTransactionsApiResourceSwagger.SavingsAccountTransactionsSearchResponse.class))) })
+    public String searchTransactions(@PathParam("savingsId") @Parameter(description = "savingsId") final Long savingsId,
```
adamsaghy: Please use the Spring DATA Page instead of the "own"... 
The idea is to move towards the standard solutions and let the "custom" ones got deprecated slowly.

Mk9894: Hi Adam,

I had also considered utilizing the Spring Data Page instead of the legacy response. However, Peter expressed a preference for the response to resemble the legacy implementation. If it is necessary for me to accommodate the legacy response fields while incorporating the spring data Page, it will require a refactoring of all the transaction-related entities.
I am concerned that making those changes may potentially disrupt the existing implementations.

adamsaghy: @Mk9894 I see.. would you please change than only to be this v1?

The idea is v2 is using the new way of generating the swagger definitions and using the Spring data page and some other enhancements, if this is to support the legacy pagination, it should remain v1.

is that okay?

Mk9894: Done.

adamsaghy (dismissed):
LGTM, but need to resolve conflicts.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/search/SavingsTransactionSearch.java
```diff
@@ -0,0 +1,80 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.math.BigDecimal;
+import java.time.LocalDate;
+import java.util.List;
+import lombok.Data;
+import org.apache.fineract.portfolio.savings.SavingsAccountTransactionType;
+
+@Data
+public class SavingsTransactionSearch {
```
galovics, line 29: I don't agree with this structure at all.
If I translate this to JSON, looks like this:

```
{
  "filters": {
     "transactionDate": {
        "gte": "2022-01-01"
     }
  }
}
```
While it should be something like:
```
{
  "filters": {
     "transactionDate": {
        "operator": "gte",
        "value": "2022-01-01"
     }
  }
}
```

Much more readable, maintainable and extendable this way.

galovics, line 29: I don't get why you need the JsonProperty annotations everywhere. Why don't you just name the attributes appropriately?

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/search/SavingsTransactionsSearchRepositoryImpl.java
```diff
@@ -0,0 +1,176 @@
...
+            if (Objects.nonNull(dateFilters)) {
+                if (Objects.nonNull(dateFilters.getLte())) {
+                    queryBuilder.append(" AND tr.dateOf <= :dateLte ");
+                    parameterMap.put("dateLte", dateFilters.getLte());
+                }
+                if (Objects.nonNull(dateFilters.getGte())) {
+                    queryBuilder.append(" AND tr.dateOf >= :dateGte ");
+                    parameterMap.put("dateGte", dateFilters.getGte());
+                }
+                // We prioritize the lte/gte filter by applying it first. If lte/gte is not provided, only then we
```
galovics: All this can be removed if you do the structure I mentioned above.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/search/SavingsTransactionsSearchRepositoryImpl.java
```diff
@@ -0,0 +1,176 @@
...
+import org.springframework.stereotype.Repository;
+
+@Repository
+@RequiredArgsConstructor
+public class SavingsTransactionsSearchRepositoryImpl implements SavingsTransactionsSearchRepository {
+
+    private final EntityManager entityManager;
+
+    @Override
+    public Page<SavingsTransactionSearchResult> searchTransactions(Long savingsId, Integer depositAccountType, Filters filters,
```
galovics: Let's work with types, why don't we use a DepositAccountType enum here?

galovics: Why don't we use a Parameter Object here instead of the individual params?

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/search/SavingsTransactionsSearchRepositoryImpl.java
```diff
@@ -0,0 +1,176 @@
...
+                        JOIN ApplicationCurrency currency ON (currency.code = tr.savingsAccount.currency.code)
+                        LEFT JOIN AccountTransferTransaction fromtran ON (fromtran.fromSavingsTransaction = tr)
+                        LEFT JOIN AccountTransferTransaction totran ON (totran.toSavingsTransaction = tr)
+                        LEFT JOIN tr.notes nt ON (nt.savingsTransaction = tr)
+                """;
+    }
+
+    private Long getTotalElements(StringBuilder queryBuilder, Map<String, Object> parameterMap) {
+        // ORDER BY Clause not required for Count Query
+        int orderByIndex = queryBuilder.indexOf("ORDER BY");
```
galovics: This is a super no-go. Why construct the order by in there in the first place? What if the ORDER BY clause is all lowercase? This is extremely fragile.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/search/SavingsTransactionsSearchRepositoryImpl.java
```diff
@@ -0,0 +1,176 @@
...
+        if (orderByIndex != -1) {
+            queryBuilder.replace(orderByIndex, queryBuilder.length(), "");
+        }
+        String countJPQL = "SELECT COUNT(tr) " + queryBuilder.substring(queryBuilder.indexOf("FROM"), queryBuilder.length());
```
galovics: No, please no. No string manipulation. As soon as you start doing it, there's never a stop to it. Really difficult to refactor and maintain + as I said above, extremely fragile.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/search/SavingsTransactionsSearchRepositoryImpl.java
```diff
@@ -0,0 +1,176 @@
...
+@Repository
+@RequiredArgsConstructor
+public class SavingsTransactionsSearchRepositoryImpl implements SavingsTransactionsSearchRepository {
+
+    private final EntityManager entityManager;
+
+    @Override
+    public Page<SavingsTransactionSearchResult> searchTransactions(Long savingsId, Integer depositAccountType, Filters filters,
+            Pageable pageable) {
+        String jpql = buildJpqlQuery();
```
galovics: The whole implementation lacks structure.

I'd imagine on a high level this:

1. Build base query with filters but without the selection
2. Attach the selection
3. Attach the ordering
4. Execute the query
5. Build base query with filters
6. Attach the COUNT selection
7. Execute the query (with the PageExecutionUtils..)
8. Construct the page object

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/SavingsAccountTransactionsApiResource.java
```diff
@@ -118,6 +124,19 @@ public String retrieveOne(@PathParam("savingsId") final Long savingsId, @PathPar
...
     }
 
+    @POST
+    @Path("search")
+    @Consumes({ MediaType.APPLICATION_JSON })
+    @Produces({ MediaType.APPLICATION_JSON })
+    @Operation(summary = "Search Savings Account Transactions")
+    @ApiResponses({
+            @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = SavingsAccountTransactionsApiResourceSwagger.SavingsAccountTransactionsSearchResponse.class))) })
+    public String searchTransactions(@PathParam("savingsId") @Parameter(description = "savingsId") final Long savingsId,
```
galovics, line 134: Is there any reason we must return a String from this instead of the Page itself?

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/search/SavingsAccountTransactionSearchService.java
```diff
@@ -0,0 +1,28 @@
...
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.portfolio.savings.service.search;
+
+import org.apache.fineract.infrastructure.core.service.PagedRequest;
+import org.apache.fineract.portfolio.savings.domain.search.SavingsTransactionSearch;
+
+public interface SavingsAccountTransactionSearchService {
```
galovics, line 26: Interface is not used.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/search/SavingsAccountTransactionsSearchServiceImpl.java
```diff
@@ -0,0 +1,74 @@
...
+import org.apache.fineract.portfolio.savings.domain.search.SavingsTransactionSearch.Filters;
+import org.springframework.data.domain.Pageable;
+import org.springframework.data.domain.Sort.Order;
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+
+@Service
+@Transactional(readOnly = true)
+@RequiredArgsConstructor
+public class SavingsAccountTransactionsSearchServiceImpl {
```
galovics: Class shouldn't be named as Impl since there's no interface behind it.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/search/SavingsAccountTransactionsSearchServiceImpl.java
```diff
@@ -0,0 +1,74 @@
...
+        Optional<SavingsTransactionSearch> request = searchRequest.getRequest();
+        Pageable pageable = searchRequest.toPageableWithDefaultSortOrders(getDefaultOrders());
+        Filters searchFilters = request.map(SavingsTransactionSearch::getFilters).orElse(null);
+        org.springframework.data.domain.Page<SavingsAccountTransactionData> pageResult = savingsTransactionRepository
+                .searchTransactions(savingsId, depositType.getValue(), searchFilters, pageable)
+                .map(SavingsTransactionSearchResult::toSavingsAccountTransactionData);
+        return new Page<>(pageResult.getContent(), Long.valueOf(pageResult.getTotalElements()).intValue());
+    }
+
+    private List<Order> getDefaultOrders() {
```
galovics: I don't think putting the default ordering on is the responsibility of the service but rather the repository layer since it has to make sure for a paged query, there must be a consistent ordering in place.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/PagedRequest.java
```diff
@@ -53,6 +55,18 @@ public Pageable toPageable() {
         }
     }
 
+    public Pageable toPageableWithDefaultSortOrders(List<Order> defaultSortOrders) {
```
galovics: This is mostly copy paste and as I said on the service class, the default ordering shouldn't be on this level but rather on the repo.

integration-tests/src/test/java/org/apache/fineract/integrationtests/common/savings/SavingsAccountHelper.java
```diff
@@ -646,6 +653,47 @@ public HashMap getSavingsTransaction(final Integer savingsID, final Integer savi
         return Utils.performServerGet(requestSpec, responseSpec, URL, "");
     }
 
+    public SavingsAccountTransactionsSearchResponse searchTransactions(Integer savingsId, Filters filters) {
```
galovics: These methods are copy paste. Why don't you reuse them properly?

integration-tests/src/test/java/org/apache/fineract/integrationtests/SavingsAccountTransactionsSearchIntegrationTest.java
```diff
@@ -0,0 +1,386 @@
...
+        amountFilters.setGte(BigDecimal.valueOf(100));
+        amountFilters.setLte(BigDecimal.valueOf(200));
+        filters.setAmount(amountFilters);
+        SavingsAccountTransactionsSearchResponse transactionsResponse = this.savingsAccountHelper.searchTransactions(savingsId, filters);
+        Assertions.assertNotNull(transactionsResponse);
+        assertEquals(1, transactionsResponse.getTotalFilteredRecords());
+        Assertions.assertNotNull(transactionsResponse.getPageItems());
+        Set<GetSavingsAccountTransactionsPageItem> pageItems = transactionsResponse.getPageItems();
+        pageItems.forEach(data -> {
+            BigDecimal expectedAmount = BigDecimal.valueOf(100).setScale(6);
```
galovics: BigDecimals shouldn't be compared with equals but rather compareTo.

integration-tests/src/test/java/org/apache/fineract/integrationtests/SavingsAccountTransactionsSearchIntegrationTest.java
```diff
@@ -0,0 +1,386 @@
...
+        this.savingsAccountHelper.withdrawalFromSavingsAccount(savingsId, "100", withdrawDate, CommonConstants.RESPONSE_RESOURCE_ID);
+        this.savingsAccountHelper.postInterestForSavings(savingsId);
+        Filters filters = new Filters();
+        int page = 0;
+        int size = 10;
+        SavingsAccountTransactionsSearchResponse transactionsResponse = this.savingsAccountHelper.searchTransactions(savingsId, filters,
+                null, page, size);
+        Assertions.assertNotNull(transactionsResponse);
+        Assertions.assertNotNull(transactionsResponse.getPageItems());
+        assertEquals(10, transactionsResponse.getPageItems().size());
```
galovics: This test is not following the FIRST principles, specifically the I and R are violated since you're expecting 10 savings transactions but you're only saving 3 transactions + the interest posting which I assume is only 1 transaction.
So if this test is being ran by itself, this will fail.

Note: unless the post savings interest thingy creates 7+ transactions, then just disregard my comment.

integration-tests/src/test/java/org/apache/fineract/integrationtests/SavingsAccountTransactionsSearchIntegrationTest.java
```diff
@@ -0,0 +1,386 @@
...
+        Filters filters = new Filters();
+        filters.setTransactionType(List.of(Filters.TransactionTypeEnum.DEPOSIT));
+        SavingsAccountTransactionsSearchResponse transactionsResponse = this.savingsAccountHelper.searchTransactions(savingsId, filters);
+        Assertions.assertNotNull(transactionsResponse);
+        assertEquals(2, transactionsResponse.getTotalFilteredRecords());
+        Assertions.assertNotNull(transactionsResponse.getPageItems());
+        assertEquals(2, transactionsResponse.getPageItems().size());
+        List<LocalDate> transactionDates = transactionsResponse.getPageItems().stream().map(data -> data.getDate())
+                .collect(Collectors.toList());
+        assertTrue(isDateListOrdered(transactionDates, "desc"));
```
galovics: Why don't you check the individual items one by one in the proper order?

integration-tests/src/test/java/org/apache/fineract/integrationtests/SavingsAccountTransactionsSearchIntegrationTest.java
```diff
@@ -0,0 +1,386 @@
...
+        List<SortOrder> sortOrders = List.of(sortOrder);
+        SavingsAccountTransactionsSearchResponse transactionsResponse = this.savingsAccountHelper.searchTransactions(savingsId, filters,
+                sortOrders);
+        Assertions.assertNotNull(transactionsResponse);
+        assertEquals(2, transactionsResponse.getTotalFilteredRecords());
+        Assertions.assertNotNull(transactionsResponse.getPageItems());
+        assertEquals(2, transactionsResponse.getPageItems().size());
+        List<BigDecimal> amountList = transactionsResponse.getPageItems().stream().map(data -> data.getAmount())
+                .collect(Collectors.toList());
+        assertTrue(isListOrdered(amountList, "asc"));
```
galovics: Same as above.

galovics (changes requested):

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/SavingsAccountTransactionsApiResource.java
Mk9894: This is done to match the legacy response of savings account transactions

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/search/SavingsTransactionSearch.java
Mk9894: Done

Mk9894: Done

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/search/SavingsAccountTransactionSearchService.java
Mk9894: Implemented the interface

galovics (approved):

jdailey:
Let's figure out a solution to the auto checks failing.  Gradle upgrade could be part of this.

PR #3204, opened 2023-05-25
[FINERACT-1926] loan transfer validation error fix

taskain7:
## Description

Exception is thrown when a field validation fails on the request body


adamsaghy (approved):
LGTM

PR #3228, opened 2023-06-05
[FINERACT-1936] fixing SQL queries that was failing on Maria DB

taskain7:
## Description

SQL queries was failing on Maria DB, caused by lowercase-uppercase differencies.
Multiple inner query alias have been renamed.


adamsaghy (approved):
LGTM

reluxa (approved):

PR #3252, opened 2023-06-13
fix(build): make logging warnings clearer

chtompki:
Starting with small non-breaking contributions to clean the code up.

### DIFF
--- a/buildSrc/src/main/groovy/org/apache/fineract/gradle/service/GpgService.groovy
+++ b/buildSrc/src/main/groovy/org/apache/fineract/gradle/service/GpgService.groovy
@@ -87,7 +87,7 @@ class GpgService {
                 this.privateKey = secretKey.extractPrivateKey(new JcePBESecretKeyDecryptorBuilder().setProvider(BouncyCastleProvider.PROVIDER_NAME).build(config.password.toCharArray()))
             }
         } else {
-            log.warn("If preparing a fineract release: could not open the public key ring.")
+            log.warn("WARN: If preparing a fineract release: could not open the public key ring.")
         }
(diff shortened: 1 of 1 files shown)


jdailey (commented):
LGTM

galovics:
@chtompki can you please conform the contribution requirements for Fineract? Let's squash your commits, have a FINERACT-XYZ ticket in the title of the PR that your changes are referring to and the same thing for the commit message.
Thanks.

reluxa (commented):
I am just wondering if it would be possible to decide if we are building a release or we are in a regular build. In case we are in a regular build the initialization of this class could be omitted.

buildSrc/src/main/groovy/org/apache/fineract/gradle/service/GpgService.groovy
```diff
@@ -87,7 +87,7 @@ class GpgService {
                 this.privateKey = secretKey.extractPrivateKey(new JcePBESecretKeyDecryptorBuilder().setProvider(BouncyCastleProvider.PROVIDER_NAME).build(config.password.toCharArray()))
             }
         } else {
-            log.warn("Could not open the public key ring.")
+            log.warn("WARN: If preparing a fineract release: could not open the public key ring.")
```
reluxa, line 90: The GPGService is only used in step7, so we may defer the initialization of the GpgService when that step is executed. 

https://github.com/apache/fineract/blob/a90506c7e0841e16932af1c3c69c95166388ad96/buildSrc/src/main/groovy/org/apache/fineract/gradle/FineractPlugin.groovy#L335

PR #3278, opened 2023-06-29
FINERACT-1724: Fineract Idempotency status null fix

b0c1:
FINERACT-1724: Fineract Idempotency status null fix
- [x] Return 500 when the idempotency status is inconsistent in the database
- [x] Unit test for the mapper to avoid NullPointerException


adamsaghy (approved):
LGTM

PR #3308, opened 2023-07-14
Mitigation of Log Injection Vulnerabilities in Sender.java

Onyx2406:
## Description

This PR mitigates potential log injection vulnerabilities found in the Sender.java file. Log Injection vulnerabilities occur when unsanitized user input is written directly to a log entry. A malicious user could potentially exploit this to forge log entries or inject malicious content.

For example, consider the following code:
```
String responseBody = getUserInput();
LOG.log(Level.WARNING, "Unrecognized response: " + responseBody);
```
If a malicious user provides a responseBody that contains newline characters like "Valid response\nSomething malicious", it would be logged as two separate log entries. This could be exploited to forge log entries or to inject malicious content into the logs.

To mitigate this, a new private method sanitize was introduced that replaces newline and carriage return characters in the logs' input strings with their literal equivalents, i.e., \n and \r. This method was then used to sanitize responseBody variable in the relevant parts of the code.

Reference: [CWE-117](https://github.com/github/codeql/pull/6182)
## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/gcm/domain/Sender.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/gcm/domain/Sender.java
@@ -116,6 +116,13 @@ public class Sender {
         return endpoint;
     }
 
+    public static String sanitize(String message) {
+        if (message == null) {
+            return null;
+        }
+        return message.replace("\n", "\\n").replace("\r", "\\r");
+    }
+
     /**
      * Set the underlying URLConnection's connect timeout (in milliseconds). A timeout value of 0 specifies an infinite
      * timeout.
@@ -270,7 +277,7 @@ public class Sender {
                     String error = jsonResponse.get(JSON_ERROR).getAsString();
                     resultBuilder.errorCode(error);
                 } else {
-                    LOG.log(Level.WARNING, "Expected " + JSON_MESSAGE_ID + " or " + JSON_ERROR + " found: " + responseBody);
+                    LOG.log(Level.WARNING, "Expected " + JSON_MESSAGE_ID + " or " + JSON_ERROR + " found: " + sanitize(responseBody));
                     return null;
                 }
             } else if (jsonResponse.has(JSON_SUCCESS) && jsonResponse.has(JSON_FAILURE)) {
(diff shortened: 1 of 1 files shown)


vidakovic:
I think there are a couple of things wrong here:

-  this issue seems to be inspired by a python project; you can't just make the same assumptions here without looking at the libs we use for logging (logback)... which has no known log injection issue (like log4j had)
- are multiple lines (the only thing I possibly see here happening) really that much of an issue?
-  let's say yes for a moment... then why do string concatenation here instead of using proper log placeholders? concatenation is just wrong
-  when we use placeholders then the input passes through the log encoder; if you still think that multiple lines are that much of an issue then configure JSON log encoding and voila the problem is contained
- personally I would discourage people to log content like that in the first place; in production this just bloats the log files unnecessarily; better use debug log level if you really need to investigate then this perceived issue won't happen in production and you have full control when and if the input reaches your logs

Note: at first sight this might look like a security issue... we have a proper process in place for this... if you reveal everything immediately in an open PR then we have no chance of fixing things before they get exploited.

vidakovic:
I don't recommend to merge this. The sanitize method is taking care of one specific use case. I would at least expect that we have the ability to add more use cases (regex patterns). And again, overall this is a non-issue from my perspective.

vidakovic:
One more thing: we should not use java.util.logging.* ... probably that is a historic left over... we should always use Slf4j.

Onyx2406:
Thanks for reviewing this PR in detail.

I understand your viewpoint about this being a potential non-issue, considering the nature of the logging libraries being used in the project. You're correct that logback doesn't have the same log injection vulnerabilities as log4j and there is not any security impact due to this.

You bring up an excellent point about using proper log placeholders instead of string concatenation. This is indeed a better practice as it leverages the library's built-in protection mechanisms, and I apologize for the oversight on my part.

In regards to the logging level, your comment about using debug instead of warning to control what gets logged in production makes perfect sense. It's certainly a balance to strike between logging necessary data for debugging purposes and avoiding unnecessary bloating of log files.

Also, your recommendation to migrate from java.util.logging.* to Slf4j is noted, and I agree that this could be a positive step for the project. I would like to implement the same.

I will do a more thorough analysis and assure you that I will send such reports via email instead of opening PR's in the future. So, closing this for now.

PR #3312, opened 2023-07-18
[FINERACT-1953] Matured on date fix

taskain7:
## Description

If the loan is closed or overpaid and becomes active again, the `Actual Maturity Date` needs to be the same as the `Expected Maturiy Date`

### DIFF
--- a/fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
+++ b/fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
@@ -7124,7 +7124,7 @@ public class Loan extends AbstractAuditableWithUTCDateTimeCustom {
     }
 
     public void handleMaturityDateActivate() {
-        if (this.expectedMaturityDate != null && this.actualMaturityDate == null) {
+        if (this.expectedMaturityDate != null && !this.expectedMaturityDate.equals(this.actualMaturityDate)) {
             this.actualMaturityDate = this.expectedMaturityDate;
         }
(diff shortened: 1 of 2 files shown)


fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/DefaultLoanLifecycleStateMachine.java
```diff
@@ -42,6 +42,10 @@ public LoanStatus dryTransition(final LoanEvent loanEvent, final Loan loan) {
     public void transition(final LoanEvent loanEvent, final Loan loan) {
         LoanStatus newStatus = getNextStatus(loanEvent, loan);
         if (newStatus != null) {
+            if (isNotLoanCreation(loanEvent) && (loan.isClosed() || LoanStatus.OVERPAID.equals(loan.getStatus()))
```
adamsaghy, line 45: Please either use the loan methods to decide on the statuses or use the equality check everywhere, but do not mix them.
Example:

if (isNotLoanCreation(loanEvent) && (loan.isClosed() || **loan.isOverpaid())** && LoanStatus.ACTIVE.equals(newStatus))

adamsaghy (approved):
LGTM

adamsaghy (changes requested):
You might wanna check `LoanStatusChangePlatformServiceImpl.onBusinessEvent` which already doing something similar. It might worth to enhance that one.

adamsaghy (approved):
LGTM

PR #3320, opened 2023-07-21
[FINERACT-1760] Client transaction by transaction external ID

taskain7:
## Description

 - Pay charge with transaction external ID
 - retrieve client transaction by transaction external ID
 - undo client transaction by transaction external ID

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/ClientTransaction.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/ClientTransaction.java
@@ -39,6 +39,7 @@ import java.util.Set;
 import org.apache.fineract.accounting.glaccount.domain.GLAccount;
 import org.apache.fineract.infrastructure.core.data.EnumOptionData;
 import org.apache.fineract.infrastructure.core.domain.AbstractAuditableWithUTCDateTimeCustom;
+import org.apache.fineract.infrastructure.core.domain.ExternalId;
 import org.apache.fineract.infrastructure.core.service.DateUtils;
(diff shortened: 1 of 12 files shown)


adamsaghy (changes requested):
@taskain7 Please rebase this PR, there is a merge commit!

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/ClientTransaction.java
```diff
@@ -98,24 +99,23 @@ public class ClientTransaction extends AbstractAuditableWithUTCDateTimeCustom {
...
-        final String externalId = null;
         return new ClientTransaction(client, office, paymentDetail, ClientTransactionType.PAY_CHARGE.getValue(), transactionDate, amount,
                 isReversed, externalId, currencyCode);
     }
 
     public static ClientTransaction waiver(final Client client, final Office office, final LocalDate transactionDate, final Money amount,
             final String currencyCode) {
         final boolean isReversed = false;
-        final String externalId = null;
+        final ExternalId externalId = null;
```
adamsaghy: ExternalId by design should never be null. Please use ExternalId.empty() in this case.

adamsaghy (changes requested):
Please see my comment!

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/ClientTransaction.java
taskain7: fixed

adamsaghy (approved):
LGTM

PR #3321, opened 2023-07-21
Fineract-1910 data table query advanced

marta-jankovics:
## Description

## Checklist

### DIFF
--- a/fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/JdbcJavaType.java
+++ b/fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/JdbcJavaType.java
@@ -30,14 +30,14 @@ public enum JdbcJavaType {
 
         @Override
         public Object toJdbcValueImpl(@NotNull DatabaseType dialect, Object value) {
-            return Boolean.TRUE.equals(value) ? 1 : 0;
+            return value == null ? null : (Boolean.TRUE.equals(value) ? 1 : 0);
         }
     },
     BOOLEAN(JavaType.BOOLEAN, new DialectType(JDBCType.BIT), new DialectType(JDBCType.BOOLEAN, null, "BOOL")) { //
 
         @Override
         public Object toJdbcValueImpl(@NotNull DatabaseType dialect, Object value) {
-            return dialect.isMySql() ? (Boolean.TRUE.equals(value) ? 1 : 0) : super.toJdbcValueImpl(dialect, value);
+            return (value != null && dialect.isMySql()) ? (Boolean.TRUE.equals(value) ? 1 : 0) : super.toJdbcValueImpl(dialect, value);
         }
     },
     SMALLINT(JavaType.SHORT, new DialectType(JDBCType.SMALLINT, true), new DialectType(JDBCType.SMALLINT, null, "INT2")), //
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/api/DatatablesApiResource.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/api/DatatablesApiResource.java
@@ -52,11 +52,14 @@ import org.apache.fineract.infrastructure.core.api.ApiParameterHelper;
 import org.apache.fineract.infrastructure.core.data.CommandProcessingResult;
 import org.apache.fineract.infrastructure.core.data.CommandProcessingResultBuilder;
 import org.apache.fineract.infrastructure.core.serialization.ToApiJsonSerializer;
+import org.apache.fineract.infrastructure.core.service.PagedLocalRequest;
 import org.apache.fineract.infrastructure.dataqueries.data.DatatableData;
 import org.apache.fineract.infrastructure.dataqueries.data.GenericResultsetData;
 import org.apache.fineract.infrastructure.dataqueries.service.GenericDataService;
 import org.apache.fineract.infrastructure.dataqueries.service.ReadWriteNonCoreDataService;
 import org.apache.fineract.infrastructure.security.service.PlatformSecurityContext;
+import org.apache.fineract.portfolio.search.data.AdvancedQueryData;
+import org.springframework.data.domain.Page;
 import org.springframework.stereotype.Component;
 
 @Path("/v1/datatables")
@@ -222,6 +225,19 @@ public class DatatablesApiResource {
         final List<JsonObject> result = this.readWriteNonCoreDataService.queryDataTable(datatable, columnFilter, valueFilter,
                 resultColumns);
 
+        return this.toApiJsonSerializer.serializePretty(ApiParameterHelper.prettyPrint(uriInfo.getQueryParameters()), result);
+    }
+
+    @POST
+    @Path("{datatable}/query")
+    @Consumes({ MediaType.APPLICATION_JSON })
+    @Produces({ MediaType.APPLICATION_JSON })
+    @Operation(summary = "Query Data Table values", description = "Query values from a registered data table.")
+    @ApiResponses({
+            @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = List.class))) })
+    public String advancedQuery(@PathParam("datatable") @Parameter(description = "datatable") final String datatable,
+            PagedLocalRequest<AdvancedQueryData> queryRequest, @Context final UriInfo uriInfo) {
+        final Page<JsonObject> result = this.readWriteNonCoreDataService.queryDataTableAdvanced(datatable, queryRequest);
         return this.toApiJsonSerializer.serializePretty(ApiParameterHelper.prettyPrint(uriInfo.getQueryParameters()), result);
     }
 
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/data/EntityTables.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/data/EntityTables.java
@@ -56,15 +56,20 @@ public enum EntityTables {
     private static final Map<String, EntityTables> BY_ENTITY_NAME = Arrays.stream(VALUES)
             .collect(Collectors.toMap(EntityTables::getName, e -> e));
 
+    @NotNull
     private final String name;
+    @NotNull
     private final String apptableName;
 
+    @NotNull
     private final String foreignKeyColumnNameOnDatatable;
+    @NotNull
     private final String refColumn; // referenced column name on apptable
 
     private final ImmutableList<StatusEnum> checkStatuses;
 
-    EntityTables(String name, String apptableName, String foreignKeyColumnNameOnDatatable, String refColumn, StatusEnum... statuses) {
+    EntityTables(@NotNull String name, @NotNull String apptableName, @NotNull String foreignKeyColumnNameOnDatatable,
+            @NotNull String refColumn, StatusEnum... statuses) {
         this.name = name;
         this.apptableName = apptableName;
         this.foreignKeyColumnNameOnDatatable = foreignKeyColumnNameOnDatatable;
@@ -72,22 +77,26 @@ public enum EntityTables {
         this.checkStatuses = statuses == null ? ImmutableList.of() : ImmutableList.copyOf(statuses);
     }
 
-    EntityTables(String name, String foreignKeyColumnNameOnDatatable, String refColumn, StatusEnum... statuses) {
+    EntityTables(@NotNull String name, @NotNull String foreignKeyColumnNameOnDatatable, @NotNull String refColumn, StatusEnum... statuses) {
         this(name, name, foreignKeyColumnNameOnDatatable, refColumn, statuses);
     }
 
+    @NotNull
     public String getName() {
         return name;
     }
 
+    @NotNull
     public String getApptableName() {
         return apptableName;
     }
 
+    @NotNull
     public String getForeignKeyColumnNameOnDatatable() {
         return this.foreignKeyColumnNameOnDatatable;
     }
 
+    @NotNull
     public String getRefColumn() {
         return refColumn;
     }
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
@@ -83,6 +83,7 @@ import org.apache.fineract.infrastructure.core.exception.PlatformServiceUnavaila
 import org.apache.fineract.infrastructure.core.serialization.DatatableCommandFromApiJsonDeserializer;
 import org.apache.fineract.infrastructure.core.serialization.FromJsonHelper;
 import org.apache.fineract.infrastructure.core.service.MathUtil;
+import org.apache.fineract.infrastructure.core.service.PagedLocalRequest;
 import org.apache.fineract.infrastructure.core.service.database.DatabaseSpecificSQLGenerator;
 import org.apache.fineract.infrastructure.core.service.database.DatabaseType;
 import org.apache.fineract.infrastructure.core.service.database.DatabaseTypeResolver;
@@ -101,11 +102,17 @@ import org.apache.fineract.infrastructure.security.service.PlatformSecurityConte
 import org.apache.fineract.infrastructure.security.service.SqlInjectionPreventerService;
 import org.apache.fineract.infrastructure.security.utils.ColumnValidator;
 import org.apache.fineract.infrastructure.security.utils.SQLInjectionValidator;
+import org.apache.fineract.portfolio.search.data.AdvancedQueryData;
+import org.apache.fineract.portfolio.search.data.ColumnFilterData;
 import org.apache.fineract.portfolio.search.service.SearchUtil;
 import org.apache.fineract.useradministration.domain.AppUser;
 import org.springframework.dao.DataAccessException;
 import org.springframework.dao.DataIntegrityViolationException;
 import org.springframework.dao.EmptyResultDataAccessException;
+import org.springframework.data.domain.Page;
+import org.springframework.data.domain.PageRequest;
+import org.springframework.data.domain.Sort;
+import org.springframework.data.support.PageableExecutionUtils;
 import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 import org.springframework.jdbc.support.GeneratedKeyHolder;
@@ -199,20 +206,16 @@ public class ReadWriteNonCoreDataServiceImpl implements ReadWriteNonCoreDataServ
     public List<JsonObject> queryDataTable(@NotNull String datatable, @NotNull String columnName, String columnValueString,
             @NotNull String resultColumnsString) {
         datatable = validateDatatableRegistered(datatable);
-        Map<String, ResultsetColumnHeaderData> columnHeaders = SearchUtil
+        Map<String, ResultsetColumnHeaderData> headersByName = SearchUtil
                 .mapHeadersToName(genericDataService.fillResultsetColumnHeaders(datatable));
 
-        List<ApiParameterError> errors = new ArrayList<>();
-        columnName = SearchUtil.validateToJdbcColumn(columnName, columnHeaders, errors, false);
         List<String> resultColumns = asList(resultColumnsString.split(","));
-        List<String> selectColumns = SearchUtil.validateToJdbcColumns(resultColumns, columnHeaders, errors, false);
-        if (!errors.isEmpty()) {
-            throw new PlatformApiDataValidationException(errors);
-        }
+        List<String> selectColumns = SearchUtil.validateToJdbcColumnNames(resultColumns, headersByName, false);
+        ResultsetColumnHeaderData column = SearchUtil.validateToJdbcColumn(columnName, headersByName, false);
 
-        Object columnValue = SearchUtil.parseAndValidateJdbcColumnValue(columnName, columnValueString, columnHeaders, false, sqlGenerator);
+        Object columnValue = SearchUtil.parseJdbcColumnValue(column, columnValueString, null, null, null, false, sqlGenerator);
         String sql = sqlGenerator.buildSelect(selectColumns, null, false) + " " + sqlGenerator.buildFrom(datatable, null, false) + " WHERE "
-                + EQ.formatPlaceholder(sqlGenerator, columnName, null);
+                + EQ.formatPlaceholder(sqlGenerator, column.getColumnName(), 1, null);
         SqlRowSet rowSet = jdbcTemplate.queryForRowSet(sql, columnValue);
 
         List<JsonObject> results = new ArrayList<>();
@@ -220,6 +223,110 @@ public class ReadWriteNonCoreDataServiceImpl implements ReadWriteNonCoreDataServ
             SearchUtil.extractJsonResult(rowSet, selectColumns, resultColumns, results);
         }
         return results;
+    }
+
+    @Override
+    public Page<JsonObject> queryDataTableAdvanced(@NotNull String datatable, @NotNull PagedLocalRequest<AdvancedQueryData> pagedRequest) {
+        datatable = validateDatatableRegistered(datatable);
+        context.authenticatedUser().validateHasDatatableReadPermission(datatable);
+
+        AdvancedQueryData request = pagedRequest.getRequest().orElseThrow();
+        dataTableValidator.validateTableSearch(request);
+
+        Map<String, ResultsetColumnHeaderData> headersByName = SearchUtil
+                .mapHeadersToName(genericDataService.fillResultsetColumnHeaders(datatable));
+        String pkColumn = SearchUtil.getFiltered(headersByName.values(), ResultsetColumnHeaderData::getIsColumnPrimaryKey).getColumnName();
+
+        List<ColumnFilterData> columnFilters = request.getNonNullFilters();
+        columnFilters.forEach(e -> e.setColumn(SearchUtil.validateToJdbcColumnName(e.getColumn(), headersByName, false)));
+
+        List<String> resultColumns = request.getNonNullResultColumns();
+        List<String> selectColumns;
+        if (resultColumns.isEmpty()) {
+            resultColumns.add(pkColumn);
+            selectColumns = new ArrayList<>();
+            selectColumns.add(pkColumn);
+        } else {
+            selectColumns = SearchUtil.validateToJdbcColumnNames(resultColumns, headersByName, false);
+        }
+        PageRequest pageable = pagedRequest.toPageable();
+        PageRequest sortPageable;
+        if (pageable.getSort().isSorted()) {
+            List<Sort.Order> orders = pageable.getSort().toList();
+            sortPageable = pageable.withSort(Sort.by(orders.stream()
+                    .map(e -> e.withProperty(SearchUtil.validateToJdbcColumnName(e.getProperty(), headersByName, false))).toList()));
+        } else {
+            pageable = pageable.withSort(Sort.Direction.DESC, pkColumn);
+            sortPageable = pageable;
+        }
+
+        String dateFormat = pagedRequest.getDateFormat();
+        String dateTimeFormat = pagedRequest.getDateTimeFormat();
+        Locale locale = pagedRequest.getLocaleObject();
+
+        String select = sqlGenerator.buildSelect(selectColumns, null, false);
+        String from = " " + sqlGenerator.buildFrom(datatable, null, false);
+        StringBuilder where = new StringBuilder();
+        ArrayList<Object> params = new ArrayList<>();
+        SearchUtil.buildQueryCondition(columnFilters, where, params, null, headersByName, dateFormat, dateTimeFormat, locale, false,
+                sqlGenerator);
+
+        List<JsonObject> results = new ArrayList<>();
+        Object[] args = params.toArray();
+
+        // Execute the count Query
+        String countQuery = "SELECT COUNT(*)" + from + where;
+        Integer totalElements = jdbcTemplate.queryForObject(countQuery, Integer.class, args);
+        if (totalElements == null || totalElements == 0) {
+            return PageableExecutionUtils.getPage(results, pageable, () -> 0);
+        }
+
+        StringBuilder query = new StringBuilder().append(select).append(from).append(where);
+        query.append(" ").append(sqlGenerator.buildOrderBy(sortPageable.getSort().toList(), null, false));
+        if (pageable.isPaged()) {
+            query.append(" ").append(sqlGenerator.limit(pageable.getPageSize(), (int) pageable.getOffset()));
+        }
+
+        SqlRowSet rowSet = jdbcTemplate.queryForRowSet(query.toString(), args);
+
+        while (rowSet.next()) {
+            SearchUtil.extractJsonResult(rowSet, selectColumns, resultColumns, results);
+        }
+        return PageableExecutionUtils.getPage(results, pageable, () -> totalElements);
+    }
+
+    @Override
+    public boolean buildDataQueryEmbedded(@NotNull EntityTables entityTable, @NotNull String datatable, @NotNull AdvancedQueryData request,
+            @NotNull List<String> selectColumns, @NotNull StringBuilder select, @NotNull StringBuilder from, @NotNull StringBuilder where,
+            @NotNull List<Object> params, String mainAlias, String alias, String dateFormat, String dateTimeFormat, Locale locale) {
+        List<String> resultColumns = request.getResultColumns();
+        List<ColumnFilterData> columnFilters = request.getColumnFilters();
+        if ((resultColumns == null || resultColumns.isEmpty()) && (columnFilters == null || columnFilters.isEmpty())) {
+            return false;
+        }
+
+        datatable = validateDatatableRegistered(datatable);
+        context.authenticatedUser().validateHasDatatableReadPermission(datatable);
+
+        Map<String, ResultsetColumnHeaderData> headersByName = SearchUtil
+                .mapHeadersToName(genericDataService.fillResultsetColumnHeaders(datatable));
+
+        List<String> thisSelectColumns = SearchUtil.validateToJdbcColumnNames(resultColumns, headersByName, true);
+        if (columnFilters != null) {
+            columnFilters.forEach(e -> e.setColumn(SearchUtil.validateToJdbcColumnName(e.getColumn(), headersByName, false)));
+        }
+
+        select.append(sqlGenerator.buildSelect(thisSelectColumns, alias, true));
+        selectColumns.addAll(thisSelectColumns);
+
+        String joinType = "LEFT";
+        if (SearchUtil.buildQueryCondition(columnFilters, where, params, alias, headersByName, dateFormat, dateTimeFormat, locale, true,
+                sqlGenerator)) {
+            joinType = null; // INNER
+        }
+        from.append(sqlGenerator.buildJoin(datatable, alias, entityTable.getForeignKeyColumnNameOnDatatable(), mainAlias,
+                entityTable.getRefColumn(), joinType));
+        return true;
     }
 
     private void logAsErrorUnexpectedDataIntegrityException(final Exception dve) {
@@ -1830,7 +1937,7 @@ public class ReadWriteNonCoreDataServiceImpl implements ReadWriteNonCoreDataServ
                         columnHeaderUnderscored = this.genericDataService.replace(columnHeader.getColumnName(), space, underscore);
                         if (queryParamColumnUnderscored.equalsIgnoreCase(columnHeaderUnderscored)) {
                             pValue = queryParams.get(key);
-                            validatedValue = SearchUtil.parseAndValidateColumnValue(columnHeader, pValue, dateFormat,
+                            validatedValue = SearchUtil.parseColumnValue(columnHeader, pValue, dateFormat, dateFormat,
                                     clientApplicationLocale, true, sqlGenerator);
                             affectedColumns.put(columnHeader.getColumnName(), validatedValue);
                             notFound = false;
@@ -1890,7 +1997,7 @@ public class ReadWriteNonCoreDataServiceImpl implements ReadWriteNonCoreDataServ
                 + " JOIN m_entity_datatable_check edc ON edc.x_registered_table_name = xrt.registered_table_name"
                 + " WHERE edc.x_registered_table_name = '" + datatableName + "'";
         final Long count = this.jdbcTemplate.queryForObject(sql, Long.class);
-        return count > 0;
+        return count != null && count > 0;
     }
 
     // --- DbUtils ---
@@ -1940,8 +2047,8 @@ public class ReadWriteNonCoreDataServiceImpl implements ReadWriteNonCoreDataServ
 
     private boolean isRegisteredDataTable(final String datatable) {
         final String sql = "SELECT COUNT(application_table_name) FROM " + TABLE_REGISTERED_TABLE + " WHERE registered_table_name = ?";
-        final int count = jdbcTemplate.queryForObject(sql, Integer.class, datatable);
-        return count > 0;
+        final Integer count = jdbcTemplate.queryForObject(sql, Integer.class, datatable);
+        return count != null && count > 0;
     }
 
     private void validateDataTableExists(final String datatableName) {
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/SavingsAccountTransactionsApiResource.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/SavingsAccountTransactionsApiResource.java
@@ -18,6 +18,7 @@
  */
 package org.apache.fineract.portfolio.savings.api;
 
+import com.google.gson.JsonObject;
 import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.Parameter;
 import io.swagger.v3.oas.annotations.media.ArraySchema;
@@ -39,12 +40,14 @@ import jakarta.ws.rs.core.MediaType;
 import jakarta.ws.rs.core.UriInfo;
 import java.math.BigDecimal;
 import java.util.Collection;
+import java.util.List;
 import java.util.Locale;
 import lombok.RequiredArgsConstructor;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.fineract.commands.domain.CommandWrapper;
 import org.apache.fineract.commands.service.CommandWrapperBuilder;
 import org.apache.fineract.commands.service.PortfolioCommandSourceWritePlatformService;
+import org.apache.fineract.infrastructure.core.api.ApiParameterHelper;
 import org.apache.fineract.infrastructure.core.api.ApiRequestParameterHelper;
 import org.apache.fineract.infrastructure.core.data.CommandProcessingResult;
 import org.apache.fineract.infrastructure.core.exception.PlatformDataIntegrityException;
@@ -52,6 +55,7 @@ import org.apache.fineract.infrastructure.core.exception.UnrecognizedQueryParamE
 import org.apache.fineract.infrastructure.core.serialization.ApiRequestJsonSerializationSettings;
 import org.apache.fineract.infrastructure.core.serialization.DefaultToApiJsonSerializer;
 import org.apache.fineract.infrastructure.core.serialization.JsonParserHelper;
+import org.apache.fineract.infrastructure.core.service.PagedLocalRequest;
 import org.apache.fineract.infrastructure.security.service.PlatformSecurityContext;
 import org.apache.fineract.portfolio.paymenttype.data.PaymentTypeData;
 import org.apache.fineract.portfolio.paymenttype.service.PaymentTypeReadPlatformService;
@@ -60,6 +64,7 @@ import org.apache.fineract.portfolio.savings.SavingsApiConstants;
 import org.apache.fineract.portfolio.savings.data.SavingsAccountTransactionData;
 import org.apache.fineract.portfolio.savings.service.SavingsAccountReadPlatformService;
 import org.apache.fineract.portfolio.savings.service.search.SavingsAccountTransactionSearchService;
+import org.apache.fineract.portfolio.search.data.AdvancedQueryRequest;
 import org.apache.fineract.portfolio.search.data.TransactionSearchRequest;
 import org.springframework.dao.CannotAcquireLockException;
 import org.springframework.data.domain.Page;
@@ -157,6 +162,19 @@ public class SavingsAccountTransactionsApiResource {
                 .pageable(offset, limit, orderBy, sortOrder);
         Page<SavingsAccountTransactionData> transactionsData = transactionsSearchService.searchTransactions(savingsId, searchParameters);
         return toApiJsonSerializer.serialize(transactionsData);
+    }
+
+    @POST
+    @Path("query")
+    @Consumes({ MediaType.APPLICATION_JSON })
+    @Produces({ MediaType.APPLICATION_JSON })
+    @Operation(summary = "Advanced search Savings Account Transactions")
+    @ApiResponses({
+            @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = List.class))) })
+    public String advancedQuery(@PathParam("savingsId") @Parameter(description = "savingsId") final Long savingsId,
+            PagedLocalRequest<AdvancedQueryRequest> queryRequest, @Context final UriInfo uriInfo) {
+        final Page<JsonObject> result = transactionsSearchService.queryAdvanced(savingsId, queryRequest);
+        return this.toApiJsonSerializer.serializePretty(ApiParameterHelper.prettyPrint(uriInfo.getQueryParameters()), result);
     }
 
     @POST
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/search/service/SearchUtil.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/search/service/SearchUtil.java
@@ -25,7 +25,6 @@ import static org.apache.fineract.infrastructure.core.service.DateUtils.DEFAULT_
 import static org.apache.fineract.infrastructure.dataqueries.api.DataTableApiConstant.API_FIELD_MANDATORY;
 import static org.apache.fineract.portfolio.search.SearchConstants.API_PARAM_COLUMN;
 
-import com.google.common.base.Predicate;
 import com.google.gson.JsonObject;
 import jakarta.validation.constraints.NotNull;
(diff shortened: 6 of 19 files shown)


marta-jankovics:
Review only, no tests yet, to be added - tests added

galovics:
Squash your commits pls and format the PR title properly (FINERACT-1234 Something)

galovics:
@marta-jankovics please change back the javadoc/comments where you've only changed formatting. It's polluting the PR and makes it hard to review.

Also, do we need a 4k line change in this specific PR, just to extend the datatable functionality? Let's make the changes as small as possible unless we have a good reason to change it in its core.

Until that, it's way too big to review.

marta-jankovics:
Sorry I did not know that merge the PRs can not be done with squash and merge here. I will then squash the commits (most of them are merge).
'just to extend the datatable functionality' it is a questionable statement, maybe better to ask first. 
Anyway I see some merge commits are represented as changed files which should not be, I'll check.

About the formatting: I did not change anything, spotless did. I reverted 300+ changes after spotless run but I do not check it again. Would be nice to create a check-style which does not touch the existing code (and does not ruin the formatted enums, does not add empty line after each method start, allows 200 chars line length...)

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/DatabaseSpecificSQLGenerator.java
```diff
@@ -36,14 +42,26 @@ public DatabaseSpecificSQLGenerator(DatabaseTypeResolver databaseTypeResolver) {
     }
 
     public String escape(String arg) {
-        if (databaseTypeResolver.isMySQL()) {
+        return escape(databaseTypeResolver.databaseType(), arg);
+    }
+
+    public static String escape(DatabaseType dialect, String arg) {
```
galovics: Why do we need to pass the dialect here but not everywhere else?

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/DatabaseSpecificSQLGenerator.java
```diff
@@ -189,4 +207,49 @@ public String castJson(String sql) {
             throw new IllegalStateException("Database type is not supported for casting to json " + databaseTypeResolver.databaseType());
         }
     }
+
+    public static String alias(@NotNull String field, String alias) {
```
galovics: How are these methods below specific to the database type?

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/DatabaseSpecificSQLGenerator.java
```diff
@@ -36,14 +42,26 @@ public DatabaseSpecificSQLGenerator(DatabaseTypeResolver databaseTypeResolver) {
...
             return format("\"%s\"", arg);
         }
         return arg;
     }
 
+    public String formatValue(JdbcJavaType columnType, String value) {
+        return formatValue(databaseTypeResolver.databaseType(), columnType, value);
+    }
+
+    public static String formatValue(DatabaseType dialect, JdbcJavaType columnType, String value) {
```
galovics: How is this specific again to a dialect?

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/DatabaseTypeResolver.java
```diff
@@ -58,10 +58,10 @@ public DatabaseType databaseType() {
     }
 
     public boolean isPostgreSQL() {
-        return DatabaseType.POSTGRESQL.equals(currentDatabaseType.get());
+        return DatabaseType.POSTGRESQL == currentDatabaseType.get();
```
galovics: Why was this necessary? It's not an issue, I'm just curious.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/data/SqlOperator.java
```diff
@@ -0,0 +1,197 @@
...
+
+    protected String formatPlaceholderImpl(String definition, String placeholder) {
+        if (!isPlaceholderSupported()) {
+            throw new UnsupportedOperationException("Placeholder is not supported for this operator");
+        }
+        return paramCount == 0 ? format("%s %s", definition, symbol) : format("%s %s %s", definition, symbol, placeholder);
+    }
+
+    public void validateValues(String... values) {
+        if (values == null ? paramCount != 0 : (paramCount < 0 ? values.length < -paramCount : values.length != paramCount)) {
```
galovics: This condition is hard to understand unfortunately, can't we do it differently?

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/data/SqlOperator.java
```diff
@@ -0,0 +1,197 @@
...
+            throw new PlatformServiceUnavailableException("error.msg.database.operator.invalid",
+                    "Number of parameters " + Arrays.toString(values) + " must be " + Math.abs(paramCount) + " on " + this);
+        }
+    }
+
+    public void validateValues(List<String> values) {
+        int size = values == null ? 0 : values.size();
+        if (paramCount < 0 ? size < -paramCount : size != paramCount) {
```
galovics: This condition is hard to understand unfortunately, can't we do it differently?

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/data/SqlOperator.java
```diff
@@ -0,0 +1,197 @@
...
+import java.util.stream.Collectors;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import org.apache.fineract.infrastructure.core.exception.PlatformServiceUnavailableException;
+import org.apache.fineract.infrastructure.core.service.database.DatabaseType;
+import org.apache.fineract.infrastructure.core.service.database.JdbcJavaType;
+
+@AllArgsConstructor
+@Getter
+public enum SqlOperator {
```
galovics: I think this class is way too complicated and trying to solve a lot of things which you don't need to solve your needs.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/JavaType.java
```diff
@@ -0,0 +1,352 @@
...
+import java.util.Date;
+import java.util.Locale;
+import java.util.Map;
+import java.util.stream.Collectors;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.fineract.infrastructure.core.exception.PlatformServiceUnavailableException;
+import org.apache.fineract.organisation.monetary.domain.MoneyHelper;
+import org.springframework.data.domain.Persistable;
+
+public enum JavaType {
```
galovics: Just like I said with the SqlOperator class, this is way way way too complex for a single change. Let's shoot for the minimum you need instead of trying to solve all the world problems in a single breath.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/JdbcJavaType.java
```diff
@@ -0,0 +1,321 @@
+package org.apache.fineract.infrastructure.core.service.database;
+
+import jakarta.validation.constraints.NotNull;
+import java.sql.JDBCType;
+import org.apache.fineract.infrastructure.core.exception.PlatformServiceUnavailableException;
+
+public enum JdbcJavaType {
```
galovics: Same as for JavaType.

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
```diff
@@ -558,8 +558,8 @@ private Loan(final String accountNo, final Client client, final Group group, fin
         this.createStandingInstructionAtDisbursement = createStandingInstructionAtDisbursement;
 
         /*
-         * During loan origination stage and before loan is approved principal_amount, approved_principal and
-         * principal_amount_demanded will same amount and that amount is same as applicant loan demanded amount.
+         * During loan origination stage and before loan is approved principal_amount, approved_principal and principal_amount_demanded will same amount and that amount is same as applicant loan
```
galovics: Please revert these unnecessary changes. I'm fine with javadoc changes if needed but these are polluting the PR.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/serialization/DatatableCommandFromApiJsonDeserializer.java
```diff
@@ -34,161 +42,101 @@
...
-            Arrays.asList(NAME, TYPE, LENGTH, MANDATORY, CODE, UNIQUE, INDEXED));
-    private static final Set<String> SUPPORTED_PARAMETERS_FOR_UPDATE = new HashSet<>(
-            Arrays.asList(APPTABLE_NAME, ENTITY_SUB_TYPE, CHANGE_COLUMNS, ADD_COLUMNS, DROP_COLUMNS));
-    private static final Set<String> SUPPORTED_PARAMETERS_FOR_ADD_COLUMNS = new HashSet<>(
-            Arrays.asList(NAME, TYPE, LENGTH, MANDATORY, AFTER, CODE, UNIQUE, INDEXED));
-    private static final Set<String> SUPPORTED_PARAMETERS_FOR_CHANGE_COLUMNS = new HashSet<>(
-            Arrays.asList(NAME, NEW_NAME, LENGTH, MANDATORY, AFTER, CODE, NEW_CODE, UNIQUE, INDEXED));
-    private static final Set<String> SUPPORTED_PARAMETERS_FOR_DROP_COLUMNS = new HashSet<>(List.of(NAME));
-    private static final Object[] SUPPORTED_COLUMN_TYPES = { STRING, NUMBER, BOOLEAN, DECIMAL, DATE, DATETIME, TEXT, DROPDOWN };
+    private static final Set<String> SUPPORTED_PARAMETERS_FOR_CREATE = Set.of(API_PARAM_DATATABLE_NAME, API_PARAM_SUBTYPE,
```
galovics: Why was it necessary to rename it? Your PR is already way too big, this is just bringing more unnecessary changes into scope.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/serialization/DatatableCommandFromApiJsonDeserializer.java
```diff
@@ -34,161 +42,101 @@
...
     private final FromJsonHelper fromApiJsonHelper;
+    private final DatabaseTypeResolver databaseTypeResolver;
 
     @Autowired
-    public DatatableCommandFromApiJsonDeserializer(final FromJsonHelper fromApiJsonHelper) {
+    public DatatableCommandFromApiJsonDeserializer(final FromJsonHelper fromApiJsonHelper, DatabaseTypeResolver databaseTypeResolver) {
         this.fromApiJsonHelper = fromApiJsonHelper;
-    }
-
-    private void validateType(final DataValidatorBuilder baseDataValidator, final JsonElement column) {
```
galovics: Why was this method moved down in the class? Unnecessary change again.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/serialization/DatatableCommandFromApiJsonDeserializer.java
```diff
@@ -34,161 +42,101 @@
...
+                        false);
+                baseDataValidator.reset().parameter(API_FIELD_INDEXED).value(indexed).ignoreIfNull().notBlank().isOneOfTheseValues(true,
+                        false);
             }
         }
 
         throwExceptionIfValidationWarningsExist(dataValidationErrors);
     }
 
-    private void validateEntitySubType(final DataValidatorBuilder baseDataValidator, final JsonElement element, final String apptableName) {
```
galovics: Same with this. This has been moved down unnecessarily.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/api/DatatablesApiResource.java
```diff
@@ -225,6 +228,18 @@ public String queryValues(@PathParam("datatable") @Parameter(description = "data
...
+    @POST
+    @Path("{datatable}/query")
+    @Produces({ MediaType.APPLICATION_JSON })
+    @Operation(summary = "Query Data Table values", description = "Query values from a registered data table.")
+    @ApiResponses({
+            @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = List.class))) })
+    public String advancedQuery(@PathParam("datatable") @Parameter(description = "datatable") final String datatable,
+            PagedRequest<AdvancedQueryData> queryRequest, @Context final UriInfo uriInfo) {
+        final Page<JsonObject> result = this.readWriteNonCoreDataService.queryDataTableAdvanced(datatable, queryRequest);
+        return this.toApiJsonSerializer.serializePretty(ApiParameterHelper.prettyPrint(uriInfo.getQueryParameters()), result);
```
galovics, line 241: Since this is a new API, you should simply return a Page<....> and it'll be serialized automatically. That way you can drop all the openAPI annotations as well.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/EntityDatatableChecksReadPlatformServiceImpl.java
```diff
@@ -174,20 +174,22 @@ private List<DatatableChecksData> getDataTables() {
...
             final String tableName = rs.getString("tableName");
 
             return new DatatableChecksData(entity, tableName);
         }
 
         public String schema() {
-            return " t.application_table_name as entity, t.registered_table_name as tableName " + " from x_registered_table t "
-                    + " where application_table_name IN( 'm_client','m_group','m_savings_account','m_loan')";
+            String[] values = EntityTables.getFiltered(EntityTables::hasCheck).stream().map(EntityTables::getName).toArray(String[]::new);
+            return SELECT_FROM + SqlOperator.IN.formatSql(databaseTypeResolver.databaseType(), JdbcJavaType.VARCHAR,
```
galovics: This SqlOperator doesn't seem like to be the best idea.. very confusing to use. Why don't we simply do the "IN" condition as a string here? Nothing seems to hard about that as far as I can see.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/EntityDatatableChecksWritePlatformServiceImpl.java
```diff
@@ -105,11 +104,11 @@ public CommandProcessingResult createCheck(final JsonCommand command) {
                     throw new EntityDatatableCheckAlreadyExistsException(entity, status, datatableName);
                 }
             } else {
-                EntityTables entityTable = EntityTables.fromName(entity);
-                if (EntityTables.LOAN == entityTable) {
+                EntityTables entityTable = EntityTables.fromEntityName(entity);
+                if (entityTable == EntityTables.LOAN) {
```
galovics: So why did we flip the condition? To be honest what you did is also my preference but it's unnecessary. The result is the same yet it's adding an extra change to your PR.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadReportingServiceImpl.java
```diff
@@ -383,10 +382,8 @@ public String schema(final Long reportId) {
             return sql;
 
             /*
-             * used to only return reports that the use can run as done in report UI but not necessary as there is a
-             * read_report permission which should give user access to look all reports + " where exists" +
-             * " (select 'f'" + " from m_appuser_role ur " + " join m_role r on r.id = ur.role_id" +
-             * " left join m_role_permission rp on rp.role_id = r.id" +
+             * used to only return reports that the use can run as done in report UI but not necessary as there is a read_report permission which should give user access to look all reports +
```
galovics: Why these changes again?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
```diff
@@ -19,31 +19,50 @@
...
-import java.time.LocalDate;
-import java.time.LocalDateTime;
-import java.time.format.DateTimeFormatter;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
galovics: I'm not sure how this didn't break but let's not use * imports. Spotless should break on these.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanAccountDomainServiceJpa.java
```diff
@@ -173,9 +173,8 @@ public LoanTransaction makeRepayment(final LoanTransactionType repaymentTransact
         // TODO: Is it required to validate transaction date with meeting dates
         // if repayments is synced with meeting?
         /*
-         * if(loan.isSyncDisbursementWithMeeting()){ // validate actual disbursement date against meeting date
-         * CalendarInstance calendarInstance = this.calendarInstanceRepository.findCalendarInstaneByLoanId
-         * (loan.getId(), CalendarEntityType.LOANS.getValue()); this.loanEventApiJsonValidator
+         * if(loan.isSyncDisbursementWithMeeting()){ // validate actual disbursement date against meeting date CalendarInstance calendarInstance =
```
galovics: Not needed. Please revert.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/transactionprocessor/AbstractLoanRepaymentScheduleTransactionProcessor.java
```diff
@@ -48,8 +48,7 @@
 import org.apache.fineract.portfolio.loanaccount.domain.transactionprocessor.impl.InterestPrincipalPenaltyFeesOrderLoanRepaymentScheduleTransactionProcessor;
 
 /**
- * Abstract implementation of {@link LoanRepaymentScheduleTransactionProcessor} which is more convenient for concrete
- * implementations to extend.
+ * Abstract implementation of {@link LoanRepaymentScheduleTransactionProcessor} which is more convenient for concrete implementations to extend.
```
galovics: Not needed. Please revert.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/SavingsAccountTransactionsApiResource.java
```diff
@@ -137,6 +141,20 @@ public String searchTransactions(@PathParam("savingsId") @Parameter(description
...
+    @Consumes({ MediaType.APPLICATION_JSON })
+    @Produces({ MediaType.APPLICATION_JSON })
+    @Operation(summary = "Search Savings Account Transactions")
+    @ApiResponses({
+            @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = List.class))) })
+    public String advancedQuery(@PathParam("savingsId") @Parameter(description = "savingsId") final Long savingsId,
+            PagedRequest<AdvancedQueryRequest> queryRequest, @Context final UriInfo uriInfo) {
+        final org.springframework.data.domain.Page<JsonObject> result = transactionsSearchServiceImpl.queryAdvanced(savingsId,
+                queryRequest);
+        return this.toApiJsonSerializer.serializePretty(ApiParameterHelper.prettyPrint(uriInfo.getQueryParameters()), result);
```
galovics, line 177: As I said for the datatables API, you can go with Page<...> since this is a new API.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/search/SavingsTransactionSearch.java
```diff
@@ -43,7 +43,7 @@ public static class Filters {
     @Data
     public static class RangeFilter<T> {
 
-        private RangeOperator operator;
+        private SqlOperator operator;
```
galovics: Now this I don't like. We're exposing "SqlOperators" to higher layers (i.e.API) which shouldn't happen.

fineract-provider/src/main/java/org/apache/fineract/portfolio/search/service/SearchUtil.java
```diff
@@ -0,0 +1,266 @@
...
+import org.apache.fineract.infrastructure.core.service.database.DatabaseType;
+import org.apache.fineract.infrastructure.core.service.database.JdbcJavaType;
+import org.apache.fineract.infrastructure.dataqueries.data.ResultsetColumnHeaderData;
+import org.apache.fineract.infrastructure.security.utils.SQLInjectionValidator;
+import org.apache.fineract.portfolio.search.data.AdvancedQueryData;
+import org.apache.fineract.portfolio.search.data.ColumnFilterData;
+import org.apache.fineract.portfolio.search.data.FilterData;
+import org.springframework.jdbc.support.rowset.SqlRowSet;
+
+public class SearchUtil {
```
galovics: Woah, this class is really really complex and impossible to review. Can we separate pieces out according to responsibilities?

galovics (changes requested):
There are a few things still to do. Let's make the PR less complex, less polluted with unnecessary changes.

Also, I see zero tests added yet there are thousands of line changes in the codebase. Please cover those. 

Thanks.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/data/EntityTables.java
```diff
@@ -18,88 +18,114 @@
...
-            lookup.put(d.getName(), d);
-        }
-    }
+    CLIENT("m_client", "client_id", "id", CREATE, ACTIVATE, CLOSE), //
+    GROUP("m_group", "group_id", "id", CREATE, ACTIVATE, CLOSE), //
+    CENTER("m_center", "m_group", "center_id", "id"), //
+    OFFICE("m_office", "office_id", "id"), //
+    LOAN_PRODUCT("m_product_loan", "product_loan_id", "id"), //
+    LOAN("m_loan", "loan_id", "id", CREATE, APPROVE, DISBURSE, WITHDRAWN, REJECTED, WRITE_OFF), //
+    SAVING_PRODUCT("m_savings_product", "savings_product_id", "id"), //
```
adamsaghy: Please correct the typos for savings. It should be:
SAVINGS
SAVINGS_PRODUCT
SAVINGS_TRANSACTION

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/DatabaseSpecificSQLGenerator.java
marta-jankovics: because it is static

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/data/EntityTables.java
marta-jankovics: I was thinking on this, but SAVING and SAVING_PRODUCT were already here, and if I would change this, then it would mean several other classes to be changed, which I wanted to avoid. So I just followed the naming with SAVING_TRANSACTION.
I can rename of course, please let me know if I should do that and modify the depending classes.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/DatabaseSpecificSQLGenerator.java
marta-jankovics: Do not understand this question, of course formatting a value is dialect dependent.

fineract-provider/src/main/java/org/apache/fineract/portfolio/search/service/SearchUtil.java
marta-jankovics: What do you mean? A util class with 10 static methods, all search related is not complex.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanAccountDomainServiceJpa.java
marta-jankovics: Spotless

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/transactionprocessor/AbstractLoanRepaymentScheduleTransactionProcessor.java
marta-jankovics: spotless

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/data/EntityTables.java
adamsaghy: It was SAVINGS_PRODUCT before your change :/
I would not change them if not necessary, but if you want to have them on the same naming convention, i am fine.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/search/SavingsTransactionSearch.java
marta-jankovics: I like this. Why can't sql operators be exposed to higher levels??? (Actually it was exposed before but as simple string or RangeOparator)
I was thinking on even using graphQL, but this was not complex enough for that.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/data/EntityTables.java
marta-jankovics: It was SAVING and SAVING_PRODUCT, but Muthu just renamed them with the PR which was merged lately. Anyway I'll check how many classes are involved, probably not too much if the other PR renamed them already.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/DatabaseSpecificSQLGenerator.java
galovics: This class is a spring bean, let's not make static methods on it but use it as a proper component.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/DatabaseSpecificSQLGenerator.java
galovics: I dont see that, the parameter is not even used.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/SavingsAccountTransactionsApiResource.java
marta-jankovics: I do not know where you said this already, but this is a v1 API, and I learned that only v2 APIs can go with Object response. Additionally the response here is dynamic.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/DatabaseSpecificSQLGenerator.java
marta-jankovics: I do not agree with your opinion.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
```diff
@@ -2274,11 +2039,72 @@ public Long countDatatableEntries(final String datatableName, final Long appTabl
...
+        if (name == null || name.isEmpty()) {
+            throw new PlatformDataIntegrityException("error.msg.datatables.datatable.null.name", "Data table name must not be blank.");
+        } else if (!name.matches(DATATABLE_NAME_REGEX_PATTERN)) {
+            throw new PlatformDataIntegrityException("error.msg.datatables.datatable.invalid.name.regex", "Invalid data table name.", name);
+        }
+        SQLInjectionValidator.validateSQLInput(name);
+    }
+
+    private String validateDatatable(String datatable) {
+        datatable = SearchUtil.camelToSnake(datatable);
```
adamsaghy: I dont think this is okay. Having automatically convert datatable name, column names are breaking backward changes. If there is a datatable that already existing or there are column names which was not following the snake case naming convention, they will not be accessible / supported anymore..
What do you think?

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/JavaType.java
marta-jankovics: Well, this class is very useful and I could not solve this issue so quickly without this. Maybe you could a bit value what I did. This enum is not complex, I think it is very clear and easy to understand.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/JdbcJavaType.java
marta-jankovics: Same answer, thanks

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/api/DatatablesApiResource.java
marta-jankovics: I answered this request once, please let me know the rules, because so far I learned that only v2 APIs can use object response by decision, and this one was added as v1 API.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/EntityDatatableChecksReadPlatformServiceImpl.java
marta-jankovics: I do not know what to answer to that use string literals was better. I do not agree at all, but I am open to discuss.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/EntityDatatableChecksWritePlatformServiceImpl.java
marta-jankovics: This was a conflict between me and Muthu, we were working on the same code, and I resolved this way.

marta-jankovics (commented):
@vidakovic Could you please review org.apache.fineract.infrastructure.security.utils.SQLInjectionValidator
SQL_PATTERN extended by these characters: {,},[,],+
We should support json value in dynamic insert statements.
Please confirm that it is ok from security perspective.

Tried to use your solution from
https://github.com/apache/fineract/blame/1.8.4/fineract-provider/src/main/java/org/apache/fineract/infrastructure/security/utils/SQLInjectionValidator.java
and introduced INJECTION_PATTERNS as black patterns, but ReadWriteNonCoreDataServiceImplTest failed, as "1; DROP TABLE m_loan; SELECT" in a select query was not detected as injection.
Added a TODO FINERACT-1870 so you will add the final solution.
Thank you

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/DateUtils.java
```diff
@@ -86,4 +89,106 @@ public static LocalDate getBusinessLocalDate() {
     public static long getDifferenceInDays(final LocalDate localDateBefore, final LocalDate localDateAfter) {
         return DAYS.between(localDateBefore, localDateAfter);
     }
+
+    public static int compareToBusinessDate(LocalDate date) {
```
adamsaghy: unused

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/DateUtils.java
```diff
@@ -86,4 +89,106 @@ public static LocalDate getBusinessLocalDate() {
...
+        return compare(date, getBusinessLocalDate());
+    }
+
+    public static boolean isEqualBusinessDate(LocalDate date) {
```
adamsaghy: unused

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/DateUtils.java
```diff
@@ -86,4 +89,106 @@ public static LocalDate getBusinessLocalDate() {
...
+        return isEqual(date, getBusinessLocalDate());
+    }
+
+    public static boolean isBeforeBusinessDate(LocalDate date) {
```
adamsaghy: unused

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/DateUtils.java
```diff
@@ -86,4 +89,106 @@ public static LocalDate getBusinessLocalDate() {
...
+        return isBefore(date, getBusinessLocalDate());
+    }
+
+    public static boolean isAfterBusinessDate(LocalDate date) {
```
adamsaghy: unused

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/DateUtils.java
```diff
@@ -86,4 +89,106 @@ public static LocalDate getBusinessLocalDate() {
...
+        return isAfter(date, getBusinessLocalDate());
+    }
+
+    public static int compare(LocalDate first, LocalDate second) {
```
adamsaghy: unused, also it is wrong: if we are comparing two null LocalDate it will says the first is earlier which is not true.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/DateUtils.java
```diff
@@ -86,4 +89,106 @@ public static LocalDate getBusinessLocalDate() {
...
+        return first == null ? -1 : (second == null ? 1 : first.compareTo(second));
+    }
+
+    public static boolean isEqual(LocalDate first, LocalDate second) {
```
adamsaghy: unused

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/DateUtils.java
```diff
@@ -86,4 +89,106 @@ public static LocalDate getBusinessLocalDate() {
...
+        return first == null ? second == null : (second != null && first.isEqual(second));
+    }
+
+    public static boolean isBefore(LocalDate first, LocalDate second) {
```
adamsaghy: unused

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/DateUtils.java
```diff
@@ -86,4 +89,106 @@ public static LocalDate getBusinessLocalDate() {
...
+        return first == null ? second != null : (second != null && first.isBefore(second));
+    }
+
+    public static boolean isAfter(LocalDate first, LocalDate second) {
```
adamsaghy: unused

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/SavingsAccountTransactionsApiResourceSwagger.java
```diff
@@ -182,8 +182,8 @@ private GetSavingsAccountChargesPaidByData() {}
         }
 
         @Schema(example = "2")
-        public Long totalFilteredRecords;
```
adamsaghy: This is breaking change.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/DatabaseSpecificSQLGenerator.java
vidakovic: I'd also vote to remove the "static"... we have a lot of "utility" classes that have static methods and are just creating problems down the road (example: just have a look at how a tenant's timezone is resolved). Globally there is also no added value in terms of saving mem or similar (by not instantiating a full object); effectively "@Component"s are singletons... and in that form they are composable... with static they are not... there are plenty of examples where this leads to bloated static functions (more than 500 lines of code). The fix is very easy, the class is already instantiated as a singleton, so there's added value by using the static keyword.

vidakovic: ... and on top of that: if we don't declare static then we don't have to pass the dialect parameter... and if we don't need to pass that parameter then we don't need to declare a new function which makes this class more explicit and leaves less to guess for the dev.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/DatabaseSpecificSQLGenerator.java
vidakovic: Same arguments here, see my previous comments "escape" function.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/DatabaseSpecificSQLGenerator.java
vidakovic: Just looking at all these functions in general (not only Marta's): this is getting more and more complex... a fight we cannot win.... and in the end we have compatibility for 2 database systems. A lot of work (also future maintenance). Why not hit the breaks here and use e. g. JOOQ. Slowly, but steadily we are recreating it here with less features and less supported databases. For another day to decide, but I think it will be an important improvement.... and all this code can go away.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/DatabaseType.java
```diff
@@ -19,5 +19,16 @@
 package org.apache.fineract.infrastructure.core.service.database;
 
 public enum DatabaseType {
-    MYSQL, POSTGRESQL
+
+    MYSQL, //
+    POSTGRESQL, //
+    ;
+
+    public boolean isMySql() {
```
vidakovic: Is there any added value of having these "convenience functions"? I think they hide more than they actually help... what happens if we add support for 10 more databases? We have other examples in the code base where these things quickly went out of hand and people started adding all kind of functionality to enums... whereas these files should be really simple and readable. My 2 cents.
Is a simple `MYSQL.equals(dbType)` so much worse (for me it's more readable... imagine static imports of those enums)?
BTW: we should really use "equals" instead "==".

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/DatabaseTypeResolver.java
vidakovic: I would dump these functions too (see my previous comment on DatabaseType). What are we saving here? Again, let's say we add support for 10 more databases...

vidakovic: I guess my point is: you can iterate over all values of an enum or use it for example in a switch statement... those "is..." functions you can only use in (ugly) if-then-else-if statements.... I think this style has a tendency to bloat code...

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/JavaType.java
vidakovic: I'm getting more and more convinced that we should use JOOQ. With below code we are getting closer and closer to create some kind of ORM layer...

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/JdbcJavaType.java
vidakovic: See previous comment for JavaType, looks like a job for JOOQ.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/DatabaseType.java
marta-jankovics: This enum is not new. 
For enums the == comparison is better because 
- enums are singleton classes and it never happens that the same enum is not identical with itself
- implementation of equals for enums is actually ==
- == is null safe
- == shows what it is, does not give the impression as if it could be not t
In my opinion enum is better than static string because
- it can have its own functionality
- type safe and value safe
- singleton and identical
- ability to use as list type
- serialised by name (not by value)

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/DatabaseTypeResolver.java
marta-jankovics: These isX functions are not new and sometimes really useful, when you do not want to know which enum(s) belong to a category.
For identity please see my previous comment.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/JavaType.java
marta-jankovics: You are right, we are getting closer. Datatables are dynamic low level object to database relations and before everything was handled as string like type.equals("int") || type.equals("int4")... so this solution brings it closer to be type safe. Maybe we will enhance this to use an existing framework as next step.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/DatabaseSpecificSQLGenerator.java
marta-jankovics: The dialect parameter might not be used for the currently supported dialects but will be needed for the next supported one, that is why this class has a reference to the dialect.
You can not instantiate a component, only from a component (mainly) so if not static then a parameter of the SqlGenerator component for every method in the call chain. That would be another solution to this, right.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/JdbcJavaType.java
```diff
@@ -30,14 +30,14 @@ public enum JdbcJavaType {
 
         @Override
         public Object toJdbcValueImpl(@NotNull DatabaseType dialect, Object value) {
-            return Boolean.TRUE.equals(value) ? 1 : 0;
+            return value == null ? null : (Boolean.TRUE.equals(value) ? 1 : 0);
```
adamsaghy, line 33: Previously the default was false. Are you sure it wise to change?

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/JdbcJavaType.java
```diff
@@ -30,14 +30,14 @@ public enum JdbcJavaType {
...
         }
     },
     BOOLEAN(JavaType.BOOLEAN, new DialectType(JDBCType.BIT), new DialectType(JDBCType.BOOLEAN, null, "BOOL")) { //
 
         @Override
         public Object toJdbcValueImpl(@NotNull DatabaseType dialect, Object value) {
-            return dialect.isMySql() ? (Boolean.TRUE.equals(value) ? 1 : 0) : super.toJdbcValueImpl(dialect, value);
+            return (value != null && dialect.isMySql()) ? (Boolean.TRUE.equals(value) ? 1 : 0) : super.toJdbcValueImpl(dialect, value);
```
adamsaghy, line 40: same here

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/api/DatatablesApiResource.java
adamsaghy: The rule is
- for V2 it is mandatory to be object response
- for V1 **new** API it is okay to be object response

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
```diff
@@ -1940,8 +2047,8 @@ private String validateDatatableRegistered(String datatable) {
 
     private boolean isRegisteredDataTable(final String datatable) {
         final String sql = "SELECT COUNT(application_table_name) FROM " + TABLE_REGISTERED_TABLE + " WHERE registered_table_name = ?";
-        final int count = jdbcTemplate.queryForObject(sql, Integer.class, datatable);
-        return count > 0;
+        final Integer count = jdbcTemplate.queryForObject(sql, Integer.class, datatable);
```
adamsaghy, line 2050: count function can never return null. using `int` is more than fine and no need for null check

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
```diff
@@ -1890,7 +1997,7 @@ public boolean isDatatableAttachedToEntityDatatableCheck(final String datatableN
                 + " JOIN m_entity_datatable_check edc ON edc.x_registered_table_name = xrt.registered_table_name"
                 + " WHERE edc.x_registered_table_name = '" + datatableName + "'";
         final Long count = this.jdbcTemplate.queryForObject(sql, Long.class);
-        return count > 0;
+        return count != null && count > 0;
```
adamsaghy, line 2000: count function can never return null. using `int` is more than fine and no need for null check

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/SavingsAccountTransactionsApiResource.java
adamsaghy: The rule is

for V2 it is mandatory to be object response
for V1 new API it is okay to be object response

fineract-provider/src/main/java/org/apache/fineract/portfolio/search/service/SearchUtil.java
```diff
@@ -116,47 +116,75 @@ public static void extractJsonResult(@NotNull SqlRowSet rowSet, @NotNull List<St
     }
 
     @NotNull
-    public static List<String> validateToJdbcColumns(List<String> columns, Map<String, ResultsetColumnHeaderData> columnHeaders,
-            List<ApiParameterError> errors, boolean allowEmpty) {
-        List<String> result = new ArrayList<>();
+    public static List<String> validateToJdbcColumnNames(List<String> columns, Map<String, ResultsetColumnHeaderData> headersByName,
+            boolean allowEmpty) {
+        List<ResultsetColumnHeaderData> columnHeaders = validateToJdbcColumns(columns, headersByName, allowEmpty);
+        return columnHeaders.stream().map(e -> e == null ? null : e.getColumnName()).toList();
```
adamsaghy, line 122: in what situation can the `columnHeaders` element be null?

integration-tests/src/test/java/org/apache/fineract/integrationtests/datatable/DatatableIntegrationTest.java
```diff
@@ -499,37 +499,40 @@ public void validateInsertNullValues() {
...
-        assertNull(((List) ((Map) ((List) items.get("data")).get(1)).get("row")).get(2));
-        assertNull(((List) ((Map) ((List) items.get("data")).get(1)).get("row")).get(3));
-        assertNull(((List) ((Map) ((List) items.get("data")).get(1)).get("row")).get(4));
-        assertNull(((List) ((Map) ((List) items.get("data")).get(1)).get("row")).get(5));
-        assertNull(((List) ((Map) ((List) items.get("data")).get(1)).get("row")).get(6));
-        assertNull(((List) ((Map) ((List) items.get("data")).get(1)).get("row")).get(7));
-        assertNull(((List) ((Map) ((List) items.get("data")).get(1)).get("row")).get(8));
-        assertNull(((List) ((Map) ((List) items.get("data")).get(1)).get("row")).get(9));
+        List headers = (List) items.get("columnHeaders");
+        List values0 = (List) ((Map) ((List) items.get("data")).get(0)).get("row");
```
adamsaghy: Can you please rename this to `values`?

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/JdbcJavaType.java
marta-jankovics: there is no place for default on this level, previously it was not correct

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/JdbcJavaType.java
marta-jankovics: there is no place for default on this level, previously it was not correct

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
marta-jankovics: for me it was a checkstyle warning, if it can not return null, it is fine, but lets be safe

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/ReadWriteNonCoreDataServiceImpl.java
marta-jankovics: for me it was a checkstyle warning, if it can not return null, it is fine, but lets be safe

integration-tests/src/test/java/org/apache/fineract/integrationtests/datatable/DatatableIntegrationTest.java
marta-jankovics: no, because its the first table row values, and there are also values from the second table row, see line 525

adamsaghy: I see. `FirstRowValues`, `secondRowValues` then?

fineract-provider/src/main/java/org/apache/fineract/portfolio/search/service/SearchUtil.java
marta-jankovics: columnHeaders can not be null, but the elements in it could be (in case there are null values in the columns list and allowEmpty is true)

PR #3340, opened 2023-07-31
FINERACT-1724: Fix Multi Disburse internal server error

b0c1:
- [x] Fix wrong mapping
- [x] Add empty hashmap for initialization


adamsaghy (approved):
LGTM

fineract-loan/src/main/java/org/apache/fineract/portfolio/repaymentwithpostdatedchecks/domain/PostDatedChecks.java
```diff
@@ -40,7 +39,7 @@ public class PostDatedChecks extends AbstractPersistableCustom {
     @ManyToOne(optional = false)
     @JoinColumn(name = "loan_id", referencedColumnName = "id", nullable = false)
     private Loan loan;
-    @OneToOne(optional = false)
+    @ManyToOne(optional = false)
```
adamsaghy, line 42: This was the problem? :o

b0c1: Yeah... because the other side is `@OneToMany`

PR #3351, opened 2023-08-02
FINERACT-1943 Fix submitted on date references

Mk9894:
## Description

## Checklist

### DIFF
--- a/fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountTransactionData.java
+++ b/fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountTransactionData.java
@@ -112,24 +112,26 @@ public final class SavingsAccountTransactionData implements Serializable {
             final Money amount, final boolean isManualTransaction) {
         final boolean isReversed = false;
         final Boolean lienTransaction = false;
+        final LocalDate submittedOnDate = DateUtils.getBusinessLocalDate();
         final SavingsAccountTransactionType savingsAccountTransactionType = SavingsAccountTransactionType.INTEREST_POSTING;
         SavingsAccountTransactionEnumData savingsAccountTransactionEnumData = new SavingsAccountTransactionEnumData(
                 savingsAccountTransactionType.getValue().longValue(), savingsAccountTransactionType.getCode(),
                 savingsAccountTransactionType.getValue().toString());
         return new SavingsAccountTransactionData(amount.getAmount(), date, savingsAccount.getId(), savingsAccountTransactionEnumData,
-                isReversed, null, isManualTransaction, lienTransaction);
+                isReversed, null, isManualTransaction, lienTransaction, submittedOnDate);
     }
 
     public static SavingsAccountTransactionData overdraftInterest(final SavingsAccountData savingsAccount, final LocalDate date,
             final Money amount, final boolean isManualTransaction) {
         final boolean isReversed = false;
         final Boolean lienTransaction = false;
+        final LocalDate submittedOnDate = DateUtils.getBusinessLocalDate();
         final SavingsAccountTransactionType savingsAccountTransactionType = SavingsAccountTransactionType.OVERDRAFT_INTEREST;
         SavingsAccountTransactionEnumData savingsAccountTransactionEnumData = new SavingsAccountTransactionEnumData(
                 savingsAccountTransactionType.getValue().longValue(), savingsAccountTransactionType.getCode(),
                 savingsAccountTransactionType.getValue().toString());
         return new SavingsAccountTransactionData(amount.getAmount(), date, savingsAccount.getId(), savingsAccountTransactionEnumData,
-                isReversed, null, isManualTransaction, lienTransaction);
+                isReversed, null, isManualTransaction, lienTransaction, submittedOnDate);
     }
 
     public boolean isInterestPostingAndNotReversed() {
@@ -303,13 +305,13 @@ public final class SavingsAccountTransactionData implements Serializable {
 
     private SavingsAccountTransactionData(final Long savingsId, final Long officeId, final PaymentDetailData paymentDetailData,
             final SavingsAccountTransactionEnumData savingsAccountTransactionType, final LocalDate transactionDate,
-            final LocalDate createdDate, final BigDecimal amount, final boolean isReversed, final Long userId,
+            final LocalDate submittedOnDate, final BigDecimal amount, final boolean isReversed, final Long userId,
             final boolean isManualTransaction, final Boolean lienTransaction) {
         this.savingsAccountId = savingsId;
         this.paymentDetailData = paymentDetailData;
         this.transactionType = savingsAccountTransactionType;
         this.transactionDate = transactionDate;
-        this.submittedOnDate = createdDate;
+        this.submittedOnDate = submittedOnDate;
         this.amount = amount;
         this.isManualTransaction = isManualTransaction;
         this.lienTransaction = lienTransaction;
@@ -432,12 +434,13 @@ public final class SavingsAccountTransactionData implements Serializable {
         final boolean isReversed = false;
         final boolean isManualTransaction = false;
         final Boolean lienTransaction = false;
+        final LocalDate submittedOnDate = DateUtils.getBusinessLocalDate();
         SavingsAccountTransactionType savingsAccountTransactionType = SavingsAccountTransactionType.WITHHOLD_TAX;
         SavingsAccountTransactionEnumData transactionType = new SavingsAccountTransactionEnumData(
                 savingsAccountTransactionType.getValue().longValue(), savingsAccountTransactionType.getCode(),
                 savingsAccountTransactionType.getValue().toString());
         SavingsAccountTransactionData accountTransaction = new SavingsAccountTransactionData(amount.getAmount(), date,
-                savingsAccount.getId(), transactionType, isReversed, null, isManualTransaction, lienTransaction);
+                savingsAccount.getId(), transactionType, isReversed, null, isManualTransaction, lienTransaction, submittedOnDate);
         updateTaxDetails(taxDetails, accountTransaction);
         return accountTransaction;
     }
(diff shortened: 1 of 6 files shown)


fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccountTransaction.java
```diff
@@ -149,6 +149,46 @@ public final class SavingsAccountTransaction extends AbstractPersistableCustom {
...
+            final String refNo) {
+        initializeCommonFields(typeOf, transactionLocalDate, createdDate, amount, isReversed, isManualTransaction, lienTransaction, refNo);
+
+        this.savingsAccount = savingsAccount;
+        this.office = office;
+        this.paymentDetail = paymentDetail;
+        this.appUser = appUser;
+    }
+
+    private SavingsAccountTransaction(final Integer transactionTypeEnum, final LocalDate transactionDate, final BigDecimal amount,
```
marta-jankovics: AFAIS this constructor is only called from one static initialiser 'from' method, which is not used, so probably you can remove that method and so this constructor, and then no need for initializeCommonFields. What do you think?

marta-jankovics (approved):

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountTransactionData.java
```diff
@@ -117,7 +117,7 @@ public static SavingsAccountTransactionData interestPosting(final SavingsAccount
                 savingsAccountTransactionType.getValue().longValue(), savingsAccountTransactionType.getCode(),
                 savingsAccountTransactionType.getValue().toString());
         return new SavingsAccountTransactionData(amount.getAmount(), date, savingsAccount.getId(), savingsAccountTransactionEnumData,
-                isReversed, null, isManualTransaction, lienTransaction);
+                isReversed, null, isManualTransaction, lienTransaction, date);
```
adamsaghy: This looks wrong to me. We should not set the submitted on date by the transaction date.

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountTransactionData.java
```diff
@@ -129,7 +129,7 @@ public static SavingsAccountTransactionData overdraftInterest(final SavingsAccou
...
```
adamsaghy: This looks wrong to me. We should not set the submitted on date by the transaction date.

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountTransactionData.java
```diff
@@ -437,7 +437,7 @@ public static SavingsAccountTransactionData withHoldTax(final SavingsAccountData
                 savingsAccountTransactionType.getValue().longValue(), savingsAccountTransactionType.getCode(),
                 savingsAccountTransactionType.getValue().toString());
         SavingsAccountTransactionData accountTransaction = new SavingsAccountTransactionData(amount.getAmount(), date,
-                savingsAccount.getId(), transactionType, isReversed, null, isManualTransaction, lienTransaction);
+                savingsAccount.getId(), transactionType, isReversed, null, isManualTransaction, lienTransaction, date);
```
adamsaghy: This looks wrong to me. We should not set the submitted on date by the transaction date.

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountTransactionData.java
```diff
@@ -782,7 +783,8 @@ public static SavingsAccountTransactionData withWithDrawalTransactionDetails(
                 savingsAccountTransactionData.runningBalance, savingsAccountTransactionData.reversed,
                 savingsAccountTransactionData.transfer, savingsAccountTransactionData.paymentTypeOptions,
                 savingsAccountTransactionData.interestedPostedAsOn, savingsAccountTransactionData.submittedByUsername,
-                savingsAccountTransactionData.note, savingsAccountTransactionData.lienTransaction);
+                savingsAccountTransactionData.note, savingsAccountTransactionData.lienTransaction,
+                savingsAccountTransactionData.getSubmittedOnDate());
```
adamsaghy, line 791: if all the other fields are referenced directly instead of calling the getter, i think it's okay to follow that pattern: `savingsAccountTransactionData.submittedOnDate`

adamsaghy (changes requested):
Please add tests to cover the proper submitted on date behaviour ( = actual business date)!

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountTransactionData.java
marta-jankovics: This is the period end due date of an interest posting period, which means that the transactionDate and the submittedOnDate are the same (or will be the same when the interest will be posted)

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountTransactionData.java
marta-jankovics: Same interest posting period

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountTransactionData.java
marta-jankovics: Same interest posting

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountTransactionData.java
marta-jankovics: I agree with you that having the same pattern for all the parameters would be the best. But it is not a good pattern to call the fields directly, so I am not convinced that it is better, to follow a wrong pattern than break it.

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountTransactionData.java
adamsaghy: Does that mean the transaction date is end of month (31th July) or start of month (1st of Aug)?
However, my concern is with the negative use cases. What would be the outcome if the job which posts the interest will not be executed on 1st of Aug, but got executed on 2nd of Aug. I reckon the transaction date would be still 1st of Aug (if the interest posting happens on 1st day of the month), but the submitted on date should be 2nd of Aug in this case, no?

marta-jankovics: The job which run on business date 1st Aug will pick the interest postings due on 1st Aug independently on which day it is running actually. So both the transaction date and the submittedOnDate will be the same.
The problem with SavingsAccountTransactionData is that it is used for many purposes (internal transfer object, template description, get API response) and so the submittedOnDate can not be left empty where it was set before. Eg. for interest posting it will be mapped directly to insert query.

adamsaghy: i dont think it should be left empty. it should be set the actual business date. However in the above example the transaction date and submitted on date cannot be the same. So i dont think we should set the submitted on date based on the transaction date.

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountTransactionData.java
Mk9894: I think we can modify all other fields to use the getter, which would be the better way to do it.

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountTransactionData.java
marta-jankovics: AFAIK the job picks the same day interest dues, so your example would not happen. But if there is no business date set, then it can happen that the job runs days later and still posts interest from the past and in this case you are right that the submittedOnDate should not be set to the transaction date.
@Mk9894 please change posting submitted dates to getBusinessDate

marta-jankovics: Sometimes this data is a template that is shown on the UI (eg. representing future interest postings), that case the submittedOnDate would be better to keep null, because it was not submitted yet... but it is hard to differentiate when it is used for what... so better would be clean this up and use different data types, but we did not want to make a bigger refactor...

adamsaghy: if it is used for template purpose and the submitted on date is not null, thats not a big problem, however whenever this data object is used as a result of an API call or as the data object which will be used to create savings transaction, the submitted on date should be correct!

marta-jankovics: This one was not the question here

adamsaghy: i might misunderstood the `the submittedOnDate would be better to keep null` part... mea culpa

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/DepositAccountReadPlatformServiceImpl.java
```diff
@@ -1664,7 +1668,7 @@ public SavingsAccountTransactionData mapRow(final ResultSet rs, @SuppressWarning
             final String note = null;
             return SavingsAccountTransactionData.create(savingsId, transactionType, paymentDetailData, savingsId, accountNo, duedate,
                     currency, dueamount, outstandingChargeAmount, runningBalance, false, transfer, postInterestAsOn, submittedByUsername,
-                    note);
+                    note, duedate);
```
marta-jankovics: Here I think still the saving schedule duedate is passed in as submittedOnDate

marta-jankovics (approved):

PR #3360, opened 2023-08-05
FINERACT-1939: Hold Amount - should be debit

marta-jankovics:
## Description

## Checklist

### DIFF
--- a/fineract-core/src/main/java/org/apache/fineract/portfolio/savings/SavingsAccountTransactionType.java
+++ b/fineract-core/src/main/java/org/apache/fineract/portfolio/savings/SavingsAccountTransactionType.java
@@ -18,34 +18,60 @@
  */
 package org.apache.fineract.portfolio.savings;
 
+import jakarta.validation.constraints.NotNull;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import org.apache.fineract.portfolio.TransactionEntryType;
+
 /**
  * An enumeration of different transactions that can occur on a {@link SavingsAccount}.
  */
 public enum SavingsAccountTransactionType {
 
     INVALID(0, "savingsAccountTransactionType.invalid"), //
-    DEPOSIT(1, "savingsAccountTransactionType.deposit"), //
-    WITHDRAWAL(2, "savingsAccountTransactionType.withdrawal"), //
-    INTEREST_POSTING(3, "savingsAccountTransactionType.interestPosting"), //
-    WITHDRAWAL_FEE(4, "savingsAccountTransactionType.withdrawalFee"), //
-    ANNUAL_FEE(5, "savingsAccountTransactionType.annualFee"), //
+    DEPOSIT(1, "savingsAccountTransactionType.deposit", TransactionEntryType.CREDIT), //
+    WITHDRAWAL(2, "savingsAccountTransactionType.withdrawal", TransactionEntryType.DEBIT), //
+    INTEREST_POSTING(3, "savingsAccountTransactionType.interestPosting", TransactionEntryType.CREDIT), //
+    WITHDRAWAL_FEE(4, "savingsAccountTransactionType.withdrawalFee", TransactionEntryType.DEBIT), //
+    ANNUAL_FEE(5, "savingsAccountTransactionType.annualFee", TransactionEntryType.DEBIT), //
     WAIVE_CHARGES(6, "savingsAccountTransactionType.waiveCharge"), //
-    PAY_CHARGE(7, "savingsAccountTransactionType.payCharge"), //
-    DIVIDEND_PAYOUT(8, "savingsAccountTransactionType.dividendPayout"), //
+    PAY_CHARGE(7, "savingsAccountTransactionType.payCharge", TransactionEntryType.DEBIT), //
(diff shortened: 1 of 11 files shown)


fineract-core/src/main/java/org/apache/fineract/portfolio/savings/SavingsAccountTransactionType.java
```diff
@@ -56,158 +82,115 @@ public String getCode() {
...
     }
 
     public boolean isAmountRelease() {
-        return this.value.equals(SavingsAccountTransactionType.AMOUNT_RELEASE.getValue());
+        return this == AMOUNT_RELEASE;
+    }
+
+    public boolean isCredit() {
+        // AMOUNT_RELEASE is not credit, because the account balance is not changed
+        return isCreditEntryType() && !isAmountRelease();
```
adamsaghy, line 184: I am a little confused here. The AMOUNT_RELEASE is marked as CREDIT in line 52, but here it got excluded.

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/SavingsAccountTransactionType.java
```diff
@@ -56,158 +82,115 @@ public String getCode() {
...
     }
 
     public boolean isDebit() {
-        return isWithdrawal() || isWithdrawalFee() || isAnnualFee() || isPayCharge() || isIncomeFromInterest() || isWithHoldTax()
-                || isEscheat();
+        // AMOUNT_HOLD, ESCHEAT are not debit, because the account balance is not changed
+        return isDebitEntryType() && !isAmountOnHold() && !isEscheat();
```
adamsaghy, line 189: I am a little confused here. The AMOUNT_HOLD and ESCHEAT is marked as DEBIT in line 50-51, but here it got excluded

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
```diff
@@ -434,7 +434,7 @@ protected void recalculateDailyBalances(final Money openingAccountBalance, final
                 }
 
                 runningBalance = runningBalance.plus(transactionAmount);
-                if (!transaction.getRunningBalance(transactionAmount.getCurrency()).isEqualTo(transactionAmount)) {
+                if (!transaction.getRunningBalance(transactionAmount.getCurrency()).isEqualTo(runningBalance)) {
```
adamsaghy, line 437: Nice catch! Can you please add a test case which covers this?

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountTransactionData.java
```diff
@@ -144,9 +147,12 @@ public boolean isOverdraftInterestAndNotReversed() {
         return this.transactionType.isIncomeFromInterest() && isNotReversed();
     }
 
+    public boolean isCredit() {
+        return transactionType.getTransactionTypeEnum().isCredit() && isNotReversed() && !isReversalTransaction();
```
adamsaghy, line 151: I was wondering if there is a reversal transaction which is DEBIT, neither the isCredit, neither the isDebit function returns true. Is this valid?

adamsaghy (commented):
Please see my questions.

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/SavingsAccountTransactionType.java
marta-jankovics: I agree with you and but this is how fineract currently works, and if I wanted to be backward compatible, I had to keep this. I would like very much to clean up all these, but it was not the scope of this story.
onHold transaction:
-  should be shown as DEBIT on the UI
- currently included in the transaction runningBalance field calculation
- not included in the account runningBalance calculation

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
marta-jankovics: No, this is a simple typo and a very edge case. I can not set up a use-case where the calculated runningBalance was changed, it is equal to the transaction amount, but does not equal to the running balance. Actually the whole condition could be removed.

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountTransactionData.java
marta-jankovics: Again I would love to clean this up, but wanted to change as less as I could. (isDebit should return true or false independently from the reversal/reversed status and the place of usage should decide what to do with these transactions)
But if the reversed transactions are not considered to be neither credit nor debit, then the reversal transaction (which reversed it) should behave the same way, otherwise the balance would not be correct.

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/SavingsAccountTransactionType.java
marta-jankovics: See my comment on isDebit. Actually I've added the comment to avoid these question.
amountRelease transaction:

- should be shown as CREDIT on the UI
- currently included in the transaction runningBalance field calculation
- not included in the account runningBalance calculation

PR #3390, opened 2023-08-21
[FINERACT-1958] Repayment schedule for down payment get loan details fix

taskain7:
## Description

Fixing the get loan details when down paymnet is enabled


adamsaghy (approved):
LGTM

PR #3406, opened 2023-08-28
[FINERACT-1958] Charge cannot be added to down payment installment

taskain7:
## Description

Loan charge cannot be addoed to down payment installment.

### DIFF
--- a/fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanRepaymentScheduleProcessingWrapper.java
+++ b/fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanRepaymentScheduleProcessingWrapper.java
@@ -42,37 +42,45 @@ public class LoanRepaymentScheduleProcessingWrapper {
         LocalDate startDate = disbursementDate;
         for (final LoanRepaymentScheduleInstallment period : repaymentPeriods) {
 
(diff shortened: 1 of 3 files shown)


fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanRepaymentScheduleProcessingWrapper.java
```diff
@@ -42,37 +42,44 @@ public void reprocess(final MonetaryCurrency currency, final LocalDate disbursem
...
-                    period.getDueDate(), loanCharges, currency, !period.isRecalculatedInterestComponent(), period.isFirstPeriod());
-
-            period.updateChargePortion(feeChargesDueForRepaymentPeriod, feeChargesWaivedForRepaymentPeriod,
-                    feeChargesWrittenOffForRepaymentPeriod, penaltyChargesDueForRepaymentPeriod, penaltyChargesWaivedForRepaymentPeriod,
-                    penaltyChargesWrittenOffForRepaymentPeriod);
-
-            startDate = period.getDueDate();
+            if (!period.isDownPayment()) {
+
+                boolean isFirstPeriod = repaymentPeriods.stream()
```
adamsaghy: Please use a more describing name for this
It is not the first (repayment) period, rather the first (normal / non-down repayment period).

taskain7: renamed

adamsaghy (approved):
LGTM

PR #3441, opened 2023-09-10
FinancialActivityData class bug fixed

SubhamKrGuptaDev:
Hi, I have seen FinancialActivityData This class doesn't have a constructor but 'convertToFinancialActivityData(final FinancialActivity type)' this method is try to inject value by constructor so I have created customer please merge this changes 
Thank you

### DIFF
--- a/fineract-core/src/main/java/org/apache/fineract/accounting/financialactivityaccount/data/FinancialActivityData.java
+++ b/fineract-core/src/main/java/org/apache/fineract/accounting/financialactivityaccount/data/FinancialActivityData.java
@@ -27,4 +27,10 @@ public class FinancialActivityData {
     private final Integer id;
     private final String name;
     private final GLAccountType mappedGLAccountType;
+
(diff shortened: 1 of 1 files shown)


adamsaghy (commented):
@Data lombok annotation contains @RequiredArgsContructor which is equal with the constructor you have created, no?

vidakovic:
@SubhamKrGuptaDev please provide the Fineract Jira ticket as a prefix in your PR title ("FINERACT-1234: ...")

vidakovic:
@SubhamKrGuptaDev please look at the other PRs and follow the same pattern for the PR title... "FINERACT-XXXX: Title of the Jira ticket"

vidakovic:
The code change seems to have no consequences (i. e. is not used anywhere). Note: we are already using Lombok annotations here; if we really wanted  an all args constructor we should use the Lombok annotation (one liner).

PR #3459, opened 2023-09-18
FINERACT-1979: Better versioning plugin for Gradle builds

kjozsa:
use qoomon versioning plugin to properly version built artifacts based on the last git tag

## Description

Got rid of the jgitver `0.10.0-rc03` plugin and replaced it with the more configurable qoomon git versioning plugin.

Behaviour: 
- when HEAD has a git tag, the artifacts get a release version named the same as the tag (eg. tag 1.8.4 will result artifacts like `fineract-client/build/libs/fineract-client-1.8.4.jar`
- when HEAD has no git tag, artifacts use a SNAPSHOT version after the last minor version. Eg. if last tag was 1.8.4, the built artifacts will have version `1.8.5-SNAPSHOT`.


vidakovic (approved):
LGTM

PR #3555, opened 2023-10-31
FINERACT-1997: Log failed requests of a BatchAPI request in audit trail

mariiaKraievska:
## Description

Fixed the issue when failed requests of a BatchAPI request were not logged in audit trail.

JIRA ticket - https://issues.apache.org/jira/browse/FINERACT-1997

## Checklist


adamsaghy:
@b0c1 @marta-jankovics Please review!

fineract-core/src/main/java/org/apache/fineract/commands/service/SynchronousCommandProcessingService.java
```diff
@@ -139,7 +139,8 @@ public CommandProcessingResult executeCommand(final CommandWrapper wrapper, fina
             commandSource.setCommandJson(toApiJsonSerializer.serializeResult(result.getChanges()));
         }
 
-        commandSource = commandSourceService.saveResultSameTransaction(commandSource);
+        commandSource = sameTransaction ? commandSourceService.saveResultSameTransaction(commandSource)
+                : commandSourceService.saveResultNewTransaction(commandSource);
```
mariiaKraievska, line 148: I made this fix to prevent a "403 - Dublicate entry key" exception when the 'sameTransaction' variable in this method is "true".

fineract-core/src/main/java/org/apache/fineract/batch/service/BatchApiServiceImpl.java
```diff
@@ -116,7 +116,7 @@ public List<BatchResponse> handleBatchRequestsWithoutEnclosingTransaction(final
      */
     @Override
     public List<BatchResponse> handleBatchRequestsWithEnclosingTransaction(final List<BatchRequest> requestList, final UriInfo uriInfo) {
-        return callInTransaction(Function.identity()::apply, () -> handleBatchRequests(true, requestList, uriInfo));
+        return callInTransaction(Function.identity()::apply, () -> handleBatchRequests(true, requestList, uriInfo), true);
```
marta-jankovics, line 119: Could you please explain why this was needed? The enclosing transaction was put both cases but
withEnclosingTransaction around the whole batch and withoutEnclosingTransaction around each separate execution.
If you change this then there is a use-case with withoutEnclosingTransaction which might not work: when the parent request fails then all the child requests should also fail, in method parentRequestFailedRecursive we should mark the child transactions to be rolled back. Is this working fine after your change?
Additionally this story is about to handle both with- and without enclosing transaction cases - to save the failed command - and I do not see how this helps in case of withEnclosing case. 
With this change, as I see, the problem is not solved, but postponed.

fineract-core/src/main/java/org/apache/fineract/commands/service/SynchronousCommandProcessingService.java
marta-jankovics: Can you please explain why you had "403 - Dublicate entry key" and how save in new transaction helps?
As this is the success branch, it actually does not matter if we save in the same or in new transaction, in both cases it will be saved.
Please make sure that you do not save new instance for each retry.
What is not happening right now is to save the command even if it was failed, so the code that you should change is in the catch (Throwable t) branch: instead of saveResultSameTransaction for 'sameTransaction' we should call 'saveResultNewTransaction' each time. But what is missing though, is to set the failed code at the end.

marta-jankovics (changes requested):

fineract-core/src/main/java/org/apache/fineract/commands/service/SynchronousCommandProcessingService.java
marta-jankovics: I was thinking... saving in a new transaction is wrong, because we do NOT want to save success for each sub-request, only when the whole batch transaction was success. This can not be merged.

mariiaKraievska: But if I understand what you mean correctly, every successful sub-request was saved in the audit before my changes (when enclosingTransaction=false), the problem was that unsuccessful sub-requests were not saved. When enclosingTransaction=true, if there is at least one unsuccessful sub-request, then both: before and after my changes, no sub-request is saved in the audit.

adamsaghy (changes requested):
Based on the comments of @marta-jankovics

fineract-core/src/main/java/org/apache/fineract/commands/service/SynchronousCommandProcessingService.java
mariiaKraievska: > Can you please explain why you had "403 - Dublicate entry key" and how save in new transaction helps? As this is the success branch, it actually does not matter if we save in the same or in new transaction, in both cases it will be saved. Please make sure that you do not save new instance for each retry. What is not happening right now is to save the command even if it was failed, so the code that you should change is in the catch (Throwable t) branch: instead of saveResultSameTransaction for 'sameTransaction' we should call 'saveResultNewTransaction' each time. But what is missing though, is to set the failed code at the end.

Due to the fact that the saveInitialSameTransaction was called instead of saveInitialNewTransaction, unsuccessful sub-requests were not logged to the audit trail (we are talking about the case when enclosingTransaction=false when calling BatchAPI, since this is exactly the case described in the task, as far as I understand). When I made it so that when enclosingTransaction=false, saveInitialNewTransaction was called, unsuccessful sub-requests began to be logged to the audit trail, but a new problem arose: after executing saveInitialNewTransaction, the database already contained a transaction with a unique key binding (action_name, entity_name, idempotency_key) in table "m_portfolio_command_source". Therefore, when calling the saveResultSameTransaction below, the error "403 - Dublicate entry key" began to return, because the database already had a record with the same key. You can't use saveInitialNewTransaction first and saveResultSameTransaction below on the same entity - this can lead to data inconsistency or re-saving data (as it happened to me). What about the duplication, there is no duplication, because saveAndFlush works as follows: if there is a record in the database, it updates it, if not, it creates it (let's ignore the data inconsistency problems that I described earlier).

mariiaKraievska: I also tried to do as you described ("Тhe code that you should change is in the catch (Throwable t) branch: instead of saveResultSameTransaction for 'sameTransaction' we should call 'saveResultNewTransaction' each time."). But with this approach, unsuccessful sub-requests are not logged to the audit trail :(

fineract-core/src/main/java/org/apache/fineract/batch/service/BatchApiServiceImpl.java
mariiaKraievska: I did this so that when calling the withoutEnclosingTransaction method, the 'sameTransaction' flag in the executeCommand method of the SynchronousCommandProcessingService was false and then saveInitialNewTransaction was called, because this guarantees that unsuccessful sub-requests are recorded in the trail audit logs. As for the use-case `when the parent request fails then all the child requests should also fail, in method parentRequestFailedRecursive we should mark the child transactions to be rolled back`, after my changes nothing seems to be broken. I tested it again today.

fineract-core/src/main/java/org/apache/fineract/commands/service/SynchronousCommandProcessingService.java
marta-jankovics: "When enclosingTransaction=true, if there is at least one unsuccessful sub-request, then both: before and after my changes, no sub-request is saved in the audit."
Exactly this is the problem to be fixed.

"Due to the fact that the saveInitialSameTransaction was called instead of saveInitialNewTransaction, unsuccessful sub-requests were not logged to the audit trail (we are talking about the case when enclosingTransaction=false when calling BatchAPI, since this is exactly the case described in the task, as far as I understand)."
The task is about both with- and without enclosing transaction use-cases so we should fix both.
SaveInitial just saves a skeleton, that we cannot leave in the database as such (maybe we can completely remove this step), but we should extend this - for both failed and successful cases - with the error code or response and we should make sure that this extension is also saved. Without this, saving the skeleton is not a solution.

"a new problem arose: after executing saveInitialNewTransaction, the database already contained a transaction with a unique key binding (action_name, entity_name, idempotency_key) in table "m_portfolio_command_source""
This cannot happen, only if you create new instance and try to save in new transaction, which is a bug. That is why we check if the command already exists for retry. 
If someone called the same request with the same idempotency key, that should fail and return the appropriate error code.

"You can't use saveInitialNewTransaction first and saveResultSameTransaction below on the same entity".
Why not? Either every sub-requests are successful and then will be saved with the same transaction, or when a sub-request fails, you should go and update all the already ran sub requests as failed in the catch branch. With your solution we might save sub-request commands as successful, although these were rolled back with a later failed sub-request.

"I also tried to do as you described ("Тhe code that you should change is in the catch (Throwable t) branch: instead of saveResultSameTransaction for 'sameTransaction' we should call 'saveResultNewTransaction' each time."). But with this approach, unsuccessful sub-requests are not logged to the audit trail :("
I guess the actual sub request was saved as failed, but the previous sub-requests were not saved. This is because you should go back to the already ran sub requests and mark them as failed in a new transaction. Even if we do not save these earlier sub-requests, it is better than save them as success.
Maybe we should introduce an idempotency key for the batch itself and not for the sub-requests.
This was not designed yet.

marta-jankovics (changes requested):

fineract-core/src/main/java/org/apache/fineract/commands/service/SynchronousCommandProcessingService.java
mariiaKraievska: @marta-jankovics  Thank you for your answer. However, I would like to look into this issue in more detail. Please read the attached document, in which I have described my thought process. Thank you for your patience, I look forward to your response.
[BatchAPI issue, 1997.docx](https://github.com/apache/fineract/files/13380854/BatchAPI.issue.1997.docx)

marta-jankovics: Hi Maria, appreciate your detailed description, here is my answer.
We have three different cases for command processing:
1.	Independent API calls
There is no _BatchRequestContextHolder.enclosingTransaction_ set.
It is working as expected, both successful and failed commands are saved, no issue here.
2.	Batch request without enclosing transaction
_BatchRequestContextHolder. batchAttributes_ is set – so we know that we are inside a batch.
_BatchRequestContextHolder.enclosingTransaction_ was set for each sub-request separately to a new transaction (which was different for each sub-request).
You changed this to keep _BatchRequestContextHolder.enclosingTransaction_ empty , which might be ok, I just asked one thing to check:
**When the parent request fails then all the referencing child requests should also fail, in method parentRequestFailedRecursive we should mark the child transactions to be rolled back. Is this working fine after your change?**
Another minor change: **there is already a parameter, called boolean enclosingTransaction. Please name your parameter the same way**.
After your change, this will behave the same way as the independent transactions, see point 1., so **I do not think, that we need to change anything** in _SynchronousCommandProcessingService_.
3.	Batch request with enclosing transaction
_BatchRequestContextHolder. batchAttributes_ is set – so we know that we are inside a batch.
_BatchRequestContextHolder.enclosingTransaction_ is set to one common transaction for all the sub-requests.
**Actually, the story was created to solve the problem of this use-case, because the failed requests here are not saved at all, and it is not correct from audit perspective.**
With your change, you eliminated the problem for the previous batch type by changing the _enclosingTransaction_ state, but the problem itself was not solved.
The solution needs to be designed first and shared with the community, before we change anything.
Probably the best solution would be to add an idempotency key to the batch request itself, and save only one command (with the whole request body as command json), and nothing for the sub-requests.

Please note that there are some changes coming for both _SynchronousCommandProcessingService_ and _BatchApiServiceImpl_ with this PR: https://github.com/apache/fineract/pull/3589

mariiaKraievska: @marta-jankovics Thank you for your answer. Please read the attached document in which I have answered your question on the paragraph described in your last comment "**2. Batch request without enclosing transaction**" as well as described my changes from the last commit, which relates to the paragraph "**3. Batch request with enclosing transaction**". If my solution is appropriate, I will squash the commits and add a test. Thank you, I am waiting for your reply.
[Detailed response.docx](https://github.com/apache/fineract/files/13451772/Detailed.response.docx)

fineract-core/src/main/java/org/apache/fineract/commands/service/SynchronousCommandProcessingService.java
```diff
@@ -201,6 +203,14 @@ private void setIdempotencyKeyStoreFlag(boolean flag) {
         fineractRequestContextHolder.setAttribute(IDEMPOTENCY_KEY_STORE_FLAG, flag);
     }
 
+    @Override
+    public void logFailedBatchRequestWithEnclosingTransaction(final CommandWrapper commandRequest, final BatchResponse failedBatchResult) {
+        final AppUser user = context.authenticatedUser(commandRequest);
+        final CommandSource commandSource = CommandSource.fullEntryForBatchFailed(commandRequest, failedBatchResult, user,
+                idempotencyKeyGenerator.create());
```
marta-jankovics, line 210: I do not think that generating a new idempotency key is correct, rather we should read it from the incoming request header and save that. The main purpose of the idempotency is that the caller knows it and if it tries to reuse (from same or different instance) then the same result (without perform anything) is returned.

fineract-core/src/main/java/org/apache/fineract/batch/service/BatchApiServiceImpl.java
marta-jankovics: You are right that enclosingTransaction flag should be correctly set for 'withoutEnclosingTransaction' case to false. This was a bug and will be fixed with the 4-eye principle PR 
https://github.com/apache/fineract/pull/3649
as maker-checker concept was also not working in this case.
Unfortunately this does not solve the batch api problem of this story.

marta-jankovics:
@mariiaKraievska Can you please leave this story until we carefully design the solution, discuss it with the community and wait for the developers response? Your solution might work, although I've commented some code parts that should be changed, but let us think first on the final solution we should go for. We should be careful here, because with this story, the usage of the batch API will be changed.
I really appriciate your contribution. If you have time, until this story is well designed, it will be a big help, if you could check
https://issues.apache.org/jira/browse/FINERACT-2013 
Thank you!

PR #3563, opened 2023-11-04
FINERACT-2009: Upgrade latest framework libs

vidakovic:
Upgrade latest framework libs.


adamsaghy (approved):
LGTM

galovics:
Is it still needed @vidakovic? I did a dep upgrade beginning of Dec.

vidakovic:
yes... there are a lot more upgrades... this will be the base for the release. Please leave as is. Thanks.

fineract-core/src/main/java/org/apache/fineract/accounting/glaccount/domain/GLAccountType.java
```diff
@@ -86,9 +82,25 @@ public static EnumOptionData fromString(String accountType) {
         }
     }
 
-    public static GLAccountType fromInt(final int i) {
-        final GLAccountType type = intToEnumMap.get(Integer.valueOf(i));
-        return type;
+    public static GLAccountType fromInt(final Integer v) {
```
marta-jankovics, line 85: Can you please not hardcode the mapping but do something dynamic, which works even after a new enum element is added.
Usually I do something like in SavingsAccountTransactionType
private static final Map<Integer, SavingsAccountTransactionType> BY_ID = Arrays.stream(VALUES).collect(Collectors.toMap(SavingsAccountTransactionType::getValue, v -> v));
    public static SavingsAccountTransactionType fromInt(final Integer value) {
        SavingsAccountTransactionType transactionType = BY_ID.get(value);
        return transactionType == null ? INVALID : transactionType;
    }

fineract-core/src/main/java/org/apache/fineract/accounting/journalentry/domain/JournalEntryType.java
```diff
@@ -41,16 +38,19 @@ public String getCode() {
...
-            intToEnumMap.put(type.value, type);
+    public static JournalEntryType fromInt(final Integer v) {
+        if (v == null) {
+            return null;
         }
-    }
 
-    public static JournalEntryType fromInt(final int i) {
-        return intToEnumMap.get(i);
+        switch (v) {
```
marta-jankovics, line 46: Same here as for GLAccountType

fineract-core/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/domain/PortfolioProductType.java
```diff
@@ -47,33 +44,41 @@ public String getCode() {
...
+    public static PortfolioProductType fromInt(final Integer v) {
+        if (v == null) {
+            return null;
         }
-    }
 
-    public static PortfolioProductType fromInt(final int i) {
-        final PortfolioProductType type = intToEnumMap.get(Integer.valueOf(i));
-        return type;
+        switch (v) {
```
marta-jankovics, line 52: Please do not hardcode the mapping, same as for GLAccountType. I do not comment other enums but this request applies for all of them.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/JavaType.java
```diff
@@ -91,9 +91,7 @@ public enum JavaType {
     OBJECT(Object.class), //
     ;
 
-    public static final JavaType[] VALUES = values();
-
-    private static final Map<Class<?>, JavaType> BY_TYPE = Arrays.stream(VALUES).filter(e -> e.type != null)
+    private static final Map<Class<?>, JavaType> BY_TYPE = Arrays.stream(values()).filter(e -> e.type != null)
```
marta-jankovics, line 94: Actually it does make sense to have a static VALUES array, because values() is dynamic. If it only used here for the static mapper then you can remove, but if it is used anywhere else then it worths to keep.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/JdbcJavaType.java
```diff
@@ -136,11 +136,19 @@ public static JdbcJavaType getByTypeName(@NotNull DatabaseType dialect, String n
         for (JdbcJavaType type : values()) {
             DialectType dialectType = type.getDialectType(dialect);
             if (dialectType.getNameResolved().equals(name)) {
+                // NOTE: make MySQL systems happy aka TINYINT vs BOOLEAN issue!
```
marta-jankovics, line 139: Can you please explain the use-case why do you need this and why you do not change the BOOLEAN and TINYINT behaviour for mysql?

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/JdbcJavaType.java
```diff
@@ -136,11 +136,19 @@ public static JdbcJavaType getByTypeName(@NotNull DatabaseType dialect, String n
...
+                if (type.canBooleanType(dialect)) {
+                    return BOOLEAN;
+                }
                 return type;
             }
             if (dialectType.alterNames != null) {
                 for (String alterName : dialectType.alterNames) {
                     if (alterName.equals(name)) {
+                        // NOTE: make MySQL systems happy aka TINYINT vs BOOLEAN issue!
```
marta-jankovics, line 148: Can you please explain the use-case why do you need this and why you do not change the BOOLEAN and TINYINT behaviour for mysql?

fineract-core/src/main/java/org/apache/fineract/accounting/glaccount/domain/GLAccountType.java
vidakovic: Well... the way the enums are handled is wrong on a whole different level. If you think that putting them in a map this makes it better... I don't think so. My argument here is: why use enums in the first place? If, in the end, we compare against integer values... the complaint here - I guess with "hardcoding" you mean the switch statement - goes further... why all the additional functions? Like `isInvalid()` etc. etc. ... everytime you add a new enum value a new function needs to be added... this kind of defeats the whole purpose of an enum. I investigated a bunch of these implementations (someone started this strategy 10 years ago and then many followed... unfortunately) and pretty much all of the enums have this flaw.
I'd say not relevant for the release. Let's decide then on a strategy and put this in the next release.ö

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/JavaType.java
vidakovic: See my earlier comment. Again, kind of defeats the purpose of an enum. With an enum you have type safety built-in, but then we bolt all those integers on top of it and create a parallel structure that we could have simpler with constants, why bother with enums?!?

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/JdbcJavaType.java
vidakovic: I use the code that you created. That's why I asked you for help a couple of weeks ago, because you know your implementation (and the goals you want to achieve with it) best. If you have an alternative that fixes the problem and fits better in your database mapping mechanics then please send me a snippet. Happy to replace.
But for now this is the least intrusive way that I could come up with in 5min. My goal is to create a working release.

fineract-core/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/domain/PortfolioProductType.java
vidakovic: See comment above.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/JavaType.java
vidakovic: Reading this again... I might have commented a bit off topic here. Was still in the "enum" thinking that was discussed earlier. This here is related to your database mapping mechanics. Don`'t really have an opinion here other than: we have technically 2 other database mapping mechanics in place (JPA, Spring Data JDBC) that kind of - at least partially - do a lot of what you do here (type mapping). It feels a bit redundant, but I have to admit that I didn't follow exactly what requirements you had on your list and/or if there is maybe no other way to it.
Again, not relevant for the release.

fineract-core/src/main/java/org/apache/fineract/accounting/journalentry/domain/JournalEntryType.java
vidakovic: Same as my previous comment.

galovics (approved):

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/JdbcJavaType.java
vidakovic: Have to do the release. Please propose a pull request if you see any need for further improvement. As is works.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/database/JdbcJavaType.java
vidakovic: Have to do the release. Please propose a pull request if you see any need for further improvement. As is works.

PR #3568, opened 2023-11-06
FINERACT-1484: Micrometer Monitoring

vidakovic:
Add Micrometer monitoring with default configuration for:
- Prometheus
- Loki
- Grafana
- Tempo

Make Docker Compose configuration more modular.


vidakovic:
Still need to verify that the docker-compose.yml file is working... at the moment failing with this message:

```
Caused by: org.apache.fineract.infrastructure.event.external.exception.ExternalEventConfigurationNotFoundException: All external events are not configured
```

vidakovic:
> Still need to verify that the docker-compose.yml file is working... at the moment failing with this message:
> 
> ```
> Caused by: org.apache.fineract.infrastructure.event.external.exception.ExternalEventConfigurationNotFoundException: All external events are not configured
> ```

Found the reason why this is not working. Gradle cache needs to be removed with `rm -rf .gradle` and then execute `./gradlew :fineract-provider:clean :fineract-provider:jibDockerBuild -x tes`.

fineract-core/src/main/java/org/apache/fineract/infrastructure/event/external/exception/ExternalEventConfigurationNotFoundException.java
```diff
@@ -21,7 +21,7 @@
 public class ExternalEventConfigurationNotFoundException extends RuntimeException {
 
     public ExternalEventConfigurationNotFoundException() {
-        super("All external events are not configured");
+        super("No external events configured");
```
reluxa, line 24: https://github.com/apache/fineract/blob/2b4c932ddd62a50c686be314c442aafa1bae2eee/fineract-core/src/main/java/org/apache/fineract/infrastructure/event/external/service/ExternalEventConfigurationValidationService.java#L78-L80

This exception is thrown, when number of events found on the classpath are not matching with the number of the events found in the database for the tenant. So I am not sure that the `No external events configured` is good here.

vidakovic: Renaming the exception doesn't seem to have any unintended side effect... I did it, because "All ... are not" is simply just "No...". It's just a question of brevity... I think you would say "All external events are configured" for the positive case, but "No external events configured" for the negative case (i. e. this failure/exception situation). "All external events are not configured" sounds a bit weird.

vidakovic:
@reluxa ... you were right... there was a test checking for that exception message... fixed it. Thanks for the help.

fineract-core/src/main/java/org/apache/fineract/infrastructure/event/external/exception/ExternalEventConfigurationNotFoundException.java
adamsaghy: I would rather with "Not all external events are configured" because that is the closer to the truth.

adamsaghy (approved):
LGTM

PR #3577, opened 2023-11-09
[FINERACT-1968] Refund for active loan transaction

taskain7:
## Description

Refund for active loan transaction type handling is implemented for advanced payment allocation strategy.
It is configurable. If the rule is not configured then then the order will be the reversed order of the default rule.


adamsaghy (approved):
LGTM

PR #3649, opened 2023-12-19
FINERACT-1977: 4-eye-principle, maker-checker permission

marta-jankovics:
## Description

## Checklist


IOhacker:
I have reviewed the code and it is Ok for the create client maker-checker action. 

Is there a way to include in the integration test case some coverage for Loan, Savings, Transaction (disbursement, withdrawal/deposit)? 

This is because I think that this is an important feature enhancement.

marta-jankovics:
> I have reviewed the code and it is Ok for the create client maker-checker action.
> 
> Is there a way to include in the integration test case some coverage for Loan, Savings, Transaction (disbursement, withdrawal/deposit)?
> 
> This is because I think that this is an important feature enhancement.

Thank you for your review.
I chose CREATE_CLIENT because it is one of the most complicated cases. The request and the command is made for CREATE_CLIENT which is not maker-checker, but ACTIVATE_CLIENT, which runs hidden as a side-action, is maker-checker. So when you try to approve the maker CREATE_CLIENT command request, it should be successful, even if the main action is not 'maker'.
For your request I've added now one more test case, which is more simpler: WITHDRAWAL_SAVINGSACCOUNT. It is a real maker-checker transaction level action, on Savings Account main entity.
If 4 eye is working for both of these cases then I do not think that any more tests are required, because the concept is the same for all other entities/actions.

fineract-core/src/main/java/org/apache/fineract/batch/service/BatchApiServiceImpl.java
```diff
@@ -346,6 +324,15 @@ private BatchResponse buildErrorResponse(Throwable ex, BatchRequest request) {
         return buildErrorResponse(requestId, statusCode, body, headers);
     }
 
+    private BatchResponse buildOrThrowErrorResponse(RuntimeException ex, BatchRequest request) {
+        BatchResponse response = buildErrorResponse(ex, request);
+        if (response.getStatusCode() != SC_OK && BatchRequestContextHolder.isEnclosingTransaction()) {
```
adamsaghy, line 329: Are you sure we dont need to throw error if the `enclosingTransaction = false`?

fineract-core/src/main/java/org/apache/fineract/commands/service/SynchronousCommandProcessingService.java
```diff
@@ -92,73 +87,64 @@ public CommandProcessingResult executeCommand(final CommandWrapper wrapper, fina
...
         exceptionWhenTheRequestAlreadyProcessed(wrapper, idempotencyKey, isRetry);
 
-        boolean sameTransaction = BatchRequestContextHolder.getEnclosingTransaction().isPresent();
+        AppUser user = context.authenticatedUser(wrapper);
         if (commandSource == null) {
-            AppUser user = context.authenticatedUser(wrapper);
-            if (sameTransaction) {
-                commandSource = commandSourceService.getInitialCommandSource(wrapper, command, user, idempotencyKey);
-            } else {
-                commandSource = commandSourceService.saveInitialNewTransaction(wrapper, command, user, idempotencyKey);
```
adamsaghy, line 112: This seems not correct to me. If we are not storing the initial command, anyone can call it again without any error... Am i missing something?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/bulkimport/importhandler/savings/SavingsImportHandler.java
```diff
@@ -147,7 +148,9 @@ private SavingsAccountData readSavings(final Workbook workbook, final Row row, f
         Long interestPostingPeriodTypeId = null;
         EnumOptionData interestPostingPeriodTypeEnum = null;
         if (interestPostingPeriodType != null) {
-            if (interestPostingPeriodType.equalsIgnoreCase(MONTHLY)) {
+            if (interestPostingPeriodType.equalsIgnoreCase(DAILY)) {
```
adamsaghy, line 151: This seems out of context. Would you mind to create a separate PR for this? It has not much to do with 4 eye principle and maker-checker handling

integration-tests/src/test/java/org/apache/fineract/integrationtests/BatchApiTest.java
```diff
@@ -2153,12 +2153,7 @@ public void shouldNotFindAnyDatatableEntryByQueryAPIAndFailsToUpdateItsColumn()
 
         LOG.info("Batch Response : {}", new Gson().toJson(responseOfQueryAndUpdateDatatableBatch));
 
-        final BatchResponse queryResponse = responseOfQueryAndUpdateDatatableBatch.get(0);
```
adamsaghy, line 2156: dont we have anymore the query response (with request id =1 )?

fineract-core/src/main/java/org/apache/fineract/commands/service/SynchronousCommandProcessingService.java
marta-jankovics: Any other call would fail, because the idempotency key is unique, but not with 'Too early' error, which would be the correct error code in this case.

PR #3667, opened 2024-01-06
FINERACT-2012: Fineract resource settings

vidakovic:
Improve memory configuration for Docker images.

### DIFF
--- a/custom/docker/build.gradle
+++ b/custom/docker/build.gradle
@@ -44,22 +44,6 @@ jib {
     container {
         creationTime = 'USE_CURRENT_TIMESTAMP'
         mainClass = 'org.apache.fineract.ServerApplication'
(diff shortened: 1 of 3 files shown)


custom/docker/build.gradle
```diff
@@ -111,5 +95,3 @@ jib {
         }
     }
 }
-
-// tasks.jibDockerBuild.dependsOn(':fineract-provider:bootJar')
```
adamsaghy, line 115: are you sure this need to be commented out?

vidakovic: well... this is the latest... if no one complained until now then I assume it's not needed...

adamsaghy: mea culpa. i missed you deleted the commented out code piece only 😅

PR #3685, opened 2024-01-15
FINERACT-2020: Added currency to LoanAccountSummaryData

Aboya1417:
We want the loan account summary data to contain currency as we will like to know the currencies tied to different loan products during setup.
The client account endpoint https://baseurl/fineract-provider/api/v1/clients/23/accounts returns a payload containing both savings and loan account, the savings account summary contains currency while the loan account summary does not contain currency.
The aim is to replicate what already exist on the savings account summary for the loan account summary by bringing in currency.
This Jira ticket [FINERACT-2020](https://issues.apache.org/jira/browse/FINERACT-2020) contains more vivid description.


mzhu-ai (approved):
LGTM

adamsaghy (changes requested):
Please add proper testing!
(Example: an integration test which creates a loan and checks whether the currency details are sent)

galovics (changes requested):
I assume this is exposed on the API now, I'm missing the Swagger class changes so fineract-client reflects the new attribute.

PR #3687, opened 2024-01-17
FINERACT-1971: Add urlEncoder for build.gradle file in fineract-client module

GoyalRocks007:
The line inputSpec = "file:///$swaggerFile" in build.gradle file was not able to handle spaces in the file path. Hence, encoded the URL properly to handle spaces in the file path.

Refer images below for details:

### DIFF
--- a/fineract-client/build.gradle
+++ b/fineract-client/build.gradle
@@ -33,7 +33,8 @@ openApiMeta {
 }
 
 openApiValidate {
-    inputSpec = "file:///$swaggerFile"
+    def encodedUrl = "file:///$swaggerFile".replaceAll(" ","%20")
+    inputSpec = "$encodedUrl"
     recommend = true
 }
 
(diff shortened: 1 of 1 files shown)


galovics:
hi @GoyalRocks007 , I'm not really convinced this change is needed. I see your path has a whitespace in it and that's causing the problem. I'm afraid there are other problems lying around that'll be triggered as soon as you try to use whitespace in paths.

I'd suggest please put your project into a path without whitespaces.

GoyalRocks007:
Hello @galovics , I understand. Actually my username has a space in it and it will require changes at system level. However, I faced this issue while building the project due to the same and I thought it will be good if the project is able to handle it....

galovics:
@GoyalRocks007 I understand that but as I said before, I think there are many more places where a space in a path name will simply crash the application, even during runtime, so just changing the swagger generation path is not a full-fledged fix for this problem.

GoyalRocks007:
@galovics as of now, I have faced the problem in this particular piece of code only. It also makes sense that fixing a problem in just one part of code serves no purpose. However, I see 2 paths of action from here:
1) I will keep posting changes if I face problems in future
2) I close this PR
Will do, whatever you think is the best..

PR #3746, opened 2024-02-13
JMeter test suite for Fineract banking operations

anjiltech:
## Description

Created JMeter test suits for performing Fineract based banking setup and transaction operations along with scripts for running the test suites

## Checklist

### DIFF
--- a/jmeter-suite/Fineract-Setup.jmx
+++ b/jmeter-suite/Fineract-Setup.jmx
@@ -1,0 +1,1522 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<jmeterTestPlan version="1.2" properties="5.0" jmeter="5.5">
+  <hashTree>
+    <TestPlan guiclass="TestPlanGui" testclass="TestPlan" testname="Fineract-Setup" enabled="true">
+      <stringProp name="TestPlan.comments">Fineract banking use cases</stringProp>
+      <boolProp name="TestPlan.functional_mode">false</boolProp>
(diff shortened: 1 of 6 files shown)


galovics (changes requested):
@anjiltech I'm still not convinced this is something we need at this point. 

1. JMeter is not for functional testing.
2. The AccountNum.csv file is super fishy and shouldn't be there at all. Why can't we generate account numbers?
3. By the looks of it the test suite is not reproducible; i.e. you can't run it twice for the same fineract instance since the account ids are hardcoded (see point 2)
4.  The suite assumes read/write separation which isn't always the case.

anjiltech:
hi @galovics ,

Regarding Point No. 2 above, It was kept in that format to be able to run a highly concurrent user requests all working across different account numbers at a time (Say 1000 concurrent requests with each request working on different account number at a time). The use CSV config file helps achieving the unique account numbers to be used across different threads.  This can be simplified to just create and use account numbers from within the JMeter suite for a relatively lower concurrent user requests.

Point 3.  The transactions can be performed repetitively. In its current format it expects to clear the data and re-run. By simplifying the format of usage in Point 2, this behavior will change.

Point 4. By default both read and write points to same IP and it works against the single instance Fineract without Read/Write seperation. I have added the provision to be able to work with a Read/Write seperated setup, in which case, user can specify a different IP for those two instances.

With further simplified version, we can change behavior to overcome the observations you mentioned in 2 and 3 above.

PR #3791, opened 2024-03-11
FINERACT-1964: Add functionality to calculate maturity amount before creating FD account

GoyalRocks007:
Created a functionality to calculate maturity amount. I have tried to follow all the best practices in my knowledge. Please let me know if any changes need to be done. Also, please feel free to check business logic to calculate the amount and the validation operations on the input, because I used my own knowledge and some research to write them.

**Current Business logic to calculate the interest:**

Maturity amount = P* ((1+ r/n) ^nt))

Where -

A  denotes the future valuation of the investment made

P  denotes the Principal amount

r   denotes the interest rate

n  denotes the number of times interest gets compounded per period

t   denotes the time period the money was invested

**Current Applied validations on input data:**

1) All input fields (Principal Amount, annualInterestRate, tenureInMonths, interestPostingPeriodInMonths, interestCompoundingPeriodInMonths) must be greater than 0 and cannot be null.
2) 0 < interestCompoundingPeriodInMonths <= 12

Thanks!

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/FixedDepositAccountsApiResource.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/FixedDepositAccountsApiResource.java
@@ -46,6 +46,7 @@ import java.io.InputStream;
 import java.math.BigDecimal;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Set;
 import lombok.RequiredArgsConstructor;
@@ -79,6 +80,7 @@ import org.apache.fineract.portfolio.savings.data.SavingsAccountChargeData;
 import org.apache.fineract.portfolio.savings.data.SavingsAccountTransactionData;
 import org.apache.fineract.portfolio.savings.service.DepositAccountPreMatureCalculationPlatformService;
 import org.apache.fineract.portfolio.savings.service.DepositAccountReadPlatformService;
+import org.apache.fineract.portfolio.savings.service.FixedDepositAccountInterestCalculationService;
 import org.apache.fineract.portfolio.savings.service.SavingsAccountChargeReadPlatformService;
 import org.glassfish.jersey.media.multipart.FormDataContentDisposition;
 import org.glassfish.jersey.media.multipart.FormDataParam;
@@ -102,6 +104,7 @@ public class FixedDepositAccountsApiResource {
     private final AccountAssociationsReadPlatformService accountAssociationsReadPlatformService;
     private final BulkImportWorkbookService bulkImportWorkbookService;
     private final BulkImportWorkbookPopulatorService bulkImportWorkbookPopulatorService;
+    private final FixedDepositAccountInterestCalculationService fixedDepositAccountInterestCalculationService;
 
     @GET
     @Path("template")
@@ -206,6 +209,33 @@ public class FixedDepositAccountsApiResource {
                 mandatoryResponseParameters);
         return this.toApiJsonSerializer.serialize(settings, accountTemplate,
                 DepositsApiConstants.FIXED_DEPOSIT_ACCOUNT_RESPONSE_DATA_PARAMETERS);
+    }
+
+    @GET
+    @Path("calculate-fd-interest")
+    @Consumes({ MediaType.APPLICATION_JSON })
+    @Produces({ MediaType.APPLICATION_JSON })
+    @ApiResponses({
+            @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = FixedDepositAccountsApiResourceSwagger.CalculateFixedDepositInterestResponse.class))) })
+    public String calculateFixedDepositInterest(@Context final UriInfo uriInfo,
+            @QueryParam("principalAmount") @Parameter(description = "BigDecimal principalAmount") final BigDecimal principalAmount,
+            @QueryParam("annualInterestRate") @Parameter(description = "annualInterestRate") final BigDecimal annualInterestRate,
+            @QueryParam("tenureInMonths") @Parameter(description = "tenureInMonths") final Long tenureInMonths,
+            @QueryParam("interestCompoundingPeriodInMonths") @Parameter(description = "interestCompoundingPeriodInMonths") final Long interestCompoundingPeriodInMonths,
+            @QueryParam("interestPostingPeriodInMonths") @Parameter(description = "interestPostingPeriodInMonths") final Long interestPostingPeriodInMonths) {
+        HashMap request = new HashMap<>();
+        request.put("annualInterestRate", annualInterestRate);
+        request.put("tenureInMonths", tenureInMonths);
+        request.put("interestCompoundingPeriodInMonths", interestCompoundingPeriodInMonths);
+        request.put("interestPostingPeriodInMonths", interestPostingPeriodInMonths);
+        request.put("principalAmount", principalAmount);
+        String apiRequestBodyAsJson = toApiJsonSerializer.serialize(request);
+        JsonElement jsonElement = fromJsonHelper.parse(apiRequestBodyAsJson);
+        HashMap result = fixedDepositAccountInterestCalculationService
+                .calculateInterest(new JsonQuery(apiRequestBodyAsJson, jsonElement, fromJsonHelper));
+
+        return toApiJsonSerializer.serializeResult(result);
+
     }
 
     private BigDecimal getActivationCharge(Long accountId) {
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/FixedDepositAccountsApiResourceSwagger.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/FixedDepositAccountsApiResourceSwagger.java
@@ -19,6 +19,7 @@
 package org.apache.fineract.portfolio.savings.api;
 
 import io.swagger.v3.oas.annotations.media.Schema;
+import java.math.BigDecimal;
 import java.time.LocalDate;
 import java.util.Set;
 
@@ -254,6 +255,36 @@ final class FixedDepositAccountsApiResourceSwagger {
         @Schema(example = "6")
         public Integer depositPeriod;
         public GetFixedDepositAccountsDepositPeriodFrequency depositPeriodFrequency;
+    }
+
+    @Schema(description = "CalculateFixedDepositInterestRequest")
+    public static final class CalculateFixedDepositInterestRequest {
+
+        private CalculateFixedDepositInterestRequest() {}
+
+        @Schema(example = "10000")
+        public BigDecimal principalAmount;
+        @Schema(example = "5")
+        public BigDecimal annualInterestRate;
+        @Schema(example = "12")
+        public Long tenureInMonths;
+        @Schema(example = "3")
+        public Long interestPostingPeriodInMonths;
+        @Schema(example = "1")
+        public Long interestCompoundingPeriodInMonths;
+
+    }
+
+    @Schema(description = "CalculateFixedDepositInterestResponse")
+    public static final class CalculateFixedDepositInterestResponse {
+
+        private CalculateFixedDepositInterestResponse() {}
+
+        @Schema(example = "10511.61")
+        public BigDecimal maturityAmount;
+
+        @Schema(example = "Accuracy Warning")
+        public String warning;
     }
 
     @Schema(description = "PostFixedDepositAccountsRequest")
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/data/DepositAccountDataValidator.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/data/DepositAccountDataValidator.java
@@ -20,11 +20,14 @@ package org.apache.fineract.portfolio.savings.data;
 
 import static org.apache.fineract.portfolio.savings.DepositsApiConstants.adjustAdvanceTowardsFuturePaymentsParamName;
 import static org.apache.fineract.portfolio.savings.DepositsApiConstants.allowWithdrawalParamName;
+import static org.apache.fineract.portfolio.savings.DepositsApiConstants.annualInterestRateParamName;
 import static org.apache.fineract.portfolio.savings.DepositsApiConstants.depositAmountParamName;
 import static org.apache.fineract.portfolio.savings.DepositsApiConstants.depositPeriodFrequencyIdParamName;
 import static org.apache.fineract.portfolio.savings.DepositsApiConstants.depositPeriodParamName;
 import static org.apache.fineract.portfolio.savings.DepositsApiConstants.inMultiplesOfDepositTermParamName;
 import static org.apache.fineract.portfolio.savings.DepositsApiConstants.inMultiplesOfDepositTermTypeIdParamName;
+import static org.apache.fineract.portfolio.savings.DepositsApiConstants.interestCompoundingPeriodInMonthsParamName;
+import static org.apache.fineract.portfolio.savings.DepositsApiConstants.interestPostingPeriodInMonthsParamName;
 import static org.apache.fineract.portfolio.savings.DepositsApiConstants.isCalendarInheritedParamName;
 import static org.apache.fineract.portfolio.savings.DepositsApiConstants.isMandatoryDepositParamName;
 import static org.apache.fineract.portfolio.savings.DepositsApiConstants.linkedAccountParamName;
@@ -37,8 +40,10 @@ import static org.apache.fineract.portfolio.savings.DepositsApiConstants.minDepo
 import static org.apache.fineract.portfolio.savings.DepositsApiConstants.preClosurePenalApplicableParamName;
 import static org.apache.fineract.portfolio.savings.DepositsApiConstants.preClosurePenalInterestOnTypeIdParamName;
 import static org.apache.fineract.portfolio.savings.DepositsApiConstants.preClosurePenalInterestParamName;
+import static org.apache.fineract.portfolio.savings.DepositsApiConstants.principalAmountParamName;
 import static org.apache.fineract.portfolio.savings.DepositsApiConstants.recurringFrequencyParamName;
 import static org.apache.fineract.portfolio.savings.DepositsApiConstants.recurringFrequencyTypeParamName;
+import static org.apache.fineract.portfolio.savings.DepositsApiConstants.tenureInMonthsParamName;
 import static org.apache.fineract.portfolio.savings.DepositsApiConstants.transferInterestToSavingsParamName;
 import static org.apache.fineract.portfolio.savings.DepositsApiConstants.transferToSavingsIdParamName;
 import static org.apache.fineract.portfolio.savings.SavingsApiConstants.accountNoParamName;
@@ -197,6 +202,23 @@ public class DepositAccountDataValidator {
         // validateSavingsCharges(element, baseDataValidator);
         validateWithHoldTax(element, baseDataValidator);
 
+        throwExceptionIfValidationWarningsExist(dataValidationErrors);
+
+    }
+
+    public void validateFixedDepositForInterestCalculation(final String json) {
+        if (StringUtils.isBlank(json)) {
+            throw new InvalidJsonException();
+        }
+        final Type typeOfMap = new TypeToken<Map<String, Object>>() {}.getType();
+        this.fromApiJsonHelper.checkForUnsupportedParameters(typeOfMap, json,
+                DepositsApiConstants.FIXED_DEPOSIT_ACCOUNT_INTEREST_CALCULATION_PARAMETERS);
+        final List<ApiParameterError> dataValidationErrors = new ArrayList<>();
+        final DataValidatorBuilder baseDataValidator = new DataValidatorBuilder(dataValidationErrors)
+                .resource(DepositsApiConstants.FIXED_DEPOSIT_ACCOUNT_RESOURCE_NAME);
+        final JsonElement element = this.fromApiJsonHelper.parse(json);
+
+        validateForInterestCalc(element, baseDataValidator);
         throwExceptionIfValidationWarningsExist(dataValidationErrors);
 
     }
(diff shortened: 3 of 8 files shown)


fineract-doc/src/docs/en/images/diag-28a77b45f78221c5cb0e8f5d27c6b34f.png
adamsaghy, line 1: Was there any change regarding these pictures?

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/FixedDepositAccountsApiResource.java
```diff
@@ -208,6 +211,23 @@ public String retrieveOne(@PathParam("accountId") @Parameter(description = "acco
                 DepositsApiConstants.FIXED_DEPOSIT_ACCOUNT_RESPONSE_DATA_PARAMETERS);
     }
 
+    @POST
+    @Path("calculatefdinterest")
```
adamsaghy: I think "calculate-fd-interest" would be a better API path for this.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/FixedDepositAccountsApiResource.java
```diff
@@ -208,6 +211,23 @@ public String retrieveOne(@PathParam("accountId") @Parameter(description = "acco
...
+    @Consumes({ MediaType.APPLICATION_JSON })
+    @Produces({ MediaType.APPLICATION_JSON })
+    @RequestBody(required = true, content = @Content(schema = @Schema(implementation = FixedDepositAccountsApiResourceSwagger.CalculateFixedDepositInterestRequest.class)))
+    @ApiResponses({
+            @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = FixedDepositAccountsApiResourceSwagger.CalculateFixedDepositInterestResponse.class))) })
+    public String calculateFixedDepositInterest(@Parameter(hidden = true) final String apiRequestBodyAsJson) {
+        JsonElement jsonElement = fromJsonHelper.parse(apiRequestBodyAsJson);
+        double maturityAmount = fixedDepositAccountInterestCalculationService
```
adamsaghy: I dont think it should be a POST API, a GET API hence the fixed number of parameters and the nature of this functionality would be more suitable.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/FixedDepositAccountInterestCalculationServiceImpl.java
```diff
@@ -0,0 +1,56 @@
...
+    public double calculateInterest(JsonQuery query) {
+        depositAccountDataValidator.validateFixedDepositForInterestCalculation(query.json());
+        JsonElement element = query.parsedJson();
+        Long principalAmount = this.fromApiJsonHelper.extractLongNamed(principalAmountParamName, element);
+        BigDecimal annualInterestRate = this.fromApiJsonHelper.extractBigDecimalWithLocaleNamed(annualInterestRateParamName, element);
+        Long tenureInMonths = this.fromApiJsonHelper.extractLongNamed(tenureInMonthsParamName, element);
+        Long interestCompoundingPeriodInMonths = this.fromApiJsonHelper.extractLongNamed(interestCompoundingPeriodInMonthsParamName,
+                element);
+
+        double n = (12 / (double) interestCompoundingPeriodInMonths);
```
adamsaghy: Dont use double for calculation. Please always use BigDecimal and/or Money object for having the proper precision and rounding!

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/FixedDepositAccountInterestCalculationServiceImpl.java
```diff
@@ -0,0 +1,56 @@
...
+
+@Service
+@RequiredArgsConstructor
+public class FixedDepositAccountInterestCalculationServiceImpl implements FixedDepositAccountInterestCalculationService {
+
+    private final DepositAccountDataValidator depositAccountDataValidator;
+    private final FromJsonHelper fromApiJsonHelper;
+
+    @Override
+    public double calculateInterest(JsonQuery query) {
```
adamsaghy: It would be nice to have some test cases for this calculation!

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/data/DepositAccountDataValidator.java
```diff
@@ -582,6 +604,33 @@ private void validateDepositTermDeatilForUpdate(final JsonElement element, final
         }
     }
 
+    private void validateForInterestCalc(final JsonElement element, final DataValidatorBuilder baseDataValidator) {
+
+        Long principalAmount = this.fromApiJsonHelper.extractLongNamed(principalAmountParamName, element);
+        baseDataValidator.reset().parameter(principalAmountParamName).value(principalAmount).notNull();
```
adamsaghy: You dont need 2 of these validations. You can chain them together:
```
baseDataValidator.reset().parameter(principalAmountParamName).value(principalAmount).notNull().longGreaterThanZero();
```

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/FixedDepositAccountsApiResourceSwagger.java
```diff
@@ -256,6 +256,33 @@ private GetFixedDepositAccountsDepositPeriodFrequency() {}
...
+
+    }
+
+    @Schema(description = "CalculateFixedDepositInterestResponse")
+    public static final class CalculateFixedDepositInterestResponse {
+
+        private CalculateFixedDepositInterestResponse() {}
+
+        @Schema(example = "10511.61")
+        public float maturityAmount;
```
adamsaghy: Use BigDecimal!

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/FixedDepositAccountsApiResourceSwagger.java
```diff
@@ -256,6 +256,33 @@ private GetFixedDepositAccountsDepositPeriodFrequency() {}
...
 
+    @Schema(description = "CalculateFixedDepositInterestRequest")
+    public static final class CalculateFixedDepositInterestRequest {
+
+        private CalculateFixedDepositInterestRequest() {}
+
+        @Schema(example = "10000")
+        public long principalAmount;
+        @Schema(example = "5")
+        public float annualInterestRate;
```
adamsaghy: Use BigDecimal pls

fineract-doc/src/docs/en/images/diag-28a77b45f78221c5cb0e8f5d27c6b34f.png
GoyalRocks007: These were autogenerated during spotlessApply i guess.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/FixedDepositAccountsApiResource.java
GoyalRocks007: Will change this.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/FixedDepositAccountsApiResource.java
GoyalRocks007: Noted

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/FixedDepositAccountInterestCalculationServiceImpl.java
GoyalRocks007: Actually power function was not available for big decimal. So, I used double instead. What's the ideal work around?

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/FixedDepositAccountInterestCalculationServiceImpl.java
GoyalRocks007: I'll try and add a test case for the same.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/data/DepositAccountDataValidator.java
GoyalRocks007: noted.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/FixedDepositAccountsApiResourceSwagger.java
GoyalRocks007: Noted

GoyalRocks007:
Dear @adamsaghy ,
I have made the changes as suggested. Kindly, review.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/FixedDepositAccountsApiResource.java
```diff
@@ -208,6 +211,22 @@ public String retrieveOne(@PathParam("accountId") @Parameter(description = "acco
                 DepositsApiConstants.FIXED_DEPOSIT_ACCOUNT_RESPONSE_DATA_PARAMETERS);
     }
 
+    @GET
+    @Path("calculate-fd-interest")
+    @Consumes({ MediaType.APPLICATION_JSON })
+    @Produces({ MediaType.APPLICATION_JSON })
+    @RequestBody(required = true, content = @Content(schema = @Schema(implementation = FixedDepositAccountsApiResourceSwagger.CalculateFixedDepositInterestRequest.class)))
```
adamsaghy: GET calls dont have request body. Please use query params!

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/FixedDepositAccountInterestCalculationServiceImpl.java
```diff
@@ -0,0 +1,73 @@
...
+        result.put("warning", warning);
+
+        return result;
+    }
+
+    public BigDecimal calculateInterestInternal(Long principalAmount, BigDecimal annualInterestRate, Long tenureInMonths,
+            Long interestCompoundingPeriodInMonths) {
+        BigDecimal numberOfCompoundingsPerAnnum = BigDecimal.valueOf(12).divide(BigDecimal.valueOf(interestCompoundingPeriodInMonths));
+        Long totalNumberOfCompoundings = tenureInMonths / interestCompoundingPeriodInMonths;
+        MathContext mc = new MathContext(9);
```
adamsaghy: Please use the MathContext from the MoneyHelper!

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/FixedDepositAccountInterestCalculationServiceImpl.java
```diff
@@ -0,0 +1,73 @@
...
+public class FixedDepositAccountInterestCalculationServiceImpl implements FixedDepositAccountInterestCalculationService {
+
+    private final DepositAccountDataValidator depositAccountDataValidator;
+    private final FromJsonHelper fromApiJsonHelper;
+
+    @Override
+    public HashMap calculateInterest(JsonQuery query) {
+        depositAccountDataValidator.validateFixedDepositForInterestCalculation(query.json());
+        JsonElement element = query.parsedJson();
+        Long principalAmount = this.fromApiJsonHelper.extractLongNamed(principalAmountParamName, element);
```
adamsaghy: Principal could be fraction, please use BigDecimal!

fineract-provider/src/test/java/org/apache/fineract/portfolio/savings/service/FixedDepositAccountInterestCalculationServiceImplTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Test
+    public void testCalculateInterestInternal1() {
+
+        // Calculate interest
+        double expectedInterest = 10455.0624; // Expected interest calculated based on provided values
+        double calculatedInterest = service.calculateInterestInternal(10000L, BigDecimal.valueOf(4.5), 12L, 6L).doubleValue();
+
+        // Verify the result
+        assertEquals(expectedInterest, calculatedInterest, 0.001);
```
adamsaghy: We should not have any delta...

fineract-provider/src/test/java/org/apache/fineract/portfolio/savings/service/FixedDepositAccountInterestCalculationServiceImplTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @InjectMocks
+    FixedDepositAccountInterestCalculationServiceImpl service;
+
+    @Test
+    public void testCalculateInterestInternal1() {
+
+        // Calculate interest
+        double expectedInterest = 10455.0624; // Expected interest calculated based on provided values
+        double calculatedInterest = service.calculateInterestInternal(10000L, BigDecimal.valueOf(4.5), 12L, 6L).doubleValue();
```
adamsaghy: Dont work with doubles. Use BigDecimal!

integration-tests/src/test/java/org/apache/fineract/integrationtests/FixedDepositTest.java
```diff
@@ -126,6 +137,65 @@ public void setup() {
...
+        jsonObject.addProperty("tenureInMonths", 12);
+        jsonObject.addProperty("interestPostingPeriodInMonths", 3);
+        jsonObject.addProperty("interestCompoundingPeriodInMonths", 7);
+        JsonParser parser = new JsonParser();
+        String apiRequestBodyAsJson = jsonObject.toString();
+        JsonElement element = parser.parse(apiRequestBodyAsJson);
+        fixedDepositAccountInterestCalculationServiceImpl = new FixedDepositAccountInterestCalculationServiceImpl(
+                new DepositAccountDataValidator(new FromJsonHelper(), null), new FromJsonHelper());
+
+        assertThrows(PlatformApiDataValidationException.class, () -> fixedDepositAccountInterestCalculationServiceImpl
```
adamsaghy: Please validate the thrown error message!

integration-tests/src/test/java/org/apache/fineract/integrationtests/FixedDepositTest.java
```diff
@@ -126,6 +137,65 @@ public void setup() {
...
+        jsonObject.addProperty("annualInterestRate", 5);
+        jsonObject.addProperty("tenureInMonths", 12);
+        jsonObject.addProperty("interestPostingPeriodInMonths", 3);
+        jsonObject.addProperty("interestCompoundingPeriodInMonths", 1);
+        JsonParser parser = new JsonParser();
+        String apiRequestBodyAsJson = jsonObject.toString();
+        JsonElement element = parser.parse(apiRequestBodyAsJson);
+        fixedDepositAccountInterestCalculationServiceImpl = new FixedDepositAccountInterestCalculationServiceImpl(
+                new DepositAccountDataValidator(new FromJsonHelper(), null), new FromJsonHelper());
+        Double expectedResult = 10511.61897;
```
adamsaghy: Please dont work with doubles! Use BigDecimal!

adamsaghy (changes requested):
Please kindly check my comments!

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/FixedDepositAccountInterestCalculationServiceImpl.java
```diff
@@ -0,0 +1,73 @@
...
+        String warning = "This is an approximate calculated amount - it may vary slightly when the account is created";
+
+        HashMap result = new HashMap<>();
+        result.put("maturityAmount", maturityAmount);
+        result.put("warning", warning);
+
+        return result;
+    }
+
+    public BigDecimal calculateInterestInternal(Long principalAmount, BigDecimal annualInterestRate, Long tenureInMonths,
```
adamsaghy: it its internal, use private or protected for visibility!

fineract-provider/src/test/java/org/apache/fineract/portfolio/savings/service/FixedDepositAccountInterestCalculationServiceImplTest.java
GoyalRocks007: Hello, delta of .001 was given to provide a little room to adjust the code and not result in test case failing with slightest change in code since the calculated value is highly precise.
What's your suggestion?

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/FixedDepositAccountInterestCalculationServiceImpl.java
GoyalRocks007: If we make it private/protected, it won't be testable anymore. Hence, the public access modifier.

fineract-provider/src/test/java/org/apache/fineract/portfolio/savings/service/FixedDepositAccountInterestCalculationServiceImplTest.java
adamsaghy: I dont think we need delta here. Just simply check whether they are equal or not...

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/FixedDepositAccountInterestCalculationServiceImpl.java
GoyalRocks007: Hello @adamsaghy,
Using MoneyHelper is proving to challenging to be handled while writing testcases. Since, in normal code the dependencies are managed by spring but in test cases they have to be injected manually. Since, moneyhelper has a lot of dependencies, its proving to be difficult to handle it. What shall I do? Shall I use a simple math context as used earlier?

adamsaghy: Hi 

You can see an example of mocking the MoneyHelper for unit tests in the `DefaultLoanLifecycleStateMachineTest` class.

GoyalRocks007: This uses a mocked version of MoneyHelper. However, in integration test we shouldn't be using a mock. Can you suggest? 🤔

GoyalRocks007: However on second thought, this seems to be the best viable solution for the situation. Another solution would be to use MathContext directly like I used earlier, but it is not a good solution. I'll proceed with this only.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/data/DepositAccountDataValidator.java
```diff
@@ -606,9 +606,9 @@ private void validateDepositTermDeatilForUpdate(final JsonElement element, final
 
     private void validateForInterestCalc(final JsonElement element, final DataValidatorBuilder baseDataValidator) {
 
-        Long principalAmount = this.fromApiJsonHelper.extractLongNamed(principalAmountParamName, element);
-        baseDataValidator.reset().parameter(principalAmountParamName).value(principalAmount).notNull();
-        baseDataValidator.reset().parameter(principalAmountParamName).value(principalAmount).longGreaterThanZero();
+        BigDecimal principalAmount = this.fromApiJsonHelper.extractBigDecimalWithLocaleNamed(principalAmountParamName, element);
+        baseDataValidator.reset().parameter(principalAmountParamName).value(principalAmount).notNull()
+                .notLessThanMin(BigDecimal.valueOf(0));
```
adamsaghy: This can be changed to GreaterThanZero or isPositiveAmount. Accepting 0 principal seems not valid.

integration-tests/src/test/java/org/apache/fineract/integrationtests/FixedDepositTest.java
```diff
@@ -153,9 +153,13 @@ public void testFixedDepositInterestCalculationWithWrongCompoundingPeriod() {
         JsonElement element = parser.parse(apiRequestBodyAsJson);
         fixedDepositAccountInterestCalculationServiceImpl = new FixedDepositAccountInterestCalculationServiceImpl(
                 new DepositAccountDataValidator(new FromJsonHelper(), null), new FromJsonHelper());
-
-        assertThrows(PlatformApiDataValidationException.class, () -> fixedDepositAccountInterestCalculationServiceImpl
```
adamsaghy: This is also valid, but assertThrows can return the exception object which was raised and you can assert it, but like i said. this is fine also ;)

adamsaghy (changes requested):
Kindly see my only remaining comment:
https://github.com/apache/fineract/pull/3791/files/bf90e8a64a38b3dcd2e575641324551ad6e68e8b..1c8ab8104941ca680b80c25d2e839fc04b712a12#r1535297134

adamsaghy:
@GoyalRocks007 Please squash your commits and fix the type of principal in the swagger class

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/api/FixedDepositAccountsApiResourceSwagger.java
```diff
@@ -256,6 +257,36 @@ private GetFixedDepositAccountsDepositPeriodFrequency() {}
         public GetFixedDepositAccountsDepositPeriodFrequency depositPeriodFrequency;
     }
 
+    @Schema(description = "CalculateFixedDepositInterestRequest")
+    public static final class CalculateFixedDepositInterestRequest {
+
+        private CalculateFixedDepositInterestRequest() {}
+
+        @Schema(example = "10000")
+        public Long principalAmount;
```
adamsaghy: Please change this to be BigDecimal

GoyalRocks007:
> @GoyalRocks007 Please squash your commits and fix the type of principal in the swagger class

Yes, just running the tests. Will squash the commits after that.

GoyalRocks007:
Hello @adamsaghy,
I have made the required changes. Please review.

adamsaghy (approved):
LGTM

PR #3761, opened 2024-02-20
FINERACT-1971: Enhanced error logging: Optimistic lock exception handling + log most specific exception

adamsaghy:
## Description

Describe the changes made and why they were made.

Ignore if these details are present on the associated [Apache Fineract JIRA ticket](https://github.com/apache/fineract/pull/1284).

## Checklist

Please make sure these boxes are checked before submitting your pull request - thanks!

- [ ] Write the commit message as per https://github.com/apache/fineract/#pull-requests

- [ ] Acknowledge that we will not review PRs that are not passing the build _("green")_ - it is your responsibility to get a proposed PR to pass the build, not primarily the project's maintainers.

- [ ] Create/update unit or integration tests for verifying the changes made.

- [ ] Follow coding conventions at https://cwiki.apache.org/confluence/display/FINERACT/Coding+Conventions.

- [ ] Add required Swagger annotation and update API documentation at fineract-provider/src/main/resources/static/legacy-docs/apiLive.htm with details of any API changes

- [ ] Submission is not a "code dump".  (Large changes can be made "in repository" via a branch.  Ask on the developer mailing list for guidance, if required.)

FYI our guidelines for code reviews are at https://cwiki.apache.org/confluence/display/FINERACT/Code+Review+Guide.

PR #3854, opened 2024-04-08
FINERACT-2022: Rewrite account domain, services and jobs using QueryDSL.

mariiaKraievska:
## Description

JIRA ticket - https://issues.apache.org/jira/browse/FINERACT-2022

In this PR, I implemented QueryDSL to generate code for type-safe SQL queries for the Account entity.
I used QueryDSL for complex JPA queries, removed all custom "@Query" annotations from repository classes, and refused to use JdbcTemplate.

I'd like to explain the reason for using the `@Transactional(propagation = Propagation.REQUIRES_NEW)` annotation for methods that use QueryDSL for "select" queries.

Previously, using `JdbcTemplate` to work with the database, we did not catch errors related to optimistic locking. This was due to the fact that `JdbcTemplate` works directly with SQL and does not take into account the optimistic locking mechanisms built into JPA. But they were still present and it could be seen in the logs.

After switching to QueryDSL, we started to face `OptimisticLockException` on read operations. This indicated that we had concurrent changes in the database and these changes were being checked at the time of the read requests, which are now performed via JPA. To avoid such cases and ensure data consistency, I applied the `REQUIRES_NEW` annotation. This allows each read request to be executed in a new transaction, isolating it from other operations that may be modifying the data at the same time.

Please let me know if you have any questions or suggestions.

## Checklist


vidakovic:
@mariiaKraievska ... I'll check it out as soon as possible... will ping you... FYI

fineract-provider/src/main/java/org/apache/fineract/portfolio/account/jobs/executestandinginstructions/ExecuteStandingInstructionsTasklet.java
```diff
@@ -113,8 +120,12 @@ public RepeatStatus execute(StepContribution contribution, ChunkContext chunkCon
                 final boolean transferCompleted = transferAmount(errors, accountTransferDTO, data.getId());
 
                 if (transferCompleted) {
-                    final String updateQuery = "UPDATE m_account_transfer_standing_instructions SET last_run_date = ? where id = ?";
-                    jdbcTemplate.update(updateQuery, transactionDate, data.getId());
+                    final JPAQueryFactory queryFactory = new JPAQueryFactory(this.entityManager);
```
adamsaghy: I would move this whole logic into an AccountTransferStandingInstruction repository. We can change our repository interfaces to an impl where queryDSL is used.

fineract-provider/src/main/java/org/apache/fineract/portfolio/account/jobs/executestandinginstructions/ExecuteStandingInstructionsTasklet.java
```diff
@@ -151,17 +160,21 @@ private boolean transferAmount(final List<Throwable> errors, final AccountTransf
...
+
+        final AccountTransferStandingInstructionsHistory newHistory = new AccountTransferStandingInstructionsHistory();
+        newHistory.setAccountTransferStandingInstruction(standingInstructionRepository.getReferenceById(instructionId));
+        newHistory.setStatus(transferCompleted ? "success" : "failed");
+        newHistory.setAmount(accountTransferDTO.getTransactionAmount());
+        newHistory.setExecutionTime(LocalDateTime.now(DateUtils.getDateTimeZoneOfTenant()));
+        newHistory.setErrorLog(String.valueOf(errorLog));
+
+        entityManager.persist(newHistory);
+        entityManager.flush();
```
adamsaghy: I dont think we need explicit flush.

fineract-provider/src/main/java/org/apache/fineract/portfolio/account/jobs/executestandinginstructions/ExecuteStandingInstructionsTasklet.java
```diff
@@ -151,17 +160,21 @@ private boolean transferAmount(final List<Throwable> errors, final AccountTransf
...
-            updateQuery.append("'failed'").append(",");
-        } else {
-            updateQuery.append("'success'").append(",");
         }
-        updateQuery.append(accountTransferDTO.getTransactionAmount().doubleValue());
-        updateQuery.append(", now(),");
-        updateQuery.append("'").append(errorLog).append("')");
-        jdbcTemplate.update(updateQuery.toString());
+
+        final AccountTransferStandingInstructionsHistory newHistory = new AccountTransferStandingInstructionsHistory();
```
adamsaghy: I would move this whole logic into a repository

fineract-provider/src/main/java/org/apache/fineract/portfolio/account/service/AccountAssociationsReadPlatformServiceImpl.java
```diff
@@ -18,86 +18,101 @@
...
+import org.springframework.transaction.annotation.Transactional;
 
 @RequiredArgsConstructor
 @Slf4j
 public class AccountAssociationsReadPlatformServiceImpl implements AccountAssociationsReadPlatformService {
 
-    private final JdbcTemplate jdbcTemplate;
+    private final EntityManager entityManager;
 
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
```
adamsaghy: I dont really understand the new transaction for reading... I believe you are referring to Optimistic read locking but I am afraid you are failing to sort the explained problem by this.

If I understand the problem you are explaining in the description of the PR is the following:
Transaction isolation= READ COMMITTED
TX1 started
- Read Account (by name) -> Account (id=1, name=Adam, version=1, ref=1234)
TX2 started
- Read Account (by id) -> Account (id=1, name=Adam, version=1, ref=1234)
- Modify Account name -> Account (id=1, name=Josh, version=**2**, ref=**3456**)
TX2 ended
TX1:
- Read Accounts -> [Account(id=1, name=Josh, version=2)] -> JPA realize with the same ID a modified entry it has in its persistence context, so it is not in sync with the DB anymore and trigger Optimistic lock!

Imagine if you are reading the entity in new transactions when the read completed, the read entity became detached and removed from the persistence context and the JPA optimistic locking would only be triggered if this detached entity to be written back to the DB, but not when it was read again...

What do you think?

fineract-provider/src/main/java/org/apache/fineract/portfolio/account/service/AccountAssociationsReadPlatformServiceImpl.java
```diff
@@ -18,86 +18,101 @@
...
-        final String sql = "select " + mapper.schema() + " where aa.loan_account_id = ? and aa.association_type_enum = ?";
-        try {
-            final AccountAssociationsData accountAssociationsData = this.jdbcTemplate.queryForObject(sql, mapper, loanId, // NOSONAR
-                    AccountAssociationType.LINKED_ACCOUNT_ASSOCIATION.getValue());
-            if (accountAssociationsData != null) {
-                linkedAccount = accountAssociationsData.linkedAccount();
-            }
-        } catch (final EmptyResultDataAccessException e) {
-            log.debug("Linking account is not configured");
+        final QAccountAssociations qAccountAssociations = QAccountAssociations.accountAssociations;
```
adamsaghy: It would be better to have for this a Repository

adamsaghy (changes requested):
I am really happy to see we are moving away from building queries "manually" toward QueryDSL!!

In the other hand please take a look on my review comments and let me know about your thoughts!

fineract-provider/src/main/java/org/apache/fineract/portfolio/account/jobs/executestandinginstructions/ExecuteStandingInstructionsTasklet.java
mariiaKraievska: Done

fineract-provider/src/main/java/org/apache/fineract/portfolio/account/jobs/executestandinginstructions/ExecuteStandingInstructionsTasklet.java
mariiaKraievska: You right, done

fineract-provider/src/main/java/org/apache/fineract/portfolio/account/jobs/executestandinginstructions/ExecuteStandingInstructionsTasklet.java
mariiaKraievska: Done

fineract-provider/src/main/java/org/apache/fineract/portfolio/account/service/AccountAssociationsReadPlatformServiceImpl.java
mariiaKraievska: Done

fineract-provider/src/main/java/org/apache/fineract/portfolio/account/service/AccountAssociationsReadPlatformServiceImpl.java
mariiaKraievska: Thank you for your comments and concerns regarding the use of @Transactional with Propagation.REQUIRES_NEW for read-only operations. Initially, this approach was adopted to manage specific issues with OptimisticLockException observed when using QueryDSL instead of JdbcTemplate. However, improvements made in the handling of OptimisticLockException by the recent updates https://github.com/apache/fineract/pull/3761 have resolved these issues effectively. As a result, the additional transactional annotation is no longer necessary, and our tests confirm that the system performs well without it. I appreciate your input and agree that adhering to these updated practices ensures cleaner and more efficient code.

vidakovic:
Just a quick note... QueryDSL (the project) seems to have hit a bit of a slow down due to members occupied with work and studies... the good news: it seems that the people from OpenFeign jumped in and already made quite some significant improvements. Also: there is also a better way to generate the metadata sources with the Infobip extensions.

I have contributed some improvements to both projects and hope they get accepted. The one for the Infobip extensions is a fairly small one, just trying to convince them to switch from `com.querydsl:*` to `io.github.openfeign.querydsl:*`; package names etc. are all the same.

The other PR on OpenFeign's QueryDSL fork is not relevant for what we need, just supporting a newer geo library (JTS) and the latest Postgis driver extensions.

The PR that would be interesting for us is the Infobip one (replacing `com.querydsl:*` to `io.github.openfeign.querydsl:*`).

FYI

https://github.com/infobip/infobip-spring-data-querydsl/pull/101
https://github.com/OpenFeign/querydsl/pull/416

adamsaghy:
@mariiaKraievska Please kindly check @vidakovic recommendation and resolve the conflicting files and squash the commits! Thank you! ;)

adamsaghy:
@mariiaKraievska Kindly asking you to squash your commits and resolve the conflicts. Feel free to reach out if you are facing any issues with that.

mariiaKraievska:
@adamsaghy I'm sorry for the delay. (I thought I could quickly deal with the problem and then write back, but no).
I'm a bit stuck with these two tests right now:
`ClientLoanIntegrationTest - testLoanCharges_INSTALLATION_FEE`.
`SchedulerJobsTestResults - testInterestTransferForSavings`.

They started to fail for me after this commit https://github.com/apache/fineract/pull/3854/commits/c267bfd61902c9313c2a1c4dc5f598104d5c3188, they do not pass without an annotation @Transactional(propagation = Propagation.REQUIRES_NEW)
The first one does not pass without this annotation on the `fetchPostInterestTransactionIds` method in the `AccountTransfersReadPlatformServiceImpl` class.
And the second without it on the `retrieveLoanLinkedAssociation` method in the `AccountAssociationsCustomRepositoryImpl` class.

I will work further and try to solve this ASAP.

borikas:
> @adamsaghy I'm sorry for the delay. (I thought I could quickly deal with the problem and then write back, but no). I'm a bit stuck with these two tests right now: `ClientLoanIntegrationTest - testLoanCharges_INSTALLATION_FEE`. `SchedulerJobsTestResults - testInterestTransferForSavings`.
> 
> They started to fail for me after this commit [c267bfd](https://github.com/apache/fineract/commit/c267bfd61902c9313c2a1c4dc5f598104d5c3188), they do not pass without an annotation @transactional(propagation = Propagation.REQUIRES_NEW) The first one does not pass without this annotation on the `fetchPostInterestTransactionIds` method in the `AccountTransfersReadPlatformServiceImpl` class. And the second without it on the `retrieveLoanLinkedAssociation` method in the `AccountAssociationsCustomRepositoryImpl` class.
> 
> I will work further and try to solve this ASAP.

@adamsaghy It looks like we have a problem with caching sql queries when migrating from jdbcTemplate to queryDSL, because queryDSL uses EntityManager Cache, at the same time as jdbcTemplate uses Spring Cache.

adamsaghy:
> > @adamsaghy I'm sorry for the delay. (I thought I could quickly deal with the problem and then write back, but no). I'm a bit stuck with these two tests right now: `ClientLoanIntegrationTest - testLoanCharges_INSTALLATION_FEE`. `SchedulerJobsTestResults - testInterestTransferForSavings`.
> > They started to fail for me after this commit [c267bfd](https://github.com/apache/fineract/commit/c267bfd61902c9313c2a1c4dc5f598104d5c3188), they do not pass without an annotation @transactional(propagation = Propagation.REQUIRES_NEW) The first one does not pass without this annotation on the `fetchPostInterestTransactionIds` method in the `AccountTransfersReadPlatformServiceImpl` class. And the second without it on the `retrieveLoanLinkedAssociation` method in the `AccountAssociationsCustomRepositoryImpl` class.
> > I will work further and try to solve this ASAP.
> 
> @adamsaghy It looks like we have a problem with caching sql queries when migrating from jdbcTemplate to queryDSL, because queryDSL uses EntityManager Cache, at the same time as jdbcTemplate uses Spring Cache.

I might not yet fully understand the issue. Are we talking about the prepared statement cache? Or the entity cache, rather when something was already fetched, it does not go to the database to fetch again?

borikas:
> > > @adamsaghy I'm sorry for the delay. (I thought I could quickly deal with the problem and then write back, but no). I'm a bit stuck with these two tests right now: `ClientLoanIntegrationTest - testLoanCharges_INSTALLATION_FEE`. `SchedulerJobsTestResults - testInterestTransferForSavings`.
> > > They started to fail for me after this commit [c267bfd](https://github.com/apache/fineract/commit/c267bfd61902c9313c2a1c4dc5f598104d5c3188), they do not pass without an annotation @transactional(propagation = Propagation.REQUIRES_NEW) The first one does not pass without this annotation on the `fetchPostInterestTransactionIds` method in the `AccountTransfersReadPlatformServiceImpl` class. And the second without it on the `retrieveLoanLinkedAssociation` method in the `AccountAssociationsCustomRepositoryImpl` class.
> > > I will work further and try to solve this ASAP.
> > 
> > 
> > @adamsaghy It looks like we have a problem with caching sql queries when migrating from jdbcTemplate to queryDSL, because queryDSL uses EntityManager Cache, at the same time as jdbcTemplate uses Spring Cache.
> 
> I might not yet fully understand the issue. Are we talking about the prepared statement cache? Or the entity cache, rather when something was already fetched, it does not go to the database to fetch again?

This problem is similar to an incorrect sequence of sql queries within the same transaction 
Снимок экрана 2024-06-04 в 15 21 38
Снимок экрана 2024-06-04 в 15 21 58
Снимок экрана 2024-06-04 в 15 24 22

mariiaKraievska:
@adamsaghy Related to the problem of having in the same transaction queries that use `jdbcTemplate` and new, rewritten queries that use `QueryDSL'. I propose to go the way of rewriting within transactions or as now within modules (as in this PR `account domain`), but duplicating methods related to other modules, written with `jdbcTemplate`, rewriting them with `QueryDSL`. Or the third option, which I am inclined to and have already started working on, is to take another module, more independent from other modules, and present PR on its example. What do you think?

adamsaghy:
> @adamsaghy Related to the problem of having in the same transaction queries that use `jdbcTemplate` and new, rewritten queries that use `QueryDSL'. I propose to go the way of rewriting within transactions or as now within modules (as in this PR `account domain`), but duplicating methods related to other modules, written with `jdbcTemplate`, rewriting them with `QueryDSL`. Or the third option, which I am inclined to and have already started working on, is to take another module, more independent from other modules, and present PR on its example. What do you think?

I like the idea to do it within module boundaries in this case. I wont mind small steps towards the final goal ;)

@vidakovic Thoughts?

vidakovic:
@mariiaKraievska ... maybe some inspiration: https://github.com/vidakovic/fineract/tree/feature/APACHE_EU_2024/custom/apachecon/note ... if you want we can talk this through

PR #3870, opened 2024-04-19
chore: Add systemProp.sonar.gradle.skipCompile property to gradle.properties

hartou:


vidakovic:
No Jira ticket. Does more than the subject line indicates (you are also upgrading Github Actions). Also: rebase 4 commits into one otherwise PR will not be accepted.

PR #3919, opened 2024-06-07
FINERACT-1981: EMI Calculation

janez89:
## Description

EMI Calculation with 100% code coverage and without caller side logic.
[https://issues.apache.org/jira/browse/FINERACT-1981](https://issues.apache.org/jira/browse/FINERACT-1981)


fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanproduct/calc/EMICalculationResult.java
```diff
@@ -16,29 +16,28 @@
...
-     *
-     * @param currentRateFactor
-     *
-     * @param mathContext
-     *
-     */
-    public static BigDecimal fnValue(final BigDecimal previousFnValue, final BigDecimal currentRateFactor, final MathContext mc) {
-        return BigDecimal.ONE.add(previousFnValue.multiply(currentRateFactor, mc));
+@AllArgsConstructor(access = AccessLevel.PACKAGE)
+public class EMICalculationResult {
```
adamsaghy, line 28: I am not fully understand this class. Why do we need a `getRateFactorMinus1ForPeriod` method? Also i dont really like the idea of having the `getNumberOfPeriods` method which returns the value based on the size of the ratefactor list. Down payments are particular periods where we dont calculate ratefactor, neither interest... but it is still a period.

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanproduct/calc/EMICalculator.java
```diff
@@ -0,0 +1,160 @@
...
+     * @param mc
+     * @return
+     */
+    public EMICalculationResult calculateEMI(final List<Integer> repaymentPeriodDays,
+                                             final BigDecimal principal,
+                                             final BigDecimal interestRate,
+                                             final Integer daysInYear,
+                                             final MathContext mc) {
+        final List<BigDecimal> rateFactorList = getRateFactorList(repaymentPeriodDays, interestRate, daysInYear, mc);
+        final List<BigDecimal> rateFactorMinus1List = getRateFactorMinus1List(rateFactorList, mc);
```
adamsaghy: Is this a different `getRateFactorMinus1List` list or the same with `EMICalculationResult. getRateFactorMinus1List`?

fineract-progressive-loan/src/test/java/org/apache/fineract/portfolio/loanproduct/calc/EMICalculatorTest.java
```diff
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License. You may obtain a copy of the License at
- *
+ * <p>
```
adamsaghy: Please remove these changes from the license

janez89: Oh, I applied a code reformat before I committed the code and as I see the IDE reformatted the license. I fix it.

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanproduct/calc/EMICalculator.java
janez89: These two are the same, but the reason why this was highlighted in a row is because I simplified the calculation. Originally, there was an extra plus 1 and a minus 1 in the fn calculation.

adamsaghy: yeah, but it is not really the same... the first one is deducting 1 from the calculated rate factor amount, the other one just simply taking the period -1 from the list, but not deducting anything.

I found it a little misleading or not straightforward...

I was wondering whether it would make sense to calculate and store the ratefactor (which is bigger than 1) and when the calculation requires us to take the ratefactor - 1 just simply do the calculation... I am not a big fan of having `getRateFactorMinus1List` methods which sometimes give you a the period -1 ratefactor and sometimes it is calculating and having the ratefactor -1 as value.

What do you think?

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanproduct/calc/EMICalculationResult.java
janez89: Based on our review talks, I updated the PR.

adamsaghy (approved):
LGTM

PR #3946, opened 2024-06-21
FINERACT-1981: EMI Calculation, Daily, Weekly, Monthly support plug to Repayment Schedule

janez89:
## Description

Implement EMI Calculation and Plug to Repayment Schedule

https://issues.apache.org/jira/browse/FINERACT-1981


fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanproduct/calc/EMICalculator.java
```diff
@@ -20,47 +20,162 @@
...
+import org.apache.fineract.portfolio.common.domain.DaysInMonthType;
+import org.apache.fineract.portfolio.common.domain.DaysInYearType;
+import org.apache.fineract.portfolio.common.domain.PeriodFrequencyType;
+import org.apache.fineract.portfolio.loanaccount.loanschedule.data.LoanScheduleParams;
+import org.apache.fineract.portfolio.loanaccount.loanschedule.domain.LoanScheduleModelPeriod;
+import org.apache.fineract.portfolio.loanproduct.domain.LoanProductRelatedDetail;
 
-@Component
-public class EMICalculator {
+public final class EMICalculator {
```
adamsaghy: Is there a reason to change this component to be static? It would be easier to override (if needed) if it is not static. Thoughts?

janez89: We talked and we decided to follow the component direction. It changed back to the component.

janez89:
Screen Shot 2024-06-24 at 17 11 36

PR #3951, opened 2024-06-27
FINERACT-2101: Loan product using global config values

alberto-art3ch:
## Description

There is an issue when loan product with "enabled" state for using global config values for `dueDaysForRepaymentEvent` and `overDueDaysForRepaymentEvent`  is edited, when opening edit screen, use global values checkbox is not checked, but global values are in the fields (so basically the values are ok, but this point by submitting the loan product the “use globals” will be disabled

[FINERACT-2101](https://issues.apache.org/jira/browse/FINERACT-2101)

## Checklist

### DIFF
--- a/fineract-loan/src/main/java/org/apache/fineract/portfolio/loanproduct/api/LoanProductsApiResourceSwagger.java
+++ b/fineract-loan/src/main/java/org/apache/fineract/portfolio/loanproduct/api/LoanProductsApiResourceSwagger.java
@@ -157,6 +157,8 @@ final class LoanProductsApiResourceSwagger {
         public Long delinquencyBucketId;
         @Schema(example = "false")
         public Boolean enableInstallmentLevelDelinquency;
+        @Schema(example = "false")
+        public Boolean useDueRepaymentGlobalConfigs;
         @Schema(example = "3")
         public Integer dueDaysForRepaymentEvent;
         @Schema(example = "3")
@@ -1260,6 +1262,8 @@ final class LoanProductsApiResourceSwagger {
         public Boolean enableInstallmentLevelDelinquency;
         @Schema(example = "true")
         public Boolean disallowExpectedDisbursements;
+        @Schema(example = "false")
+        public Boolean useDueRepaymentGlobalConfigs;
         @Schema(example = "3")
         public Integer dueDaysForRepaymentEvent;
         @Schema(example = "3")
@@ -1402,6 +1406,8 @@ final class LoanProductsApiResourceSwagger {
         public Long delinquencyBucketId;
         @Schema(example = "false")
         public Boolean enableInstallmentLevelDelinquency;
+        @Schema(example = "false")
+        public Boolean useDueRepaymentGlobalConfigs;
         @Schema(example = "3")
         public Integer dueDaysForRepaymentEvent;
         @Schema(example = "3")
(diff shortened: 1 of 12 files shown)


fineract-loan/src/main/java/org/apache/fineract/portfolio/loanproduct/api/LoanProductsApiResourceSwagger.java
```diff
@@ -1260,6 +1262,8 @@ private GetLoanCharge() {}
         public Boolean enableInstallmentLevelDelinquency;
         @Schema(example = "true")
         public Boolean disallowExpectedDisbursements;
+        @Schema(example = "false")
```
adamsaghy, line 1265: Do we need this for GetLoanCharge?

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanproduct/domain/LoanProduct.java
```diff
@@ -232,6 +232,9 @@ public class LoanProduct extends AbstractPersistableCustom<Long> {
     @Column(name = "repayment_start_date_type_enum", nullable = false)
     private RepaymentStartDateType repaymentStartDateType;
 
+    @Column(name = "use_due_repayment_global_configs", nullable = false)
```
adamsaghy, line 235: This was not stored previously?

adamsaghy (changes requested):
`FINERACT-2101: UI data required from Fineract` seems unrelated story. Can you please use proper story to cover this changes?

Also, can you please explain a little further the reasoning behind introducing new field on loan product and so? I am not fully understand.

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanproduct/api/LoanProductsApiResourceSwagger.java
```diff
@@ -157,6 +157,8 @@ private PostLoanProductsRequest() {}
         public Long delinquencyBucketId;
         @Schema(example = "false")
         public Boolean enableInstallmentLevelDelinquency;
+        @Schema(example = "false")
+        public Boolean useDueRepaymentGlobalConfigs;
```
adamsaghy, line 161: I would prefer the following:
- We dont have a new field that contains a boolean value to decide whether something got fetched from global configuration or not.

Instead, if we have any settings on the loan product, fetch it, if not, we read from global configuration.

User can decide whether they provide any value for it or not... Basically these field would be optional ones, and what we can do when we fetch the loan product or the loan product template, we can send back the values from the global configuration and as a hint it can be shown on the UI. After that the user can decide whether they are happy with it or wanna override it. What do you think?
@bharathc27

alberto-art3ch: @adamsaghy / @bharathc27 

This boolean attribute is to let the UI know If the products uses or not the global config values. This is because in the UI exists a checkbox to indicate this. When you are creating and select that checkbox the global values are sent to Fineract. Until this, It is ok

The problem exists when you edit a Loan Product, the UI can not to know If the values of these configurations are or not copied from the global config

This issue was reported by Peter. The UI has the logic to read the value, but If Fineract does not return any value for this, then the UI can not indicate if the configurations are or not inherit from the global configurations

adamsaghy: I still believe we should not have the which indicates whether it is coming from global configuration or not. I would simply leave these fields as optional and if user does not fill them, it is coming from global configurations. If user provide a value then that will be used instead of the one from global configurations. Additionally we can enhance further and return the values from global configurations and we can show it on the UI to inform the user if nothing is sent, those value will be used, but they should be different values and do not send it back to backend! 

What are your thoughts?

PR #3915, opened 2024-06-05
FINERACT-2021: refactor the note GET end-points to use Jackson instead of GSON.

Zeyad2003:
## Description

Ignore if these details are present on the associated [Apache Fineract JIRA ticket](https://github.com/apache/fineract/pull/1284).

## Checklist

Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Write the commit message as per https://github.com/apache/fineract/#pull-requests

- [x] Acknowledge that we will not review PRs that are not passing the build _("green")_ - it is your responsibility to get a proposed PR to pass the build, not primarily the project's maintainers.

- [ ] Create/update unit or integration tests for verifying the changes made.

- [x] Follow coding conventions at https://cwiki.apache.org/confluence/display/FINERACT/Coding+Conventions.

- [x] Add required Swagger annotation and update API documentation at fineract-provider/src/main/resources/static/legacy-docs/apiLive.htm with details of any API changes

- [x] Submission is not a "code dump".  (Large changes can be made "in repository" via a branch.  Ask on the developer mailing list for guidance, if required.)

FYI our guidelines for code reviews are at https://cwiki.apache.org/confluence/display/FINERACT/Code+Review+Guide.

PR #3994, opened 2024-07-26
FINERACT-2110: Introduce type-safety API for commands (Note API)

Zeyad2003:
## Description

## Checklist

### DIFF
--- a/custom/v3/command/data/src/main/java/org/apache/fineract/v3/command/data/CommandResponse.java
+++ b/custom/v3/command/data/src/main/java/org/apache/fineract/v3/command/data/CommandResponse.java
@@ -1,0 +1,44 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.fineract.v3.command.data;
+
+import java.io.Serial;
+import java.io.Serializable;
+import java.time.OffsetDateTime;
+import lombok.Data;
+import lombok.experimental.FieldNameConstants;
+import lombok.experimental.SuperBuilder;
+
+@Data
+@SuperBuilder
+@FieldNameConstants
+public class CommandResponse<T> implements Serializable {
+
+    @Serial
+    private static final long serialVersionUID = 1L;
+
+    private String tenantId;
+    private Long userId;
+    private String errorCode;
+    private String errorMessage;
+    private OffsetDateTime createdAt;
+
+    private T data;
+}
--- a/custom/v3/command/domain/src/main/resources/db/custom-changelog/0001_create_command_table.xml
+++ b/custom/v3/command/domain/src/main/resources/db/custom-changelog/0001_create_command_table.xml
@@ -1,0 +1,62 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements. See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership. The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License. You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied. See the License for the
+    specific language governing permissions and limitations
+    under the License.
+
+-->
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.1.xsd">
+
+    <changeSet author="fineract" id="1" context="postgresql">
+        <sql>
+            CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
+        </sql>
+    </changeSet>
+
+    <!-- Create the m_command table for PostgreSQL -->
+    <changeSet author="fineract" id="2" context="postgresql">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <tableExists tableName="m_command"/>
+            </not>
+        </preConditions>
+
+        <createTable tableName="m_command">
+            <column autoIncrement="true" name="id" type="BIGINT" remarks="Internal ID">
+                <constraints nullable="false" primaryKey="true"/>
+            </column>
+            <column name="request_idempotency_key" type="VARCHAR(50)" remarks="Unique request idempotency key">
+                <constraints nullable="false" unique="true"/>
+            </column>
+            <column name="command_processing_status" type="VARCHAR(255)" remarks="Command processing status"/>
+            <column name="result" type="TEXT" remarks="Command result"/>
+            <column name="result_status_code" type="INTEGER" remarks="Result status code"/>
+            <column name="created_date" type="TIMESTAMP WITH TIME ZONE" remarks="Creation date"/>
+            <column name="processed_at" type="TIMESTAMP WITH TIME ZONE" remarks="Processing timestamp"/>
+            <column name="body" type="TEXT" remarks="Request body"/>
+        </createTable>
+    </changeSet>
+
+    <!-- Add indexes for m_command table -->
+    <changeSet author="fineract" id="3">
+        <createIndex indexName="idx_m_command_command_processing_status" tableName="m_command">
+            <column name="command_processing_status"/>
+        </createIndex>
+    </changeSet>
+</databaseChangeLog>
--- a/custom/v3/command/mapping/src/main/java/org/apache/fineract/v3/command/mapping/CommandMapper.java
+++ b/custom/v3/command/mapping/src/main/java/org/apache/fineract/v3/command/mapping/CommandMapper.java
@@ -1,0 +1,59 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.fineract.v3.command.mapping;
+
+import static org.mapstruct.NullValueCheckStrategy.ALWAYS;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.util.UUID;
+import org.apache.fineract.infrastructure.core.config.MapstructMapperConfig;
+import org.apache.fineract.v3.command.data.CommandRequest;
+import org.apache.fineract.v3.command.domain.Command;
+import org.mapstruct.Mapper;
+import org.mapstruct.Mapping;
+import org.mapstruct.Named;
+
+@Mapper(config = MapstructMapperConfig.class, nullValueCheckStrategy = ALWAYS)
+public interface CommandMapper {
+
+    @Mapping(target = "id", ignore = true)
+    @Mapping(target = "requestIdempotencyKey", source = "requestIdempotencyKey", qualifiedByName = "mapRequestIdempotencyKey")
+    @Mapping(target = "body", source = "body", qualifiedByName = "mapBody")
+    Command map(CommandRequest<?> source);
+
+    @Named("mapRequestIdempotencyKey")
+    default String mapRequestIdempotencyKey(UUID requestIdempotencyKey) {
+        return requestIdempotencyKey != null ? requestIdempotencyKey.toString() : null;
+    }
+
+    @Named("mapBody")
+    default JsonNode mapBody(Object body) {
+        if (body instanceof JsonNode) {
+            return (JsonNode) body;
+        }
+        try {
+            ObjectMapper objectMapper = new ObjectMapper();
+            return objectMapper.valueToTree(body);
+        } catch (Exception e) {
+            throw new IllegalArgumentException("Error mapping body to JsonNode: " + body.getClass().getName(), e);
+        }
+    }
+}
--- a/custom/v3/note/data/src/main/java/org/apache/fineract/v3/note/data/NoteRequestBody.java
+++ b/custom/v3/note/data/src/main/java/org/apache/fineract/v3/note/data/NoteRequestBody.java
@@ -1,0 +1,38 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.fineract.v3.note.data;
+
+import jakarta.validation.constraints.NotBlank;
+import jakarta.validation.constraints.Size;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class NoteRequestBody {
+
+    @NotBlank(message = "Note cannot be null or blank")
+    @Size(max = 1000, message = "Note cannot exceed 1000 characters")
+    private String note;
+}
--- a/custom/v3/note/data/src/main/java/org/apache/fineract/v3/note/data/NoteResponseData.java
+++ b/custom/v3/note/data/src/main/java/org/apache/fineract/v3/note/data/NoteResponseData.java
@@ -1,0 +1,41 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.fineract.v3.note.data;
+
+import java.util.Map;
+import lombok.Data;
+import lombok.experimental.FieldNameConstants;
+import lombok.experimental.SuperBuilder;
+
+@Data
+@SuperBuilder
+@FieldNameConstants
+public class NoteResponseData {
+
+    private final Long resourceId;
+    private final Long officeId;
+    private final Long clientId;
+    private final Long groupId;
+    private final Long savingsId;
+    private final Long loanId;
+    private final Long loanTransactionId;
+
+    Map<String, Object> changes;
+}
--- a/custom/v3/note/service/src/main/java/org/apache/fineract/v3/note/service/NoteWritePlatformServiceImpl.java
+++ b/custom/v3/note/service/src/main/java/org/apache/fineract/v3/note/service/NoteWritePlatformServiceImpl.java
@@ -1,0 +1,211 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.fineract.v3.note.service;
+
+import java.util.HashMap;
+import java.util.Map;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.fineract.portfolio.client.domain.Client;
+import org.apache.fineract.portfolio.client.domain.ClientRepositoryWrapper;
+import org.apache.fineract.portfolio.group.domain.Group;
+import org.apache.fineract.portfolio.group.domain.GroupRepositoryWrapper;
+import org.apache.fineract.portfolio.loanaccount.domain.Loan;
+import org.apache.fineract.portfolio.loanaccount.domain.LoanRepositoryWrapper;
+import org.apache.fineract.portfolio.loanaccount.domain.LoanTransaction;
+import org.apache.fineract.portfolio.loanaccount.domain.LoanTransactionRepository;
+import org.apache.fineract.portfolio.loanaccount.exception.LoanTransactionNotFoundException;
+import org.apache.fineract.portfolio.note.domain.Note;
+import org.apache.fineract.portfolio.note.domain.NoteRepository;
+import org.apache.fineract.portfolio.note.exception.NoteNotFoundException;
+import org.apache.fineract.portfolio.note.exception.NoteResourceNotSupportedException;
+import org.apache.fineract.portfolio.savings.domain.SavingsAccount;
+import org.apache.fineract.portfolio.savings.domain.SavingsAccountRepositoryWrapper;
+import org.apache.fineract.v3.note.data.NoteCreateRequest;
+import org.apache.fineract.v3.note.data.NoteCreateResponse;
+import org.apache.fineract.v3.note.data.NoteDeleteRequest;
+import org.apache.fineract.v3.note.data.NoteDeleteResponse;
+import org.apache.fineract.v3.note.data.NoteResponseData;
+import org.apache.fineract.v3.note.data.NoteUpdateRequest;
+import org.apache.fineract.v3.note.data.NoteUpdateResponse;
+import org.springframework.stereotype.Service;
+
+@Slf4j
+@Service
+@RequiredArgsConstructor
+public class NoteWritePlatformServiceImpl implements NoteWritePlatformService {
+
+    private final ClientRepositoryWrapper clientRepository;
+    private final GroupRepositoryWrapper groupRepository;
+    private final LoanRepositoryWrapper loanRepository;
+    private final SavingsAccountRepositoryWrapper savingsAccountRepository;
+    private final LoanTransactionRepository loanTransactionRepository;
+    private final NoteRepository noteRepository;
+
+    @Override
+    public NoteCreateResponse createNote(NoteCreateRequest createRequest) {
+
+        Note newNoteEntity;
+        Long officeId = null;
+        Long clientId = null;
+        Long groupId = null;
+        Long loanId = null;
+        Long savingsId = null;
+
+        switch (createRequest.getNoteType()) {
+            case CLIENT -> {
+                clientId = createRequest.getResourceId();
+                Client client = clientRepository.findOneWithNotFoundDetection(clientId);
+                newNoteEntity = new Note(client, createRequest.getBody().getNote());
+                officeId = client.officeId();
+            }
+            case GROUP -> {
+                groupId = createRequest.getResourceId();
+                Group group = groupRepository.findOneWithNotFoundDetection(groupId);
+                newNoteEntity = new Note(group, createRequest.getBody().getNote());
+                officeId = group.officeId();
+            }
+            case LOAN -> {
+                loanId = createRequest.getResourceId();
+                Loan loan = loanRepository.findOneWithNotFoundDetection(loanId);
+                newNoteEntity = Note.loanNote(loan, createRequest.getBody().getNote());
+                officeId = loan.getOfficeId();
+            }
+            case SAVING_ACCOUNT -> {
+                savingsId = createRequest.getResourceId();
+                SavingsAccount savingAccount = savingsAccountRepository.findOneWithNotFoundDetection(savingsId);
+                officeId = savingAccount.getClient().getOffice().getId();
+                newNoteEntity = Note.savingNote(savingAccount, createRequest.getBody().getNote());
+            }
+            case LOAN_TRANSACTION -> {
+                Long loanTransactionId = createRequest.getResourceId();
+                LoanTransaction loanTransaction = loanTransactionRepository.findById(loanTransactionId)
+                        .orElseThrow(() -> new LoanTransactionNotFoundException(createRequest.getResourceId()));
+                Loan loan = loanTransaction.getLoan();
+                newNoteEntity = Note.loanTransactionNote(loan, loanTransaction, createRequest.getBody().getNote());
+                officeId = loan.getOfficeId();
+                loanId = loan.getId();
+            }
+            default -> throw new NoteResourceNotSupportedException(createRequest.getNoteType().name());
+        }
+
+        noteRepository.saveAndFlush(newNoteEntity);
+        return buildNoteCreateResponse(newNoteEntity.getId(), clientId, officeId, groupId, loanId, savingsId);
+    }
+
+    @Override
+    public NoteUpdateResponse updateNote(NoteUpdateRequest updateRequest) {
+
+        Long officeId = null;
+        Long clientId = null;
+        Long groupId = null;
+        Long loanId = null;
+        Long savingsId = null;
+        Long noteId = updateRequest.getNoteId();
+        Note persistedNote = noteRepository.findById(noteId).orElseThrow(() -> new NoteNotFoundException(noteId));
+
+        switch (updateRequest.getNoteType()) {
+            case CLIENT -> {
+                clientId = updateRequest.getResourceId();
+                Client client = clientRepository.findOneWithNotFoundDetection(clientId);
+                officeId = client.officeId();
+            }
+            case GROUP -> {
+                groupId = updateRequest.getResourceId();
+                Group group = groupRepository.findOneWithNotFoundDetection(groupId);
+                officeId = group.officeId();
+            }
+            case LOAN -> {
+                loanId = updateRequest.getResourceId();
+                Loan loan = loanRepository.findOneWithNotFoundDetection(loanId);
+                officeId = loan.getOfficeId();
+            }
+            case LOAN_TRANSACTION -> {
+                Long loanTransactionId = updateRequest.getResourceId();
+                LoanTransaction loanTransaction = loanTransactionRepository.findById(loanTransactionId)
+                        .orElseThrow(() -> new LoanTransactionNotFoundException(loanTransactionId));
+                Loan loan = loanTransaction.getLoan();
+                officeId = loan.getOfficeId();
+                loanId = loan.getId();
+            }
+            case SAVING_ACCOUNT -> {
+                savingsId = updateRequest.getResourceId();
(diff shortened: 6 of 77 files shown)


custom/v3/commands/data/src/main/java/org/apache/fineract/v3/commands/data/JsonTypedCommand.java
```diff
@@ -0,0 +1,43 @@
...
+
+@Getter
+@Builder
+public class JsonTypedCommand<T> {
+
+    private final T requestBody;
+    private final Long commandId;
+    private final Long resourceId;
+    private final Long subresourceId;
+    private final Long groupId;
```
vidakovic: Why are these IDs necessary here? This is mimicking exactly the same implementation as the current production one... in other words: if you have a command that has IDs belonging to different entities then it's not really type safe.

custom/v3/commands/data/src/main/java/org/apache/fineract/v3/commands/data/JsonTypedCommand.java
```diff
@@ -0,0 +1,43 @@
...
+ * under the License.
+ */
+package org.apache.fineract.v3.commands.data;
+
+import lombok.Builder;
+import lombok.Getter;
+
+@Getter
+@Builder
+public class JsonTypedCommand<T> {
```
vidakovic: Why is "JSON" necessary here? Tomorrow we'll use GRPC or AVRO to serialize and then the naming will be off.

custom/v3/commands/handler/src/main/java/org/apache/fineract/v3/commands/handler/NewTypedCommandSourceHandler.java
```diff
@@ -0,0 +1,27 @@
...
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.v3.commands.handler;
+
+import org.apache.fineract.infrastructure.core.data.CommandProcessingResult;
+import org.apache.fineract.v3.commands.data.JsonTypedCommand;
+
+public interface NewTypedCommandSourceHandler<T> {
```
vidakovic: Have to find a better name... tomorrow we might have another implementation... do we call it then "NewNewTypedCommandSourceHandler"? If in doubt always try to keep it as short and simple as possible: "CommandHandler"

custom/v3/commands/provider/dependencies.gradle
```diff
@@ -0,0 +1,26 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+dependencies {
+    implementation(project(':fineract-core'))
+    implementation(project(':custom:v3:commands:data'))
+    implementation(project(':custom:v3:commands:handler'))
+    implementation('org.apache.commons:commons-lang3')
+    implementation('com.google.guava:guava')
```
vidakovic: Why do we need this dependency (spoiler alert: assertions/validation of input values)? See comment below how to avoid.

custom/v3/commands/provider/src/main/java/org/apache/fineract/v3/commands/provider/NewCommandHandlerProvider.java
```diff
@@ -0,0 +1,95 @@
...
+     * found.
+     *
+     * @param entity
+     *            the entity to lookup the handler, must be given.
+     * @param action
+     *            the action to lookup the handler, must be given.
+     */
+    @SuppressWarnings("unchecked")
+    public <T> NewTypedCommandSourceHandler<T> getHandler(final String entity, final String action) {
+        Preconditions.checkArgument(StringUtils.isNoneEmpty(entity), "An entity must be given!");
```
vidakovic: You don't need 2 libs (commons-lang, guava) to do these checks/assertions. Just use "Objects.requireNonNull()".

custom/v3/commands/service/src/main/java/org/apache/fineract/v3/commands/service/CommandTypedWrapperBuilder.java
```diff
@@ -0,0 +1,115 @@
...
+    private T requestBody;
+    private String transactionId;
+    private Long productId;
+    private Long templateId;
+    private Long creditBureauId;
+    private Long organisationCreditBureauId;
+    private String jobName;
+    private String idempotencyKey;
+
+    public CommandTypedWrapper<T> buildCommandWrapper() {
```
vidakovic: Why is this function needed? How many times is it used (my bet: only once)? Why the slashes "//" at the end of the line?

custom/v3/commands/service/src/main/java/org/apache/fineract/v3/commands/service/CommandTypedWrapperBuilder.java
```diff
@@ -0,0 +1,115 @@
...
+                .productId(this.productId) //
+                .templateId(this.templateId) //
+                .creditBureauId(this.creditBureauId) //
+                .organisationCreditBureauId(this.organisationCreditBureauId) //
+                .jobName(this.jobName) //
+                .idempotencyKey(this.idempotencyKey) //
+                .build(); //
+    }
+
+    public CommandTypedWrapperBuilder<T> createNote(final CommandTypedWrapper<T> resourceDetails, final String resourceType,
```
vidakovic: Is  it really necessary to have these functions here? How many times are they used... my bet is again: only once... in that case: don't put them here... we follow similar strategies all over the code and end up with 7000 LoC with utility functions like that... anti-pattern.

custom/v3/commands/service/src/main/java/org/apache/fineract/v3/commands/service/SynchronousTypedCommandProcessingService.java
```diff
@@ -0,0 +1,102 @@
...
+import org.apache.fineract.infrastructure.core.exception.ErrorHandler;
+import org.apache.fineract.infrastructure.jobs.service.SchedulerJobRunnerReadService;
+import org.apache.fineract.infrastructure.security.service.PlatformSecurityContext;
+import org.apache.fineract.useradministration.domain.AppUser;
+import org.apache.fineract.v3.commands.data.CommandTypedWrapper;
+import org.springframework.stereotype.Service;
+
+@Service
+@RequiredArgsConstructor
+public class SynchronousTypedCommandProcessingService implements TypedCommandProcessingService {
```
vidakovic: Naming can be improved. In other words: this is just too long... we don't have to reflect every feature of this implementation in the name.

custom/v3/commands/service/src/main/java/org/apache/fineract/v3/commands/service/SynchronousTypedCommandProcessingService.java
```diff
@@ -0,0 +1,102 @@
...
+    public static final String COMMAND_SOURCE_ID = "commandSourceId";
+
+    private final PlatformSecurityContext context;
+    private final CommandSourceService commandSourceService;
+    private final IdempotencyKeyResolver idempotencyKeyResolver;
+    private final FineractRequestContextHolder fineractRequestContextHolder;
+    private final SchedulerJobRunnerReadService schedulerJobRunnerReadService;
+
+    @Override
+    public <T> CommandProcessingResult executeCommand(CommandTypedWrapper<T> wrapper) {
```
vidakovic: This still looks overly complicated and very similar to the current implementation. This is still very hard to reason about. I think we can do better and avoid these nested calls.

custom/v3/commands/service/src/main/java/org/apache/fineract/v3/commands/service/TypedCommandProcessingService.java
```diff
@@ -0,0 +1,28 @@
...
+ * under the License.
+ */
+package org.apache.fineract.v3.commands.service;
+
+import org.apache.fineract.infrastructure.core.data.CommandProcessingResult;
+import org.apache.fineract.v3.commands.data.CommandTypedWrapper;
+
+public interface TypedCommandProcessingService {
+
+    <T> CommandProcessingResult executeCommand(CommandTypedWrapper<T> wrapper);
```
vidakovic: Where is the type safety for the "CommandProcessingResult"?

custom/v3/note/api/src/main/java/org/apache/fineract/v3/note/api/NotesApiController.java
```diff
@@ -0,0 +1,172 @@
...
+import org.apache.fineract.v3.commands.service.CommandTypedWrapperBuilder;
+import org.apache.fineract.v3.note.data.NoteRequestBody;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
```
vidakovic: You are mixing read and write here... read related stuff is one PR, write related stuff is a separate PR... this cannot be approved, because this code is duplicated in 2 PRs!

vidakovic: A cleaner approach would be to have 1 controller for read and a separate one for write...

custom/v3/note/service/src/main/java/org/apache/fineract/v3/note/service/NoteWritePlatformServiceImpl.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.apache.fineract.infrastructure.core.data.CommandProcessingResult;
+import org.apache.fineract.portfolio.client.domain.Client;
+import org.apache.fineract.v3.commands.data.JsonTypedCommand;
+import org.apache.fineract.v3.note.data.NoteRequestBody;
+
+public class NoteWritePlatformServiceImpl implements NoteWritePlatformService {
+
+    @Override
+    public CommandProcessingResult createNote(JsonTypedCommand<NoteRequestBody> command) {
```
vidakovic: Anything "JSON" is just wrong here...  this is what we wanted to avoid in the first place. ONLY DTOs!

fineract-core/src/main/java/org/apache/fineract/commands/service/PortfolioCommandSourceWritePlatformServiceImpl.java
```diff
@@ -53,9 +53,10 @@ public class PortfolioCommandSourceWritePlatformServiceImpl implements Portfolio
     @Override
     public CommandProcessingResult logCommandSource(final CommandWrapper wrapper) {
```
vidakovic: This looks again like the changes from another of your PRs... can't be accepted, because you duplicated code!!! One PR, one concern!

fineract-provider/src/main/resources/application.properties
```diff
@@ -44,6 +44,9 @@ fineract.tenant.read-only-password=${FINERACT_DEFAULT_TENANTDB_RO_PWD:}
 fineract.tenant.read-only-parameters=${FINERACT_DEFAULT_TENANTDB_RO_CONN_PARAMS:}
 fineract.tenant.read-only-name=${FINERACT_DEFAULT_TENANTDB_RO_NAME:}
 
+# backwards compatibility for 'org.apache.fineract.infrastructure.security.filter.TenantAwareBasicAuthenticationFilter'
+baseUrl=https://localhost:8443${FINERACT_SERVER_SERVLET_CONTEXT_PATH:/fineract-provider}
```
vidakovic: We need to use a variable for host and port... those are available already with Spring Boot (just don't remember the names). Should be used here instead of hardcoded values.

vidakovic (changes requested):

fineract-provider/src/main/resources/application.properties
Zeyad2003: > We need to use a variable for host and port... those are available already with Spring Boot (just don't remember the names). Should be used here instead of hardcoded values.

How to use spring built-in vars?
I tried this but didn't work:
```properties
# FINERACT-914
server.forward-headers-strategy=framework
server.port=${FINERACT_SERVER_PORT:8443}
server.address=${FINERACT_SERVER_ADDRESS:localhost}
server.servlet.context-path=${FINERACT_SERVER_SERVLET_CONTEXT_PATH:/fineract-provider}
server.compression.enabled=${FINERACT_SERVER_COMPRESSION_ENABLED:true}

# backwards compatibility for 'org.apache.fineract.infrastructure.security.filter.TenantAwareBasicAuthenticationFilter
baseUrl=https://${server.address}:${server.port}${server.servlet.context-path}
```

and the checks failed as following:
```
Run curl -f -k --retry 10 --retry-connrefused --connect-timeout 30 --retry-delay 30 https://localhost:8443/fineract-provider/actuator/health
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (35) error:0A000[12](https://github.com/Zeyad2003/fineract/actions/runs/10129914159/job/28010527803#step:8:13)6:SSL routines::unexpected eof while reading
Error: Process completed with exit code 35.
```

any thoughts?

custom/v3/commands/service/src/main/java/org/apache/fineract/v3/commands/service/CommandTypedWrapperBuilder.java
Zeyad2003: The **//** at the end is a trick to prevent the code formatter from merging multiple assignments onto the same line, for better readability.

custom/v3/commands/service/src/main/java/org/apache/fineract/v3/commands/service/CommandTypedWrapperBuilder.java
Zeyad2003: The entire `CommandTypedWrapperBuilder` class has been removed. Now, I've implemented the `NoteUtilityMethods` class, which only contains the specialized methods for notes.

custom/v3/commands/service/src/main/java/org/apache/fineract/v3/commands/service/SynchronousTypedCommandProcessingService.java
Zeyad2003: Noted.
Became `CommandProcessingServiceImpl`

custom/v3/commands/service/src/main/java/org/apache/fineract/v3/commands/service/SynchronousTypedCommandProcessingService.java
Zeyad2003: I have no idea how I could make it more organized .. just split the logic into methods and invoke them.
What's your suggestion about resolving this flow to be more organized (handle idempotency, initializing `CommandSource`, maker-checker, locating handler, publishing the event...etc.)?

custom/v3/note/api/src/main/java/org/apache/fineract/v3/note/api/NotesApiController.java
Zeyad2003: I can split the controller into 2 files, but the Commands' PR depends on Queries PR (same structure, share packages, and dependencies ...etc).
If the #3915 got merged then this one can be rebased with it seamlessly (no duplication).

Am I missing something?

custom/v3/commands/data/src/main/java/org/apache/fineract/v3/commands/data/JsonTypedCommand.java
Zeyad2003: removed.
Now we depend on `resourceId` and `subresourceId` only for representing IDs

fineract-provider/src/main/resources/application.properties
Zeyad2003: This approach has passed the checks:
```properties
server.forward-headers-strategy=framework
server.port=${FINERACT_SERVER_PORT:8443}
server.domain=${FINERACT_SERVER_DOMAIN:localhost}
server.servlet.context-path=${FINERACT_SERVER_SERVLET_CONTEXT_PATH:/fineract-provider}
server.compression.enabled=${FINERACT_SERVER_COMPRESSION_ENABLED:true}

# backwards compatibility for 'org.apache.fineract.infrastructure.security.filter.TenantAwareBasicAuthenticationFilter'
baseUrl=https://${server.domain}:${server.port}${server.servlet.context-path}
```

It appears that the `server.address` property causes some issues. The exact reason is ambiguous to me.

custom/v3/commands/data/src/main/java/org/apache/fineract/v3/commands/data/CommandTypedWrapper.java
```diff
@@ -62,9 +49,9 @@ public CommandWrapper getCommandWrapper() {
             throw new RuntimeException(e);
         }
 
-        return new CommandWrapper(officeId, groupId, clientId, loanId, savingsId, actionName, entityName, entityId,
-                subentityId, href, json, transactionId, productId, null, creditBureauId, organisationCreditBureauId,
-                jobName, idempotencyKey);
+        return new CommandWrapper(null, null, null, null, null, actionName, entityName, entityId,
```
vidakovic: No... this whole function is written in a way that goes against all best practices of dependency injection; "ObjectMapper" should be re-used/injected (with all the global configurations)... here we create it with default configurations over and over again... but even without that: this function is acts like copy function... how many times do we use this? My bet is again: once... this programming style is obscuring more things than it saves or helps...

custom/v3/commands/data/src/main/java/org/apache/fineract/v3/commands/data/TypedCommand.java
```diff
@@ -57,7 +47,7 @@ public JsonCommand getJsonCommand() {
         }
 
         return new JsonCommand(null, json, null, null, entityName, resourceId,
```
vidakovic: We have a "CommandWrapper" and a "JsonCommand"... why? Can't we just express what we need in one class?

custom/v3/note/util/src/main/java/org/apache/fineract/v3/note/util/NoteUtilityMethods.java
```diff
@@ -68,35 +68,18 @@ public static NoteType getNoteType(final String resourceType) {
...
-            }
-            case GROUP -> {
-                resourceNameForPermissions = GROUPNOTE;
-                resourceDetails = resourceDetails.toBuilder().groupId(resourceId).build();
-            }
-            default -> resourceNameForPermissions = INVALIDNOTE;
-        }
-
-        return resourceDetails.toBuilder().entityName(resourceNameForPermissions).build();
+        String resourceNameForPermissions = switch (type) {
```
vidakovic: Without even reading more details: very likely the wrong place for ANYTHING permission related! Note: we don't need to squeeze every concept into one class... we are going to run out of space if we create another class... remidner: this belongs to Spring Security (remember when we talked about "Access Desicion Voters"?).

custom/v3/commands/data/src/main/java/org/apache/fineract/v3/commands/data/TypedCommand.java
Zeyad2003: Yes, I've thought about that, there's no major differences. (just action & entity names).
Thanks for mentioning it!

custom/v3/commands/data/src/main/java/org/apache/fineract/v3/commands/data/CommandTypedWrapper.java
Zeyad2003: It's just temporary. Not implemented there to remain.
Actually, I'm breaking some rules to get things to work first. 
Then it can be improved or even removed.

custom/v3/command/data/src/main/java/org/apache/fineract/v3/command/data/CommandResponse.java
```diff
@@ -0,0 +1,48 @@
...
+    private static final long serialVersionUID = 1L;
+
+    private UUID requestId;
+    private String tenantId;
+    private Long userId;
+    private String errorCode;
+    private String errorMessage;
+    private Instant createdAt;
+
+    @JsonUnwrapped
```
vidakovic: I don't think you want to do this... JsonUnwrapped is flattening out the attributes of an embedded object.

custom/v3/command/domain/src/main/resources/db/changelog/parts/0001_create_command_table.xml
```diff
@@ -0,0 +1,29 @@
...
+    KIND, either express or implied. See the License for the
+    specific language governing permissions and limitations
+    under the License.
+
+-->
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.1.xsd">
+    <changeSet author="fineract" id="1" context="postgresql">
+<!--        TODO: Complete the changeSet for `Command` table-->
```
vidakovic: Detailed column definitions missing.

custom/v3/command/service/src/main/java/org/apache/fineract/v3/command/service/IdempotencyService.java
```diff
@@ -0,0 +1,29 @@
...
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.fineract.v3.command.service;
+
+import java.util.UUID;
+
+public interface IdempotencyService {
```
vidakovic: Instead of creating a new one why not use for now the existing idempotency implementation from upstream. Let's stay conservative with the changes we introduce here. Otherwise we'll open too many cans of worms.

custom/v3/common/configuration/src/main/java/org/apache/fineract/v3/common/configuration/CommonCacheConfiguration.java
```diff
@@ -0,0 +1,57 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.data.redis.cache.RedisCacheConfiguration;
+import org.springframework.data.redis.serializer.GenericJackson2JsonRedisSerializer;
+import org.springframework.data.redis.serializer.RedisSerializationContext;
+
+@Slf4j
+@EnableCaching
+@Configuration
+@RequiredArgsConstructor
+public class CommonCacheConfiguration {
```
vidakovic: While I think this would be a good thing to have (see my presentation sources): not our concern now. No new stuff when we already have a cache implementation. Let's stay focused.

custom/v3/note/data/src/main/java/org/apache/fineract/v3/note/data/NoteRequestBody.java
```diff
@@ -0,0 +1,28 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.fineract.v3.note.data;
+
+import jakarta.validation.constraints.NotBlank;
+import jakarta.validation.constraints.Size;
+
+public record NoteRequestBody(
```
vidakovic: Either we use records everywhere or we stay with the classic Java beans. Given that some community members have difficulties already with maintaining proper Java beans then let's stay with the bean + Lombok combination for now. Records are relatively new and introducing them will require a separate discussion on mailing list.

custom/v3/note/data/src/main/java/org/apache/fineract/v3/note/data/NoteResponseData.java
```diff
@@ -0,0 +1,43 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude;
+import java.util.Map;
+import lombok.Data;
+import lombok.experimental.FieldNameConstants;
+import lombok.experimental.SuperBuilder;
+
+@Data
+@SuperBuilder
+@FieldNameConstants
+@JsonInclude(JsonInclude.Include.NON_NULL)
```
vidakovic: Let's define these behaviors of the JSON parser when we configure Jackson on a global level... otherwise you'll have to be consistent and use that annotation everywhere... let's avoid boilerplate code where we can (read: other people will forget these things very easily and then the code quality suffers again).

custom/v3/note/service/src/main/java/org/apache/fineract/v3/note/service/NoteWritePlatformServiceImpl.java
```diff
@@ -0,0 +1,227 @@
...
+    private final GroupRepositoryWrapper groupRepository;
+    private final LoanRepositoryWrapper loanRepository;
+    private final SavingsAccountRepositoryWrapper savingsAccountRepository;
+    private final LoanTransactionRepository loanTransactionRepository;
+    private final NoteRepository noteRepository;
+
+    // =========================================== Create Note ===========================================
+
+    @Override
+    public NoteCreateResponse createNote(NoteCreateRequest createRequest) {
```
vidakovic, line 63: Order of function scopes.... first PUBLIC then PRIVATE... these additional comments for create, update, delete blocks are not necessary if people can clearly read where the public functions are (on top of the class). Note: sometimes private functions make the code more readable, but if you look at the details you'll see that this is a lot of repetitive code... I would argue that the private functions are not helping the readability here and you might consider inlining that code in the public functions. My bet would be that the overall length of this class will be also shorter and people don't have to jump back and forth to understand what happens here.

custom/v3/note/util/src/main/java/org/apache/fineract/v3/note/util/NoteUtilityMethods.java
```diff
@@ -0,0 +1,36 @@
...
+package org.apache.fineract.v3.note.util;
+
+import org.apache.fineract.portfolio.note.domain.NoteType;
+import org.apache.fineract.portfolio.note.exception.NoteResourceNotSupportedException;
+
+public final class NoteUtilityMethods {
+
+    private NoteUtilityMethods() {}
+
+    public static NoteType getNoteType(final String resourceType) {
```
vidakovic: How many times do you use this class? I bet only 2-3 times... and only in the API layer... doesn't seem worth it and again makes readers jump back and forth when they read the controller classes. If you are not saving a hundred lines of code (you won't) then please inline.

fineract-provider/src/main/java/org/apache/fineract/portfolio/note/domain/Note.java
```diff
@@ -56,6 +57,7 @@ public class Note extends AbstractAuditableWithUTCDateTimeCustom<Long> {
     @JoinColumn(name = "loan_transaction_id", nullable = true)
     private LoanTransaction loanTransaction;
 
+    @Setter
```
vidakovic, line 62: You need also @Getter to comply with the JPA specification.

fineract-provider/src/main/java/org/apache/fineract/portfolio/note/domain/Note.java
```diff
@@ -119,7 +121,7 @@ public Note(final Client client, final String note) {
         this.noteTypeId = NoteType.CLIENT.getValue();
     }
 
-    private Note(final Group group, final String note) {
+    public Note(final Group group, final String note) {
```
vidakovic, line 126: Use @RequiredArgsConstructor. Let's be consistent.

custom/v3/note/util/src/main/java/org/apache/fineract/v3/note/util/NoteUtilityMethods.java
Zeyad2003: Noted.

fineract-provider/src/main/java/org/apache/fineract/portfolio/note/domain/Note.java
Zeyad2003: There's another PR, #4020, that addresses the `Note` instantiation issues by implementing the **Factory** pattern. 
You're welcome to review it.

custom/v3/common/configuration/src/main/java/org/apache/fineract/v3/common/configuration/CommonCacheConfiguration.java
Zeyad2003: Noted.

custom/v3/command/data/src/main/java/org/apache/fineract/v3/command/data/CommandResponse.java
Zeyad2003: Yes, I used it to return the same response structure as the upstream API does.
It's been removed now.

> [!NOTE]
>
> The difference between responses:
>
> - **Current Response:**
> ```json
> {
>     // other attributes
>     "data": {
>         "resourceId": 1
>         // data attributes
>     }
> }
> ```
> - **The Upstream Response:**
> ```json
> {
>     "resourceId": 1
> }
> ```

custom/v3/note/data/src/main/java/org/apache/fineract/v3/note/data/NoteRequestBody.java
Zeyad2003: Noted.

custom/v3/note/service/src/main/java/org/apache/fineract/v3/note/service/NoteWritePlatformServiceImpl.java
Zeyad2003: Noted.

custom/v3/note/data/src/main/java/org/apache/fineract/v3/note/data/NoteResponseData.java
Zeyad2003: It's already configured globally in [`application-v3.properties`](https://github.com/Zeyad2003/fineract/blob/18c2d660373657fa7a55a652d9088eaeae75c45c/custom/v3/common/configuration/src/main/resources/application-v3.properties#L29)
Forgot to remove it from other data classes. 
Removed now.

fineract-provider/src/main/java/org/apache/fineract/portfolio/note/domain/Note.java
Zeyad2003: Noted.

Zeyad2003: Noted.

custom/v3/command/service/src/main/java/org/apache/fineract/v3/command/service/IdempotencyService.java
Zeyad2003: Could you please clarify what specific logic from the existing idempotency implementation you'd recommend using here?

vidakovic:
Please rebase into 1 commit.

custom/v3/command/data/src/main/java/org/apache/fineract/v3/command/data/CommandResponse.java
vidakovic: Yes, we want the first thing! Remove that Unwrapped thing please... and yes, the data structure looks a little bit different than upstream... but that's why we do this proof of concept and cleanup!

custom/v3/command/domain/src/main/resources/db/custom-changelog/0001_create_command_table.xml
```diff
@@ -0,0 +1,62 @@
...
+            </column>
+            <column name="request_idempotency_key" type="VARCHAR(50)" remarks="Unique request idempotency key">
+                <constraints nullable="false" unique="true"/>
+            </column>
+            <column name="command_processing_status" type="VARCHAR(255)" remarks="Command processing status"/>
+            <column name="command_result" type="TEXT" remarks="Command result"/>
+            <column name="result_status_code" type="INTEGER" remarks="Result status code"/>
+            <column name="created_date" type="TIMESTAMP WITH TIME ZONE" remarks="Creation date"/>
+            <column name="processed_at" type="TIMESTAMP WITH TIME ZONE" remarks="Processing timestamp"/>
+            <column name="request_body" type="TEXT" remarks="Request body"/>
```
vidakovic: First: keep attributes simple... rename to `body` (we know that it's from a request... and there will never be another body). Alternative name `payload`.

custom/v3/command/domain/src/main/resources/db/custom-changelog/0001_create_command_table.xml
```diff
@@ -0,0 +1,62 @@
...
+        <createTable tableName="m_command">
+            <column autoIncrement="true" name="id" type="BIGINT" remarks="Internal ID">
+                <constraints nullable="false" primaryKey="true"/>
+            </column>
+            <column name="request_idempotency_key" type="VARCHAR(50)" remarks="Unique request idempotency key">
+                <constraints nullable="false" unique="true"/>
+            </column>
+            <column name="command_processing_status" type="VARCHAR(255)" remarks="Command processing status"/>
+            <column name="command_result" type="TEXT" remarks="Command result"/>
+            <column name="result_status_code" type="INTEGER" remarks="Result status code"/>
```
vidakovic, line 49: Let's name this just `status`. Simpler and says everything.

custom/v3/command/domain/src/main/resources/db/custom-changelog/0001_create_command_table.xml
```diff
@@ -0,0 +1,62 @@
...
+
+        <createTable tableName="m_command">
+            <column autoIncrement="true" name="id" type="BIGINT" remarks="Internal ID">
+                <constraints nullable="false" primaryKey="true"/>
+            </column>
+            <column name="request_idempotency_key" type="VARCHAR(50)" remarks="Unique request idempotency key">
+                <constraints nullable="false" unique="true"/>
+            </column>
+            <column name="command_processing_status" type="VARCHAR(255)" remarks="Command processing status"/>
+            <column name="command_result" type="TEXT" remarks="Command result"/>
```
vidakovic: Let's name this just `result`. Simpler and says everything.

custom/v3/command/domain/src/main/resources/db/custom-changelog/0001_create_command_table.xml
```diff
@@ -0,0 +1,62 @@
...
+        </preConditions>
+
+        <createTable tableName="m_command">
+            <column autoIncrement="true" name="id" type="BIGINT" remarks="Internal ID">
+                <constraints nullable="false" primaryKey="true"/>
+            </column>
+            <column name="request_idempotency_key" type="VARCHAR(50)" remarks="Unique request idempotency key">
+                <constraints nullable="false" unique="true"/>
+            </column>
+            <column name="command_processing_status" type="VARCHAR(255)" remarks="Command processing status"/>
```
vidakovic, line 47: Why do we need yet another status... and in a different type? Let's do either integer or string or enum status...

custom/v3/command/domain/src/main/resources/db/custom-changelog/0001_create_command_table.xml
vidakovic: Also: type `TEXT` is not really the way to go... it should be in the database `jsonb` and Java side `com.fasterxml.jackson.databind.JsonNode`. That way we can keep the structure and the parsing is done automatically.

Check https://www.baeldung.com/spring-boot-jpa-storing-postgresql-jsonb section 3.

custom/v3/command/domain/src/main/resources/db/custom-changelog/0001_create_command_table.xml
vidakovic: Also: type `TEXT` is not really the way to go... it should be in the database `jsonb` and Java side `com.fasterxml.jackson.databind.JsonNode`. That way we can keep the structure and the parsing is done automatically.

Check https://www.baeldung.com/spring-boot-jpa-storing-postgresql-jsonb section 3.

custom/v3/command/domain/src/main/resources/db/custom-changelog/0001_create_command_table.xml
Zeyad2003: Because they should represent different things.

The integer => Represents the http status code.(200, 201, 403, 404 ...etc)

The enum => Represents the command execution status within the system. (PROCESSED, UNDER_PROCESSING, ERROR ...etc)

custom/v3/command/mapping/src/main/java/org/apache/fineract/v3/command/mapping/CommandMapper.java
```diff
@@ -0,0 +1,59 @@
...
+    @Mapping(target = "requestBody", source = "requestBody", qualifiedByName = "mapRequestBody")
+    Command map(CommandRequest<?> source);
+
+    @Named("mapRequestIdempotencyKey")
+    default String mapRequestIdempotencyKey(UUID requestIdempotencyKey) {
+        return requestIdempotencyKey != null ? requestIdempotencyKey.toString() : null;
+    }
+
+    @Named("mapRequestBody")
+    default JsonNode mapRequestBody(Object body) {
```
vidakovic: This should be done by a Spring Converter. See my previous comments above and the Baeldung link. If implemented correctly then this default function is not necessary (Spring Data JPA will automatically take care of it).

custom/v3/command/domain/src/main/resources/db/custom-changelog/0001_create_command_table.xml
Zeyad2003: Actually, I checked it, but couldn't resolve the mapping because the ORM needs to know how to map the attribute to a `jsonb` column, and the way explaied in the tutorial uses hibernate (which doesn't work with eclipselink).

custom/v3/note/starter/src/main/java/org/apache/fineract/v3/note/starter/NoteV3AutoConfiguration.java
```diff
@@ -0,0 +1,31 @@
...
+import org.springframework.boot.autoconfigure.AutoConfiguration;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.ComponentScan;
+
+@AutoConfiguration
+@EnableConfigurationProperties({ NoteProperties.class })
+@ComponentScan("org.apache.fineract.v3.note")
+@ConditionalOnProperty(value = "fineract.note.enabled", havingValue = "true")
+public class NoteV3AutoConfiguration {}
```
vidakovic, line 31: Just `NoteAutoConfiguration` will do.

Zeyad2003: I tried it but there's a conflict between it and the other configuration in the upstream code.

https://github.com/apache/fineract/blob/24c83b5f9f97eebe27d5c521673403ababa80d0f/fineract-provider/src/main/java/org/apache/fineract/portfolio/note/starter/NoteAutoConfiguration.java#L38

Here's the error logs:
```java
fineract-fineract-1  | org.springframework.beans.factory.BeanDefinitionStoreException: Failed to process import candidates for configuration class [org.apache.fineract.ServerApplication$Configuration]: Annotation-specified bean name 'noteAutoConfiguration' for bean class [org.apache.fineract.portfolio.note.starter.NoteAutoConfiguration] conflicts with existing, non-compatible bean definition of same name and class [org.apache.fineract.v3.note.starter.NoteAutoConfiguration]
```

custom/v3/command/domain/src/main/resources/db/custom-changelog/0001_create_command_table.xml
adamsaghy: @vidakovic I dont think jsonb would work with mysql or mariadb... isnt it postgres specific?

I reckon we can differentiate json and jsonb based on which engine is used if needed

Zeyad2003: @adamsaghy

You're right, `jsonb` is specific to PostgreSQL, but I think it's possible to configure the column as `json` for MySQL (and MariaDB) and `jsonb` for PostgreSQL.

The challenge I'm facing is how to map a `jsonb` attribute using EclipseLink.

PR #4020, opened 2024-08-10
FINERACT-2081: introduce factory pattern to Note entity instantiation.

Zeyad2003:
- Removed all redundant static instantiation methods and constructors from Note.java.
- Implemented `NoteFactory` class in the domain package.
- Used Builder pattern to make it easy to instantiate the Note object.
- Changed all places to use the new Factor implementation.

## Description

## Checklist

- [X] Write the commit message as per https://github.com/apache/fineract/#pull-requests

- [X] Create/update unit or integration tests for verifying the changes made.

- [X] Follow coding conventions at https://cwiki.apache.org/confluence/display/FINERACT/Coding+Conventions.

- [X] Submission is not a "code dump".  (Large changes can be made "in repository" via a branch.  Ask on the developer mailing list for guidance, if required.)

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/note/domain/Note.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/note/domain/Note.java
@@ -23,10 +23,12 @@ import jakarta.persistence.Entity;
 import jakarta.persistence.JoinColumn;
 import jakarta.persistence.ManyToOne;
 import jakarta.persistence.Table;
-import java.util.LinkedHashMap;
-import java.util.Map;
-import org.apache.commons.lang3.StringUtils;
-import org.apache.fineract.infrastructure.core.api.JsonCommand;
+import lombok.AccessLevel;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
 import org.apache.fineract.infrastructure.core.domain.AbstractAuditableWithUTCDateTimeCustom;
 import org.apache.fineract.portfolio.client.domain.Client;
 import org.apache.fineract.portfolio.group.domain.Group;
@@ -37,25 +39,30 @@ import org.apache.fineract.portfolio.savings.domain.SavingsAccountTransaction;
 import org.apache.fineract.portfolio.shareaccounts.domain.ShareAccount;
 
 @Entity
+@Getter
+@NoArgsConstructor(access = AccessLevel.PROTECTED)
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+@Builder(access = AccessLevel.PACKAGE)
 @Table(name = "m_note")
 public class Note extends AbstractAuditableWithUTCDateTimeCustom<Long> {
 
     @ManyToOne
-    @JoinColumn(name = "client_id", nullable = true)
+    @JoinColumn(name = "client_id")
     private Client client;
 
     @ManyToOne
-    @JoinColumn(name = "group_id", nullable = true)
+    @JoinColumn(name = "group_id")
     private Group group;
 
     @ManyToOne
-    @JoinColumn(name = "loan_id", nullable = true)
+    @JoinColumn(name = "loan_id")
     private Loan loan;
 
     @ManyToOne
-    @JoinColumn(name = "loan_transaction_id", nullable = true)
+    @JoinColumn(name = "loan_transaction_id")
     private LoanTransaction loanTransaction;
 
+    @Setter
     @Column(name = "note", length = 1000)
     private String note;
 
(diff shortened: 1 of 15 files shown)


fineract-provider/src/main/java/org/apache/fineract/portfolio/note/domain/NoteFactory.java
```diff
@@ -0,0 +1,96 @@
...
+
+import org.apache.fineract.portfolio.client.domain.Client;
+import org.apache.fineract.portfolio.group.domain.Group;
+import org.apache.fineract.portfolio.loanaccount.domain.Loan;
+import org.apache.fineract.portfolio.loanaccount.domain.LoanTransaction;
+import org.apache.fineract.portfolio.savings.domain.SavingsAccount;
+import org.apache.fineract.portfolio.savings.domain.SavingsAccountTransaction;
+import org.apache.fineract.portfolio.shareaccounts.domain.ShareAccount;
+
+public final class NoteFactory {
```
kjozsa, line 30: hi @Zeyad2003, I came across your PR today. I would suggest either using the builder, or using the factory, but putting a factory layer facading a builder layer seems to be an overkill to me.

Please give it a try to refactor going with either of the approaches and see if the code gets simpler and more straightforward.. Thanks!

kjozsa (changes requested):

fineract-provider/src/main/java/org/apache/fineract/portfolio/note/domain/NoteFactory.java
Zeyad2003: Hi @kjozsa, Thanks for your feedback.

I understand your point, and I'd agree but in different scenarios that don't involve JPA entities.

Using public `@Builder` may cause some violations like bypassing some validation checks, or improperly updating fields that should remain immutable after instantiation. 
> That's why I limited the access to `AccessLevel.PACKAGE`

Relying solely on *Factory*  would push us towards less flexible solutions like:
- Gigantic constructor with +30 parameter or multiple parameterized constructors definition (not flexible).

- Using some static methods that have their own problems but also violate the Single Responsibility Principle (SRP) and at the end of the day we might face the same constructor problem as well.

You can review the earlier discussion on using `@Builder` with JPA entities [here](https://lists.apache.org/thread/l2xdp07984bl4x1h34hyymr9q3o6qxxo).

Please let me know your thoughts.

Thank you!
Zeyad

fineract-provider/src/main/java/org/apache/fineract/portfolio/note/domain/Note.java
```diff
@@ -37,23 +42,27 @@
 import org.apache.fineract.portfolio.shareaccounts.domain.ShareAccount;
 
 @Entity
+@Getter
+@NoArgsConstructor
```
marta-jankovics: No-one should call the NoArgsConstructor, JPA is fine with protected access level, so could you please change it to be protected.

fineract-provider/src/main/java/org/apache/fineract/portfolio/note/domain/Note.java
```diff
@@ -37,23 +42,27 @@
...
+@AllArgsConstructor
```
marta-jankovics: AllArgsConstructor does not make sense for this entity, because from the several entity-type references only one (or in some special cases maybe more but never all) can be set.
The Builder will generate a private AllArgsConstructor anyway. Could you please remove this annotation?

fineract-provider/src/main/java/org/apache/fineract/portfolio/note/domain/Note.java
```diff
@@ -63,107 +72,17 @@ public class Note extends AbstractAuditableWithUTCDateTimeCustom<Long> {
     private Integer noteTypeId;
```
kjozsa, line 70: Could we use the enum itself in the entity and map it as @Enumerated(ORDINAL)? Wouldn't it achieve the same results, but allow to use the enum from the Java code side?

fineract-provider/src/main/java/org/apache/fineract/portfolio/note/domain/NoteFactory.java
kjozsa: Ok, I agree with your reasons, but in that case, I would enforce everyone to use the dedicated Factory for Note instantiation, so we'd need to get rid of the public constructors. 

I would also consider moving the `Note#update(JsonCommand)` function's functionality away from the entity, as this makes a direct dependency between the entity layer and the web layer..

Zeyad2003: Done.
It has been replaced by a private `update()` method within the `NoteWritePlatformService`.

fineract-provider/src/main/java/org/apache/fineract/portfolio/note/domain/Note.java
Zeyad2003: Noted.

fineract-provider/src/main/java/org/apache/fineract/portfolio/note/domain/Note.java
Zeyad2003: Actually, the `@Builder` annotation requires a proper constructor for initializing the entity, so I've limited the access to private instead.

fineract-provider/src/main/java/org/apache/fineract/portfolio/note/domain/Note.java
Zeyad2003: Yes, I think it would be better, and also I would recommend using `EnumType.STRING` instead to avoid any conflict if we change the order or something.

But I think it's out of the scope of this PR for now.

fineract-provider/src/main/java/org/apache/fineract/portfolio/note/service/NoteWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -48,6 +50,8 @@
 @Slf4j
 public class NoteWritePlatformServiceJpaRepositoryImpl implements NoteWritePlatformService {
 
+    private static final String NOTE = "note";
```
kjozsa: I'm very much against this, in such trivial cases. Especially if you are writing `"note"` just in the same file at line 84.

Zeyad2003: Noted. It has been removed.

kjozsa:
LGTM

vidakovic:
Why do we need a "NoteFactory" in the first place? This is just hiding the builder pattern for instantiating "Note" classes behind these static functions with endless parameters that are really not a good developer experience (especially when reading/reviewing code). I see no improvement over the original implementation... why not just use the builder pattern when instantiating the Note? If we just use the builders then everything is just perfectly clear to the reader and we don't have to carry around yet more boilerplate code... my 2 cents.

Zeyad2003:
@vidakovic 

Please refer to [this reply](https://github.com/apache/fineract/pull/4020#discussion_r1713451113) for an explanation of why I chose to use the **Factory** pattern as a facade for the **Builder** pattern.

We also had an earlier discussion [here](https://lists.apache.org/thread/l2xdp07984bl4x1h34hyymr9q3o6qxxo) about using `@Builder` with JPA entities, where @marta-jankovics shared valuable insights on the appropriate access level for employing the builder approach.

Looking forward to your thoughts.

Best regards,
Zeyad

fineract-provider/src/main/java/org/apache/fineract/portfolio/note/domain/Note.java
marta-jankovics: If a class is annotated with @Builder, then a package-private constructor is generated with all fields as arguments (as if @AllArgsConstructor(access = AccessLevel.PACKAGE) is present on the class), and it is as if this constructor has been annotated with @Builder instead. Note that this constructor is only generated if you haven't written any constructors and also haven't added any explicit @XArgsConstructor annotations. In those cases, lombok will assume an all-args constructor is present and generate code that uses it; this means you'd get a compiler error if this constructor is not present.
So yes, you need to add the @AllArgsConstructor, just because @NoArgsConstructor is there. Private is fine.

fineract-provider/src/main/java/org/apache/fineract/portfolio/note/domain/Note.java
marta-jankovics: Unfortunately we cant change this right now, because this Integer is not the ordinal of the enum, but a custom number associated with each enum type. This topic would worth a discussion and then a refactor of the more hundred enum-like properties not using real enum types, but this is really not the scope of this PR.

fineract-provider/src/main/java/org/apache/fineract/portfolio/note/service/NoteWritePlatformServiceJpaRepositoryImpl.java
marta-jankovics: I've found several (> 10) static (or not static :) ) "note" parameter definition, everywhere in Fineract. Something like this:
public static final String NOTE_PARAM = "note";
If you refactor the notes anyway, it would be nice to add this to a central place only once and everywhere (all the account types) could use this.
Do you think it fits this PR?

marta-jankovics:
LGTM
Added some suggestions that would be nice to implement, but not blocking the PR to be merged.

fineract-provider/src/main/java/org/apache/fineract/portfolio/note/service/NoteWritePlatformServiceJpaRepositoryImpl.java
Zeyad2003: Actually, the definitions of "note" appear in many places across the codebase. You can use the following command to locate them:

```bash
grep -rnw --include="*.java" '.' -e '[A-Za-z_][A-Za-z0-9_]*\s*=\s*"note"'
```

However, addressing this is something that can be tackled later. For now, I need to focus on other tasks related to implementing a type-safe API.

vidakovic:
We are mixing unrelated things... JPA and entity classes are a whole different thing... but there too I would qualify the static functions as an anti pattern....

Zeyad2003:
@vidakovic 

I understand your point, but I believe relying solely on the builder pattern can introduce challenges.

For example, consider this method:

```java
public static Note createLoanTransactionNote(Loan loan, LoanTransaction loanTransaction, String note) {
    return Note.builder()
            .loan(loan)
            .client(loan.client())
            .loanTransaction(loanTransaction)
            .noteTypeId(NoteType.LOAN_TRANSACTION.getValue())
            .note(note)
            .build();
}
```

This method is used in 14 places. Without a factory, each instance would need to know the full structure of the `Note` entity and the necessary attributes to set, (e.g. when creating `LoanTransactionNote` we should set `client` attribute as well).

Additionally, a factory method allows us to encapsulate extra logic, such as fetching required data, applying business rules, or performing validation ...etc 
BTW, I understand your concerns about validation at the domain level, and your suggestion to enforce it at the database schema. However, as @marta-jankovics  pointed out earlier "it's too late, and the error message is too ugly".

Regarding static methods, perhaps we could establish guidelines to ensure they’re only created when necessary and in an acceptable form.

I'd like to know your thoughts on how we might balance these concerns while keeping the codebase as maintainable and clear as possible.

Best regards!

vidakovic:
I don't agree. And I can show why, but this here is not the main task. So, let's move on with the write requests, make them work and document everything for the gsoc project.

This here doesn't contribute and is a distraction.

kjozsa:
> Why do we need a "NoteFactory" in the first place? This is just hiding the builder pattern for instantiating "Note" classes behind these static functions with endless parameters that are really not a good developer experience (especially when reading/reviewing code). I see no improvement over the original implementation... why not just use the builder pattern when instantiating the Note? If we just use the builders then everything is just perfectly clear to the reader and we don't have to carry around yet more boilerplate code... my 2 cents.

That was my very first thought as well :slightly_smiling_face: The slight adventage with having both might be that the client doesn't need to know the mandatory fields for each "note type", instead the factory forces it to do the right thing... 

I'm still not sold it's completely worth the additional LOC though.

vidakovic:
Let's close this one for now. I don't think the factory pattern introduces any improvements here. And Zeyad is occupied finishing his GSoC project anyway... which has higher priority.

Zeyad2003:
I've opened another PR (#4071) for using builder pattern only and removed all other static instantiation methods.
If you have time, please take a look.

PR #4021, opened 2024-08-12
FINERACT-2114: Interest Rate Modification - adjust reschedule validation

janez89:
## Description

Adjust progressive loan reschedule validation logic to accept interest rate change for any day in the schedule.

Progressive loan Reschedule validations:
- accepts only one change at the same time (due date change and interest change should be two reschedule operation on the loan)
- due date change accept only period due date
- interest change accept any day in the schedule
- validation take care and reject any change that is not supported yet on reschedule

https://issues.apache.org/jira/browse/FINERACT-2114

### DIFF
--- a/fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/rescheduleloan/data/LoanRescheduleRequestDataValidator.java
+++ b/fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/rescheduleloan/data/LoanRescheduleRequestDataValidator.java
@@ -42,6 +42,7 @@ import org.apache.fineract.infrastructure.core.service.DateUtils;
 import org.apache.fineract.portfolio.loanaccount.domain.Loan;
 import org.apache.fineract.portfolio.loanaccount.domain.LoanCharge;
 import org.apache.fineract.portfolio.loanaccount.domain.LoanRepaymentScheduleInstallment;
+import org.apache.fineract.portfolio.loanaccount.loanschedule.domain.LoanScheduleType;
 import org.apache.fineract.portfolio.loanaccount.rescheduleloan.RescheduleLoansApiConstants;
 import org.apache.fineract.portfolio.loanaccount.rescheduleloan.domain.LoanRescheduleRequest;
 import org.springframework.stereotype.Component;
@@ -140,11 +141,77 @@ public class LoanRescheduleRequestDataValidator {
         dataValidatorBuilder.reset().parameter(RescheduleLoansApiConstants.rescheduleReasonCommentParamName).value(rescheduleReasonComment)
                 .ignoreIfNull().notExceedingLengthOf(500);
 
-        if (this.fromJsonHelper.parameterExists(RescheduleLoansApiConstants.emiParamName, jsonElement)
-                || this.fromJsonHelper.parameterExists(RescheduleLoansApiConstants.endDateParamName, jsonElement)) {
+        final LocalDate adjustedDueDate = this.fromJsonHelper.extractLocalDateNamed(RescheduleLoansApiConstants.adjustedDueDateParamName,
+                jsonElement);
+
+        if (adjustedDueDate != null && DateUtils.isBefore(adjustedDueDate, rescheduleFromDate)) {
+            dataValidatorBuilder.reset().parameter(RescheduleLoansApiConstants.rescheduleFromDateParamName).failWithCode(
+                    "adjustedDueDate.before.rescheduleFromDate", "Adjusted due date cannot be before the reschedule from date");
+        }
+
+        if (loan.getLoanProduct().getLoanProductRelatedDetail().getLoanScheduleType() == LoanScheduleType.CUMULATIVE) {
             final LocalDate endDate = jsonCommand.localDateValueOfParameterNamed(RescheduleLoansApiConstants.endDateParamName);
             final BigDecimal emi = jsonCommand.bigDecimalValueOfParameterNamed(RescheduleLoansApiConstants.emiParamName);
+            validateForCumulativeLoan(dataValidatorBuilder, loan, jsonElement, rescheduleFromDate, endDate, emi);
+        } else {
+            validateForProgressiveLoan(dataValidatorBuilder, loan, jsonElement, rescheduleFromDate, adjustedDueDate);
+        }
 
+        if (!dataValidationErrors.isEmpty()) {
+            throw new PlatformApiDataValidationException(dataValidationErrors);
+        }
+    }
+
+    private void validateForProgressiveLoan(final DataValidatorBuilder dataValidatorBuilder, final Loan loan, final JsonElement jsonElement,
+            final LocalDate rescheduleFromDate, final LocalDate adjustedDueDate) {
+        final var unsupportedFields = List.of(RescheduleLoansApiConstants.graceOnPrincipalParamName, //
+                RescheduleLoansApiConstants.graceOnInterestParamName, //
+                RescheduleLoansApiConstants.extraTermsParamName, //
+                RescheduleLoansApiConstants.emiParamName//
+        );
+
+        for (var unsupportedField : unsupportedFields) {
+            if (this.fromJsonHelper.parameterHasValue(unsupportedField, jsonElement)) {
+                dataValidatorBuilder.reset().parameter(unsupportedField).failWithCode(
+                        RescheduleLoansApiConstants.rescheduleSelectedOperationNotSupportedErrorCode,
+                        "Selected operation is not supported by Progressive Loan at a time during Loan Rescheduling");
+                return;
+            }
+        }
+
+        final LocalDate businessDate = DateUtils.getBusinessLocalDate();
+        LoanRepaymentScheduleInstallment installment = null;
+        if (rescheduleFromDate != null) {
+            boolean hasInterestRateChange = this.fromJsonHelper.parameterHasValue(RescheduleLoansApiConstants.newInterestRateParamName,
+                    jsonElement);
+            if (hasInterestRateChange && adjustedDueDate != null) {
+                dataValidatorBuilder.reset().parameter(RescheduleLoansApiConstants.adjustedDueDateParamName).failWithCode(
+                        RescheduleLoansApiConstants.rescheduleMultipleOperationsNotSupportedErrorCode,
+                        "Only one operation is supported at a time during Loan Rescheduling");
+                return;
+            }
+
+            if (hasInterestRateChange && !rescheduleFromDate.isAfter(businessDate)) {
+                dataValidatorBuilder.reset().parameter(RescheduleLoansApiConstants.rescheduleFromDateParamName).failWithCode(
+                        "loan.reschedule.interestratechange.reschedulefrom.shouldbefuture",
+                        "Loan Reschedule From date should be in the future.");
+            }
+
+            if (adjustedDueDate != null) {
+                installment = loan.getRepaymentScheduleInstallment(rescheduleFromDate);
+            } else if (hasInterestRateChange) {
+                installment = loan.getRelatedRepaymentScheduleInstallment(rescheduleFromDate);
+            }
+
+            validateReschedulingInstallment(dataValidatorBuilder, installment);
+        }
+
+        validateForOverdueCharges(dataValidatorBuilder, loan, installment);
+    }
+
+    private void validateForCumulativeLoan(final DataValidatorBuilder dataValidatorBuilder, final Loan loan, final JsonElement jsonElement,
+            final LocalDate rescheduleFromDate, final LocalDate endDate, final BigDecimal emi) {
+        if (emi != null || endDate != null) {
             dataValidatorBuilder.reset().parameter(RescheduleLoansApiConstants.endDateParamName).value(endDate).notNull();
             dataValidatorBuilder.reset().parameter(RescheduleLoansApiConstants.emiParamName).value(emi).notNull().positiveAmount();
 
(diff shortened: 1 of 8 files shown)


fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/rescheduleloan/data/LoanRescheduleRequestDataValidator.java
```diff
@@ -140,11 +141,85 @@ public void validateForCreateAction(final JsonCommand jsonCommand, final Loan lo
...
+                RescheduleLoansApiConstants.graceOnInterestParamName, //
+                RescheduleLoansApiConstants.extraTermsParamName, //
+                RescheduleLoansApiConstants.emiParamName//
+        );
+
+        for (var unsupportedField : unsupportedFields) {
+            if (this.fromJsonHelper.parameterHasValue(unsupportedField, jsonElement)) {
+                dataValidatorBuilder.reset().parameter(unsupportedField).failWithCode(
+                        RescheduleLoansApiConstants.rescheduleSelectedOperationNotSupportedErrorCode,
+                        "Selected operation is not supported at a time during Loan Rescheduling");
```
adamsaghy: This is progressive loan specific restrictions. We might wanna mention it as part of the error message.

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/rescheduleloan/data/LoanRescheduleRequestDataValidator.java
```diff
@@ -140,11 +141,85 @@ public void validateForCreateAction(final JsonCommand jsonCommand, final Loan lo
...
+    private void validateForProgressiveLoan(final DataValidatorBuilder dataValidatorBuilder, final Loan loan, final JsonElement jsonElement,
+            final LocalDate rescheduleFromDate, final LocalDate adjustedDueDate) {
+        final var unsupportedFields = List.of(RescheduleLoansApiConstants.graceOnPrincipalParamName, //
+                RescheduleLoansApiConstants.graceOnInterestParamName, //
+                RescheduleLoansApiConstants.extraTermsParamName, //
+                RescheduleLoansApiConstants.emiParamName//
+        );
+
+        for (var unsupportedField : unsupportedFields) {
+            if (this.fromJsonHelper.parameterHasValue(unsupportedField, jsonElement)) {
```
adamsaghy, line 174: Why do we need this new check? Was it sending empty strings?

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/rescheduleloan/data/LoanRescheduleRequestDataValidator.java
```diff
@@ -140,11 +141,85 @@ public void validateForCreateAction(final JsonCommand jsonCommand, final Loan lo
         dataValidatorBuilder.reset().parameter(RescheduleLoansApiConstants.rescheduleReasonCommentParamName).value(rescheduleReasonComment)
                 .ignoreIfNull().notExceedingLengthOf(500);
 
-        if (this.fromJsonHelper.parameterExists(RescheduleLoansApiConstants.emiParamName, jsonElement)
-                || this.fromJsonHelper.parameterExists(RescheduleLoansApiConstants.endDateParamName, jsonElement)) {
+        final LocalDate adjustedDueDate = this.fromJsonHelper.extractLocalDateNamed(RescheduleLoansApiConstants.adjustedDueDateParamName,
+                jsonElement);
+
+        if (adjustedDueDate != null && DateUtils.isBefore(adjustedDueDate, rescheduleFromDate)) {
```
adamsaghy, line 147: I dont think we need this. This validation does not related to interest rate change.

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/rescheduleloan/data/LoanRescheduleRequestDataValidator.java
```diff
@@ -140,11 +141,85 @@ public void validateForCreateAction(final JsonCommand jsonCommand, final Loan lo
...
+                dataValidatorBuilder.reset().parameter(RescheduleLoansApiConstants.adjustedDueDateParamName).failWithCode(
+                        RescheduleLoansApiConstants.rescheduleMultipleOperationsNotSupportedErrorCode,
+                        "Only one operation is supported at a time during Loan Rescheduling");
+                return;
+            }
+
+            if (hasInterestRateChange && !rescheduleFromDate.isAfter(businessDate)) {
+                dataValidatorBuilder.reset().parameter(RescheduleLoansApiConstants.rescheduleFromDateParamName).failWithCode(
+                        "loan.reschedule.interestratechange.reschedulefrom.shouldstartnextday",
+                        "Loan Reschedule From date should start on the next business day.");
```
adamsaghy: Misleading error message. It must be future date, but does not necessarily need to be the next business day...

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/rescheduleloan/data/LoanRescheduleRequestDataValidator.java
```diff
@@ -156,15 +231,6 @@ public void validateForCreateAction(final JsonCommand jsonCommand, final Loan lo
                             .failWithCode("repayment.schedule.installment.does.not.exist", "Repayment schedule installment does not exist");
                 }
             }
-
-        }
```
adamsaghy, line 160: Why do you remove this validation?

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/rescheduleloan/data/LoanRescheduleRequestDataValidator.java
```diff
@@ -140,11 +141,85 @@ public void validateForCreateAction(final JsonCommand jsonCommand, final Loan lo
...
+                installment = loan.getRelatedRepaymentScheduleInstallment(rescheduleFromDate);
+            }
+
+            if (installment == null) {
+                dataValidatorBuilder.reset().parameter(RescheduleLoansApiConstants.rescheduleFromDateParamName)
+                        .failWithCode("repayment.schedule.installment.does.not.exist", "Repayment schedule installment does not exist");
+            }
+
+            if (installment != null && installment.isObligationsMet()) {
+                dataValidatorBuilder.reset().parameter(RescheduleLoansApiConstants.rescheduleFromDateParamName)
```
adamsaghy: We should have the same validation and same error when due date was changed for progressive loan...

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/rescheduleloan/service/LoanRescheduleRequestWritePlatformServiceImpl.java
```diff
@@ -195,29 +196,24 @@ public CommandProcessingResult create(JsonCommand jsonCommand) {
...
-
-                    // update the value of the "rescheduleFromDate" variable
-                    rescheduleFromDate = localDate;
-                }
+            if (rescheduleFromDate != null) {
+                final boolean isProgressiveLoanInterestRateChangeOperation = jsonCommand
+                        .hasParameterValue(RescheduleLoansApiConstants.newInterestRateParamName)
+                        && loan.getLoanProductRelatedDetail().getLoanScheduleType() == LoanScheduleType.PROGRESSIVE;
+                // get installment by due date
+                LoanRepaymentScheduleInstallment installment = isProgressiveLoanInterestRateChangeOperation
```
adamsaghy: I reckon we can use `loan.getRelatedRepaymentScheduleInstallment` for both case, no?

adamsaghy (changes requested):
Please kindly see my review!

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/rescheduleloan/data/LoanRescheduleRequestDataValidator.java
janez89: Good point

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/rescheduleloan/data/LoanRescheduleRequestDataValidator.java
janez89: the question is why I use `parameterHasValue` to check or just why I use the check here?
The answer to the first question is yes. It sends an empty string and the built-in validation utility recognizes the client sent the parameter.

On the other side, we ensure that the validation catches those parameters that are not supported by progressive loan.

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/rescheduleloan/data/LoanRescheduleRequestDataValidator.java
janez89: Not removed, just moved out of the method because of it's common with progressive loan

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/rescheduleloan/service/LoanRescheduleRequestWritePlatformServiceImpl.java
janez89: we have to ensure if we do due date rescheduling than the provided date should be on due date.
We can use this method also, but in that case we have to check the provided date is an due date.

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/rescheduleloan/data/LoanRescheduleRequestDataValidator.java
janez89: yes could be, we can find a better way to handle it.
for me the next business date is that day which is specified by the organisation. If no holiday setup then every day is a business day. The first message that I used was exactly same "Should be in the future".

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/rescheduleloan/service/LoanRescheduleRequestWritePlatformServiceImpl.java
adamsaghy: I reckon we already doing it within validations, since if date is not an installment due date then it will be null  and NPE will be triggered in the next line.

But it is okay, we can have this different logic :)

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/rescheduleloan/data/LoanRescheduleRequestDataValidator.java
janez89: This validation is related to due date change and this logic needed for both cases.
We have data-driven operation here and it contains a minor refactor that why we have this

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/rescheduleloan/data/LoanRescheduleRequestDataValidator.java
adamsaghy: I am not fully sure how holidays play a part here... Are you happy to change the error message or you wanna keep this one?

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/rescheduleloan/data/LoanRescheduleRequestDataValidator.java
adamsaghy: ooh... it was moved... it is not a new validation... i see

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/rescheduleloan/service/LoanRescheduleRequestWritePlatformServiceImpl.java
janez89: I totally lost in this GitHub code, I just see this is not the Validator class.
This is historical here, because originally the `getRelatedRepaymentScheduleInstallment` method excluded the due date, but the updated logic already accepts that. So you are right, here we can just use that method.

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/rescheduleloan/data/LoanRescheduleRequestDataValidator.java
janez89: No, I will change it. It should be more self-expressive to avoid misunderstanding.

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/rescheduleloan/data/LoanRescheduleRequestDataValidator.java
janez89: yes, currently we have the same for cumulative and progressive loans.
Just this fragment duplicated. May would much more nice extract into a new method.

adamsaghy (approved):
LGTM

adamsaghy (approved):
LGTM

PR #4042, opened 2024-08-30
FINERACT-2081: Validation error in mandatory loan account parameters

alberto-art3ch:
## Description

In CreateLoan request, missing mandatory fields which were being returned as 400 in the past are currently being returned as 403 ? Is this an expected change? Please confirm. The following fields are being returned with 403 code when missing

- clientId
- productId
- principal
- transactionProcessingStrategyCode
- expectedDisbursementDate
- submittedOnDate

[FINERACT-2081](https://issues.apache.org/jira/browse/FINERACT-2081)

Screenshot 2024-08-30 at 2 17 35 p m

## Checklist

### DIFF
--- a/fineract-core/src/main/java/org/apache/fineract/infrastructure/core/data/DataValidatorBuilder.java
+++ b/fineract-core/src/main/java/org/apache/fineract/infrastructure/core/data/DataValidatorBuilder.java
@@ -1056,4 +1056,12 @@ public class DataValidatorBuilder {
             throw new PlatformApiDataValidationException(dataValidationErrors);
         }
     }
+
+    public static ApiParameterError buildValidationParameterApiError(final String resource, final String parameterName,
+            final String errorCode, final String errorMessage, final Object... defaultUserMessageArgs) {
+        final String validationErrorCode = "validation.msg." + resource + "." + parameterName + errorCode;
+        String defaultEnglishMessage = "The parameter `" + parameterName + "` " + errorMessage;
+        return ApiParameterError.parameterError(validationErrorCode, defaultEnglishMessage, parameterName, defaultUserMessageArgs);
+    }
+
 }
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/serialization/LoanApplicationValidator.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/serialization/LoanApplicationValidator.java
@@ -248,6 +248,9 @@ public final class LoanApplicationValidator {
         boolean isMeetingMandatoryForJLGLoans = configurationDomainService.isMeetingMandatoryForJLGLoans();
 
         final Long productId = this.fromApiJsonHelper.extractLongNamed(LoanApiConstants.productIdParameterName, element);
+        if (productId == null) {
+            throwMandatoryParameterError(LoanApiConstants.productIdParameterName);
+        }
         final LoanProduct loanProduct = this.loanProductRepository.findById(productId)
                 .orElseThrow(() -> new LoanProductNotFoundException(productId));
 
@@ -257,7 +260,6 @@ public final class LoanApplicationValidator {
         final Group group = groupId != null ? this.groupRepository.findOneWithNotFoundDetection(groupId) : null;
 
         validateClientOrGroup(client, group, productId);
-
         validateOrThrow("loan", baseDataValidator -> {
             final String loanTypeStr = this.fromApiJsonHelper.extractStringNamed(LoanApiConstants.loanTypeParameterName, element);
             baseDataValidator.reset().parameter(LoanApiConstants.loanTypeParameterName).value(loanTypeStr).notNull();
@@ -394,7 +396,6 @@ public final class LoanApplicationValidator {
             baseDataValidator.reset().parameter(LoanApiConstants.interestCalculationPeriodTypeParameterName)
                     .value(interestCalculationPeriodType).notNull().inMinMaxRange(0, 1);
 
-            boolean isInterestBearing = false;
             if (loanProduct.isLinkedToFloatingInterestRate()) {
                 if (isEqualAmortization) {
                     throw new EqualAmortizationUnsupportedFeatureException("floating.interest.rate", "floating interest rate");
@@ -430,7 +431,6 @@ public final class LoanApplicationValidator {
                 baseDataValidator.reset().parameter(interestRateDifferentialParameterName).value(interestRateDifferential).notNull()
                         .zeroOrPositiveAmount().inMinAndMaxAmountRange(loanProduct.getFloatingRates().getMinDifferentialLendingRate(),
                                 loanProduct.getFloatingRates().getMaxDifferentialLendingRate());
-                isInterestBearing = true;
             } else {
 
                 if (this.fromApiJsonHelper.parameterExists(LoanApiConstants.isFloatingInterestRate, element)) {
@@ -448,7 +448,6 @@ public final class LoanApplicationValidator {
                         .extractBigDecimalWithLocaleNamed(LoanApiConstants.interestRatePerPeriodParameterName, element);
                 baseDataValidator.reset().parameter(LoanApiConstants.interestRatePerPeriodParameterName).value(interestRatePerPeriod)
                         .notNull().zeroOrPositiveAmount();
-                isInterestBearing = interestRatePerPeriod != null && interestRatePerPeriod.compareTo(BigDecimal.ZERO) > 0;
             }
 
             final Integer amortizationType = this.fromApiJsonHelper
@@ -511,7 +510,6 @@ public final class LoanApplicationValidator {
 
             final LocalDate submittedOnDate = this.fromApiJsonHelper.extractLocalDateNamed(LoanApiConstants.submittedOnDateParameterName,
                     element);
-
             baseDataValidator.reset().parameter(LoanApiConstants.submittedOnDateParameterName).value(submittedOnDate).notNull();
 
             if (this.fromApiJsonHelper.parameterExists(LoanApiConstants.submittedOnNoteParameterName, element)) {
@@ -523,8 +521,8 @@ public final class LoanApplicationValidator {
 
             final String transactionProcessingStrategy = this.fromApiJsonHelper
                     .extractStringNamed(LoanApiConstants.transactionProcessingStrategyCodeParameterName, element);
-
-            validateTransactionProcessingStrategy(transactionProcessingStrategy, loanProduct, baseDataValidator);
+            baseDataValidator.reset().parameter(LoanApiConstants.transactionProcessingStrategyCodeParameterName)
+                    .value(transactionProcessingStrategy).notNull();
 
             validateLinkedSavingsAccount(element, baseDataValidator);
 
@@ -727,9 +725,6 @@ public final class LoanApplicationValidator {
 
             validateBorrowerCycle(element, loanProduct, clientId, groupId, baseDataValidator);
 
-            loanProductDataValidator.fixedLengthValidations(transactionProcessingStrategy, isInterestBearing, numberOfRepayments,
-                    repaymentEvery, element, baseDataValidator);
-
             // Validate If the externalId is already registered
             final String externalIdStr = this.fromApiJsonHelper.extractStringNamed("externalId", element);
             ExternalId externalId = ExternalIdFactory.produce(externalIdStr);
@@ -744,12 +739,11 @@ public final class LoanApplicationValidator {
             loanScheduleValidator.validateDownPaymentAttribute(loanProduct.getLoanProductRelatedDetail().isEnableDownPayment(), element);
 
             checkForProductMixRestrictions(element);
-            validateSubmittedOnDate(element, null, null, loanProduct);
             validateDisbursementDetails(loanProduct, element);
             validateCollateral(element);
             // validate if disbursement date is a holiday or a non-working day
             validateDisbursementDateIsOnNonWorkingDay(expectedDisbursementDate);
-            Long officeId = client != null ? client.getOffice().getId() : group.getOffice().getId();
+            Long officeId = resolveOfficeId(client, group);
             validateDisbursementDateIsOnHoliday(expectedDisbursementDate, officeId);
             final Integer recurringMoratoriumOnPrincipalPeriods = this.fromApiJsonHelper
                     .extractIntegerWithLocaleNamed("recurringMoratoriumOnPrincipalPeriods", element);
@@ -758,6 +752,32 @@ public final class LoanApplicationValidator {
                 loanProductDataValidator.validateRepaymentPeriodWithGraceSettings(numberOfRepayments, graceOnPrincipalPayment,
                         graceOnInterestPayment, graceOnInterestCharged, recurringMoratoriumOnPrincipalPeriods, baseDataValidator);
             }
+        });
+
+        validateSubmittedOnDate(element, null, null, loanProduct);
+
+        final String transactionProcessingStrategy = this.fromApiJsonHelper
+                .extractStringNamed(LoanApiConstants.transactionProcessingStrategyCodeParameterName, element);
+        validateTransactionProcessingStrategy(transactionProcessingStrategy, loanProduct);
+
+        fixedLengthValidations(element);
+    }
+
+    private void fixedLengthValidations(final JsonElement element) {
+        validateOrThrow("loan", baseDataValidator -> {
+            boolean isInterestBearing = false;
+            final String transactionProcessingStrategy = this.fromApiJsonHelper
+                    .extractStringNamed(LoanApiConstants.transactionProcessingStrategyCodeParameterName, element);
+            final Integer numberOfRepayments = this.fromApiJsonHelper
+                    .extractIntegerWithLocaleNamed(LoanApiConstants.numberOfRepaymentsParameterName, element);
+            final Integer repaymentEvery = this.fromApiJsonHelper
+                    .extractIntegerWithLocaleNamed(LoanApiConstants.repaymentEveryParameterName, element);
+
+            final BigDecimal interestRatePerPeriod = this.fromApiJsonHelper
+                    .extractBigDecimalWithLocaleNamed(LoanApiConstants.interestRatePerPeriodParameterName, element);
+            isInterestBearing = interestRatePerPeriod != null && interestRatePerPeriod.compareTo(BigDecimal.ZERO) > 0;
+            loanProductDataValidator.fixedLengthValidations(transactionProcessingStrategy, isInterestBearing, numberOfRepayments,
+                    repaymentEvery, element, baseDataValidator);
         });
     }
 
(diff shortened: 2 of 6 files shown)


fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/serialization/LoanApplicationValidator.java
```diff
@@ -2127,4 +2139,25 @@ public static void validateOrThrow(String resource, Consumer<DataValidatorBuilde
...
+
+    private void setOfficeId(Long officeId, Client client, Group group) {
+        if (client != null) {
+            officeId = client.getOffice().getId();
+        }
+        if (group != null) {
+            officeId = group.getOffice().getId();
+        }
+
+        if (officeId == null) {
```
adamsaghy: I dont really understand why should we throw `clientId` is mandatory if the `officeId` is null. Please review this logic and rewrite to be more straightforward... also setOfficeId method which does not set anything is weird.

Also please be aware if you override the officeId inside this method it has absolutely no any effect outside of this method...

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/data/DataValidatorBuilder.java
```diff
@@ -1056,4 +1056,12 @@ public void throwValidationErrors() throws PlatformApiDataValidationException {
             throw new PlatformApiDataValidationException(dataValidationErrors);
         }
     }
+
+    public static ApiParameterError buildValidationParameterApiError(final String resource, final String parameterName,
```
adamsaghy, line 1060: I dont think we need this. You can just simply force return and the `baseDataValidator` would be processed till that point...

adamsaghy (changes requested):
Please kindly check my review!
Also please add test cases to cover the "correct" behaviour

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/serialization/LoanApplicationValidator.java
alberto-art3ch: Done! update it

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/data/DataValidatorBuilder.java
alberto-art3ch: I've moved almost all the cases, just one to use this new method

jdailey:
I just note that this is being done on an open ended ticket for "small bug fixes" (2081).... is this a new functionality or a bug fix and if a bug fix the idea should be to make sure the approach is clear, maintainable by the next person coming along.  If the logic and approach is not clear then that defeats the purpose of the changes.   The definition of "DONE" needs to be better than we found it.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/serialization/LoanApplicationValidator.java
```diff
@@ -1750,9 +1774,6 @@ private void officeSpecificLoanProductValidation(final Long productId, final Lon
     private void validateTransactionProcessingStrategy(final String transactionProcessingStrategy, final LoanProduct loanProduct,
             final DataValidatorBuilder baseDataValidator) {
 
-        baseDataValidator.reset().parameter(LoanApiConstants.transactionProcessingStrategyCodeParameterName)
```
adamsaghy, line 1753: why to remove this validation?

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/serialization/LoanApplicationValidator.java
```diff
@@ -2127,4 +2148,22 @@ public static void validateOrThrow(String resource, Consumer<DataValidatorBuilde
                     dataValidationErrors);
         }
     }
+
+    private Long setOfficeId(Client client, Group group) {
```
adamsaghy: It still not sets anything. Please rename it! Example: `fetchOfficeId` or something would be more descriptive...

adamsaghy (changes requested):
Please kindly check my review!
Also please add test cases to cover the "correct" behaviour

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/serialization/LoanApplicationValidator.java
```diff
@@ -466,6 +469,9 @@ private void validateForCreate(final JsonElement element) {
                     .extractLocalDateNamed(LoanApiConstants.expectedDisbursementDateParameterName, element);
             baseDataValidator.reset().parameter(LoanApiConstants.expectedDisbursementDateParameterName).value(expectedDisbursementDate)
                     .notNull();
+            if (expectedDisbursementDate == null) {
```
galovics: Alberto, can you pls explain this logic? One line above we're requiring the expectedDisbursementDate to be not null and here we are checking whether it's null and completely ignoring what's coming after that.

Isn't this a bypass to the validation logic (which would follow these lines)?

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/serialization/LoanApplicationValidator.java
```diff
@@ -511,8 +517,10 @@ private void validateForCreate(final JsonElement element) {
 
             final LocalDate submittedOnDate = this.fromApiJsonHelper.extractLocalDateNamed(LoanApiConstants.submittedOnDateParameterName,
                     element);
-
             baseDataValidator.reset().parameter(LoanApiConstants.submittedOnDateParameterName).value(submittedOnDate).notNull();
+            if (submittedOnDate == null) {
```
galovics: Same as above.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/serialization/LoanApplicationValidator.java
```diff
@@ -523,7 +531,11 @@ private void validateForCreate(final JsonElement element) {
 
             final String transactionProcessingStrategy = this.fromApiJsonHelper
                     .extractStringNamed(LoanApiConstants.transactionProcessingStrategyCodeParameterName, element);
-
+            baseDataValidator.reset().parameter(LoanApiConstants.transactionProcessingStrategyCodeParameterName)
+                    .value(transactionProcessingStrategy).notNull();
+            if (transactionProcessingStrategy == null) {
```
galovics: Same as above.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/serialization/LoanApplicationValidator.java
galovics: Very much agreed. I wouldn't even say "fetch" because it suggests fetching from an external source (like the DB).

I'd rather call it something like "resolveOfficeId". Also, probably this wouldn't be the place to put such a method because in my opinion this is a very generic way to get the office ID for a specific client.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/serialization/LoanApplicationValidator.java
```diff
@@ -1410,7 +1431,10 @@ public void validateForModify(final JsonCommand command, final Loan loan) {
 
             // validate if disbursement date is a holiday or a non-working day
             validateDisbursementDateIsOnNonWorkingDay(expectedDisbursementDate);
-            Long officeId = client != null ? client.getOffice().getId() : group.getOffice().getId();
+            final Long officeId = setOfficeId(client, group);
+            if (officeId == null) {
```
galovics: Same as above.

galovics (changes requested):
In general I'm missing some tests to cover these changes.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/serialization/LoanApplicationValidator.java
alberto-art3ch: If we keep that validation we get a different error when the the `transactionProcessingStrategy` is null

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/serialization/LoanApplicationValidator.java
alberto-art3ch: Done! method name and logic updated

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/serialization/LoanApplicationValidator.java
alberto-art3ch: The reason to add this return  there, is because previously we were generating a HTTP 400 Bad Request when some required field like expectedDisbursementDate or principal or transactionProcessingStrategy, etc are missed. 

Now with the new way, we have those schema validations and other business logic mixed, then the business validations are returning first than schema validations.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/serialization/LoanApplicationValidator.java
alberto-art3ch: The reason to add this return  there, is because previously we were generating a HTTP 400 Bad Request when some required field like expectedDisbursementDate or principal or transactionProcessingStrategy, etc are missed. 

Now with the new way, we have those schema validations and other business logic mixed, then the business validations are returning first than schema validations.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/serialization/LoanApplicationValidator.java
alberto-art3ch: The reason to add this return  there, is because previously we were generating a HTTP 400 Bad Request when some required field like expectedDisbursementDate or principal or transactionProcessingStrategy, etc are missed. 

Now with the new way, we have those schema validations and other business logic mixed, then the business validations are returning first than schema validations.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/serialization/LoanApplicationValidator.java
alberto-art3ch: The reason to add this return  there, is because previously we were generating a HTTP 400 Bad Request when some required field like expectedDisbursementDate or principal or transactionProcessingStrategy, etc are missed. 

Now with the new way, we have those schema validations and other business logic mixed, then the business validations are returning first than schema validations.

alberto-art3ch (commented):
Please review

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/serialization/LoanApplicationValidator.java
galovics: Wait @alberto-art3ch. Are we running business validations before schema validations? Shouldn't it be the other way around? 

If we see schema-wise that something is missing, has negative value, etc, how can a business validator even run?

galovics: One more thing, can you point me to this "new way" schema validations and business validations?

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanOriginationValidationTest.java
```diff
@@ -172,6 +172,83 @@ public void uc2() {
...
+                    .loanTermFrequencyType(2)//
+                    .transactionProcessingStrategyCode(LoanProductTestBuilder.ADVANCED_PAYMENT_ALLOCATION_STRATEGY)//
+                    .repaymentEvery(1)//
+                    .repaymentFrequencyType(2)//
+            ;//
+
+            callFailedRuntimeException = Assertions.assertThrows(CallFailedRuntimeException.class,
+                    () -> loanTransactionHelper.applyLoan(applicationRequest03));
+
+            Assertions.assertTrue(callFailedRuntimeException.getMessage().contains("The parameter `submittedOnDate` is mandatory."));
```
adamsaghy, line 245: Would you mind checking the HTTP code?

adamsaghy (changes requested):
Please see my review!

alberto-art3ch:
> Please see my review!

Done! Comments reviewed and attended

galovics (changes requested):

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/serialization/LoanApplicationValidator.java
adamsaghy: Let me join to the conversation. Before the changes there were validations all over the place, "randomly" through the lifecycle of an action: 
- Some validations for repayment when the service was called first steps
- Some validations in the loan
- Some validations later in the service call
etc.

It was started to rearrange for a couple actions (like loan creation, approval, disbursal) but seems one of the validations type was just collecting "validations issues" in a list and at the end it was evaluated and all errors were thrown eith Http 400.

Now all the validations are in one place seems they got mixed and there were some validations that were throwing http 403 (domain rule exceptions) and this kind of error is immediately thrown, not collected.

To avoid breaking changes our intention was to not change any of the "legacy" validations and which exception is to be thrown, but unfortunately there was some regression (see Alberto's reply), so now his intention is to fix it.

On long run it would be best to have a conversarion on the DEV mail list regarding exception handling but for now, we are aiming to fix the regression.

i hope it helps to gain some extra insight.

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanOriginationValidationTest.java
```diff
@@ -172,6 +172,80 @@ public void uc2() {
...
+                    operationDate, 100.0, 5).numberOfRepayments(6)//
+                    .submittedOnDate(null) //
+                    .loanTermFrequency(6)//
+                    .loanTermFrequencyType(2)//
+                    .transactionProcessingStrategyCode(LoanProductTestBuilder.ADVANCED_PAYMENT_ALLOCATION_STRATEGY)//
+                    .repaymentEvery(1)//
+                    .repaymentFrequencyType(2)//
+            ;//
+            loanTransactionHelper.applyLoanWithError(applicationRequest03, 400);
+            callFailedRuntimeException = Assertions.assertThrows(CallFailedRuntimeException.class,
```
adamsaghy, line 227: Please dont do this.
Use the following to check: `assertEquals(400, callFailedRuntimeException.getResponse().code());`

integration-tests/src/test/java/org/apache/fineract/integrationtests/AdvancedPaymentAllocationLoanRepaymentScheduleTest.java
```diff
@@ -3158,7 +3158,7 @@ public void uc121() {
                     .repaymentEvery(15).repaymentFrequencyType(0).numberOfRepayments(3).loanTermFrequency(45).loanTermFrequencyType(0)
                     .principal(BigDecimal.valueOf(1000.0)).maxOutstandingLoanBalance(BigDecimal.valueOf(35000))
                     .transactionProcessingStrategyCode(FineractStyleLoanRepaymentScheduleTransactionProcessor.STRATEGY_CODE)
-                    .loanScheduleProcessingType(LoanScheduleProcessingType.HORIZONTAL.name()), 403);
+                    .loanScheduleProcessingType(LoanScheduleProcessingType.HORIZONTAL.name()), 400);
```
adamsaghy: It seems to me a business domain rule exception. I dont think this should be changed!

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/serialization/LoanApplicationValidator.java
```diff
@@ -1765,15 +1777,15 @@ private void validateTransactionProcessingStrategy(final String transactionProce
             if (LoanScheduleType.PROGRESSIVE.equals(loanProduct.getLoanProductRelatedDetail().getLoanScheduleType())) {
                 if (!LoanProductConstants.ADVANCED_PAYMENT_ALLOCATION_STRATEGY.equals(transactionProcessingStrategy)) {
                     // TODO: GeneralPlatformDomainRuleException vs PlatformApiDataValidationException
-                    throw new GeneralPlatformDomainRuleException(
```
adamsaghy: This is a breaking change! Also I dont think it is correct! Please revert it

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/serialization/LoanApplicationValidator.java
```diff
@@ -759,6 +759,14 @@ private void validateForCreate(final JsonElement element) {
                         graceOnInterestPayment, graceOnInterestCharged, recurringMoratoriumOnPrincipalPeriods, baseDataValidator);
             }
         });
+
+        validateOrThrow("loan", baseDataValidator -> {
```
adamsaghy: This is  not needed hence the validator methods throws its own exceptions.

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanOriginationValidationTest.java
```diff
@@ -172,6 +174,80 @@ public void uc2() {
...
+                    .transactionProcessingStrategyCode(LoanProductTestBuilder.ADVANCED_PAYMENT_ALLOCATION_STRATEGY)//
+                    .repaymentEvery(1)//
+                    .repaymentFrequencyType(2)//
+            ;//
+            callFailedRuntimeException = Assertions.assertThrows(CallFailedRuntimeException.class,
+                    () -> loanTransactionHelper.applyLoan(applicationRequest04));
+            assertEquals(400, callFailedRuntimeException.getResponse().code());
+            Assertions
+                    .assertTrue(callFailedRuntimeException.getMessage().contains("The parameter `expectedDisbursementDate` is mandatory."));
+        });
```
adamsaghy, line 262: @alberto-art3ch Please add test case that covers if loan product id is null, we got "The parameter `loanProductId` is mandatory."

integration-tests/src/test/java/org/apache/fineract/integrationtests/AdvancedPaymentAllocationLoanRepaymentScheduleTest.java
```diff
@@ -3784,8 +3784,7 @@ public void uc130() {
                             .loanScheduleProcessingType(LoanScheduleProcessingType.HORIZONTAL.name())));
 
             assertEquals(403, exception.getResponse().code());
-            assertTrue(exception.getMessage()
-                    .contains("error.msg.loan.repayment.strategy.can.not.be.different.than.advanced.payment.allocation"));
+            assertTrue(exception.getMessage().contains("error.msg.fixed.length.only.supported.for.advanced.payment.allocation"));
```
adamsaghy: This is incorrect! Fixed length validation has nothing to do here!

adamsaghy (dismissed):
LGTM

kjozsa:
@galovics do you have any concerns left with this PR?

adamsaghy (approved):
LGTM

PR #4048, opened 2024-09-05
FINERACT-2081: Fix backdated multi disbursement on Progressive Loan

janez89:
This P.R. fixes that case when there is a multi-disbursement progressive loan and the second disbursement date is earlier than the first.

### DIFF
--- a/fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/ProgressiveLoanScheduleGenerator.java
+++ b/fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/ProgressiveLoanScheduleGenerator.java
@@ -26,6 +26,7 @@ import java.math.RoundingMode;
 import java.time.LocalDate;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/ProgressiveLoanScheduleGenerator.java
```diff
@@ -103,11 +104,15 @@ public LoanScheduleModel generate(final MathContext mc, final LoanApplicationTer
 
         prepareDisbursementsOnLoanApplicationTerms(loanApplicationTerms);
 
+        final ArrayList<DisbursementData> disbursementDataList = new ArrayList<>(loanApplicationTerms.getDisbursementDatas());
+        disbursementDataList.sort(Comparator.comparing(DisbursementData::disbursementDate));
```
galovics, line 108: Is disbursementDate a Date or a Timestamp? Is it possible that they get same "disbursementDate"s? If so, we should fall back to a secondary sort attribute like created date/etc.

galovics (changes requested):
I'm missing tests for this that covers the scenario you're fixing. Please add them.

Thanks.

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/domain/ProgressiveLoanScheduleGenerator.java
janez89: The `DisbursementData::disbursementDate` is a LocalDate. This method will handle the fallback situation.

galovics: Can you pls explain how it would handle a secondary sorting in case there's a collision in date?

janez89: I'm not sure what you mean by "secondary sorting" ? If this logic runs again?

janez89: It could be easier to talk about the purpose of that code.
It's sorting the disbursement based on their effect date. Until there is no "actual date" it will use the expected date, and if disbursed it will use the actual date. The goal is just to apply disbursements in chronological order

galovics: I mean what if there are two disbursements on the same date? Which one comes first, do we know?

janez89: Okay, I understand.
The secondary sorting occurs in the original order in the submitted JSON array, just like before. 
If the date matches, we will retrieve the identical order as previously submitted by the client.
 
From a calculation perspective, if we have multiple disbursements on the same date, the amounts of those disbursements are accumulated. The "addition" operation is associative, so it doesn't matter which disbursement came first on the same day.

galovics: noted, do you think we could create a test-case to cover the "same disbursement day" behavior?

janez89: yes, I added an extra case to the integration test and the EMI calculator to cover the same date scenario.

adamsaghy (approved):
LGTM

PR #4081, opened 2024-09-30
FINERACT-1981:  Interest bearing EMI calculation enhancements

janez89:
This P.R. includes a new EMI calculation method where the EMI Calculator carries the interest amounts to the next repayment period if they don't fit the current repayment period.


fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/data/EmiInterestPeriod.java
```diff
@@ -22,12 +22,19 @@
 import java.time.LocalDate;
 import lombok.AllArgsConstructor;
 import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
 import org.apache.fineract.organisation.monetary.domain.Money;
 import org.jetbrains.annotations.NotNull;
 
 @AllArgsConstructor
+@ToString()
```
adamsaghy: Why do we need toString for this?

PR #4099, opened 2024-10-15
FINERACT-1981: Interest Bearing loan - EMI calculation rework

janez89:
This P.R. has a total rework of the EMI Calculator.

Includes:
- interest calculation with EMI amount quota (countable interest in a period)
- overdue cases
- early repayment
- payoff


adamsaghy (approved):
LGTM

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/transactionprocessor/impl/AdvancedPaymentScheduleTransactionProcessor.java
```diff
@@ -1527,6 +1507,74 @@ private Money processAllocationsHorizontally(LoanTransaction loanTransaction, Tr
...
+            emiCalculator.payPrincipal(model, inAdvanceInstallment.getDueDate(), payDate, paidPortion);
+            updateRepaymentPeriods(loanTransaction, ctx, model);
+        } else if (INTEREST.equals(paymentAllocationType.getAllocationType())) {
+            emiCalculator.payInterest(model, inAdvanceInstallment.getDueDate(), payDate, paidPortion);
+            updateRepaymentPeriods(loanTransaction, ctx, model);
+        }
+        return paidPortion;
+    }
+
+    private static void updateRepaymentPeriods(LoanTransaction loanTransaction, ProgressiveTransactionCtx ctx,
```
galovics: Shouldn't be static I'd assume.

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/transactionprocessor/impl/AdvancedPaymentScheduleTransactionProcessor.java
```diff
@@ -1527,6 +1507,74 @@ private Money processAllocationsHorizontally(LoanTransaction loanTransaction, Tr
...
+
+        switch (paymentAllocationType) {
+            case IN_ADVANCE_INTEREST -> inAdvanceInstallment.updateInterestCharged(payableDetails.getPayableInterest().getAmount());
+            case IN_ADVANCE_PRINCIPAL -> inAdvanceInstallment.updatePrincipal(payableDetails.getPayablePrincipal().getAmount());
+            default -> {
+            }
+        }
+    }
+
+    private static LocalDate calculateNewPayDateInCaseOfInAdvancePayment(LoanTransaction loanTransaction,
```
galovics: Same here, probably shouldn't be static.

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/data/InterestPeriod.java
```diff
@@ -0,0 +1,110 @@
...
+    public InterestPeriod(RepaymentPeriod repaymentPeriod, LocalDate fromDate, LocalDate dueDate, BigDecimal rateFactor,
+            Money disbursementAmount, Money balanceCorrectionAmount, Money outstandingLoanBalance) {
+        this.repaymentPeriod = repaymentPeriod;
+        this.fromDate = fromDate;
+        this.dueDate = dueDate;
+        this.rateFactor = rateFactor;
+        this.disbursementAmount = disbursementAmount;
+        this.balanceCorrectionAmount = balanceCorrectionAmount;
+        this.outstandingLoanBalance = outstandingLoanBalance;
+    }
```
galovics, line 57: I think this is easily could be converted into a Builder rather, because we can mess up the order of the params.

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/data/InterestPeriod.java
```diff
@@ -0,0 +1,110 @@
...
+
+    public InterestPeriod(RepaymentPeriod repaymentPeriod, InterestPeriod interestPeriod) {
+        this(repaymentPeriod, interestPeriod.getFromDate(), interestPeriod.getDueDate(), interestPeriod.getRateFactor(),
+                interestPeriod.getDisbursementAmount(), interestPeriod.getBalanceCorrectionAmount(),
+                interestPeriod.getOutstandingLoanBalance());
+    }
+
+    @Override
+    public int compareTo(@NotNull InterestPeriod o) {
+        return dueDate.compareTo(o.dueDate);
```
galovics, line 67: Isn't thisrelying on the fact that the dueDate will be not null? As far as the code is concerned above, there's a setter generated by Lombok for the dueDate so somebody can rather easily mess up the not null invariant.

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/data/InterestPeriod.java
```diff
@@ -0,0 +1,110 @@
...
+                    .plus(previousInterestPeriod.getDisbursementAmount()); //
+        }
+    }
+
+    private boolean isFirstInterestPeriod() {
+        return getRepaymentPeriod().getInterestPeriods().get(0).equals(this);
+    }
+
+    @Override
+    public String toString() {
```
galovics: Why not using a Lombok toString?

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/data/InterestPeriod.java
```diff
@@ -0,0 +1,110 @@
...
+            int index = getRepaymentPeriod().getInterestPeriods().indexOf(this);
+            InterestPeriod previousInterestPeriod = getRepaymentPeriod().getInterestPeriods().get(index - 1);
+            this.outstandingLoanBalance = previousInterestPeriod.getOutstandingLoanBalance() //
+                    .plus(previousInterestPeriod.getBalanceCorrectionAmount()) //
+                    .plus(previousInterestPeriod.getDisbursementAmount()); //
+        }
+    }
+
+    private boolean isFirstInterestPeriod() {
+        return getRepaymentPeriod().getInterestPeriods().get(0).equals(this);
```
galovics, line 104: Doesn't this rely on a false invariant again that the periods are ordered? Is this enforced somewhere?

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/data/ProgressiveLoanInterestScheduleModel.java
```diff
@@ -19,45 +19,167 @@
...
+
+    public ProgressiveLoanInterestScheduleModel deepCopy() {
+        return new ProgressiveLoanInterestScheduleModel(repaymentPeriods, interestRates, loanProductRelatedDetail,
+                installmentAmountInMultiplesOf);
     }
 
-    public void addInterestRate(final LocalDate newInterestDueDate, final BigDecimal newInterestRate) {
-        interestRates.add(new ProgressiveLoanInterestRate(newInterestDueDate, newInterestDueDate.plusDays(1), newInterestRate));
-        interestRates.sort(Collections.reverseOrder());
+    private List<RepaymentPeriod> copyRepaymentPeriods(final List<RepaymentPeriod> repaymentPeriods) {
```
galovics, line 63: I'm not sure that this is the right place for this logic. Can we extract it?

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/data/ProgressiveLoanInterestScheduleModel.java
```diff
@@ -19,45 +19,167 @@
...
-            LoanProductRelatedDetail loanProductRelatedDetail, Integer installmentAmountInMultiplesOf, MathContext mc) {
-        this(repayments, new ArrayList<>(1), loanProductRelatedDetail, installmentAmountInMultiplesOf, mc);
+    private final List<RepaymentPeriod> repaymentPeriods;
+    private final List<InterestRate> interestRates;
+    private final LoanProductRelatedDetail loanProductRelatedDetail;
+    private final Integer installmentAmountInMultiplesOf;
+
+    public ProgressiveLoanInterestScheduleModel(List<RepaymentPeriod> repaymentPeriods, LoanProductRelatedDetail loanProductRelatedDetail,
+            Integer installmentAmountInMultiplesOf) {
+        this.repaymentPeriods = repaymentPeriods;
```
galovics, line 44: What's the reason we're not copying the repayment periods here?

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/data/ProgressiveLoanInterestScheduleModel.java
```diff
@@ -19,45 +19,167 @@
...
 
     public int getLoanTermInDays() {
-        if (repayments.isEmpty()) {
+        if (repaymentPeriods.isEmpty()) {
             return 0;
         }
-        final var firstPeriod = repayments.get(0);
-        final var lastPeriod = repayments.size() > 1 ? repayments.get(repayments.size() - 1) : firstPeriod;
+        final RepaymentPeriod firstPeriod = repaymentPeriods.get(0);
+        final RepaymentPeriod lastPeriod = repaymentPeriods.size() > 1 ? repaymentPeriods.get(repaymentPeriods.size() - 1) : firstPeriod;
```
galovics, line 106: Seems like again relying on a non-existing invariant.

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/data/RepaymentPeriod.java
```diff
@@ -0,0 +1,176 @@
...
+import lombok.ToString;
+import org.apache.fineract.infrastructure.core.service.MathUtil;
+import org.apache.fineract.organisation.monetary.domain.Money;
+
+@ToString
+@EqualsAndHashCode(exclude = { "previous", "next" })
+public class RepaymentPeriod {
+
+    @ToString.Exclude
+    private final RepaymentPeriod previous;
```
galovics, line 37: Are you sure you want to create a linked list from this? 

What's the reasoning behind the recursive structure?

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/data/RepaymentPeriod.java
```diff
@@ -0,0 +1,176 @@
...
+    @Getter
+    private final LocalDate fromDate;
+    @Getter
+    private final LocalDate dueDate;
+    @Getter
```
galovics, line 42: By exposing the getter, you're making it possible to change the contents of the List without using the addInterestPeriod method which is the only thing taking care of the sorting, hence the sorted invariant cannot be fully true.

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/data/RepaymentPeriod.java
```diff
@@ -0,0 +1,176 @@
...
+        this.paidInterest = repaymentPeriod.paidInterest;
+        // There is always at least 1 interest period, by default with same from-due date as repayment period
+        for (InterestPeriod interestPeriod : repaymentPeriod.interestPeriods) {
+            interestPeriods.add(new InterestPeriod(this, interestPeriod));
+        }
+    }
+
+    public void addInterestPeriod(InterestPeriod interestPeriod) {
+        interestPeriods.add(interestPeriod);
+        Collections.sort(interestPeriods);
```
galovics: Why don't we use a sorted data structure instead? What about a TreeSet?

galovics (changes requested):

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/transactionprocessor/impl/AdvancedPaymentScheduleTransactionProcessor.java
janez89: Yes, I didn't see this, thanks.

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/transactionprocessor/impl/AdvancedPaymentScheduleTransactionProcessor.java
janez89: I didn't see this, thanks.

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/data/InterestPeriod.java
janez89: Yes, someone can set it to null, but the Interest Period "dueDate" has value and is not intended to override from outside, except the EMI Calculator Service. We can also flag the setter with a "NotNull" annotation to indicate it.

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/data/InterestPeriod.java
janez89: Originally it was part of a debug purpose and now I will remove.

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/data/InterestPeriod.java
janez89: Ordering used by old EMI logic, I remove now and not necessary anymore.

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/data/ProgressiveLoanInterestScheduleModel.java
janez89: Which condition leads to non-existing invariant?

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/data/RepaymentPeriod.java
janez89: Yes, the algorithm is simpler and more performant if there is a reference to the previous reference.
We also store these repayment periods in an ArrayList to utilize List features.
When this list constructed then it has constant elements.
Unfortunately we can't use java's LinkedList implementation because it hides the important pointer references, so you can't use a real linked list.

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/data/RepaymentPeriod.java
janez89: No, sort is not needed anymore. It was part of old logic, I remove it.

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/data/ProgressiveLoanInterestScheduleModel.java
janez89: The model has a "deepCopy" method and that method makes a copy from the whole nested model.
This method contains an extracted copy logic which responsible for making a copy from repayment periods including their interest periods. Each class (RepaymentPeriod, InterestPeriod) has its own copy constructor.

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/data/ProgressiveLoanInterestScheduleModel.java
janez89: The 1st constructor is just for initialization, the 2nd constructor does. 
That constructor calls the `copyRepaymentPeriods ` and just the logic itself is isolated well and extracted into a method.

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/loanschedule/data/RepaymentPeriod.java
galovics: I'm not necessarily in agreement with your point - at least when it comes to performance.
Do you have measurements on how much "more performant" it is?

Let's say we save 10ms by using this recursive approach - even though from maintenance and debugging point of view it's adding a huge complexity - compared to an imperative approach which takes 20ms. We do multiple queries to the DB which takes let's say 300ms. That means with the imperative approach, we have 320ms execution time overall and with the recursive it's 310ms. We saved 3% of the overall executing while having an error-prone and hard to debug approach.

I'd go imperative, unless we have a super good reasoning - other than the recursive vs imperative performance difference.

adamsaghy: I am not sure i am following you guys.  We are storing a pointer to the previous repayment period because there is recalculation logic which read some of the values from the previous repayment period and this way it was super simple... nothing else... we are not using it to iterate on repayment periods or anything else...
I see no issue here

adamsaghy (approved):
LGTM

PR #4108, opened 2024-10-18
FINERACT-1981: Fix Better EMI check in EMI Calculator

janez89:
This P.R. fixes that logic where we calculate and try to search for a better value for EMI.


galovics (changes requested):
I'm missing a test case for this change. Can you pls add one at least?

janez89:
Yes, it was part of an other repository, but now I included an EMI calculation version.

adamsaghy (approved):
LGTM

PR #4141, opened 2024-10-31
FINERACT-2081: Update password policy

leksinomi:
## Description

Update password validation to ensure consistency across application layers.

Ignore if these details are present on the associated [Apache Fineract JIRA ticket](https://github.com/apache/fineract/pull/1284).

## Checklist

Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Write the commit message as per https://github.com/apache/fineract/#pull-requests

- [x] Acknowledge that we will not review PRs that are not passing the build _("green")_ - it is your responsibility to get a proposed PR to pass the build, not primarily the project's maintainers.

- [x] Create/update unit or integration tests for verifying the changes made.

- [x] Follow coding conventions at https://cwiki.apache.org/confluence/display/FINERACT/Coding+Conventions.

- [x] Add required Swagger annotation and update API documentation at fineract-provider/src/main/resources/static/legacy-docs/apiLive.htm with details of any API changes

- [x] Submission is not a "code dump".  (Large changes can be made "in repository" via a branch.  Ask on the developer mailing list for guidance, if required.)

FYI our guidelines for code reviews are at https://cwiki.apache.org/confluence/display/FINERACT/Code+Review+Guide.

PR #4155, opened 2024-11-08
FINERACT-2125: implement server side rate limit on Login Panel

kjozsa:
Fineract has a reported security vulnerability, because it does not limit login attempts, effectively allowing brute force attacks on the login screen. These changes enforce a configurable limitation of allowed requests on certain API requests (currently applied for the login request only).

Note that the enforcement happens in each JVM separately, so for clustered deployment each node enforces limits separately. That should be considered when determining the configuration value on critical deployments.

## Description

## Checklist

- [X] Write the commit message as per https://github.com/apache/fineract/#pull-requests

- [X] Create/update unit or integration tests for verifying the changes made.

- [X] Follow coding conventions at https://cwiki.apache.org/confluence/display/FINERACT/Coding+Conventions.

- [X] Submission is not a "code dump".  (Large changes can be made "in repository" via a branch.  Ask on the developer mailing list for guidance, if required.)

### DIFF
--- a/fineract-provider/dependencies.gradle
+++ b/fineract-provider/dependencies.gradle
@@ -50,6 +50,7 @@ dependencies {
             )
     implementation(
             'org.springframework.boot:spring-boot-starter-web',
+            'org.springframework.boot:spring-boot-starter-aop',
             'org.springframework.boot:spring-boot-starter-security',
             'org.springframework.boot:spring-boot-starter-cache',
             'org.springframework.boot:spring-boot-starter-oauth2-resource-server',
@@ -85,6 +86,7 @@ dependencies {
             'com.squareup.okhttp3:okhttp',
             'com.squareup.okhttp3:okhttp-urlconnection',
 
+            'com.github.ben-manes.caffeine:caffeine:3.1.8',
             'org.apache.commons:commons-lang3',
             'commons-io:commons-io',
             'org.apache.poi:poi',
--- a/fineract-provider/src/main/java/org/apache/fineract/ServerApplication.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/ServerApplication.java
@@ -21,9 +21,11 @@ package org.apache.fineract;
 import java.io.IOException;
 import org.apache.fineract.infrastructure.core.boot.FineractLiquibaseOnlyApplicationConfiguration;
 import org.apache.fineract.infrastructure.core.boot.FineractWebApplicationConfiguration;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.boot.autoconfigure.jdbc.DataSourceProperties;
 import org.springframework.boot.builder.SpringApplicationBuilder;
 import org.springframework.boot.web.servlet.support.SpringBootServletInitializer;
+import org.springframework.context.annotation.EnableAspectJAutoProxy;
 import org.springframework.context.annotation.Import;
 
 /**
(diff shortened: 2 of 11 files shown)


vidakovic:
@kjozsa ... how about using Resilience4j and not implementing our own? See [here](https://bootcamptoprod.com/resilience4j-rate-limiter/) how this can be done by adding one dependency, using one annotation and a handful of application.properties entries...

Today we have this use case... tomorrow we might have more like these (not only for login)... or we might need to protect the whole REST API...

kjozsa:
doh, you're right.. I did not recall Resilience4j covering that usecase just as fine.. Thanks!

kjozsa:
@vidakovic I investigated the Resilience4j path, but I'm not sure it supports the usecase of rate limiting per source address. Are you sure this would work?

galovics:
@kjozsa are you sure this is the right way to approach this security vulnerability? I mean if Fineract is being put behind a CDN/LB, the remote IP address will not be the actual client address but the CDN/LB's address. 

How about just enforcing Fineract to use BCrypt with a minimum 8 characters, symbols, letters (upper,lowercase) and numbers as passwords? 

image

IOhacker:
@galovics the CDN/LB must be configured to forward the real IP. I think that nowadays they do it. @kjozsa why this ticket was closed? @meonkeys

meonkeys:
> the CDN/LB must be configured to forward the real IP

"must", well, I hope so, but there's no way Fineract could guarantee that

---

I understand the idea with closing this unmerged is because wrapping infra could/should handle it instead. Rate limiting is built into most ingress/reverse proxy/SIEM solutions.

Honestly, it seems silly, but maybe even a ~2sec pause / spinner after clicking "Login" would be sufficient. Right now it's a little jarring how quickly I'm logged in. :laughing: That is one means of rate limit, even if it is a bit silly. Hmm, maybe using bcrypt would basically achieve this? It usually incurs a pause when a password is crunched.

re: passwords in general, see: #4141

IOhacker:
@meonkeys if the rate limit is reached in the API call, then the telecom/infra guy will do an update on the CDN/LB settings. Apache Fineract must then include the instruction about how to enable/disable that rate limit. The pause... well that must be implemented by the UI and if it exposed to internet, then at leas a recaptcha should be enabled for any click that invokes the Apache Fineract Api call.

IOhacker:
@meonkeys  If this is not implemented here... then we are treating this as any other API and not as a Financial API....  https://openid.net/wg/fapi/

meonkeys:
I think we should keep it simple. How about this:

1. pause 5sec from the UI side after a failed login attempt, this is low-hanging fruit
2. add rate-limiting guidance to [Securing Fineract](https://cwiki.apache.org/confluence/display/FINERACT/Securing+Fineract) -- this covers traffic going directly to the API

> Financial API

The fact that it's a a financial API is all the more reason to keep it simple. Simpler is easier to secure, and we only want to take responsibility for the security we're able to enforce.

> exposed to internet

That's certainly not a case we're intending to support, is it? Our hardening advice should all but forbid opening ports to a Fineract API on the internet.

(separately, that gives me another idea: maybe we instead/also try to detect if Fineract is wide open and issue warnings, refuse to start, etc.)

IOhacker:
1. Agree on it.
2. We delegate the solution to another component. I think that is the
reason for raising the ticket. DM

meonkeys:
1. great!
2. component? just to be clear, I'm proposing a documentation update. I know this PR was opened for a specific purpose, but noone is saying we *must* implement it a certain way, and I actually don't think we need a complex server-side rate-limiter (even one provided by a 3rd party). I'd say let's just open a new PR for client-side rate limit and go ahead and do the doc update separately

IOhacker:
2. External Component is a software/hardware that implements the rate limit functionality that prevents the denial of service attacks. If we go for delegating this to that external component. Either in a explicit (documentation) or implicit way, we are considering that the APIs exposed by Apache Fineract doen't not specify a default limit on the number of API calls that a program can make; however, it is still possible to overwhelm the system with an excessive number of calls during a given period. User failure to conform an upper limit can result in system failure, with high numbers of request failures and other unexpected results.

meonkeys:
Ah yes, classic denial of service. Agreed!

PR #4175, opened 2024-11-20
FINERACT-2114: Fix EMI adjustment when Interest Rate change has with Early Repayment Adjust Last case

janez89:
This P.R. fixes an EMI adjustment bug which happens if there is an early repayment adjust last installment setup.


adamsaghy (approved):
LGTM

PR #4177, opened 2024-11-20
FINERACT-2149: disable liquibase phone home

meonkeys:
## Description

This patch globally disables Liquibase's phone home / telemetry using [their recommended method](https://docs.liquibase.com/analytics/home.html).

I'm new and naïve... please see my commits and commit logs, scrutinize carefully, and expect beginner mistakes.

See related discussion at https://lists.apache.org/thread/5s7zjsxdcqk1qvjnc1wbhqqv66shpl54

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/migration/TenantDatabaseUpgradeService.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/service/migration/TenantDatabaseUpgradeService.java
@@ -64,6 +64,10 @@ public class TenantDatabaseUpgradeService implements InitializingBean {
     // DO NOT REMOVE! Required for liquibase custom task initialization
     private final List<CustomTaskChange> customTaskChangesForDependencyInjection;
 
+    static {
+        System.setProperty("liquibase.analytics.enabled", "false");
+    }
+
     @Override
     public void afterPropertiesSet() throws Exception {
         if (notLiquibaseOnlyMode()) {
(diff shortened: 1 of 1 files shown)


IOhacker:
Squash your commits

fineract-provider/src/main/java/org/apache/fineract/ServerApplication.java
```diff
@@ -54,6 +54,7 @@ private static SpringApplicationBuilder configureApplication(SpringApplicationBu
     }
 
     public static void main(String[] args) throws IOException {
+        System.setProperty("liquibase.analytics.enabled", "false");
```
kjozsa: wouldn't it be placed more strategically at the TenantDatabaseUpgradeService itself, in a static initializer? Can you test it if it works the same way there?

IOhacker: @meonkeys I think it can be placed as a Fineract Property set to false by default and then this property, (instead of having the System Property), can be placed in the TenantDatabaseUpgradeService as suggested by @kjozsa in the TenantDatabaseUpgradeService as static initializer

meonkeys: @kjozsa: yes, it does.

@kjozsa and @IOhacker: how about we go with just this simple static initializer in TenantDatabaseUpgradeService:

```java
static {
    System.setProperty("liquibase.analytics.enabled", "false");
}
```

I can't imagine anyone *wanting* to set the flag to `true`, and it would be many fewer lines of code just to hardcode it to `"false"` in this manner rather than provide a configurable setting. If there's demand to make it configurable later, we could of course do that (especially once it becomes [easier to do so](https://github.com/spring-projects/spring-boot/issues/43067)). I feel like our use of Liquibase is up to us, though. An under the hood / upstream decision we get to make. Thoughts?

Also, is it clear that we *must* use `System.setProperty`? Our only other alternative is to use the environment setting `LIQUIBASE_ANALYTICS_ENABLED=false`, visible to the Java VM before/as it initializes. I don't see any way in the Liquibase API to disable phoning home at runtime.

I re-worked my commit and force-pushed this suggestion -- please see below and let me know your thoughts. No rush--please don't let me get in the way of more important work.

PR #4191, opened 2024-11-26
FINERACT-1981: Introduce Interest should not be calculated on past due principal amount

janez89:
This P.R. introduces a new flag on loan and loan product.

UPDATE: now this P.R. includes the business logic and integration tests also.


adamsaghy:
@janez89 Can you please attach some related test cases?
- Set value on loan product, fetch value from loan product
- Loan inherit it from loan product, fetch value from loan... etc

adamsaghy (approved):
LGTM

PR #4212, opened 2024-12-09
FINERACT-1981: Embeddable Progressive Loan Schedule generator

janez89:
This P.R. introduces an embeddable Progressive Loan Schedule generator.

Now we introduce a new fineract module: fineract-progressive-loan-embeddable-schedule-generator. This module generates a dependency-free portable Jar.
Introduced a new CI Test to ensure the built Jar is runnable and dependency-free.
This P.R. contains refactors mostly related to the Money and MonetaryMoney classes.

If someone includes a dependency that is not part of the sample application then the CI job will fail like this:
https://github.com/janez89/fineract/actions/runs/12240841911/job/34144606937


build.gradle
```diff
@@ -109,6 +111,7 @@ plugins {
     id 'se.thinkcode.cucumber-runner' version '0.0.11' apply false
     id "com.github.davidmc24.gradle.plugin.avro-base" version "1.9.1" apply false
     id 'org.openapi.generator' version '7.8.0' apply false
+    id 'com.gradleup.shadow' version '9.0.0-beta4' apply false
```
adamsaghy: Having beta dependency is not nice. Would you give a try with latest stable version? 8.3.5?

janez89: Downgraded to 8.3.5. Let us see

PR #4236, opened 2024-12-24
FINERACT-2148: Process monetary transaction after zero interest charge off 

mariiaKraievska:
## Description

Continue work made in the PR - https://github.com/apache/fineract/pull/4227

"In case a repayment or any other monetary activity happens, after the charge-off, we dont need to reprocess all the transactions and fetch the ProgressiveLoanInterestScheduleModel and use the EmiCalculator, we can simply use the existing repayment periods and process the transaction

In the AdvancedPaymentScheduleTransactionProcessor should handle like the loan does not do interest recalculation!"

And fix Reverse-replayed issue.

## Checklist

### DIFF
--- a/fineract-e2e-tests-runner/src/test/resources/features/LoanChargeOff.feature
+++ b/fineract-e2e-tests-runner/src/test/resources/features/LoanChargeOff.feature
@@ -1920,4 +1920,208 @@ Feature: Charge-off
       | 01 January 2023  | Disbursement     | 100.0   | 0.0       | 0.0      | 0.0   | 0.0       | 100.0        |
       | 14 February 2023 | Charge-off       | 100.85  | 100.0     | 0.85     | 0.0   | 0.0       | 0.0          |
 
+  Scenario: Charge-off on due date when loan behaviour is zero-interest and interestRecalculation - repayment after charge off
+    When Admin sets the business date to "1 January 2024"
+    And Admin creates a client with random data
+    And Admin creates a fully customized loan with the following data:
+      | LoanProduct                                                                            | submitted on date | with Principal | ANNUAL interest rate % | interest type     | interest calculation period | amortization type  | loanTermFrequency | loanTermFrequencyType | repaymentEvery | repaymentFrequencyType | numberOfRepayments | graceOnPrincipalPayment | graceOnInterestPayment | interest free period | Payment strategy            |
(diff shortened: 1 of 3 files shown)


kulminsky (approved):

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
```diff
@@ -3569,4 +3569,8 @@ public LoanRepaymentScheduleTransactionProcessor getTransactionProcessor() {
     public boolean isProgressiveSchedule() {
         return getLoanProductRelatedDetail().getLoanScheduleType() == PROGRESSIVE;
     }
+
+    public boolean isTransactionBeforeChargeOff(final boolean isChargedOff) {
+        return !isChargedOff || !LoanChargeOffBehaviour.ZERO_INTEREST.equals(this.getLoanProductRelatedDetail().getChargeOffBehaviour());
```
adamsaghy: Why do we need this `!LoanChargeOffBehaviour.ZERO_INTEREST.equals(this.getLoanProductRelatedDetail().getChargeOffBehaviour()` condition?

I think this whole method can be switched with `!ctx.isChargedOff()`. 

Am I missing something?

mariiaKraievska: This condition `!LoanChargeOffBehaviour.ZERO_INTEREST.equals(this.getLoanProductRelatedDetail().getChargeOffBehaviour())` was added as a safeguard for future scenarios where someone might set `isChargedOff` to `true` when the `chargeOffBehaviour` is not `ZERO_INTEREST`.

By including this specific check, we avoid potential issues where the logic could break or produce unexpected results due to changes or incorrect assumptions about the relationship between `chargeOffBehaviour` and the `isChargedOff` flag. Replacing this with `!ctx.isChargedOff()` could simplify the code but would remove this safeguard, potentially introducing bugs if this edge case arises.

adamsaghy: When the loan got charged-off from that moment we dont wanna recalculate interest for any charge-off behaviours, so i would say we are safe to go without the 2nd condition ;)

mariiaKraievska: Done, PTAL

PR #4281, opened 2025-01-23
FINERACT-2169: New command processing infrastructure

vidakovic:
More details at...

Jira: https://issues.apache.org/jira/browse/FINERACT-2169

Wiki: https://cwiki.apache.org/confluence/display/FINERACT/FSIP-5%3A+New+command+processing+infrastructure

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -28,6 +28,7 @@ buildscript {
             [
                 'fineract-api',
                 'fineract-core',
+                'fineract-command',
                 'fineract-accounting',
                 'fineract-provider',
                 'fineract-branch',
@@ -103,7 +104,7 @@ plugins {
     id 'com.gorylenko.gradle-git-properties' version '2.4.2' apply false
     id 'org.asciidoctor.jvm.convert' version '3.3.2' apply false
     id 'org.asciidoctor.jvm.pdf' version '3.3.2' apply false
-    id 'com.google.cloud.tools.jib' version '3.4.4' apply false
+    id 'com.google.cloud.tools.jib' version '3.4.5' apply false
     id 'org.sonarqube' version '6.0.1.5171'
     id 'com.github.andygoossens.modernizer' version '1.10.0' apply false
     // TODO: upgrade to 6.0.4
@@ -112,6 +113,7 @@ plugins {
     id "com.github.davidmc24.gradle.plugin.avro-base" version "1.9.1" apply false
     id 'org.openapi.generator' version '7.8.0' apply false
     id 'com.gradleup.shadow' version '8.3.5' apply false
+    id 'me.champeau.jmh' version '0.7.1' apply false
 }
 
 apply from: "${rootDir}/buildSrc/src/main/groovy/org.apache.fineract.release.gradle"
--- a/buildSrc/src/main/groovy/org.apache.fineract.dependencies.gradle
+++ b/buildSrc/src/main/groovy/org.apache.fineract.dependencies.gradle
@@ -37,6 +37,7 @@ dependencyManagement {
         mavenBom 'org.mockito:mockito-bom:5.14.2'
         mavenBom 'software.amazon.awssdk:bom:2.29.9'
         mavenBom 'io.github.resilience4j:resilience4j-bom:2.2.0'
+        mavenBom 'org.testcontainers:testcontainers-bom:1.20.4'
     }
 
     dependencies {
@@ -235,6 +236,8 @@ dependencyManagement {
         //v42.7.5: performance issue: https://github.com/pgjdbc/pgjdbc/issues/3511#issuecomment-2637277977
         dependency 'org.postgresql:postgresql:42.7.4'
 
+        dependency 'com.mysql:mysql-connector-j:9.2.0'
+
         dependency 'org.assertj:assertj-core:3.26.3'
 
         dependency 'org.apache.commons:commons-math3:3.6.1'
@@ -249,5 +252,18 @@ dependencyManagement {
         dependency 'io.cucumber:cucumber-spring:7.20.1'
 
         dependency 'org.reflections:reflections:0.10.2'
+
+        dependency 'org.openjdk.jmh:jmh-core:1.37'
+        dependency 'org.openjdk.jmh:jmh-generator-annprocess:1.37'
+
+        dependency 'org.springframework.restdocs:spring-restdocs-asciidoctor:3.0.3'
+        dependency 'org.springframework.restdocs:spring-restdocs-mockmvc:3.0.3'
+        dependency 'org.springframework.restdocs:spring-restdocs-webtestclient:3.0.3'
+        dependency 'org.springframework.restdocs:spring-restdocs-restassured:3.0.3'
+
+        dependency 'com.lmax:disruptor:3.4.4'
+
+        dependency 'com.ibm.icu:icu4j:76.1'
+        dependency 'org.yakworks:spring-icu4j:0.4.2'
     }
 }
--- a/fineract-command/src/main/java/org/apache/fineract/command/core/CommandProperties.java
+++ b/fineract-command/src/main/java/org/apache/fineract/command/core/CommandProperties.java
@@ -1,0 +1,57 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.command.core;
+
+import com.lmax.disruptor.dsl.ProducerType;
+import java.io.Serial;
+import java.io.Serializable;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+import lombok.experimental.FieldNameConstants;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@Builder
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@FieldNameConstants
+@ConfigurationProperties(prefix = "fineract.command")
+public final class CommandProperties implements Serializable {
+
+    @Serial
+    private static final long serialVersionUID = 1L;
+
+    @Builder.Default
+    private Boolean enabled = true;
+
+    @Builder.Default
+    private CommandExecutorType executor = CommandExecutorType.sync;
+
+    @Builder.Default
+    private Integer ringBufferSize = 1024;
+
+    @Builder.Default
+    private ProducerType producerType = ProducerType.SINGLE;
+
+    public enum CommandExecutorType {
+        sync, async, disruptor
+    }
+}
--- a/fineract-command/src/main/java/org/apache/fineract/command/implementation/AsynchronousCommandExecutor.java
+++ b/fineract-command/src/main/java/org/apache/fineract/command/implementation/AsynchronousCommandExecutor.java
@@ -1,0 +1,58 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.command.implementation;
+
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Supplier;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.fineract.command.core.Command;
+import org.apache.fineract.command.core.CommandExecutor;
+import org.apache.fineract.command.core.CommandHandler;
+import org.apache.fineract.command.core.CommandMiddleware;
+import org.apache.fineract.command.core.CommandRouter;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.stereotype.Component;
+
+@Slf4j
+@RequiredArgsConstructor
+@Component
+@ConditionalOnProperty(value = "fineract.command.executor", havingValue = "async")
+public class AsynchronousCommandExecutor implements CommandExecutor {
+
+    private final List<CommandMiddleware> middlewares;
+
+    private final CommandRouter router;
+
+    @Override
+    public <REQ, RES> Supplier<RES> execute(Command<REQ> command) {
+        CompletableFuture<RES> future = CompletableFuture.supplyAsync(() -> {
+            for (CommandMiddleware middleware : middlewares) {
+                middleware.invoke(command);
+            }
+
+            CommandHandler<REQ, RES> handler = router.route(command);
+
+            return handler.handle(command);
+        });
+
+        return future::join;
+    }
+}
--- a/fineract-command/src/main/java/org/apache/fineract/command/persistence/converter/JsonAttributeConverter.java
+++ b/fineract-command/src/main/java/org/apache/fineract/command/persistence/converter/JsonAttributeConverter.java
@@ -1,0 +1,59 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.command.persistence.converter;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import jakarta.persistence.AttributeConverter;
+import jakarta.persistence.Converter;
+import lombok.RequiredArgsConstructor;
+import lombok.SneakyThrows;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+@RequiredArgsConstructor
+@Converter
+public class JsonAttributeConverter implements AttributeConverter<JsonNode, String> {
+    // TODO: it would be nicer to use a native JSON type on the database side, but not every system supports this;
+    // string/text are the lowest common denominator that should work on every database
+
+    private final ObjectMapper mapper;
+
+    @Override
+    @SneakyThrows
+    public String convertToDatabaseColumn(JsonNode source) {
+        if (source != null) {
+            return mapper.writeValueAsString(source);
+        }
+
+        // TODO: throw exception?
+        return null;
+    }
+
+    @Override
+    @SneakyThrows
+    public JsonNode convertToEntityAttribute(String source) {
+        if (source != null) {
+            return mapper.readTree(source);
+        }
+
+        // TODO: throw exception?
+        return null;
+    }
+}
--- a/fineract-command/src/main/java/org/apache/fineract/command/persistence/model/CommandEntity.java
+++ b/fineract-command/src/main/java/org/apache/fineract/command/persistence/model/CommandEntity.java
@@ -1,0 +1,83 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.command.persistence.model;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import jakarta.persistence.Column;
+import jakarta.persistence.Convert;
+import jakarta.persistence.Entity;
+import jakarta.persistence.GeneratedValue;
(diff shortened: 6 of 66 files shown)


fineract-command/src/test/resources/banner.txt
```diff
@@ -0,0 +1,35 @@
+⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣰⣿⣷⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
```
kjozsa: an easter egg? nice :grinning:

vidakovic: ... ah... Kristof... can't hide anything from your eyes :smile:

fineract-command/src/main/java/org/apache/fineract/command/core/CommandProperties.java
```diff
@@ -0,0 +1,53 @@
...
+    @Builder.Default
+    private Boolean enabled = true;
+
+    @Builder.Default
+    private CommandExecutorType executor = CommandExecutorType.sync;
+
+    @Builder.Default
+    private Integer ringBufferSize = 1024;
+
+    public enum CommandExecutorType {
```
VictorPavfurious, line 54: What about create all enum values on upper case? 
And also if this enum will be use on another classes, probably necessary move it to separate dir...?
I think it will the best way, because it's like standard practice

fineract-command/src/main/java/org/apache/fineract/command/implementation/AsynchronousCommandExecutor.java
```diff
@@ -0,0 +1,60 @@
...
+@ConditionalOnProperty(value = "fineract.command.executor", havingValue = "async")
+public class AsynchronousCommandExecutor implements CommandExecutor {
+
+    private final List<CommandMiddleware> middlewares;
+
+    private final CommandRouter router;
+
+    @Override
+    public <REQ, RES> Supplier<RES> execute(Command<REQ> command) {
+        CompletableFuture<RES> future = null;
```
VictorPavfurious: I think this is unnecessary😅, probably be better init on single line.. I think it will be better at least for code style

fineract-command/src/test/java/org/apache/fineract/command/sample/web/DummyApiController.java
```diff
@@ -0,0 +1,74 @@
...
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestHeader;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@Slf4j
+@RequiredArgsConstructor
+@RestController
+@RequestMapping(value = "/test/dummy", consumes = APPLICATION_JSON_VALUE, produces = { APPLICATION_JSON_VALUE,
```
VictorPavfurious: Are you need to consumes and produces values like APPLICATION_JSON_VALUE? I think in RestController has default (APPLICATION_JSON_VALUE) consumer and produces under hood, so it will be work fine without it😅

fineract-command/src/test/java/org/apache/fineract/command/sample/web/DummyExceptionHandler.java
```diff
@@ -0,0 +1,37 @@
...
+import org.springframework.web.bind.annotation.ControllerAdvice;
+import org.springframework.web.bind.annotation.ExceptionHandler;
+
+@Slf4j
+@ControllerAdvice
+public class DummyExceptionHandler {
+
+    @ExceptionHandler(Throwable.class)
+    public ResponseEntity<Map<String, String>> handleRuntimeException(Throwable ex) {
+        return ResponseEntity.status(INTERNAL_SERVER_ERROR).body(Map.of("error", ex.getMessage()));
```
VictorPavfurious: What about create universal "error class" like ApiErrorResponse and uses it for all body error response on our server. E.g this class will be contains fields as: (HttpStatus status, int status, String developerMessage, String errMessage, Map<String, String> properties, LocalDateTime timestamp)

fineract-command/src/main/java/org/apache/fineract/command/core/CommandProperties.java
vidakovic: ... you could, but these are only there for type ahead convenience with Spring Tools installed in your IDE. Just try to edit application-test.properties with IntelliJ Ultimate (if you have)... gives you a nice suggestion of possible values. Aesthetically, I think lower case makes more sense here... and then enums will never be used anywhere else (I think I could even make them package private and things would work).

fineract-command/src/main/java/org/apache/fineract/command/implementation/AsynchronousCommandExecutor.java
vidakovic: Sure... we save one line.

fineract-command/src/test/java/org/apache/fineract/command/sample/web/DummyApiController.java
vidakovic: Wrong. In Spring WebMVC if you do nothing then it will default to "*/*" which by definition accepts among other formats JSON, but also Octet Stream etc. etc. This matters even more when you turn on the screws and configure Spring to be even more pedantic about these things (which would be required when you go through any kind of certification process like SOC2). So, let's leave that.

fineract-command/src/test/java/org/apache/fineract/command/sample/web/DummyExceptionHandler.java
vidakovic: You see the "test" folder? This is just to give everyone an idea how the mechanics could be put together to avoid a ton of code (in this case our very manual error mapping). But for this PR this is kind of off topic and inconsequential to the rest of the proposal. And for all those interested:

- https://docs.spring.io/spring-framework/reference/web/webmvc/mvc-ann-rest-exceptions.html
- https://datatracker.ietf.org/doc/html/rfc7807

meonkeys:
v1.11 is coming soon... we don't want this in 1.11, right? Just trying to keep an eye on this since I'm working with James on the v1.11 release today.

vidakovic:
@meonkeys ... I'd say let's leave this one out of 1.11 ... it would have no impact, but still, let's keep this one out. Feel free to progress with the release.

meonkeys:
Perfect, thanks @vidakovic . Should be easy as long as it isn't merged into `develop` before we branch for v1.11 in about a couple weeks.

Cc: @jdailey

IOhacker:
LGTM

fineract-command/src/main/resources/db/custom-changelog/0001_command_init.xml
```diff
@@ -0,0 +1,49 @@
...
+            <column name="username" type="varchar(255)" />
+            <column name="payload" type="jsonb" />
+        </createTable>
+    </changeSet>
+    <changeSet author="fineract" id="command-00002" context="mysql">
+        <createTable tableName="command">
+            <column autoIncrement="true" name="id" type="bigint" remarks="Internal ID">
+                <constraints nullable="false" primaryKey="true"/>
+            </column>
+            <column name="created_at" type="timestamp with time zone" />
```
adamsaghy: Should we use DATETIME instead?

vidakovic: Right... will fix and adapt to the rest of the code... old habit.

adamsaghy: Unfortunately mysql timezone handling is pretty weird.. i just learnt not long ago DATETIME is the go there....at least with mysql / mariadb

build.gradle
```diff
@@ -112,6 +113,7 @@ plugins {
     id "com.github.davidmc24.gradle.plugin.avro-base" version "1.9.1" apply false
     id 'org.openapi.generator' version '7.8.0' apply false
     id 'com.gradleup.shadow' version '8.3.5' apply false
+    id 'me.champeau.jmh' version '0.7.1' apply false
```
adamsaghy, line 116: Do we need JMH?

buildSrc/src/main/groovy/org.apache.fineract.dependencies.gradle
```diff
@@ -235,6 +236,8 @@ dependencyManagement {
         //v42.7.5: performance issue: https://github.com/pgjdbc/pgjdbc/issues/3511#issuecomment-2637277977
         dependency 'org.postgresql:postgresql:42.7.4'
 
+        dependency 'mysql:mysql-connector-java:8.0.33'
```
adamsaghy: Isnt this problematic from licensing point of view?

buildSrc/src/main/groovy/org.apache.fineract.dependencies.gradle
```diff
@@ -247,5 +250,19 @@ dependencyManagement {
...
+        dependency 'org.openjdk.jmh:jmh-generator-annprocess:1.37'
+
+        dependency 'org.springframework.restdocs:spring-restdocs-asciidoctor:3.0.3'
+        dependency 'org.springframework.restdocs:spring-restdocs-mockmvc:3.0.3'
+        dependency 'org.springframework.restdocs:spring-restdocs-webtestclient:3.0.3'
+        dependency 'org.springframework.restdocs:spring-restdocs-restassured:3.0.3'
+
+        dependency 'com.lmax:disruptor:3.4.4'
+
+        dependency 'com.alibaba:transmittable-thread-local:2.14.5'
```
adamsaghy: Can you help me what are these dependencies for?

fineract-command/src/main/java/org/apache/fineract/command/persistence/converter/JsonAttributeConverter.java
```diff
@@ -0,0 +1,59 @@
...
+import jakarta.persistence.AttributeConverter;
+import jakarta.persistence.Converter;
+import lombok.RequiredArgsConstructor;
+import lombok.SneakyThrows;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+@RequiredArgsConstructor
+@Converter
+public class JsonAttributeConverter implements AttributeConverter<JsonNode, String> {
```
adamsaghy, line 32: Should we rather keep the value as String? We need to convert anyway to a 3rd object.
Right now:
DB: String -> JPA: convert String to JsonNode -> Backend: convert JsonNode to final data type (like: LoanRequest, LoanTransactionRequest, etc...)

Should we rather:
DB: Stirng -> JPA: String (no conversion) -> Backend: convert String to final data type (like: LoanRequest, LoanTransactionRequest, etc.) 

What do you think?

fineract-command/src/main/java/org/apache/fineract/command/persistence/model/CommandEntity.java
```diff
@@ -0,0 +1,83 @@
...
+import lombok.experimental.FieldNameConstants;
+import org.apache.fineract.command.persistence.converter.JsonAttributeConverter;
+
+@Getter
+@Setter
+@ToString
+@FieldNameConstants
+@Entity
+@Table(name = "m_command")
+public class CommandEntity implements Serializable {
```
adamsaghy, line 48: Extend `AbstractPersistableCustom` instead?

fineract-command/src/main/java/org/apache/fineract/command/persistence/model/CommandEntity.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @Transient
+    @Setter(value = AccessLevel.NONE)
+    private boolean isNew = true;
+
+    @Column(name = "command_id")
+    private UUID commandId;
+
+    @Column(name = "created_at")
+    private Instant createdAt;
```
adamsaghy: Use ` OffsetDateTime` instead?

fineract-command/src/main/resources/db/custom-changelog/0001_command_init.xml
```diff
@@ -0,0 +1,49 @@
...
+-->
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.1.xsd">
+    <changeSet author="fineract" id="command-00001" context="postgresql">
+        <createTable tableName="m_command">
+            <column autoIncrement="true" name="id" type="bigint" remarks="Internal ID">
+                <constraints nullable="false" primaryKey="true"/>
+            </column>
+            <column name="created_at" type="datetime" />
```
adamsaghy: Use `TIMESTAMP WITH TIME ZONE` instead?

fineract-command/src/test/resources/banner.txt
```diff
@@ -0,0 +1,35 @@
...
+⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣴⣿⣿⣿⠀⠀⢀⣠⣤⣶⣶⣶⣶⣶⣶⣶⣦⣤⣀⠀⠀⠀⠀⠀⢀⣀⣀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
+⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣠⣾⣿⣿⣿⣿⡔⣾⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⡿⢟⣫⣯⣷⣾⣿⣿⣿⣿⣿⣿⣷⣄⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
+⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣠⣾⣿⣿⣿⣿⣿⣿⣷⡙⢿⣿⣿⣿⣿⣿⣿⡿⢋⣵⣾⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣧⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
```
adamsaghy: I am not sure whether we should have this... Would you consider removing?

fineract-command/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,45 @@
...
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied. See the License for the
+    specific language governing permissions and limitations
+    under the License.
+
+-->
+<configuration>
+    <!--
```
adamsaghy: Remove if not used...

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/domain/BatchRequestContextHolder.java
```diff
@@ -18,25 +18,20 @@
...
 
 public final class BatchRequestContextHolder {
 
     private BatchRequestContextHolder() {}
 
-    private static final ThreadLocal<Map<String, Object>> batchAttributes = new NamedThreadLocal<>("batchAttributes");
+    private static final ThreadLocal<Map<String, Object>> batchAttributes = new TransmittableThreadLocal<>();
 
-    private static final ThreadLocal<Optional<TransactionStatus>> batchTransaction = new NamedThreadLocal<>("batchTransaction") {
+    private static final ThreadLocal<Optional<TransactionStatus>> batchTransaction = TransmittableThreadLocal.withInitial(Optional::empty);
```
adamsaghy: Do we need thread local value copies to child threads?

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/ThreadLocalContextUtil.java
```diff
@@ -33,11 +34,11 @@
 public final class ThreadLocalContextUtil {
 
     public static final String CONTEXT_TENANTS = "tenants";
-    private static final ThreadLocal<String> contextHolder = new ThreadLocal<>();
-    private static final ThreadLocal<FineractPlatformTenant> tenantContext = new ThreadLocal<>();
-    private static final ThreadLocal<String> authTokenContext = new ThreadLocal<>();
-    private static final ThreadLocal<HashMap<BusinessDateType, LocalDate>> businessDateContext = new ThreadLocal<>();
-    private static final ThreadLocal<ActionContext> actionContext = new ThreadLocal<>();
+    private static final ThreadLocal<String> contextHolder = new TransmittableThreadLocal<>();
```
adamsaghy: Same as above

build.gradle
vidakovic: I have added a (still simple) benchmark test to get an idea what kind of performance we could expect switching between the different "performance levels". So, yes.

buildSrc/src/main/groovy/org.apache.fineract.dependencies.gradle
vidakovic: You are right. In the meanwhile solved the persistence feature differently, not needed anymore. Good catch, thanks!

buildSrc/src/main/groovy/org.apache.fineract.dependencies.gradle
vidakovic: Alright, you know we have a ton of thread local variables (tenant identification, Spring Security and what not). When we switch the command processing to asynchronous mode then we have to make sure that all those thread local variables are still available in a potentially different thread from that one that initiated the request; actually it is certain that the business logic executing thread will be different from the one that receives the incoming API request. So, to make things down the pipe still work we have to make sure that all those variables are still accessible (if we don't do anything they will be completely empty). That library helps with that issue.

Disclosure: in this first introduction of a new command processing infrastructure the default processing mode (synchronous) is the only one we are focusing on, "async" and "disruptor" (aka non-blocking) need (very likely) more testing (and probably some additional tweaks). Which means technically we would not need to take care of the thread local variables right now, it kind of was a low hanging fruit with this library.

fineract-command/src/main/java/org/apache/fineract/command/persistence/converter/JsonAttributeConverter.java
vidakovic: Well, I did this exercise already in a different setting and the JsonNode is actually already quite convenient when the rest of your infrastructure depends on Jackson anyway (which is the ultimate goal here). The only reason why these two data structures are persistent is for a specific use case, deferred execution (aka "maker-checker"), otherwise I would have left it out anyway. Overall I am not expecting that we will poke and peek too much into/from this data structure anyway, so the type actually doesn't matter too much. By default we are NOT saving every request, only those that are deferred (which should not be many)... and as they are deferred speed is counted in hours/days not milliseconds. And given that one of the main goals of this exercise is to introduce type-safety makes string kind of a wrong choice from my perspective.

Concerning saving JSON data as a string: that is anyway a compromise. Both databases (MySQL and PostgreSQL) have specific data types for JSON (which is of course non-standard); unfortunately they have different approaches (I think PostgreSQL's is the better one) and it was impossible to abstract this away properly with JPA on the database side. That's why I fell back on the least common denominator (string). If we had only PostgreSQL I would have chosen their native JSON type (which would allow for proper querying... which we don't need). Just FYI.

In short: I think this should be fine as is.

fineract-command/src/main/java/org/apache/fineract/command/persistence/model/CommandEntity.java
vidakovic: OffsetDateTime is not compatible with Spring Data (see e.g. when you enable auditing in the Spring documentation). This is the type they support (from the `java.time.*` API).

fineract-command/src/test/resources/banner.txt
vidakovic: Of course.

buildSrc/src/main/groovy/org.apache.fineract.dependencies.gradle
vidakovic: Well, I checked again: we have that dependency already in multiple places. And we should actually replace it with the latest one which was re-located in Maven Central by Oracle:

```
https://mvnrepository.com/artifact/com.mysql/mysql-connector-j
```

That tells me that licensing is not an issue as we are doing this already for a long time (and Spring Data does it too BTW).

fineract-command/src/main/java/org/apache/fineract/command/persistence/model/CommandEntity.java
vidakovic: I am aware of that, but sometimes code duplication is your friend. By not requiring that hierarchy I can ensure that `fineract-command` is 100% independent and does not create any cross dependencies (of which we already have enough I think). The behavior is exactly the same as for all other JPA entities, so no surprises to be expected here. Just to say: this is intended and I think it makes sense, let's leave it.

fineract-command/src/main/resources/db/custom-changelog/0001_command_init.xml
vidakovic: I followed other examples in our migration scripts. Sure, can change... but what's the reason for the others to be `datetime` then?

fineract-command/src/test/resources/logback-test.xml
vidakovic: Alright

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/domain/BatchRequestContextHolder.java
vidakovic: Later for the "async" mode yes. See my previous comments for the dependency. I think I'll remove for now to keep this as small as possible and avoid distractions. But later we have to find a solution here if we want to use the full performance potential.

buildSrc/src/main/groovy/org.apache.fineract.dependencies.gradle
vidakovic: Removed

fineract-command/src/main/resources/db/custom-changelog/0001_command_init.xml
vidakovic: image

Quick search in IntelliJ reveals quite some places where exactly the same is used.

buildSrc/src/main/groovy/org.apache.fineract.dependencies.gradle
```diff
@@ -235,6 +236,8 @@ dependencyManagement {
         //v42.7.5: performance issue: https://github.com/pgjdbc/pgjdbc/issues/3511#issuecomment-2637277977
         dependency 'org.postgresql:postgresql:42.7.4'
 
+        dependency 'com.mysql:mysql-connector-j:9.2.0'
```
adamsaghy, line 239: What it the difference between this and `mysql:mysql-connector-java:8.0.33`?
However i am still unsure whether we can use mysql-connector as dependency... i acknowledge seems we already doing so having it in `fineract-provider/build.gradle` as driver...

adamsaghy (approved):
LGTM

jdailey:
Really well explained on wiki .  +1

a7med3del1973:
LGTM
+1

fineract-command/src/main/java/org/apache/fineract/command/persistence/converter/JsonAttributeConverter.java
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        // TODO: throw exception?
+        return null;
+    }
+
+    @Override
+    @SneakyThrows
+    public JsonNode convertToEntityAttribute(String source) {
+        if (source != null) {
```
VictorPavfurious, line 52: Same as below

fineract-command/src/main/java/org/apache/fineract/command/persistence/converter/JsonAttributeConverter.java
```diff
@@ -0,0 +1,59 @@
...
+public class JsonAttributeConverter implements AttributeConverter<JsonNode, String> {
+    // TODO: it would be nicer to use a native JSON type on the database side, but not every system supports this;
+    // string/text are the lowest common denominator that should work on every database
+
+    private final ObjectMapper mapper;
+
+    @Override
+    @SneakyThrows
+    public String convertToDatabaseColumn(JsonNode source) {
+        if (source != null) {
```
VictorPavfurious, line 41: What do you think about use Optional here and as return type Optional<String>, to avoid return null?

vidakovic:
... good idea. Please note somewhere and let's add it in another PR.

vidakovic:
If no one objects then I'll pull the trigger (merge) by the end of the day. Some devs are waiting for this to address more technical debt (aka type safety). Thanks for the comments so far!

VictorPavfurious:
Yeah, I'm one of the developers waiting for this PR in develop:)
LGTM +1

PR #4298, opened 2025-02-03
Fix/triggered sms campaign

Pinchez25:
# Fix SMS message ID issue in triggered message flow

## Problem
When processing triggered messages, `SmsMessageApiQueueResourceData` objects are being created before `SmsMessage` objects are saved to the database. This results in:

- `smsMessage.getId()` returning null
- Invalid or missing IDs in the queue data
- Inconsistency between direct and triggered message flows

## Solution
Modified the triggered message flow in `SmsMessageScheduledJobServiceImpl.java` to:

1. Save SMS messages to the database first
2. Create queue data objects using the database-generated IDs
3. Send messages to the gateway with proper internal IDs

## Implementation Details
- Reordered operations to ensure database persistence occurs before queue data creation
- Maintained all existing functionality while fixing the ID reference issue
- Aligned the triggered message flow with the direct message flow pattern


IOhacker:
hi, please squash an commit, also please check that checks are working  check the logs: Run './gradlew :fineract-provider:spotlessApply' to fix these violations

PR #4382, opened 2025-02-25
FINERACT-2081: Fix code quality issues in Cashier domain class

7ossam7atem1:
## Description

- Removed unused commented-out code blocks
- Introduced constant IS_FULL_DAY_PARAM_NAME to avoid string literal duplication
- Fixed variable shadowing by renaming local variable to avoid field hiding
- Changed to primitive boolean for better performance
- Fixed incorrect variable assignment in time handling
- Simplified time methods with immediate returns

## Checklist

### DIFF
--- a/fineract-branch/src/main/java/org/apache/fineract/organisation/teller/domain/Cashier.java
+++ b/fineract-branch/src/main/java/org/apache/fineract/organisation/teller/domain/Cashier.java
@@ -29,7 +29,6 @@ import jakarta.persistence.Transient;
 import jakarta.persistence.UniqueConstraint;
 import java.time.LocalDate;
 import java.util.LinkedHashMap;
-import java.util.List;
 import java.util.Map;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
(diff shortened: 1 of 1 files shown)


fineract-branch/src/main/java/org/apache/fineract/organisation/teller/domain/Cashier.java
```diff
@@ -140,19 +132,15 @@ public Map<String, Object> update(final JsonCommand command) {
             this.endDate = command.localDateValueOfParameterNamed(endDateParamName);
         }
 
-        final Boolean isFullDay = command.booleanObjectValueOfParameterNamed("isFullDay");
+        final boolean fullDayFlag = command.booleanPrimitiveValueOfParameterNamed(IS_FULL_DAY_PARAM_NAME);
```
adamsaghy: Changing from Boolean to boolean is changing the behaviour as from 3 options(true, false, null), now it became just 2 (true, false).

`isFullDay` is a nullable field in a the `m_cashiers` table, so this change is breaking change, no?

7ossam7atem1: > Changing from Boolean to boolean is changing the behaviour as from 3 options(true, false, null), now it became just 2 (true, false).
> 
> `isFullDay` is a nullable field in a the `m_cashiers` table, so this change is breaking change, no?

Ah, you're right 
The nullable aspect of the `isFullDay` field in `m_cashiers` is pretty important here. Moving from Boolean to boolean would definitely break things since we'd lose the null state option by forcing it to only 2 states (true, false).

I'll roll back that part of the change and stick to the safer improvements, Thanks for the thorough review!

PR #4475, opened 2025-03-18
FINERACT-2080: use enums instead of integers in Loan, LoanProduct and LoanTransaction

Cocoa-Puffs:
## Description

## Checklist


fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
```diff
@@ -227,12 +230,14 @@ public class Loan extends AbstractAuditableWithUTCDateTimeCustom<Long> {
     private Integer termFrequency;
 
     @Setter()
+    @Enumerated(EnumType.ORDINAL)
     @Column(name = "term_period_frequency_enum", nullable = false)
-    private Integer termPeriodFrequencyType;
+    private PeriodFrequencyType termPeriodFrequencyType;
```
oleksii-novikov-onix, line 234: 1. I see converters for LoanStatus, LoanSubStatus, and LoanTransactionType, there doesn't appear to be a converter implementation for PeriodFrequencyType. Do we need to implement a converter for PeriodFrequencyType?
2. Honestly, I don’t understand why we need a converters for LoanStatus, LoanSubStatus etc. I thought that @Enumerated would do the same job as the declared converters.

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
```diff
@@ -227,12 +230,14 @@ public class Loan extends AbstractAuditableWithUTCDateTimeCustom<Long> {
     private Integer termFrequency;
 
     @Setter()
+    @Enumerated(EnumType.ORDINAL)
```
oleksii-novikov-onix: Specifying EnumType.ORDINAL is not necessary - it's the default behavior

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanStatusConverter.java
```diff
@@ -0,0 +1,36 @@
...
+public class LoanStatusConverter implements AttributeConverter<LoanStatus, Integer> {
+
+    @Override
+    public Integer convertToDatabaseColumn(LoanStatus attribute) {
+        return attribute == null ? null : attribute.getValue();
+    }
+
+    @Override
+    public LoanStatus convertToEntityAttribute(Integer dbData) {
+        return dbData == null ? LoanStatus.INVALID : LoanStatus.fromInt(dbData);
```
oleksii-novikov-onix: Are we sure that null should be represented as INVALID?

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
Cocoa-Puffs: I written converters for enums that have integer values that differ from their ordinal values. These values were what we saved to the database previously and I didn't want to introduce new behaviour or change the data being saved. PeriodFrequencyTypes integer values correspond to the ordinal values of enums so it didn't need a converter. (LoanTransactionType is almost good, it's just missing number 11).

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
Cocoa-Puffs: I looked around the codebase a bit to see what the conventions were regarding annotations. I have noticed that EnumType.ORDINAL is written out at plenty of other places so I have written my code similarly. I can remove it though as it is not necessary as you have pointed out.

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanStatusConverter.java
Cocoa-Puffs: loan_status_id is currently non nullable on the database side so I don't think we can ever get a null there. Moreover this was the previous behaviour. If you look at getStatus in Loan.java, it just returns LoanStatus.fromInt, which returns INVALID on null values. I can change the converter to simply call the fromInt as that would be essentially the same

fineract-loan/src/main/java/org/apache/fineract/cob/data/LoanDataForExternalTransfer.java
```diff
@@ -31,4 +32,13 @@ public class LoanDataForExternalTransfer {
     Integer loanStatus;
```
adamsaghy: I think we can have the enum here. It is only used internally.

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanRepository.java
```diff
@@ -68,43 +70,43 @@ public interface LoanRepository extends JpaRepository<Loan, Long>, JpaSpecificat
     String FIND_ACTIVE_LOANS_PRODUCT_IDS_BY_GROUP = "Select loan.loanProduct.id from Loan loan where "
             + "loan.group.id = :groupId and loan.loanStatus = :loanStatus and loan.client.id is NULL group by loan.loanProduct.id";
 
-    String DOES_CLIENT_HAVE_NON_CLOSED_LOANS = "select case when (count (loan) > 0) then 'true' else 'false' end from Loan loan where loan.client.id = :clientId and loan.loanStatus in (100,200,300,303,304,700)";
```
adamsaghy, line 71: If you pass the statuses as parameters, you might wanna rename the methods to reflect the "new" behaviour. Example: no need the "closed / non-closed" as manually any status can be provided.

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
```diff
@@ -227,12 +230,14 @@ public class Loan extends AbstractAuditableWithUTCDateTimeCustom<Long> {
...
     @Column(name = "term_period_frequency_enum", nullable = false)
-    private Integer termPeriodFrequencyType;
+    private PeriodFrequencyType termPeriodFrequencyType;
 
     @Setter(AccessLevel.PACKAGE)
+    @Enumerated(EnumType.ORDINAL)
```
adamsaghy: Why to use @Enumerate(ORDINAL) if converter is to be used here? Loan status enum ordinal does not follow the original behaviour...

fineract-loan/src/main/java/org/apache/fineract/cob/data/LoanDataForExternalTransfer.java
Cocoa-Puffs: I wasn't sure that there were points where we actually transfer this object or parts of it to external systems so I decided to err on the safer side. I will change the property.

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
Cocoa-Puffs: I can remove it from enums where I ended up using a converter, I agree that they can be confusing. The converters take precedence though, so in these cases it's still the integer values that are saved, not the ordinal values.

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
oleksii-novikov-onix: Thank you Mark, I believe we can remove the @Enumerated annotation when we have a converter, as having both creates a conflict.

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
oleksii-novikov-onix: I would remove ORDINAL, but consistent code is also important. Check all similar places and decide for yourself.

PR #4533, opened 2025-04-01
FINERACT-2299: New command processing - Standing instructions history (org.apache.fineract.portfolio.account)

VictorPavfurious:
## Description

1) Used DTO instead String response and request
2) Add json ignore for avoid serializetion getter method and enhance dto;

## Checklist

### DIFF
--- a/fineract-core/src/main/java/org/apache/fineract/portfolio/account/data/PortfolioAccountData.java
+++ b/fineract-core/src/main/java/org/apache/fineract/portfolio/account/data/PortfolioAccountData.java
@@ -20,6 +20,7 @@ package org.apache.fineract.portfolio.account.data;
 
 import java.io.Serializable;
 import java.math.BigDecimal;
+import java.util.Optional;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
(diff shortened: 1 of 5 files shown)


fineract-core/src/main/java/org/apache/fineract/portfolio/account/data/PortfolioAccountData.java
```diff
@@ -69,6 +70,7 @@ public PortfolioAccountData(final Long id, final String accountNo, final String
         this.amtForTransfer = null;
     }
 
+    @JsonIgnore
     public String getCurrencyCode() {
```
adamsaghy: I think we can refactor and remove this method... it is used only a couple places, but can use `getCurrency().getCode()` instead.

VictorPavfurious: Yes, i agree with you. Or, we can rename this method  in order to jackson not serialize this field like "currencyCode" and use inside like Optional.ofNullable(getCurrency()).map(currency -> currency.getCode()).orElse(null), this allow for us avoid addition checks on currency != null and NPE

VictorPavfurious: Done

PR #4546, opened 2025-04-05
FINERACT-2237: Fix GSIM account creation to generate single parent for multiple clients

JebronLames32:
## Description

This fix corrects the logic for GSIM account creation to ensure that only a single GSIM parent account is created and the remaining accounts are listed as child accounts under the GSIM account when multiple clients are provided in the request payload (with only one client having `"isParentAccount": 1` set). The previous logic caused multiple GSIM parent accounts to be erroneously created, each associated with only a single client as the condition for checking if a parent account is to be created was wrong.

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsApplicationProcessWritePlatformServiceJpaRepositoryImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsApplicationProcessWritePlatformServiceJpaRepositoryImpl.java
@@ -168,7 +168,9 @@ public class SavingsApplicationProcessWritePlatformServiceJpaRepositoryImpl impl
 
                     Group group = this.groupRepositoryWrapper.findOneWithNotFoundDetection(groupId);
 
-                    if (command.booleanObjectValueOfParameterNamed("isParentAccount") != null) {
+                    if (command.booleanObjectValueOfParameterNamed("isParentAccount") != null
+                            && ("1".equals(command.stringValueOfParameterNamed("isParentAccount"))
+                                    || command.booleanObjectValueOfParameterNamed("isParentAccount"))) {
                         // empty table check
                         if (gsimRepository.count() != 0) {
                             // Parent-Not an empty table
@@ -193,7 +195,12 @@ public class SavingsApplicationProcessWritePlatformServiceJpaRepositoryImpl impl
                     } else {
                         if (gsimRepository.count() != 0) {
                             // Child-Not an empty table check
-                            gsimAccount = gsimRepository.findOneByIsAcceptingChildAndApplicationId(true, applicationId);
+                            if (applicationId.compareTo(BigDecimal.ZERO) == 0) {
+                                gsimAccount = gsimRepository.findOneByIsAcceptingChildAndApplicationIdAndGroupId(true, applicationId,
+                                        groupId);
+                            } else {
+                                gsimAccount = gsimRepository.findOneByIsAcceptingChildAndApplicationId(true, applicationId);
+                            }
                             accountNumber = gsimAccount.getAccountNumber() + (gsimAccount.getChildAccountsCount() + 1);
                             account.updateAccountNo(accountNumber);
                             this.gsimWritePlatformService.incrementChildAccountCount(gsimAccount);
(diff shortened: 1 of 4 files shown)


JebronLames32:
The current unit tests for GSIM account at `GroupSavingsIntegrationTest.java` are marked deprecated and also do not cover the scenario where **multiple clients need to be part of a GSIM account.** 
Can someone please assist me in creating Unit tests compliant with [FINERACT-2165](https://issues.apache.org/jira/browse/FINERACT-2165)

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsApplicationProcessWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -168,7 +168,7 @@ public CommandProcessingResult submitApplication(final JsonCommand command) {
 
                     Group group = this.groupRepositoryWrapper.findOneWithNotFoundDetection(groupId);
 
-                    if (command.booleanObjectValueOfParameterNamed("isParentAccount") != null) {
+                    if (command.booleanObjectValueOfParameterNamed("isParentAccount")) {
```
adamsaghy: How does this fix anything?

JebronLames32: It is a boolean value that is being returned by `command.booleanObjectValueOfParameterNamed("isParentAccount")`. It will always evaluate to `true` when compared to not be null.

JebronLames32: I do realise that I should make a better change here. Making the changes to also consider the case where `isParentAccount` is evaluating to null

JebronLames32:
- The API works when `isParentAccount` is set to true for one client and **omitted** (or set to **`null`**) for the others in the `clientArray`
- The API fails (makes it a parent account) when `isParentAccount` is explicitly set to **`false`** for the clients. Instead of accepting `false`, the API seems to expect the parameter to be omitted.
- I’m fixing this by ensuring the API accepts false as a valid value for isParentAccount. This avoids relying on omission (null) of this vlaue for accounts that are not the parent account for the GSIM

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsApplicationProcessWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -168,7 +168,9 @@ public CommandProcessingResult submitApplication(final JsonCommand command) {
 
                     Group group = this.groupRepositoryWrapper.findOneWithNotFoundDetection(groupId);
 
-                    if (command.booleanObjectValueOfParameterNamed("isParentAccount") != null) {
+                    if (command.booleanObjectValueOfParameterNamed("isParentAccount") != null
```
adamsaghy, line 171: Is this boolean or string now? I am confused...

JebronLames32: Screenshot From 2025-04-14 19-42-35

Originally documented as “1”
In the legacy documentation, isParentAccount is being indicated using 1 to represent true (and possibly 0 or omission for false). However, the current Fineract implementation (i.e., `command.booleanObjectValueOfParameterNamed("isParentAccount"))` uses a Boolean for logic.

I added (`"1".equals(command.stringValueOfParameterNamed("isParentAccount"))`) to handle the documentation scenario where consumers send the value as "1" instead of "true".
Meanwhile, `command.booleanObjectValueOfParameterNamed("isParentAccount")` properly interprets JSON booleans (true/false).

I felt we could accommodate both forms of input:

- Boolean: true/false (which booleanObjectValueOfParameterNamed understands), and
- Legacy string: "1" for true.

The intended usage going forward could be to be a boolean field in the API.
The code allows for backward compatibility by also accepting the string "1" as an equivalent to true.

PR #4552, opened 2025-04-08
FINERACT-2080: Extract loanLifecycleStateMachine from Loan entity

oleksii-novikov-onix:
## Description

1. Moved loan state transition logic from Loan class to LoanLifecycleStateMachine.
2. Moved addLoanCharge method to LoanChargeService
3. Removed helper methods from Loan class that were related to state transitions

## Checklist

### DIFF
--- a/fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/DefaultLoanLifecycleStateMachine.java
+++ b/fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/DefaultLoanLifecycleStateMachine.java
@@ -19,20 +19,20 @@
 package org.apache.fineract.portfolio.loanaccount.domain;
 
 import java.math.BigDecimal;
+import java.time.LocalDate;
 import java.util.List;
 import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.fineract.infrastructure.core.service.MathUtil;
 import org.apache.fineract.infrastructure.event.business.domain.loan.LoanStatusChangedBusinessEvent;
 import org.apache.fineract.infrastructure.event.business.service.BusinessEventNotifierService;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
 
 // TODO: introduce tests for the state machine
+@Slf4j
 @Component
 @RequiredArgsConstructor
 public class DefaultLoanLifecycleStateMachine implements LoanLifecycleStateMachine {
-
-    private static Logger LOG = LoggerFactory.getLogger(DefaultLoanLifecycleStateMachine.class);
 
     private static final List<LoanStatus> ALLOWED_LOAN_STATUSES = List.of(LoanStatus.values());
     private final BusinessEventNotifierService businessEventNotifierService;
@@ -46,7 +46,29 @@ public class DefaultLoanLifecycleStateMachine implements LoanLifecycleStateMachi
     @Override
     public void transition(final LoanEvent loanEvent, final Loan loan) {
         loan.updateLoanSummaryDerivedFields();
+        internalTransition(loanEvent, loan);
+    }
 
+    @Override
+    public void determineAndTransition(final Loan loan, final LocalDate transactionDate) {
+        if (loan.getStatus() == null) {
+            return;
+        }
+
+        loan.updateLoanSummaryDerivedFields();
+
+        final LoanStatusTransition transition = determineTransition(loan, loan.getStatus(), transactionDate);
+
+        if (MathUtil.isEmpty(loan.getTotalOverpaid())) {
+            loan.setOverpaidOnDate(null);
+        }
+
+        if (transition.transitionNeeded()) {
+            internalTransition(transition.event(), loan);
+        }
+    }
+
+    private void internalTransition(final LoanEvent loanEvent, final Loan loan) {
         LoanStatus oldStatus = loan.getStatus();
         LoanStatus newStatus = getNextStatus(loanEvent, loan);
         if (newStatus != null) {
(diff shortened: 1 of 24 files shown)


fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -3097,7 +3091,7 @@ public void handleChargebackTransaction(final Loan loan, LoanTransaction chargeb
                             new MoneyHolder(loan.getTotalOverpaidAsMoney()), null));
         }
         loan.updateLoanSummaryDerivedFields();
-        if (!loan.doPostLoanTransactionChecks(chargebackTransaction.getTransactionDate(), loanLifecycleStateMachine)) {
+        if (!loanLifecycleStateMachine.doPostLoanTransactionChecks(loan, chargebackTransaction.getTransactionDate())) {
```
adamsaghy: This is the only logic which do anything with the method return value, but do be honest i have no clue the underlying logic here...

Would you mind to review it and straighten a little bit? 
`loanLifecycleStateMachine.doPostLoanTransactionChecks` is moving the status to either overpaid or closed fully paid or initiate a loan status change of "LOAN REPAYMENT OR WAIVER" event, which is weird here and if all of them says no status change, then we call status change for "LOAN CHARGEBACK" event.... i find it confusing...

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanChargeWritePlatformServiceImpl.java
```diff
@@ -305,7 +306,7 @@ public CommandProcessingResult addLoanCharge(final Long loanId, final JsonComman
             } else {
                 reprocessLoanTransactionsService.reprocessTransactions(loan);
             }
-            loan.doPostLoanTransactionChecks(transactionDate, loan.getLoanLifecycleStateMachine());
+            defaultLoanLifecycleStateMachine.doPostLoanTransactionChecks(loan, transactionDate);
```
adamsaghy: Can we rewrite these actions to reflect the correct EVENT?
Adding a loan charge should not trigger loan status with "LOAN REPAYMENT OR WAIVER" event... which is the default fallback behaviour....

Either `defaultLoanLifecycleStateMachine` should be clever to figure out what to do based on the actual loan state or we can try to give it some context as what event occurred, and event + loan state to control loan status changes behaviour.

What do you think?

adamsaghy (changes requested):
Kindly review my comments!

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanChargeWritePlatformServiceImpl.java
oleksii-novikov-onix: Yes, Adam, as an improvement, I added context-specific methods to LoanLifecycleStateMachine. Please recheck my additional commit. I haven’t squashed yet to make it easier to track my changes.

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/DefaultLoanLifecycleStateMachine.java
```diff
@@ -276,4 +274,111 @@ private boolean anyOfAllowedWhenComingFrom(final LoanStatus state, final LoanSta
 
         return allowed;
     }
+
+    @Override
+    public void handleLoanRepayment(final Loan loan, final LocalDate transactionDate) {
+        final boolean statusChanged = evaluateAndTransitionLoanStatus(loan, transactionDate);
```
adamsaghy: I wonder whether there is a way to further simplify the things:
this `evaluateAndTransitionLoanStatus` is doing some calculation and it might change status and here we are checking again whether was there any status change and if not we try to change status once again.

Can we rework this and make sure:
- Calculate and change loan status just once -> no further conditions and loan status changes in multiple places...

I would like something like this:

* If we had a repayment and loan status was ACTIVE,
  * if current outstanding balance is greater than 0 -> No status change
  * if current outstanding balance is zero and overpaid amount is 0 -> CLOSED_OBLIGATIONS_MET
  * if current outstanding balance is zero and overpaid amount is greater than 0 -> OVERPAID
  
  
This logic just an example, but i really want this to be only one place where we evaluate and do the necessary loan status change. 
We should avoid evaluate and change statuses (having the logic) multiple places!

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/DefaultLoanLifecycleStateMachine.java
```diff
@@ -276,4 +274,111 @@ private boolean anyOfAllowedWhenComingFrom(final LoanStatus state, final LoanSta
...
+        if (!statusChanged) {
```
adamsaghy: I wonder whether there is a way to further simplify the things:
this `evaluateAndTransitionLoanStatus` is doing some calculation and it might change status and here we are checking again whether was there any status change and if not we try to change status once again.

Can we rework this and make sure:
- Calculate and change loan status just once -> no further conditions and loan status changes in multiple places...

I would like something like this:

* If we had a repayment and loan status was ACTIVE,
  * if current outstanding balance is greater than 0 -> No status change
  * if current outstanding balance is zero and overpaid amount is 0 -> CLOSED_OBLIGATIONS_MET
  * if current outstanding balance is zero and overpaid amount is greater than 0 -> OVERPAID
  
  
This logic just an example, but i really want this to be only one place where we evaluate and do the necessary loan status change. 
We should avoid evaluate and change statuses (having the logic) multiple places!

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanWritePlatformServiceJpaRepositoryImpl.java
oleksii-novikov-onix: Replaced the boolean-returning doPostLoanTransactionChecks with a more explicit method

adamsaghy (approved):
LGTM

PR #4608, opened 2025-04-21
FINERACT-2202: Bulk import fix

RamsesCamacho1171:
## Description

FINERACT-2202: Bulk import fix

## Checklist

- [x ] Write the commit message as per https://github.com/apache/fineract/#pull-requests

- [ x] Acknowledge that we will not review PRs that are not passing the build _("green")_ - it is your responsibility to get a proposed PR to pass the build, not primarily the project's maintainers.

- [x ] Create/update unit or integration tests for verifying the changes made.

- [x ] Follow coding conventions at https://cwiki.apache.org/confluence/display/FINERACT/Coding+Conventions.

- [x ] Add required Swagger annotation and update API documentation at fineract-provider/src/main/resources/static/legacy-docs/apiLive.htm with details of any API changes

- [x ] Submission is not a "code dump".  (Large changes can be made "in repository" via a branch.  Ask on the developer mailing list for guidance, if required.)

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/bulkimport/importhandler/ImportHandlerUtils.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/bulkimport/importhandler/ImportHandlerUtils.java
@@ -53,10 +53,17 @@ public final class ImportHandlerUtils {
         Integer noOfEntries = 0;
         // getLastRowNum and getPhysicalNumberOfRows showing false values
         // sometimes
-        while (sheet.getRow(noOfEntries + 1) != null && sheet.getRow(noOfEntries + 1).getCell(primaryColumn) != null) {
+        while (true) {
+            Row row = sheet.getRow(noOfEntries + 1);
+            if (row == null) {
+                break;
+            }
+            Cell cell = row.getCell(primaryColumn);
+            if (cell == null || cell.getCellType() == CellType.BLANK) {
+                break;
+            }
             noOfEntries++;
         }
-
         return noOfEntries;
     }
 
@@ -107,7 +114,7 @@ public final class ImportHandlerUtils {
 
                     String res = trimEmptyDecimalPortion(value.getStringValue());
 
-                    if (!StringUtils.isNotEmpty(res)) {
+                    if (StringUtils.isNotEmpty(res)) {
                         return res.trim();
                     }
                 } catch (Exception e) {
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/bulkimport/importhandler/client/ClientEntityImportHandler.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/bulkimport/importhandler/client/ClientEntityImportHandler.java
@@ -63,7 +63,7 @@ public class ClientEntityImportHandler implements ImportHandler {
     public Count process(final Workbook workbook, final String locale, final String dateFormat) {
 
         List<ClientData> clients = readExcelFile(workbook, locale, dateFormat);
-        return importEntity(workbook, clients, dateFormat);
+        return importEntity(workbook, clients, dateFormat, locale);
     }
 
     private List<ClientData> readExcelFile(final Workbook workbook, final String locale, final String dateFormat) {
@@ -195,7 +195,7 @@ public class ClientEntityImportHandler implements ImportHandler {
                 locale, dateFormat);
     }
 
-    private Count importEntity(final Workbook workbook, final List<ClientData> clients, final String dateFormat) {
+    private Count importEntity(final Workbook workbook, final List<ClientData> clients, final String dateFormat, final String locale) {
         Sheet clientSheet = workbook.getSheet(TemplatePopulateImportConstants.CLIENT_ENTITY_SHEET_NAME);
 
         int successCount = 0;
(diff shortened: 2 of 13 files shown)


fineract-provider/src/main/java/org/apache/fineract/infrastructure/bulkimport/importhandler/client/ClientEntityImportHandler.java
```diff
@@ -63,7 +63,8 @@ public class ClientEntityImportHandler implements ImportHandler {
     public Count process(final Workbook workbook, final String locale, final String dateFormat) {
 
         List<ClientData> clients = readExcelFile(workbook, locale, dateFormat);
-        return importEntity(workbook, clients, dateFormat);
+        return importEntity(workbook, clients, dateFormat, locale);
+        return importEntity(workbook, clients, dateFormat, locale);
```
oleksii-novikov-onix: Duplicated return statement

fineract-provider/src/main/java/org/apache/fineract/infrastructure/bulkimport/populator/fixeddeposits/FixedDepositTransactionWorkbookPopulator.java
```diff
@@ -203,7 +204,12 @@ private void populateSavingsTable(Sheet savingsTransactionSheet, String dateForm
             row = savingsTransactionSheet.createRow(rowIndex++);
             writeString(TransactionConstants.LOOKUP_CLIENT_NAME_COL, row,
                     savingsAccount.getClientName() + "(" + savingsAccount.getClientId() + ")");
-            writeString(TransactionConstants.LOOKUP_ACCOUNT_NO_COL, row, savingsAccount.getAccountNo());
+            try {
+                BigDecimal accountNoAsBigDecimal = new BigDecimal(savingsAccount.getAccountNo());
```
adamsaghy: This should not be BigDecimal... The account number on savings account is string...

adamsaghy:
@RamsesCamacho1171 @reriuss Would it be possible only one of you working on this? It is confusing and inefficient if we need to review and advise on the same piece of code multiple times...

fineract-provider/src/main/java/org/apache/fineract/infrastructure/bulkimport/populator/fixeddeposits/FixedDepositTransactionWorkbookPopulator.java
RamsesCamacho1171: The problem is that when I put the account number as a string in the Excel formulas, it's not working correctly. I tried changing the formulas directly, but it didn't work. What do you recommend? @adamsaghy @oleksii-novikov-onix

adamsaghy: Account number is a string... just because usually they store a number in it, it is a string... I dont really understand what means "The problem is that when I put the account number as a string in the Excel formulas, it's not working correctly"... what is not working? What are you trying to achieve? what problem occurs? etc.

If you must have a number still BigDecimal is not the way to go... you can give a try to Long then... but still... in Fineract the account number is string... usually it is a whole number, but not necessarily... also "0001" is a valid account number which will be "1" if you try to convert it to Long... and "1" is not equal to "0001".... so we are back to square one: account number is string... and it is not a number...

adamsaghy (changes requested):
Kindly see my concerns!

fineract-provider/src/main/java/org/apache/fineract/infrastructure/bulkimport/populator/fixeddeposits/FixedDepositTransactionWorkbookPopulator.java
RamsesCamacho1171: I already returned it to string, but I want to explain why I wanted to modify this, what happens is that in Excel the productname section is automatically filled with a formula and for some reason it is not working, because it has to compare the value of column c (account number) and search for the same number in the data table (column q), if it finds it it returns the value of column r (which is the name of the product), but I don't know if because in Excel the account number is a string it cannot make this comparison correctly and therefore does not return the product name

image

fineract-provider/src/main/java/org/apache/fineract/infrastructure/bulkimport/importhandler/client/ClientEntityImportHandler.java
RamsesCamacho1171: Duplication has now been eliminated

fineract-provider/src/main/java/org/apache/fineract/infrastructure/bulkimport/importhandler/ImportHandlerUtils.java
```diff
@@ -53,10 +53,17 @@ public static Integer getNumberOfRows(Sheet sheet, int primaryColumn) {
         Integer noOfEntries = 0;
         // getLastRowNum and getPhysicalNumberOfRows showing false values
         // sometimes
-        while (sheet.getRow(noOfEntries + 1) != null && sheet.getRow(noOfEntries + 1).getCell(primaryColumn) != null) {
+        while (true) {
```
adamsaghy, line 56: Please rewrite to this:
```
int maxRows = sheet.getLastRowNum();

while (noOfEntries < maxRows) {
    Row row = sheet.getRow(noOfEntries + 1);
    if (row == null) {
        break;
    }
    Cell cell = row.getCell(primaryColumn);
    if (cell == null || cell.getCellType() == CellType.BLANK) {
        break;
    }
    noOfEntries++;
}
```

This a little bit better as it never an endless loop

PR #4638, opened 2025-05-01
FINERACT-2242: New command processing - External Event Configuration (org.apache.fineract.infrastructure.event.external)

VictorPavfurious:
## Description

## Checklist


VictorPavfurious:
@vidakovic please check it

fineract-core/src/main/java/org/apache/fineract/infrastructure/event/external/service/ExternalEventConfigurationWritePlatformServiceImpl.java
```diff
@@ -22,31 +22,28 @@
...
     private final ExternalEventConfigurationRepository repository;
-    private final ExternalEventConfigurationCommandFromApiJsonDeserializer fromApiJsonDeserializer;
 
     @Transactional
     @Override
-    public CommandProcessingResult updateConfigurations(final JsonCommand command) {
-        final ExternalEventConfigurationCommand configurationCommand = fromApiJsonDeserializer.commandFromApiJson(command.json());
-        final Map<String, Boolean> commandConfigurations = configurationCommand.externalEventConfigurations();
-        final Map<String, Object> changes = new HashMap<>();
+    public ExternalEventConfigurationResponse updateConfigurations(Command<ExternalEventConfigurationRequest> command) {
```
galovics, line 43: I'm not sure I'm getting this and all these "new concept" changes.

The original issue was that the API level things (JSON, etc) are sinking into the business logic. 

Here the same things happens but instead of the raw JsonCommand class, we are dealing with the respective Request and Response classes. How is this any different than the original impl?

In my mind there should be a 3 layered structure in place to avoid leaking implementation details (as adivsed by the Dependency Inversion principle in SOLID).

API DTOs (Request/Response) -> Service (Business objects) -> Repository (DB level objects).

This means transformation in every layer and the arrows are intentionally the way they are, so no implementation depends on the upper layer.

@vidakovic @adamsaghy thoughts? what am I missing here?

galovics (changes requested):

fineract-core/src/main/java/org/apache/fineract/infrastructure/event/external/service/ExternalEventConfigurationWritePlatformServiceImpl.java
adamsaghy: I dont think you are missing anything, except these separation is not provided at any layer. Entities are used in services not only in repositories. DTOs are (re)used everywhere....

I reckon the idea was to do 1 step closer to the ideal solution:
leave the JsonCommand, JsonElement behind and work with DTOs...even if they are shared between the layers... I guess we can consider one additional step here...

Lets focus on the API - Service layer for now: 

Regarding communication from API to Service layer, what are your recommendations? 

My 2 cents if we are considering to have DTO separation between these layers, first we should move validations and some of the data transformation (resolving enums, etc) from service layer into the API layer to have real difference between the incoming request (DTO No.1) into the API layer and the incoming request (DTO No.2) into service layer.

@galovics @vidakovic what do you think?

VictorPavfurious: I think, that the main issue - we have many class JsonCommand, custom JsonHelpers and custom JsonSerialization for request/response String, and biggest class in response which use everywhere is class CommandProccesingResult, which has many unnecessary fields and etc, we should stop use it and make our architecture more clearer... Because of using this "custom" serializations - make the logic more complicated... So, in this new concept, we want to use pattern command (via handlers) for our API and use concrete DTO's and Response for API (without any custom logic, need to remove it..), that way it will be clearer in my opinion.

VictorPavfurious: Sorry, I don't fully catch the essence of the problem yet.. Do you think that concept has some leaking?

galovics: @adamsaghy Yep. The most pressing issue even with this new solution is that the API level things are sinking into the service layer.

Let's have separate DTOs, Mapstruct mappers in place so we can separate these.

Considering the validation logic, definitely. Plus there should be multi-level validations as well:
- One validator on the API layer which purely checks whether an API request is valid (like an attribute has positive numbers, notnull, etc)
- One validator on the service layer which checks against business logic (like a cumulative loan shouldn't have XYZ set as well)

These 2 validations are 100% mixed together as of today but they should be separated.

VictorPavfurious: @galovics , why we can not create custom jakarta validation for specific conditions for all request dto's, for what  multi-level validations and specific validation in layer like services? 
Service should do only business logic without any validation..

vidakovic: Let's see if we can standardize validations with Jakarta Validations... I understand the "business validation" part, but let's see if we can get this under one API... having 2 different validation concepts in place is just hard to explain (read: people will start again to do whatever comes to their minds). Let me see if I can bulldozer an example.

galovics: Jakarta Validations are fine for me on the API level but that's usually usable for single value operations, not complex business logic validations.

PR #4688, opened 2025-05-23
FINERACT-2276: Deprecate findExistingTransactionIds and fetch only the ids from database instead

mariiaKraievska:
## Description

## Checklist

### DIFF
--- a/fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
+++ b/fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/Loan.java
@@ -760,13 +760,6 @@ public class Loan extends AbstractAuditableWithUTCDateTimeCustom<Long> {
         return maturityDate;
     }
 
-    public List<Long> findExistingTransactionIds() {
-        return getLoanTransactions().stream() //
-                .filter(loanTransaction -> loanTransaction.getId() != null) //
-                .map(LoanTransaction::getId) //
-                .collect(Collectors.toList());
-    }
-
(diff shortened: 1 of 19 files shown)


galovics (changes requested):
I mean I understand the intent here and it's good but this could have side-effects to the behavior of the business logic.

The former approach (what you removed) is using Hibernate's 1st level cache to retrieve the loan transactions for a loan. That means whenever a loan is loaded with the corresponding transactions, it'll only see those transactions even if something during the time of execution changes in the database.

Another thing with the change is the fact that now whenever we need the transaction list for a loan, it'll go to the database every time resulting in performance degradation.

Thoughts @mariiaKraievska @adamsaghy ?

UPDATE: probably the first issue doesn't exist currently in Fineract because I see the tests are green.

adamsaghy:
> I mean I understand the intent here and it's good but this could have side-effects to the behavior of the business logic.
> 
> The former approach (what you removed) is using Hibernate's 1st level cache to retrieve the loan transactions for a loan. That means whenever a loan is loaded with the corresponding transactions, it'll only see those transactions even if something during the time of execution changes in the database.
> 
> Another thing with the change is the fact that now whenever we need the transaction list for a loan, it'll go to the database every time resulting in performance degradation.
> 
> Thoughts @mariiaKraievska @adamsaghy ?
> 
> UPDATE: probably the first issue doesn't exist currently in Fineract because I see the tests are green.

we are talking about loan ids here. I see performance issue by the fact every transactions of the loan is always fetched and some other entities as well alongside with the transactions. these methods sole and only purpose was to support Fineract weird journal entry creation logic!

galovics (approved):

PR #4711, opened 2025-05-29
Create Dockerfile

MYKLCS:
for deployment with railway


galovics (changes requested):
Fineract uses Jib to prepare docker images. Please use that and refrain creating PRs for raw Dockerfiles.

Thanks.

PR #4715, opened 2025-05-29
FINERACT-2306: release doc fixes

meonkeys:
## Description

* asciidoc: link directly to FINERACT-873 and shorten it for brevity / clarity
* fineract plugin:
  * print email, don't send (WIP, but I might keep this... I prefer manually sending emails for now.)
  * support different release date and release branch date

see also: FINERACT-2306

## Checklist

### DIFF
--- a/buildSrc/src/main/groovy/org/apache/fineract/gradle/FineractPlugin.groovy
+++ b/buildSrc/src/main/groovy/org/apache/fineract/gradle/FineractPlugin.groovy
@@ -194,7 +194,7 @@ class FineractPlugin implements Plugin<Project> {
 
                 String version = project.properties?['fineract.release.version']
                 String issue = project.properties?['fineract.release.issue']
-                String date = project.properties?['fineract.release.date']
+                String date = project.properties?['fineract.releaseBranch.date']
 
                 if(!version || !issue || !date) {
                     TextIO textIO = TextIoFactory.getTextIO()
@@ -225,7 +225,7 @@ class FineractPlugin implements Plugin<Project> {
 
                 this.context?.project?['fineract.release.version'] = version
                 this.context?.project?['fineract.release.issue'] = issue
-                this.context?.project?['fineract.release.date'] = date
+                this.context?.project?['fineract.releaseBranch.date'] = date
 
                 if(step.email) {
                     emailService.send( processEmailParams(step.email, this.context) )
--- a/buildSrc/src/main/resources/email/release.step01.headsup.message.ftl
+++ b/buildSrc/src/main/resources/email/release.step01.headsup.message.ftl
@@ -20,11 +20,11 @@
 -->
 Hello everyone,
 
-... based on our "How to Release Apache Fineract" process documented at https://cwiki.apache.org/confluence/x/DRwIB:
+... based on our "How to Release Apache Fineract" process (https://cwiki.apache.org/confluence/x/DRwIB) as well as the "Releases" chapter in the docs (https://fineract.apache.org/docs/current/#_releases),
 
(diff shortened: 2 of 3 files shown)


meonkeys:
FYI, I [asked](https://mifos.slack.com/archives/C028634A61L/p1748553812461529) in our `#fineract` Slack chat:

> I almost feel bad for all this ceremony (github PR, JIRA ticket, full build, review) for such a trivial patch. So I'm wondering out loud here: Should I invoke the full ceremony for trivial patches? Or just commit directly to develop and push? The former feels most [strictly "by the book"](https://github.com/apache/fineract/#pull-requests)/safest/simplest/most consistent, the latter feels like [lazy consensus](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=91554327#FAQ(frequentlyaskedquestions)-WhatisourConsensusapproach?), but may be much more efficient (particularly for me, I like to commit small patches early and often) when the risk is particularly low (documentation-only changes). I'd of course still make sure all my changes aren't breaking the build. Anyway, just thought I'd raise this up the flagpole to see if anyone salutes.

note that, in this patch, the groovy code changed only runs locally and is only used during the release process to simplify release email creation, and the rest is just in the asciidoc (which I built and visually checked)

buildSrc/src/main/groovy/org/apache/fineract/gradle/service/EmailService.groovy
```diff
@@ -76,7 +76,8 @@ class EmailService {
             msg.setSubject(params.subject, "UTF-8")
             msg.setText(params.message, "UTF-8");
 
-            Transport.send(msg);
+            //Transport.send(msg);
```
adamsaghy: Is there a reason to remove the email sending?

meonkeys: Great question. I did think about this for a bit. I should have been more transparent about my intent. Allow me to explain.

For the 1.12.0 release process I used the groovy code to generate email templates, but I wanted to edit the emails before sending so I disabled this line, and intentionally left it so, figuring I'd be the one using and [hopefully minimally] maintaining it for now.

The Fineract (release) plugin appears to be more or less unmaintained for about 2 1/2 years now (save the semiautomated dependency updates). Looking at the code and history, I assumed I was the only one that would be using it now and in the near future. It also struck me how most everything it does is primarily to work around antiquated Apache release process steps that are not necessarily what we want for Fineract. Personally I think the Fineract release plugin should be simply replaced by **a better release process**.

I think I'll add a comment there, and of course I welcome your feedback.

meonkeys: Anyway yeah, my intent here was simply to do the simplest, safest, and most useful change here so I could keep this release plugin limping along until we improve the release process itself. No disrespect to @vidakovic and @vorburger and whomever else poured love into this code over the years. Perhaps the options for releasing Apache code have improved since then?

Today we can of course do better (e.g. just let GitHub help us release with far fewer clicks/steps).

Also I am not proposing we change the release process today or even soon. Maybe in a quarter or two I'll propose changes, if/when there seems to be political will around that.

adamsaghy: hmm... i am not particularly mind if you are reworking the release process, but i think having a "send" step which was meant to sending a mail but now the actual sending got commented out is misleading, no?

meonkeys: Yep, you're right. I think I'll just back out this change because I agree with you it's bad to just leave the send() call commented out without any other changes.

PR #4819, opened 2025-06-27
FINERACT-2181: New Loan Accrual Activity transactions publish LoanAdjustTransaction event during loan reopen

mariiaKraievska:
## Description

## Checklist


fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanAccrualActivityProcessingServiceImpl.java
```diff
@@ -195,9 +195,10 @@ public void processAccrualActivityForLoanReopen(final @NonNull Loan loan) {
             return isDueBefore && isAfterOrEqualToLastAccrualDate;
         }).sorted(Comparator.comparing(LoanRepaymentScheduleInstallment::getDueDate)).toList();
 
-        installments.forEach(installment -> {
```
adamsaghy, line 198: Dont you like the stream foreach? :)

mariiaKraievska: Good question! I do like the stream forEach, but here I needed to reset the lastAccrualActivityMarkedToReverse variable after the first iteration. Since Java doesn't allow reassigning variables inside a lambda, I had to use the classic for-each loop to make this work :(

adamsaghy: I see. Thanks!

Still "lastAccrualActivityMarkedToReverse = Optional.empty();" what does this do exactly?

mariiaKraievska: Setting lastAccrualActivityMarkedToReverse = Optional.empty(); ensures that only the first installment in the loop uses the original value of lastAccrualActivityMarkedToReverse. For all subsequent installments, it will be empty, so they will not be treated as a replay of the previous accrual activity, but as new accrual activities.

This is necessary to publish the correct business event: only the first accrual activity after loan reopen should be a replay, and the rest should be new. 

By the way, this line (lastAccrualActivityMarkedToReverse = Optional.empty();) was actually present in the original implementation, but it seems it got lost during a recent refactoring :(

PR #4825, opened 2025-06-30
FINERACT-2314: IP tracking

JohnAlva:
## Description

FINERACT-2314: IP tracking

## Checklist

### DIFF
--- a/fineract-core/src/main/java/org/apache/fineract/commands/domain/CommandSource.java
+++ b/fineract-core/src/main/java/org/apache/fineract/commands/domain/CommandSource.java
@@ -35,6 +35,7 @@ import org.apache.fineract.infrastructure.core.data.CommandProcessingResult;
 import org.apache.fineract.infrastructure.core.domain.AbstractPersistableCustom;
 import org.apache.fineract.infrastructure.core.domain.ExternalId;
 import org.apache.fineract.infrastructure.core.service.DateUtils;
+import org.apache.fineract.infrastructure.core.service.IpAddressUtils;
 import org.apache.fineract.useradministration.domain.AppUser;
 
 @Entity
@@ -137,6 +138,9 @@ public class CommandSource extends AbstractPersistableCustom<Long> {
     @Column(name = "loan_external_id", length = 100)
     private ExternalId loanExternalId;
 
+    @Column(name = "client_ip", nullable = true)
+    private String clientIp;
+
     @Column(name = "is_sanitized", nullable = false)
     private boolean sanitized;
 
@@ -162,6 +166,7 @@ public class CommandSource extends AbstractPersistableCustom<Long> {
                 .transactionId(command.getTransactionId()) //
                 .creditBureauId(command.getCreditBureauId()) //
                 .organisationCreditBureauId(command.getOrganisationCreditBureauId()) //
+                .clientIp(IpAddressUtils.getClientIp()) //
                 .loanExternalId(command.getLoanExternalId()).sanitized(sanitized).build(); //
     }
 
--- a/fineract-core/src/main/java/org/apache/fineract/infrastructure/core/api/JsonCommand.java
+++ b/fineract-core/src/main/java/org/apache/fineract/infrastructure/core/api/JsonCommand.java
@@ -596,5 +596,4 @@ public final class JsonCommand {
     public void checkForUnsupportedParameters(final Type typeOfMap, final String json, final Set<String> requestDataParameters) {
         this.fromApiJsonHelper.checkForUnsupportedParameters(typeOfMap, json, requestDataParameters);
     }
-
 }
--- a/fineract-core/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java
+++ b/fineract-core/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java
@@ -50,6 +50,8 @@ public class FineractProperties {
 
     private FineractCorrelationProperties correlation;
 
+    private FineractIpTrackingProperties ipTracking;
+
     private FineractPartitionedJob partitionedJob;
 
     private FineractRemoteJobMessageHandlerProperties remoteJobMessageHandler;
@@ -151,6 +153,13 @@ public class FineractProperties {
 
         private boolean enabled;
         private String headerName;
+    }
+
+    @Getter
+    @Setter
+    public static class FineractIpTrackingProperties {
+
+        private boolean enabled;
     }
 
     @Getter
--- a/fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/CallerIpTrackingFilter.java
+++ b/fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/CallerIpTrackingFilter.java
@@ -1,0 +1,101 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.fineract.infrastructure.core.filters;
+
+import jakarta.servlet.FilterChain;
+import jakarta.servlet.ServletException;
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.fineract.infrastructure.core.config.FineractProperties;
+import org.springframework.web.filter.OncePerRequestFilter;
+
+@RequiredArgsConstructor
+@Slf4j
+public class CallerIpTrackingFilter extends OncePerRequestFilter {
+
+    private final FineractProperties fineractProperties;
+
+    /**
+     * Common headers used to get client IP from different proxies.
+     *
+     * "X-Forwarded-For", // Standard header used by proxies "Proxy-Client-IP", // Used by some Apache proxies
+     * "WL-Proxy-Client-IP", // Used by WebLogic "HTTP_X_FORWARDED_FOR", // Alternative to X-Forwarded-For
+     * "HTTP_X_FORWARDED", // Variation of X-Forwarded "HTTP_X_CLUSTER_CLIENT_IP", // Used in clustered environments
+     * "HTTP_CLIENT_IP", // Fallback, less common "HTTP_FORWARDED_FOR", // Less standard, used in some setups
+     * "HTTP_FORWARDED", // Standardized header (RFC 7239) that can include client IP, proxy info, and protocol
+     * "HTTP_VIA", // Shows intermediate proxies "REMOTE_ADDR" // Server's perceived client IP
+     */
+
+    private static final String[] IP_HEADER_CANDIDATES = { "X-Forwarded-For", "Proxy-Client-IP", "WL-Proxy-Client-IP",
+            "HTTP_X_FORWARDED_FOR", "HTTP_X_FORWARDED", "HTTP_X_CLUSTER_CLIENT_IP", "HTTP_CLIENT_IP", "HTTP_FORWARDED_FOR",
+            "HTTP_FORWARDED", "HTTP_VIA", "REMOTE_ADDR" };
+
+    public String getClientIpAddress(HttpServletRequest request) {
+        for (String header : IP_HEADER_CANDIDATES) {
+            String ip = request.getHeader(header);
+            if (ip != null && ip.length() != 0 && !ip.isEmpty()) {
+                log.trace("CALLER IP : {}", ip);
+                return ip;
+            }
+        }
+        log.trace("getRemoteAddr method : {}", request.getRemoteAddr());
+        return request.getRemoteAddr();
+    }
+
+    @Override
+    protected void doFilterInternal(final HttpServletRequest request, final HttpServletResponse response, final FilterChain filterChain)
+            throws IOException, ServletException {
+        if (fineractProperties.getIpTracking().isEnabled()) {
+            handleClientIp(request, response, filterChain);
+        } else {
+            filterChain.doFilter(request, response);
+        }
+
+    }
+
+    private void handleClientIp(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
+            throws IOException, ServletException {
+        try {
+            String clientIpAddress = getClientIpAddress(request);
+            if (StringUtils.isNotBlank(clientIpAddress)) {
+                log.trace("Found Client IP in header : {}", clientIpAddress);
+                request.setAttribute("IP", clientIpAddress);
+            }
+            filterChain.doFilter(request, response);
+        } finally {
+            request.setAttribute("IP", "");
+        }
+    }
+
+    @Override
+    protected boolean isAsyncDispatch(final HttpServletRequest request) {
+        return false;
+    }
+
+    @Override
+    protected boolean shouldNotFilterErrorDispatch() {
+        return false;
+    }
+
+}
--- a/fineract-provider/src/main/java/org/apache/fineract/commands/service/AuditReadPlatformServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/commands/service/AuditReadPlatformServiceImpl.java
@@ -110,11 +110,12 @@ public class AuditReadPlatformServiceImpl implements AuditReadPlatformService {
                     + "ck.username as checker, aud.checked_on_date as checkedOnDate, aud.checked_on_date_utc as checkedOnDateUTC,  ev.enum_message_property as processingResult "
                     + commandAsJsonString + ", "
                     + " o.name as officeName, gl.level_name as groupLevelName, g.display_name as groupName, c.display_name as clientName, "
-                    + " l.account_no as loanAccountNo, s.account_no as savingsAccountNo " + " from m_portfolio_command_source aud "
-                    + " left join m_appuser mk on mk.id = aud.maker_id" + " left join m_appuser ck on ck.id = aud.checker_id"
-                    + " left join m_office o on o.id = aud.office_id" + " left join m_group g on g.id = aud.group_id"
-                    + " left join m_group_level gl on gl.id = g.level_id" + " left join m_client c on c.id = aud.client_id"
-                    + " left join m_loan l on l.id = aud.loan_id" + " left join m_savings_account s on s.id = aud.savings_account_id"
+                    + " l.account_no as loanAccountNo, s.account_no as savingsAccountNo , aud.client_ip  as ip "
+                    + " from m_portfolio_command_source aud " + " left join m_appuser mk on mk.id = aud.maker_id"
+                    + " left join m_appuser ck on ck.id = aud.checker_id" + " left join m_office o on o.id = aud.office_id"
+                    + " left join m_group g on g.id = aud.group_id" + " left join m_group_level gl on gl.id = g.level_id"
+                    + " left join m_client c on c.id = aud.client_id" + " left join m_loan l on l.id = aud.loan_id"
+                    + " left join m_savings_account s on s.id = aud.savings_account_id"
                     + " left join r_enum_value ev on ev.enum_name = 'status' and ev.enum_id = aud.status";
 
             // data scoping: head office (hierarchy = ".") can see all audit
@@ -158,13 +159,14 @@ public class AuditReadPlatformServiceImpl implements AuditReadPlatformService {
             final String clientName = rs.getString("clientName");
             final String loanAccountNo = rs.getString("loanAccountNo");
             final String savingsAccountNo = rs.getString("savingsAccountNo");
+            final String ip = rs.getString("ip");
 
             ZonedDateTime madeOnDate = madeOnDateUTC != null ? madeOnDateUTC.toZonedDateTime() : madeOnDateTenant;
             ZonedDateTime checkedOnDate = checkedOnDateUTC != null ? checkedOnDateUTC.toZonedDateTime() : checkedOnDateTenant;
 
             return new AuditData(id, actionName, entityName, resourceId, subresourceId, maker, madeOnDate, checker, checkedOnDate,
                     processingResult, commandAsJson, officeName, groupLevelName, groupName, clientName, loanAccountNo, savingsAccountNo,
-                    clientId, loanId, resourceGetUrl);
+                    clientId, loanId, resourceGetUrl, ip);
         }
     }
 
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/SecurityConfig.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/SecurityConfig.java
@@ -32,6 +32,7 @@ import org.apache.fineract.infrastructure.businessdate.service.BusinessDateReadP
 import org.apache.fineract.infrastructure.cache.service.CacheWritePlatformService;
 import org.apache.fineract.infrastructure.configuration.domain.ConfigurationDomainService;
 import org.apache.fineract.infrastructure.core.domain.FineractRequestContextHolder;
+import org.apache.fineract.infrastructure.core.filters.CallerIpTrackingFilter;
 import org.apache.fineract.infrastructure.core.filters.CorrelationHeaderFilter;
 import org.apache.fineract.infrastructure.core.filters.IdempotencyStoreFilter;
 import org.apache.fineract.infrastructure.core.filters.IdempotencyStoreHelper;
@@ -172,7 +173,9 @@ public class SecurityConfig {
         } else {
             http.addFilterAfter(idempotencyStoreFilter(), FineractInstanceModeApiFilter.class); //
         }
-
+        if (fineractProperties.getIpTracking().isEnabled()) {
+            http.addFilterAfter(callerIpTrackingFilter(), RequestResponseFilter.class);
+        }
         if (fineractProperties.getSecurity().getTwoFactor().isEnabled()) {
             http.addFilterAfter(twoFactorAuthenticationFilter(), CorrelationHeaderFilter.class);
         } else {
@@ -217,6 +220,10 @@ public class SecurityConfig {
 
     public CorrelationHeaderFilter correlationHeaderFilter() {
         return new CorrelationHeaderFilter(fineractProperties, mdcWrapper);
+    }
+
+    public CallerIpTrackingFilter callerIpTrackingFilter() {
+        return new CallerIpTrackingFilter(fineractProperties);
     }
 
     public TenantAwareBasicAuthenticationFilter tenantAwareBasicAuthenticationFilter() throws Exception {
--- a/fineract-provider/src/main/resources/application.properties
+++ b/fineract-provider/src/main/resources/application.properties
@@ -62,6 +62,7 @@ fineract.correlation.enabled=${FINERACT_LOGGING_HTTP_CORRELATION_ID_ENABLED:fals
 fineract.correlation.header-name=${FINERACT_LOGGING_HTTP_CORRELATION_ID_HEADER_NAME:X-Correlation-ID}
 
 fineract.job.stuck-retry-threshold=${FINERACT_JOB_STUCK_RETRY_THRESHOLD:5}
+fineract.ip-tracking.enabled=${FINERACT_CLIENT_IP_TRACKING_ENABLED:false}
 fineract.job.loan-cob-enabled=${FINERACT_JOB_LOAN_COB_ENABLED:true}
 
 fineract.partitioned-job.partitioned-job-properties[0].job-name=LOAN_COB
--- a/fineract-provider/src/test/resources/application-test.properties
+++ b/fineract-provider/src/test/resources/application-test.properties
@@ -20,6 +20,7 @@
 fineract.node-id=1
 
 fineract.security.basicauth.enabled=true
+fineract.ip-tracking.enabled=false
 fineract.security.oauth.enabled=false
 fineract.security.2fa.enabled=false
 fineract.security.hsts.enabled=false
--- a/integration-tests/src/test/java/org/apache/fineract/integrationtests/IpTrackingIntegrationTest.java
+++ b/integration-tests/src/test/java/org/apache/fineract/integrationtests/IpTrackingIntegrationTest.java
@@ -1,0 +1,74 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.integrationtests;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+
+import io.restassured.builder.RequestSpecBuilder;
+import io.restassured.builder.ResponseSpecBuilder;
+import io.restassured.http.ContentType;
+import io.restassured.specification.RequestSpecification;
+import io.restassured.specification.ResponseSpecification;
+import java.util.HashMap;
+import java.util.List;
+import org.apache.fineract.integrationtests.common.AuditHelper;
+import org.apache.fineract.integrationtests.common.ClientHelper;
+import org.apache.fineract.integrationtests.common.Utils;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class IpTrackingIntegrationTest {
+
+    private AuditHelper auditHelper;
+    private static final String EXPECTED_LOCAL_IP = "127.0.0.1";
+    private RequestSpecification requestSpec;
+    private ResponseSpecification responseSpec;
+    private ResponseSpecification responseSpecForSearch;
+
+    @BeforeEach
+    public void setup() {
+        Utils.initializeRESTAssured();
+        this.requestSpec = new RequestSpecBuilder().setContentType(ContentType.JSON).build();
+        this.requestSpec.header("Fineract-Platform-TenantId", "default");
+        this.requestSpec.auth().basic("mifos", "password");
+        this.responseSpec = new ResponseSpecBuilder().expectStatusCode(200).build();
+        this.responseSpecForSearch = new ResponseSpecBuilder().expectStatusCode(200).build();
+        this.auditHelper = new AuditHelper(this.requestSpec, this.responseSpec);
+    }
+
+    @Test
+    public void capturesIpAddressWhenCreatingClient() throws Exception {
+        assumeTrue(Boolean.parseBoolean(System.getenv().getOrDefault("FINERACT_CLIENT_IP_TRACKING_ENABLED", "true")),
+                "Saltando test porque el tracking de IP está deshabilitado");
+
+        // given
+        final Integer clientId = ClientHelper.createClient(this.requestSpec, this.responseSpec);
+        ClientHelper.verifyClientCreatedOnServer(this.requestSpec, this.responseSpec, clientId);
+        List<HashMap<String, Object>> auditsRecieved = auditHelper.getAuditDetails(clientId, "CREATE", "CLIENT");
+
+        // when
+        String ip = auditsRecieved.get(0).get("ip").toString();
+
+        assumeTrue(!ip.isEmpty(), "IP not arrived: skipping capture test when enabled");
+        // then
+        assertEquals(EXPECTED_LOCAL_IP, ip, "Expected local IP when tracking is enabled");
+    }
+
+}
--- a/fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/IpAddressUtils.java
+++ b/fineract-core/src/main/java/org/apache/fineract/infrastructure/core/service/IpAddressUtils.java
@@ -1,0 +1,39 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.infrastructure.core.service;
+
+import org.springframework.web.context.request.RequestContextHolder;
+import org.springframework.web.context.request.ServletRequestAttributes;
+
+public final class IpAddressUtils {
+
+    private IpAddressUtils() {}
+
+    public static String getClientIp() {
+        ServletRequestAttributes attrs = (ServletRequestAttributes) RequestContextHolder.getRequestAttributes();
+        String clientIp = "";
+        if (attrs != null) {
+            Object ipAttr = attrs.getRequest().getAttribute("IP");
+            if (ipAttr != null) {
+                clientIp = ipAttr.toString();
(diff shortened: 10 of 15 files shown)


adamsaghy:
@JohnAlva @IOhacker Would you mind to provide some description for this PR? The FINERACT story is empty :(

IOhacker:
@adamsaghy content has been added.

adamsaghy:
@JohnAlva Please rebase your PR and make sure you run the spotlessApply and checkstyle on the PR before it got committed:
`./gradlew spotlessApply spotbugsMain spotbugsTest checkstyleMain checkstyleTest`

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.fineract.infrastructure.core.config.FineractProperties;
+import org.springframework.web.filter.OncePerRequestFilter;
+
+@RequiredArgsConstructor
+@Slf4j
+public class GeolocationHeaderFilter extends OncePerRequestFilter {
+
+    private final FineractProperties fineractProperties;
+
+    private static final String[] IP_HEADER_CANDIDATES = { "X-Forwarded-For", "Proxy-Client-IP", "WL-Proxy-Client-IP",
```
adamsaghy: How was the priority order of the IP header candidates decided?
Is there a specific reason for the ordering, or is it based on best practice or precedent?

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
```diff
@@ -0,0 +1,93 @@
...
+            "HTTP_X_FORWARDED_FOR", "HTTP_X_FORWARDED", "HTTP_X_CLUSTER_CLIENT_IP", "HTTP_CLIENT_IP", "HTTP_FORWARDED_FOR",
+            "HTTP_FORWARDED", "HTTP_VIA", "REMOTE_ADDR" };
+
+    public static String getClientIpAddress(HttpServletRequest request) {
+        for (String header : IP_HEADER_CANDIDATES) {
+            String ip = request.getHeader(header);
+            if (ip != null && ip.length() != 0 && !"unknown".equalsIgnoreCase(ip)) {
+                log.debug("SEND IP : {}", ip);
```
adamsaghy: "SEND IP" you mean "CALLER IP"?

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+    }
+
+    private void handleClientIp(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain,
+            FineractProperties.FineractGeolocationProperties geolocationProperties) throws IOException, ServletException {
+        try {
+            String clientIpAddress = getClientIpAddress(request);
+            if (StringUtils.isNotBlank(clientIpAddress)) {
+                log.info("Found Client IP in header : {}", clientIpAddress);
```
adamsaghy: Please change the log level to DEBUG

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
```diff
@@ -0,0 +1,93 @@
...
+                ClientIpHolder.setClientIp(clientIpAddress);
```
adamsaghy: Is there a reason to store in thread local the ip address?

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
```diff
@@ -0,0 +1,93 @@
...
+            }
+            filterChain.doFilter(request, response);
+        } catch (Exception e) {
+            e.printStackTrace();
```
adamsaghy: Please have correct error handling! "e.printStackTrace" is incorrect and easily hide exceptions that propagate from other filters!

fineract-provider/src/main/java/org/apache/fineract/commands/service/AuditReadPlatformServiceImpl.java
```diff
@@ -158,13 +159,14 @@ public AuditData mapRow(final ResultSet rs, @SuppressWarnings("unused") final in
             final String clientName = rs.getString("clientName");
             final String loanAccountNo = rs.getString("loanAccountNo");
             final String savingsAccountNo = rs.getString("savingsAccountNo");
+            final String ip = (rs.getString("ip") != null) ? rs.getString("ip") : "SN/IP";
```
adamsaghy: what does "SN/IP" means?

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
```diff
@@ -0,0 +1,93 @@
...
+import java.io.IOException;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.fineract.infrastructure.core.config.FineractProperties;
+import org.springframework.web.filter.OncePerRequestFilter;
+
+@RequiredArgsConstructor
+@Slf4j
+public class GeolocationHeaderFilter extends OncePerRequestFilter {
```
adamsaghy: You can make it a conditional Spring bean which only registers if it was enabled:
`@ConditionalOnProperty("fineract.geolocation.enabled")`

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/SecurityConfig.java
```diff
@@ -165,7 +166,9 @@ public SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
                 .addFilterBefore(tenantAwareBasicAuthenticationFilter(), SecurityContextHolderFilter.class) //
                 .addFilterAfter(requestResponseFilter(), ExceptionTranslationFilter.class) //
                 .addFilterAfter(correlationHeaderFilter(), RequestResponseFilter.class) //
-                .addFilterAfter(fineractInstanceModeApiFilter(), CorrelationHeaderFilter.class); //
+                .addFilterAfter(fineractInstanceModeApiFilter(), CorrelationHeaderFilter.class) //
+                .addFilterAfter(fineractInstanceModeApiFilter(), CorrelationHeaderFilter.class) //
+                .addFilterAfter(geolocationHeaderFilter(), RequestResponseFilter.class); //
```
adamsaghy: You can decide to not register this filter if the functionality is disabled... See example: `loanCOBFilterHelper`

fineract-provider/src/main/resources/application.properties
```diff
@@ -62,6 +62,7 @@ fineract.correlation.enabled=${FINERACT_LOGGING_HTTP_CORRELATION_ID_ENABLED:fals
 fineract.correlation.header-name=${FINERACT_LOGGING_HTTP_CORRELATION_ID_HEADER_NAME:X-Correlation-ID}
 
 fineract.job.stuck-retry-threshold=${FINERACT_JOB_STUCK_RETRY_THRESHOLD:5}
+fineract.geolocation.enabled=${FINERACT_GEOLOCATION_ENABLED:true}
```
adamsaghy: I dont think we should enabled this by default... but i can be convinced...

Tedyyy-Albur:
@adamsaghy

adamsaghy:
> @adamsaghy

@Tedyyy-Albur ???

JohnAlva:
@adamsaghy Hi Adam 
we already fixed the conflict, 
could you help us to review  please?

adamsaghy:
@JohnAlva the PR in conflicted state. the tests are failed and there are great many open questions on it!

JohnAlva:
@adamsaghy Hello, Can you help us review again, and about the change of tests we don't work on those.

adamsaghy:
@JohnAlva Please use the correct PR title and commit message naming convention:
`FINERACT-2314 - IP Tracking` -> `FINERACT-2314: IP tracking`

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
adamsaghy: @JohnAlva Can you please advise on this?

adamsaghy (changes requested):
Please check my comment regarding the PR title and commit message!
Please check all the unanswered comments and advise on them!
Please make sure, you are adding proper testing which ensures the functionality is working as expected!

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
Shelaslifter: Thanks, I've changed it to 'CALLER IP'.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
JohnAlva: Thanks, I've changed the log level to DEBUG.

fineract-provider/src/main/java/org/apache/fineract/commands/service/AuditReadPlatformServiceImpl.java
JohnAlva: I've changed it to 'NO IP' for better understanding

fineract-provider/src/main/resources/application.properties
JohnAlva: We've set it to false instead, as it seems more appropriate for this case.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/SecurityConfig.java
JohnAlva: We've added a function to handle the filters when they are disabled. Specifically, we implemented this for the geolocation filter, as it was not previously disabled, and we have now included this functionality.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
JohnAlva: We've added the conditional Spring bean as requested. It's now correctly registered with @ConditionalOnProperty("fineract.geolocation.enabled")

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
JohnAlva: We've implemented the correct error handling as requested. Instead of e.printStackTrace(), we've ensured that the exception is properly logged, and any exceptions are now handled in a way that doesn't hide issues from other filters.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
adamsaghy: `@ConditionalOnProperty("fineract.geolocation.enabled")` only works if this class is a Sprint Bean (which is not at the moment). 

Please make sure `GeolocationHeaderFilter` is a Spring Bean and new instance is not created manually:
No `new GeolocationHeaderFilter()` anywhere!

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
```diff
@@ -0,0 +1,95 @@
...
+    private void handleClientIp(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain,
+            FineractProperties.FineractGeolocationProperties geolocationProperties) throws IOException, ServletException {
+        try {
+            String clientIpAddress = getClientIpAddress(request);
+            if (StringUtils.isNotBlank(clientIpAddress)) {
+                log.debug("Found Client IP in header : {}", clientIpAddress);
+                ClientIpHolder.setClientIp(clientIpAddress);
+            }
+            filterChain.doFilter(request, response);
+        } catch (Exception e) {
```
adamsaghy: This try - catch is way too broad! You should move the `filterChain.doFilter(request, response)` call outside as this will catch and hide any exceptions that might come from any other filters! 

Do we need the catch case for anything? I think it should be removed all together!

fineract-provider/src/main/java/org/apache/fineract/commands/service/AuditReadPlatformServiceImpl.java
```diff
@@ -158,13 +159,14 @@ public AuditData mapRow(final ResultSet rs, @SuppressWarnings("unused") final in
             final String clientName = rs.getString("clientName");
             final String loanAccountNo = rs.getString("loanAccountNo");
             final String savingsAccountNo = rs.getString("savingsAccountNo");
+            final String ip = (rs.getString("ip") != null) ? rs.getString("ip") : "No IP assigned";
```
adamsaghy: Keep it NULL instead please!

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/SecurityConfig.java
```diff
@@ -227,6 +230,10 @@ public TenantAwareBasicAuthenticationFilter tenantAwareBasicAuthenticationFilter
         return filter;
     }
 
+    public GeolocationHeaderFilter geolocationHeaderFilter() {
+        return new GeolocationHeaderFilter(fineractProperties);
```
adamsaghy: Dont create it manually. Autowire this lazily instead!
```
@Autowired(required = false)`
 private GeolocationHeaderFilter geolocationHeaderFilter;
```

adamsaghy (changes requested):
Please kindly see my review!

Please make sure you are adding test cases which ensures the correct behaviour!

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
JohnAlva: In a typical multi-threaded server application (e.g., Spring Boot app), each HTTP request is handled by a thread (possibly from a thread pool). If you want to track the client IP address throughout the processing of that request (e.g., across multiple service layers), storing it in a ThreadLocal variable allows you to:

Access the IP anywhere in the call stack without passing it as a parameter.
Keep data isolated per request/thread.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
JohnAlva: We have used  the IP headers precedence based on these documents:
https://www.rfc-editor.org/rfc/rfc7239.txt
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Forwarded-For

adamsaghy: @JohnAlva These articules only talks about "X-Forwarded-For"... what about the rest of them???

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
JohnAlva: Converted to Spring bean (@Component) and conditional on enabled=false. No manual instantiation anymore.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
JohnAlva: Thanks for the heads-up! I removed the broad catch—now filterChain.doFilter() is only inside a try {…} finally to guarantee ClientIpHolder.clear(), and any exceptions propagate normally.

fineract-provider/src/main/java/org/apache/fineract/commands/service/AuditReadPlatformServiceImpl.java
JohnAlva: Removed the fallback value—ip now stays null when the column is null.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/SecurityConfig.java
JohnAlva: Replaced manual creation with lazy autowiring using @Autowired(required = false). The filter is now only added if the bean is present.

adamsaghy:
@JohnAlva LGTM, but testing is still missing!

JohnAlva:
@adamsaghy Do you have any model/reference/template for creating the test class?

JohnAlva:
@adamsaghy Do you have any comments or feedback?

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
```diff
@@ -0,0 +1,97 @@
...
+@Slf4j
+public class GeolocationHeaderFilter extends OncePerRequestFilter {
+
+    private final FineractProperties fineractProperties;
+
+    private static final String[] IP_HEADER_CANDIDATES = { "X-Forwarded-For", "Proxy-Client-IP", "WL-Proxy-Client-IP",
+            "HTTP_X_FORWARDED_FOR", "HTTP_X_FORWARDED", "HTTP_X_CLUSTER_CLIENT_IP", "HTTP_CLIENT_IP", "HTTP_FORWARDED_FOR",
+            "HTTP_FORWARDED", "HTTP_VIA", "REMOTE_ADDR" };
+
+    public static String getClientIpAddress(HttpServletRequest request) {
```
galovics: Shouldn't be static.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
```diff
@@ -0,0 +1,97 @@
...
+        for (String header : IP_HEADER_CANDIDATES) {
+            String ip = request.getHeader(header);
+            if (ip != null && ip.length() != 0 && !"unknown".equalsIgnoreCase(ip)) {
```
galovics: "unknown"? I'd rather not hardcode this unless we have a good reason to.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
```diff
@@ -0,0 +1,97 @@
...
+                log.debug("CALLER IP : {}", ip);
```
galovics: During development we run fineract in debug mode. This will spam the logs for sure. I'd change this to trace level.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
```diff
@@ -0,0 +1,97 @@
...
+                return ip;
+            }
+        }
+        log.debug("getRemoteAddr method : {}", request.getRemoteAddr());
```
galovics: Same, trace level.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
```diff
@@ -0,0 +1,97 @@
...
+        return request.getRemoteAddr();
+    }
+
+    @Override
+    protected void doFilterInternal(final HttpServletRequest request, final HttpServletResponse response, final FilterChain filterChain)
+            throws IOException, ServletException {
+        FineractProperties.FineractGeolocationProperties geolocationProperties = fineractProperties.getGeolocation();
```
galovics: I don't really get this check. Isn't the conditional on the bean handles this?

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
```diff
@@ -0,0 +1,97 @@
...
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.fineract.infrastructure.core.config.FineractProperties;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.stereotype.Component;
+import org.springframework.web.filter.OncePerRequestFilter;
+
+@Component
+@ConditionalOnProperty(prefix = "fineract.geolocation", name = "enabled", havingValue = "false", matchIfMissing = false)
```
galovics: I'd rather not use the string version of this but use the typed version. We have many "XYZCondition" classes.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+    }
+
+    private void handleClientIp(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain,
+            FineractProperties.FineractGeolocationProperties geolocationProperties) throws IOException, ServletException {
+
+        String clientIpAddress = getClientIpAddress(request);
+        if (StringUtils.isNotBlank(clientIpAddress)) {
+            log.debug("Found Client IP in header : {}", clientIpAddress);
```
galovics: Trace level.

fineract-provider/src/main/resources/application.properties
```diff
@@ -62,6 +62,7 @@ fineract.correlation.enabled=${FINERACT_LOGGING_HTTP_CORRELATION_ID_ENABLED:fals
 fineract.correlation.header-name=${FINERACT_LOGGING_HTTP_CORRELATION_ID_HEADER_NAME:X-Correlation-ID}
 
 fineract.job.stuck-retry-threshold=${FINERACT_JOB_STUCK_RETRY_THRESHOLD:5}
+fineract.geolocation.enabled=${FINERACT_GEOLOCATION_ENABLED:false}
```
galovics: This is not geolocation. I'm not even sure why we are calling it "geolocation" something.

It's simply client ip tracking, nothing else.

fineract-provider/src/test/resources/application-test.properties
```diff
@@ -20,6 +20,8 @@
 fineract.node-id=1
 
 fineract.security.basicauth.enabled=true
+fineract.geolocation.enabled=false
+fineract.security.two-factor.enabled=false
```
galovics: why the two-factor.enabled change?

fineract-provider/src/test/resources/application-test.properties
```diff
@@ -20,6 +20,8 @@
 fineract.node-id=1
 
 fineract.security.basicauth.enabled=true
+fineract.geolocation.enabled=false
```
galovics: Why is this needed? by default the application.properties falls back to a false value.

fineract-core/src/main/java/org/apache/fineract/commands/domain/CommandSource.java
```diff
@@ -162,6 +166,7 @@ public static CommandSource fullEntryFrom(final CommandWrapper wrapper, final Js
                 .transactionId(command.getTransactionId()) //
                 .creditBureauId(command.getCreditBureauId()) //
                 .organisationCreditBureauId(command.getOrganisationCreditBureauId()) //
+                .clientIp(ClientIpHolder.getClientIp()) //
```
galovics: Woah, this is really not good.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/ClientIpHolder.java
```diff
@@ -0,0 +1,39 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.fineract.infrastructure.core.filters;
+
+public final class ClientIpHolder {
```
galovics: I suggest not doing it this way. One more threadlocal is just an extra we don't need. Why can't we get the IP simply where we are saving the command?

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
galovics: big no. This must be explained one by one why a specific header is used. Please put the comments in the code so everybody understands the purpose.

galovics (changes requested):

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
Shelaslifter: The "static" keyword was removed and changed.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
Shelaslifter: The code was updated to avoid the "unknown" value and an `isEmpty()` was added to handle any empty value.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
Shelaslifter: The change was made from `log.debug` to `log.trace` to avoid log spam. However, Adamsaghy had previously indicated to use `log.debug`, but the level was ultimately adjusted to `trace`, as you requested.

fineract-core/src/main/java/org/apache/fineract/commands/domain/CommandSource.java
Shelaslifter: That's where it stores and saves the value of the IP. 
What should we do in that case?

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
Shelaslifter: done

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
Shelaslifter: done.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
Shelaslifter: Done, we have made the change as you requested.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
JohnAlva: You're right — since the bean is already conditionally loaded with @ConditionalOnProperty, the internal check was redundant. I've removed it.

fineract-provider/src/test/resources/application-test.properties
JohnAlva: For the two-factor.enabled and geoLocation.enabled properties:
These were explicitly added to the application-test.properties file because certain tests were failing when the values were not defined.
Although the default fallback is false, the tests seem to require these values to be explicitly set to avoid unexpected behavior.

Let me know if you'd prefer a different approach.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/ClientIpHolder.java
04Gomez: @galovics  
Hi, can you help us with an example of how it can be done?

fineract-provider/src/test/resources/application-test.properties
04Gomez: @galovics 
If we remove this, during the test execution it fails because it gets it as a null value, so, in addition, false is added to complete the test.

fineract-provider/src/main/resources/application.properties
04Gomez: @galovics 
The necessary modifications have already been made, leaving it as Client IP Tracking

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/ClientIpHolder.java
04Gomez: Hi, @galovics 
Done, we have eliminated that action leaving it with a single thread.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
JohnAlva: Done, comments added as requested.

JohnAlva:
Hi, @galovics, Do you have any comments to add?

04Gomez:
@galovics Hi, can you help us Review the commit?

Shelaslifter:
@galovics Hey, could you help us review the commit and let us know if you have any comments?

fineract-core/src/main/java/org/apache/fineract/commands/domain/CommandSource.java
```diff
@@ -224,4 +230,15 @@ public void updateForAudit(final CommandProcessingResult result) {
         this.subResourceExternalId = result.getSubResourceExternalId();
         this.loanExternalId = result.getLoanExternalId();
     }
+
+    private static String getClientIp() {
```
adamsaghy: Please move this logic outside of the JPA entity.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
adamsaghy: Thank you for the detailed comments, but please make sure you are split them into separate lines and please drop the `//`.
Example:
```
X-Forwarded-For: Standard header used by proxies,
Proxy-Client-IP: Used by some Apache proxies,
...
```

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/SecurityConfig.java
```diff
@@ -115,6 +116,9 @@ public class SecurityConfig {
     @Autowired
     private IdempotencyStoreHelper idempotencyStoreHelper;
 
+    @Autowired(required = false)
+    private GeolocationHeaderFilter geolocationHeaderFilter;
```
adamsaghy: Would `CallerIpTrackingFilter` naming make more sense here?

adamsaghy (changes requested):
Kindly review my comments!
Please make sure you are adding proper testing that ensures this new logic works as expected!

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/SecurityConfig.java
JohnAlva: Good point — renamed to CallerIpTrackingFilter for clarity.

fineract-core/src/main/java/org/apache/fineract/commands/domain/CommandSource.java
JohnAlva: I’ve moved the getClientIp() logic out of the JPA entity and placed it inside the JsonCommand class. Let me know if any other changes are needed

JohnAlva:
@adamsaghy Thanks for the reminder! I’ll make sure to add proper tests to verify the new logic behaves as expected.

JohnAlva:
@adamsaghy Just checking—do you need any input on the test scenarios or coverage, or are you all set?

adamsaghy:
> @adamsaghy Just checking—do you need any input on the test scenarios or coverage, or are you all set?

I am not sure what you mean... 🤔

JohnAlva:
Hi @adamsaghy,
Is there an existing example of the test you’re referring to? It would help us a lot. Thanks!

adamsaghy:
> Hi @adamsaghy, Is there an existing example of the test you’re referring to? It would help us a lot. Thanks!

I will be on leave for the rest of the week, but example for integration test: AuditIntegrationTest (i believe this is fetching the content of the `m_portfolio_command_source` table.

Also for E2E test, we might not yet created any AUDIT related one, but you can take a look at any of them and see how it works...

integration-tests/src/test/java/org/apache/fineract/integrationtests/IpTrackingIntegrationTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+        // when
+        String ip = auditsRecieved.get(0).get("ip").toString();
+
+        // then
+        if (!ip.isEmpty()) {
+            assertEquals(EXPECTED_LOCAL_IP, ip);
+        } else {
+            assertEquals("", ip);
+        }
```
galovics: How can there be an IF in the assertion? This seems to be wrong.

fineract-provider/src/main/resources/application.properties
```diff
@@ -62,6 +62,7 @@ fineract.correlation.enabled=${FINERACT_LOGGING_HTTP_CORRELATION_ID_ENABLED:fals
 fineract.correlation.header-name=${FINERACT_LOGGING_HTTP_CORRELATION_ID_HEADER_NAME:X-Correlation-ID}
 
 fineract.job.stuck-retry-threshold=${FINERACT_JOB_STUCK_RETRY_THRESHOLD:5}
+fineract.client-ip-tracking.enabled"=${FINERACT_CLIENT-IP-TRACKING_ENABLED:false}
```
galovics: Use underscores in the env var name.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/CallerIpTrackingFilter.java
```diff
@@ -0,0 +1,100 @@
...
+import org.apache.fineract.infrastructure.core.config.FineractProperties;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.stereotype.Component;
+import org.springframework.web.filter.OncePerRequestFilter;
+
+@Component
+@ConditionalOnProperty("fineract.client-ip-tracking.enabled")
+@RequiredArgsConstructor
+@Slf4j
+public class CallerIpTrackingFilter extends OncePerRequestFilter {
```
galovics, line 35: I don't get the whole filter approach.

If this is truly just extracting the IP info from the servlet request, why don't you simply create a utility method for it in a utility class and use it when needed? Why do you change the attributes of the request?

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java
```diff
@@ -50,6 +50,8 @@ public class FineractProperties {
 
     private FineractCorrelationProperties correlation;
 
+    private FineractGeolocationProperties geolocation;
```
galovics: You changed the property name, it's not geolocation anymore.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/api/JsonCommand.java
```diff
@@ -597,4 +599,15 @@ public void checkForUnsupportedParameters(final Type typeOfMap, final String jso
...
+    public String getClientIp() {
+        ServletRequestAttributes attrs = (ServletRequestAttributes) RequestContextHolder.getRequestAttributes();
+        String clientIp = "";
+        if (attrs != null) {
+            Object ipAttr = attrs.getRequest().getAttribute("IP");
+            if (ipAttr != null) {
+                clientIp = ipAttr.toString();
+            }
+        }
+        return clientIp;
```
galovics: I meant the utility method here.

galovics (changes requested):

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java
Shelaslifter: Done, I have changed the name to ipTracking.

fineract-provider/src/main/resources/application.properties
Shelaslifter: done.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/CallerIpTrackingFilter.java
Shelaslifter: It was implemented as a filter because it is considered an optional value. This allows the system user to decide whether or not they want to capture the IP address.

If it were implemented as a Utils class, the IP capture would be executed automatically at all times, without the possibility of skipping it.

Additionally, we believe it is not necessary to create an additional class solely for this purpose, as it would require more development effort for a functionality that can remain optional more easily through a filter.

integration-tests/src/test/java/org/apache/fineract/integrationtests/IpTrackingIntegrationTest.java
Shelaslifter: I have already made the correction by removing the if inside the assertion, as you suggested. Now the assertion is directly performed after verifying that the IP is not empty.

Shelaslifter:
@galovics  Hi, could you please help us review the commit and share any feedback or comments you may have? Thanks.

Shelaslifter:
@galovics  Hey, would you mind reviewing the commit and providing any feedback or thoughts you might have? Appreciate it.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/api/JsonCommand.java
Shelaslifter: It was implemented as a filter because it is considered an optional value. This allows the system user to decide whether or not they want to capture the IP address.

If it were implemented as a Utils class, the IP capture would be executed automatically at all times, without the possibility of skipping it.

Additionally, we believe it is not necessary to create an additional class solely for this purpose, as it would require more development effort for a functionality that can remain optional more easily through a filter.

Shelaslifter: As mentioned, the filter keeps IP tracking optional. Using a utility class would make capture automatic, with no option to skip. Creating a separate class for this would add unnecessary overhead, as the filter already fulfills the required functionality.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/filters/GeolocationHeaderFilter.java
Shelaslifter: Apologies for the delayed response, the change has already been made.

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/api/JsonCommand.java
adamsaghy: @Shelaslifter Fetching data from `RequestContextHolder.getRequestAttributes()` has nothing to do with JsonCommand class. Please Move this while logic into a utility class as @galovics recommended.

We should not mix business logics together...

fineract-provider/src/test/resources/application-test.properties
adamsaghy: @JohnAlva Seems there is a bug here as we have `fineract.security.2fa.enabled=false` but not have `fineract.security.two-factor.enabled`, but in Fineract properties the field name is `twoFactor`

adamsaghy: I think we should raise an another PR to resolve this.

integration-tests/src/test/java/org/apache/fineract/integrationtests/IpTrackingIntegrationTest.java
```diff
@@ -0,0 +1,74 @@
...
+import io.restassured.specification.ResponseSpecification;
+import java.util.HashMap;
+import java.util.List;
+import org.apache.fineract.integrationtests.common.AuditHelper;
+import org.apache.fineract.integrationtests.common.ClientHelper;
+import org.apache.fineract.integrationtests.common.Utils;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class IpTrackingIntegrationTest {
```
adamsaghy, line 37: Would you mind to write a negative test case, where functionality is not enabled, so there will be no IP fetched and stored?

Also Please make sure, all of the IP header options are tested one by one... ;)

fineract-provider/src/test/resources/application-test.properties
adamsaghy: @JohnAlva It's provided via "2fa", and so far those tests were running just fine.. PLease remove the `fineract.security.two-factor.enabled=false`

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/api/JsonCommand.java
Shelaslifter: @adamsaghy hi, the IpAddressUtils.java class was created, and the JsonCommand was removed from the class.

integration-tests/src/test/java/org/apache/fineract/integrationtests/IpTrackingIntegrationTest.java
Shelaslifter: The one for the headers was already added, and the functionality was moved to IpAddressUtils.java.

adamsaghy (approved):
LGTM

PR #4861, opened 2025-07-15
FINERACT-2181: Moneyhelper multitenant configuration fix

budaidev:
## Description

There were possible rounding issue during the multitenant configuration, when different tenants use different rounding modes. This fix targets to fix that

## Checklist

### DIFF
--- a/fineract-core/src/main/java/org/apache/fineract/organisation/monetary/domain/MoneyHelper.java
+++ b/fineract-core/src/main/java/org/apache/fineract/organisation/monetary/domain/MoneyHelper.java
@@ -18,54 +18,176 @@
  */
 package org.apache.fineract.organisation.monetary.domain;
 
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
-import jakarta.annotation.PostConstruct;
 import java.math.MathContext;
 import java.math.RoundingMode;
+import java.util.concurrent.ConcurrentHashMap;
 import lombok.extern.slf4j.Slf4j;
-import org.apache.fineract.infrastructure.configuration.domain.ConfigurationDomainService;
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.stereotype.Component;
+import org.apache.fineract.infrastructure.core.domain.FineractPlatformTenant;
+import org.apache.fineract.infrastructure.core.service.ThreadLocalContextUtil;
 
+/**
+ * Pure utility class for monetary calculations and rounding operations. This class does not depend on Spring components
+ * or configuration services. All rounding modes are initialized at startup and cached per tenant.
+ */
 @Slf4j
-@Component
-public class MoneyHelper {
+public final class MoneyHelper {
 
-    private static RoundingMode roundingMode = null;
-    private static MathContext mathContext;
     public static final int PRECISION = 19;
+    private static final RoundingMode DEFAULT_ROUNDING_MODE = RoundingMode.HALF_EVEN;
 
-    private static ConfigurationDomainService staticConfigurationDomainService;
+    private static final ConcurrentHashMap<String, RoundingMode> roundingModeCache = new ConcurrentHashMap<>();
+    private static final ConcurrentHashMap<String, MathContext> mathContextCache = new ConcurrentHashMap<>();
 
-    @Autowired
-    private ConfigurationDomainService configurationDomainService;
-
-    // This is a hack, but fixing this is not trivial, because some @Entity
-    // domain classes use this helper
-    @PostConstruct
-    @SuppressFBWarnings("ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD")
-    public void initialize() {
-        staticConfigurationDomainService = configurationDomainService;
+    // Private constructor to prevent instantiation
+    private MoneyHelper() {
+        throw new UnsupportedOperationException("MoneyHelper is a utility class and cannot be instantiated");
     }
 
+    /**
+     * Initialize rounding mode for a specific tenant. This method should be called during application startup for each
+     * tenant.
+     *
+     * @param tenantIdentifier
+     *            the tenant identifier
+     * @param roundingModeValue
+     *            the rounding mode value (0-6)
+     */
+    public static void initializeTenantRoundingMode(String tenantIdentifier, int roundingModeValue) {
+        if (tenantIdentifier == null) {
+            throw new IllegalArgumentException("Tenant identifier cannot be null");
+        }
+
+        RoundingMode roundingMode = validateAndConvertRoundingMode(roundingModeValue);
+        roundingModeCache.put(tenantIdentifier, roundingMode);
+        // Clear math context cache to force recreation with new rounding mode
+        mathContextCache.remove(tenantIdentifier);
+
+        log.info("Initialized rounding mode for tenant {}: {}", tenantIdentifier, roundingMode.name());
+    }
+
+    /**
+     * Get the rounding mode for the current tenant context.
+     *
+     * @return the tenant-specific rounding mode
+     * @throws IllegalStateException
+     *             if no tenant context is available or tenant is not initialized
+     */
     public static RoundingMode getRoundingMode() {
+        String tenantId = getTenantIdentifier();
+        RoundingMode roundingMode = roundingModeCache.get(tenantId);
+
         if (roundingMode == null) {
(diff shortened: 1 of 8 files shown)


fineract-core/src/main/java/org/apache/fineract/organisation/monetary/domain/MoneyHelper.java
```diff
@@ -49,23 +52,42 @@ public void initialize() {
...
+        mathContextCache.remove(tenantId); // Force recreation with new rounding mode
+    }
+
+    private static String getTenantIdentifier() {
+        FineractPlatformTenant tenant = ThreadLocalContextUtil.getTenant();
+        if (tenant != null) {
+            return tenant.getTenantIdentifier();
+        }
+        // Fallback for cases where tenant context is not available
+        return "default";
```
galovics: I'd really not hardcode something like this. You can't be sure that there will be a default tenant in the system.

Throw an exception if we think this is an invalid case rather than hiding the problem.

galovics (changes requested):

fineract-core/src/main/java/org/apache/fineract/organisation/monetary/domain/MoneyHelper.java
budaidev: fixed

fineract-core/src/main/java/org/apache/fineract/organisation/monetary/domain/MoneyHelper.java
```diff
@@ -22,17 +22,20 @@
...
 @Component
 public class MoneyHelper {
 
-    private static RoundingMode roundingMode = null;
-    private static MathContext mathContext;
+    private static final ConcurrentHashMap<String, RoundingMode> roundingModeCache = new ConcurrentHashMap<>();
+    private static final ConcurrentHashMap<String, MathContext> mathContextCache = new ConcurrentHashMap<>();
     public static final int PRECISION = 19;
 
     private static ConfigurationDomainService staticConfigurationDomainService;
```
adamsaghy: I was wondering whether we could remove the `ConfigurationDomainService` all together and remove `@Component` also.

This could be a utility class which is instantiated with values during the startup:
- Not managed bean
- No any injections
- Value holder only
- Once instance got started, for each tenants automatically we set the value

What do you think?

galovics: But if we setup the values on startup, the global configuration gets kinda meaningless, doesn't it? Its whole purpose is to be able to dynamically override values in fineract without a redeployment/restart.

adamsaghy: rounding-mode is a trap door configuration, you are not allowed to change once it was set.

budaidev: Agreed, I think a validation could be useful for that use case. Changing the rounding would cause troubles for previous calculation consistency

galovics: good, let's proceed with it then as Adam suggested. Although please put a comment there so anybody coming after us knows why it's only fetched once at startup.

adamsaghy:
@budaidev 
```
/home/runner/work/fineract/fineract/fineract-provider/src/main/java/org/apache/fineract/infrastructure/configuration/api/InternalConfigurationsApiResource.java:34: error: cannot find symbol
import org.apache.fineract.organisation.monetary.service.MoneyHelperInitializationService;
                                                        ^
  symbol:   class MoneyHelperInitializationService
  location: package org.apache.fineract.organisation.monetary.service
/home/runner/work/fineract/fineract/fineract-provider/src/main/java/org/apache/fineract/infrastructure/configuration/api/InternalConfigurationsApiResource.java:47: error: cannot find symbol
    private final MoneyHelperInitializationService moneyHelperInitializationService;
                  ^
  symbol:   class MoneyHelperInitializationService
  location: class InternalConfigurationsApiResource
/home/runner/work/fineract/fineract/fineract-provider/src/main/java/org/apache/fineract/infrastructure/configuration/api/InternalConfigurationsApiResource.java:42: error: cannot find symbol
@RequiredArgsConstructor
^
  symbol:   class MoneyHelperInitializationService
  location: class InternalConfigurationsApiResource
3 errors
2 warnings

File: fineract-core/src/main/java/org/apache/fineract/organisation/monetary/domain/MoneyHelper.java [hunk 3]
```diff
@@ -18,54 +18,176 @@
...
+ */
 @Slf4j
-@Component
-public class MoneyHelper {
+public final class MoneyHelper {
 
-    private static RoundingMode roundingMode = null;
-    private static MathContext mathContext;
     public static final int PRECISION = 19;
+    private static final RoundingMode DEFAULT_ROUNDING_MODE = RoundingMode.HALF_EVEN;
```
adamsaghy: I am not a big fan of hardcoded "default" values... I would prefer throwing error if it was not set, so it will be blocker if someone tries something nasty!

fineract-provider/src/main/java/org/apache/fineract/infrastructure/configuration/service/MoneyHelperInitializationService.java
```diff
@@ -0,0 +1,115 @@
...
+    private int getRoundingModeFromConfiguration() {
+        try {
+            GlobalConfigurationProperty roundingModeProperty = globalConfigurationRepository
+                    .findOneByNameWithNotFoundDetection(GlobalConfigurationConstants.ROUNDING_MODE);
+
+            if (roundingModeProperty != null && roundingModeProperty.getValue() != null) {
+                return roundingModeProperty.getValue().intValue();
+            }
+        } catch (Exception e) {
+            log.warn("Failed to read rounding mode from configuration: {}", e.getMessage());
```
adamsaghy: I am not a big fan of swallowing this error...

fineract-provider/src/main/java/org/apache/fineract/infrastructure/configuration/service/MoneyHelperInitializationService.java
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        // Default to HALF_UP if configuration is not available
+        int defaultRoundingMode = RoundingMode.HALF_UP.ordinal();
```
adamsaghy: I dont think we should go with default, fallback value, but even if we would, it is inconsistent to use HALF-UP here and HALF-EVEN in the MoneyHelper... :/

If we are unable to fetch the details of it, then let it fails!

fineract-provider/src/main/java/org/apache/fineract/infrastructure/configuration/service/MoneyHelperStartupInitializationService.java
```diff
@@ -0,0 +1,96 @@
...
+            log.info("MoneyHelper initialization completed - Success: {}, Failures: {}, Total: {}", successCount, failureCount,
+                    tenants.size());
+
+            if (failureCount > 0) {
+                log.warn("Some tenants failed MoneyHelper initialization. "
+                        + "These tenants may experience issues with rounding mode configuration.");
+            }
+
+        } catch (Exception e) {
+            log.error("Critical error during MoneyHelper initialization for all tenants", e);
```
adamsaghy: Dont swallow the issue...let it fails!

adamsaghy: Dont swallow the issue...let it fails!

fineract-provider/src/main/java/org/apache/fineract/infrastructure/configuration/service/MoneyHelperStartupInitializationService.java
```diff
@@ -0,0 +1,96 @@
...
+                        continue;
+                    }
+
+                    // Initialize MoneyHelper for this tenant
+                    moneyHelperInitializationService.initializeTenantRoundingMode(tenant);
+                    successCount++;
+
+                } catch (Exception e) {
+                    failureCount++;
+                    log.error("Failed to initialize MoneyHelper for tenant '{}'", tenant.getTenantIdentifier(), e);
```
adamsaghy: Dont swallow the issue...let it fails!

PR #4876, opened 2025-07-21
FINERACT-2312: Adjustment to savings account products by adding a new accounting account interest receivables account.

04Gomez:
## Description

Adjustment to savings account products by adding a new accounting account interest receivables account.

Ignore if these details are present on the associated [Apache Fineract JIRA ticket](https://github.com/apache/fineract/pull/1284).

## Checklist

Please make sure these boxes are checked before submitting your pull request - thanks!

- [ ] Write the commit message as per https://github.com/apache/fineract/#pull-requests
- [ ] Acknowledge that we will not review PRs that are not passing the build _("green")_ - it is your responsibility to get a proposed PR to pass the build, not primarily the project's maintainers.
- [ ] Create/update unit or integration tests for verifying the changes made.
- [ ] Follow coding conventions at https://cwiki.apache.org/confluence/display/FINERACT/Coding+Conventions.
- [ ] Add required Swagger annotation and update API documentation at fineract-provider/src/main/resources/static/legacy-docs/apiLive.htm with details of any API changes
- [ ] Submission is not a "code dump".  (Large changes can be made "in repository" via a branch.  Ask on the developer mailing list for guidance, if required.)

FYI our guidelines for code reviews are at https://cwiki.apache.org/confluence/display/FINERACT/Code+Review+Guide.

PR #4886, opened 2025-07-24
FINERACT-2312: Post interest with adjustments for savings accounts

JohnAlva:
## Description

Post interest with adjustments for savings accounts

-Change the posting when the account is overdrawn by noting the interest receivables accounting account without affecting products that do not have this accounting account.
-The interest posting will show if the account has a different balance during the monthly period. That is, if overdrawn transactions were made and the account was closed with a positive balance, it will show the overdraft interest and the post-interest.

## Checklist

### DIFF
--- a/fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountData.java
+++ b/fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountData.java
@@ -29,6 +29,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Set;
 import lombok.Getter;
+import lombok.Setter;
 import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
 import org.apache.fineract.infrastructure.core.data.EnumOptionData;
@@ -47,6 +48,7 @@ import org.apache.fineract.portfolio.tax.data.TaxGroupData;
 /**
  * Immutable data object representing a savings account.
  */
+@Setter
 @Getter
 @JsonLocalDateArrayFormat
 public final class SavingsAccountData implements Serializable {
@@ -143,6 +145,13 @@ public final class SavingsAccountData implements Serializable {
     private transient Set<Long> existingReversedTransactionIds = new HashSet<>();
     private transient Long glAccountIdForSavingsControl;
     private transient Long glAccountIdForInterestOnSavings;
+
+    private Long glAccountIdForInterestPayable;
+    private Long glAccountIdForOverdraftPorfolio;
+    private Long glAccountIdForInterestReceivable;
+
+    private BigDecimal interestPosting;
+    private BigDecimal overdraftPosting;
 
     public static SavingsAccountData importInstanceIndividual(Long clientId, Long productId, Long fieldOfficerId, LocalDate submittedOnDate,
             BigDecimal nominalAnnualInterestRate, EnumOptionData interestCompoundingPeriodTypeEnum,
@@ -964,4 +973,5 @@ public final class SavingsAccountData implements Serializable {
     public boolean isIsDormancyTrackingActive() {
         return this.isDormancyTrackingActive;
     }
+
 }
--- a/fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountTransactionData.java
+++ b/fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountTransactionData.java
@@ -105,6 +105,10 @@ public final class SavingsAccountTransactionData implements Serializable {
     private BigDecimal overdraftAmount;
     private transient Long modifiedId;
     private transient String refNo;
+    private Boolean isOverdraft;
+
+    private Long accountCredit;
+    private Long accountDebit;
 
     private SavingsAccountTransactionData(final Long id, final SavingsAccountTransactionEnumData transactionType,
             final PaymentDetailData paymentDetailData, final Long savingsId, final String savingsAccountNo, final LocalDate transactionDate,
@@ -112,7 +116,7 @@ public final class SavingsAccountTransactionData implements Serializable {
             final boolean reversed, final AccountTransferData transfer, final Collection<PaymentTypeData> paymentTypeOptions,
             final LocalDate submittedOnDate, final boolean interestedPostedAsOn, final String submittedByUsername, final String note,
             final Boolean isReversal, final Long originalTransactionId, boolean isManualTransaction, final Boolean lienTransaction,
-            final Long releaseTransactionId, final String reasonForBlock) {
+            final Long releaseTransactionId, final String reasonForBlock, final Boolean isOverdraft) {
         this.id = id;
         this.transactionType = transactionType;
         TransactionEntryType entryType = null;
@@ -146,6 +150,7 @@ public final class SavingsAccountTransactionData implements Serializable {
         this.lienTransaction = lienTransaction;
         this.releaseTransactionId = releaseTransactionId;
         this.reasonForBlock = reasonForBlock;
+        this.isOverdraft = isOverdraft;
     }
 
     private static SavingsAccountTransactionData createData(final Long id, final SavingsAccountTransactionEnumData transactionType,
@@ -156,7 +161,7 @@ public final class SavingsAccountTransactionData implements Serializable {
             final Boolean lienTransaction) {
         return new SavingsAccountTransactionData(id, transactionType, paymentDetailData, accountId, accountNo, date, currency, amount,
                 outstandingChargeAmount, runningBalance, reversed, transfer, paymentTypeOptions, submittedOnDate, interestedPostedAsOn,
-                submittedByUsername, note, null, null, false, lienTransaction, null, null);
+                submittedByUsername, note, null, null, false, lienTransaction, null, null, false);
     }
 
     public static SavingsAccountTransactionData create(final Long id, final SavingsAccountTransactionEnumData transactionType,
@@ -167,7 +172,8 @@ public final class SavingsAccountTransactionData implements Serializable {
             final Boolean lienTransaction, final Long releaseTransactionId, final String reasonForBlock) {
         return new SavingsAccountTransactionData(id, transactionType, paymentDetailData, savingsId, savingsAccountNo, date, currency,
                 amount, outstandingChargeAmount, runningBalance, reversed, transfer, null, submittedOnDate, interestedPostedAsOn,
-                submittedByUsername, note, isReversal, originalTransactionId, false, lienTransaction, releaseTransactionId, reasonForBlock);
+                submittedByUsername, note, isReversal, originalTransactionId, false, lienTransaction, releaseTransactionId, reasonForBlock,
+                false);
     }
 
     public static SavingsAccountTransactionData create(final Long id, final SavingsAccountTransactionEnumData transactionType,
@@ -235,10 +241,10 @@ public final class SavingsAccountTransactionData implements Serializable {
     private static SavingsAccountTransactionData createImport(final SavingsAccountTransactionEnumData transactionType,
             final PaymentDetailData paymentDetailData, final Long savingsAccountId, final String accountNumber,
             final LocalDate transactionDate, final BigDecimal transactionAmount, final boolean reversed, final LocalDate submittedOnDate,
-            boolean isManualTransaction, final Boolean lienTransaction) {
+            boolean isManualTransaction, final Boolean lienTransaction, final Boolean isOverdraft) {
         SavingsAccountTransactionData data = new SavingsAccountTransactionData(null, transactionType, paymentDetailData, savingsAccountId,
                 accountNumber, transactionDate, null, transactionAmount, null, null, reversed, null, null, submittedOnDate, false, null,
-                null, null, null, isManualTransaction, lienTransaction, null, null);
+                null, null, null, isManualTransaction, lienTransaction, null, null, isOverdraft);
         // duplicated import fields
         data.savingsAccountId = savingsAccountId;
         data.accountNumber = accountNumber;
@@ -251,14 +257,14 @@ public final class SavingsAccountTransactionData implements Serializable {
         return createImport(accountTransaction.getTransactionType(), accountTransaction.getPaymentDetailData(),
                 accountTransaction.getSavingsAccountId(), null, accountTransaction.getTransactionDate(), accountTransaction.getAmount(),
                 accountTransaction.isReversed(), accountTransaction.getSubmittedOnDate(), accountTransaction.isManualTransaction(),
-                accountTransaction.getLienTransaction());
+                accountTransaction.getLienTransaction(), false);
     }
 
     public static SavingsAccountTransactionData importInstance(BigDecimal transactionAmount, LocalDate transactionDate, Long paymentTypeId,
             String accountNumber, String checkNumber, String routingCode, String receiptNumber, String bankNumber, String note,
             Long savingsAccountId, SavingsAccountTransactionEnumData transactionType, Integer rowIndex, String locale, String dateFormat) {
         SavingsAccountTransactionData data = createImport(transactionType, null, savingsAccountId, accountNumber, transactionDate,
-                transactionAmount, false, transactionDate, false, false);
+                transactionAmount, false, transactionDate, false, false, false);
         data.rowIndex = rowIndex;
         data.paymentTypeId = paymentTypeId;
         data.checkNumber = checkNumber;
@@ -272,10 +278,11 @@ public final class SavingsAccountTransactionData implements Serializable {
     }
 
     private static SavingsAccountTransactionData createImport(SavingsAccountTransactionEnumData transactionType, Long savingsAccountId,
-            LocalDate transactionDate, BigDecimal transactionAmount, final LocalDate submittedOnDate, boolean isManualTransaction) {
+            LocalDate transactionDate, BigDecimal transactionAmount, final LocalDate submittedOnDate, boolean isManualTransaction,
+            Boolean isOverdraft) {
         // import transaction
         return createImport(transactionType, null, savingsAccountId, null, transactionDate, transactionAmount, false, submittedOnDate,
-                isManualTransaction, false);
+                isManualTransaction, false, isOverdraft);
     }
 
     public static SavingsAccountTransactionData interestPosting(final SavingsAccountData savingsAccount, final LocalDate date,
@@ -285,17 +292,28 @@ public final class SavingsAccountTransactionData implements Serializable {
         SavingsAccountTransactionEnumData transactionType = new SavingsAccountTransactionEnumData(
                 savingsAccountTransactionType.getValue().longValue(), savingsAccountTransactionType.getCode(),
                 savingsAccountTransactionType.getValue().toString());
-        return createImport(transactionType, savingsAccount.getId(), date, amount.getAmount(), submittedOnDate, isManualTransaction);
+        return createImport(transactionType, savingsAccount.getId(), date, amount.getAmount(), submittedOnDate, isManualTransaction, false);
+    }
+
+    public static SavingsAccountTransactionData accrual(final SavingsAccountData savingsAccount, final LocalDate date, final Money amount,
+            final boolean isManualTransaction) {
+        final LocalDate submittedOnDate = DateUtils.getBusinessLocalDate();
+        final SavingsAccountTransactionType savingsAccountTransactionType = SavingsAccountTransactionType.ACCRUAL;
+        SavingsAccountTransactionEnumData transactionType = new SavingsAccountTransactionEnumData(
+                savingsAccountTransactionType.getValue().longValue(), savingsAccountTransactionType.getCode(),
+                savingsAccountTransactionType.getValue().toString());
+        return createImport(transactionType, savingsAccount.getId(), date, amount.getAmount(), submittedOnDate, isManualTransaction, false);
     }
 
     public static SavingsAccountTransactionData overdraftInterest(final SavingsAccountData savingsAccount, final LocalDate date,
-            final Money amount, final boolean isManualTransaction) {
+            final Money amount, final boolean isManualTransaction, final Boolean isOverdraft) {
         final LocalDate submittedOnDate = DateUtils.getBusinessLocalDate();
         final SavingsAccountTransactionType savingsAccountTransactionType = SavingsAccountTransactionType.OVERDRAFT_INTEREST;
         SavingsAccountTransactionEnumData transactionType = new SavingsAccountTransactionEnumData(
                 savingsAccountTransactionType.getValue().longValue(), savingsAccountTransactionType.getCode(),
                 savingsAccountTransactionType.getValue().toString());
-        return createImport(transactionType, savingsAccount.getId(), date, amount.getAmount(), submittedOnDate, isManualTransaction);
+        return createImport(transactionType, savingsAccount.getId(), date, amount.getAmount(), submittedOnDate, isManualTransaction,
+                isOverdraft);
     }
 
     public static SavingsAccountTransactionData withHoldTax(final SavingsAccountData savingsAccount, final LocalDate date,
@@ -306,7 +324,7 @@ public final class SavingsAccountTransactionData implements Serializable {
                 savingsAccountTransactionType.getValue().longValue(), savingsAccountTransactionType.getCode(),
                 savingsAccountTransactionType.getValue().toString());
         SavingsAccountTransactionData accountTransaction = createImport(transactionType, savingsAccount.getId(), date, amount.getAmount(),
-                submittedOnDate, false);
+                submittedOnDate, false, false);
         accountTransaction.addTaxDetails(taxDetails);
         return accountTransaction;
     }
@@ -387,10 +405,11 @@ public final class SavingsAccountTransactionData implements Serializable {
         final MonetaryCurrency currency = openingBalance.getCurrency();
         Money endOfDayBalance = openingBalance.copy();
         if (isDeposit() || isDividendPayoutAndNotReversed()) {
-            endOfDayBalance = openingBalance.plus(getAmount());
+            endOfDayBalance = Money.of(currency, this.runningBalance);
         } else if (isWithdrawal() || isChargeTransactionAndNotReversed()) {
-
-            if (openingBalance.isGreaterThanZero()) {
+            if (isWithdrawal()) {
+                endOfDayBalance = Money.of(currency, this.runningBalance);
+            } else if (openingBalance.isGreaterThanZero()) {
                 endOfDayBalance = openingBalance.minus(getAmount());
             } else {
                 endOfDayBalance = Money.of(currency, this.runningBalance);
@@ -398,6 +417,14 @@ public final class SavingsAccountTransactionData implements Serializable {
         }
 
         return EndOfDayBalance.from(getTransactionDate(), openingBalance, endOfDayBalance, this.balanceNumberOfDays);
+    }
+
+    public EndOfDayBalance toEndOfDayBalanceDates(final Money openingBalance, LocalDateInterval date) {
+        final MonetaryCurrency currency = openingBalance.getCurrency();
+        Money endOfDayBalance = Money.of(currency, this.runningBalance);
+
+        return EndOfDayBalance.from(getTransactionDate(), openingBalance, endOfDayBalance,
+                this.balanceNumberOfDays != null ? this.balanceNumberOfDays : date.endDate().getDayOfMonth());
     }
 
     public boolean isChargeTransactionAndNotReversed() {
@@ -427,7 +454,9 @@ public final class SavingsAccountTransactionData implements Serializable {
             if (isDeposit() || isDividendPayoutAndNotReversed()) {
                 endOfDayBalance = endOfDayBalance.plus(getAmount());
             } else if (isWithdrawal() || isChargeTransactionAndNotReversed()) {
-                if (endOfDayBalance.isGreaterThanZero() || isAllowOverdraft) {
+                if (endOfDayBalance.isLessThanZero() && isAllowOverdraft) {
+                    endOfDayBalance = Money.of(currency, this.runningBalance);
+                } else if (endOfDayBalance.isGreaterThanZero() || isAllowOverdraft) {
                     endOfDayBalance = endOfDayBalance.minus(getAmount());
                 } else {
                     endOfDayBalance = Money.of(currency, this.runningBalance);
@@ -657,5 +686,13 @@ public final class SavingsAccountTransactionData implements Serializable {
 
     public TransactionEntryType getEntryType() {
         return entryType;
+    }
+
+    public void setAccountCredit(Long accountCredit) {
+        this.accountCredit = accountCredit;
+    }
+
+    public void setAccountDebit(Long accountDebit) {
+        this.accountDebit = accountDebit;
     }
 }
--- a/fineract-core/src/main/java/org/apache/fineract/portfolio/savings/domain/interest/CompoundInterestHelper.java
+++ b/fineract-core/src/main/java/org/apache/fineract/portfolio/savings/domain/interest/CompoundInterestHelper.java
@@ -47,8 +47,13 @@ public class CompoundInterestHelper {
         // total interest earned in previous periods but not yet recognised
         BigDecimal compoundedInterest = BigDecimal.ZERO;
         BigDecimal unCompoundedInterest = BigDecimal.ZERO;
+        LocalDate endDay = DateUtils.getBusinessLocalDate();
         final CompoundInterestValues compoundInterestValues = new CompoundInterestValues(compoundedInterest, unCompoundedInterest);
         for (final PostingPeriod postingPeriod : allPeriods) {
+
+            if (postingPeriod.dateOfPostingTransaction().getMonth() != endDay.getMonth()) {
+                compoundInterestValues.setCompoundedInterest(interestEarned.getAmount());
+            }
 
             final BigDecimal interestEarnedThisPeriod = postingPeriod.calculateInterest(compoundInterestValues);
 
@@ -61,8 +66,9 @@ public class CompoundInterestHelper {
             // calculation.
             if (!(postingPeriod.isInterestTransfered() || !interestTransferEnabled
                     || (lockUntil != null && !DateUtils.isAfter(postingPeriod.dateOfPostingTransaction(), lockUntil)))) {
-                compoundInterestValues.setcompoundedInterest(BigDecimal.ZERO);
+                compoundInterestValues.setCompoundedInterest(BigDecimal.ZERO);
             }
+            endDay = postingPeriod.dateOfPostingTransaction();
         }
 
         return interestEarned;
--- a/fineract-core/src/main/java/org/apache/fineract/portfolio/savings/domain/interest/PostingPeriod.java
+++ b/fineract-core/src/main/java/org/apache/fineract/portfolio/savings/domain/interest/PostingPeriod.java
@@ -26,6 +26,8 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.TreeSet;
+import lombok.Getter;
+import lombok.Setter;
 import org.apache.fineract.infrastructure.core.domain.LocalDateInterval;
 import org.apache.fineract.infrastructure.core.service.DateUtils;
 import org.apache.fineract.organisation.monetary.domain.MonetaryCurrency;
@@ -34,6 +36,8 @@ import org.apache.fineract.portfolio.savings.SavingsCompoundingInterestPeriodTyp
 import org.apache.fineract.portfolio.savings.SavingsInterestCalculationType;
 import org.apache.fineract.portfolio.savings.data.SavingsAccountTransactionData;
 
+@Setter
+@Getter
 public final class PostingPeriod {
 
     private final LocalDateInterval periodInterval;
@@ -63,6 +67,8 @@ public final class PostingPeriod {
     private Money minOverdraftForInterestCalculation;
 
     private Integer financialYearBeginningMonth;
+
+    private boolean overdraftInterest = false;
 
     public void setOverdraftInterestRateAsFraction(BigDecimal overdraftInterestRateAsFraction) {
         this.overdraftInterestRateAsFraction = overdraftInterestRateAsFraction;
@@ -301,7 +307,7 @@ public final class PostingPeriod {
 
             if (compoundingPeriodEndDate.equals(compoundingPeriod.getPeriodInterval().endDate())) {
                 BigDecimal interestCompounded = compoundInterestValues.getcompoundedInterest().add(unCompoundedInterest);
-                compoundInterestValues.setcompoundedInterest(interestCompounded);
+                compoundInterestValues.setCompoundedInterest(interestCompounded);
                 compoundInterestValues.setZeroForInterestToBeUncompounded();
             }
             interestEarned = interestEarned.add(interestUnrounded);
@@ -314,7 +320,7 @@ public final class PostingPeriod {
     }
 
     public Money getInterestEarned() {
-        return this.interestEarnedRounded;
+        return this.interestEarnedRounded != null ? this.interestEarnedRounded : Money.zero(this.currency);
     }
 
     private static List<CompoundingPeriod> compoundingPeriodsInPostingPeriod(final LocalDateInterval postingPeriodInterval,
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
@@ -33,6 +33,7 @@ import java.util.stream.Collectors;
 import lombok.RequiredArgsConstructor;
 import org.apache.fineract.infrastructure.core.domain.LocalDateInterval;
 import org.apache.fineract.infrastructure.core.service.DateUtils;
+import org.apache.fineract.infrastructure.core.service.MathUtil;
 import org.apache.fineract.organisation.monetary.domain.MonetaryCurrency;
 import org.apache.fineract.organisation.monetary.domain.Money;
 import org.apache.fineract.portfolio.savings.DepositAccountType;
@@ -81,11 +82,17 @@ public class SavingsAccountInterestPostingServiceImpl implements SavingsAccountI
         for (final PostingPeriod interestPostingPeriod : postingPeriods) {
             final LocalDate interestPostingTransactionDate = interestPostingPeriod.dateOfPostingTransaction();
             final Money interestEarnedToBePostedForPeriod = interestPostingPeriod.getInterestEarned();
+            final Boolean isOverdraft = interestPostingPeriod.isOverdraftInterest();
 
             if (!DateUtils.isAfter(interestPostingTransactionDate, interestPostingUpToDate)) {
                 interestPostedToDate = interestPostedToDate.plus(interestEarnedToBePostedForPeriod);
-                final SavingsAccountTransactionData postingTransaction = findInterestPostingTransactionFor(interestPostingTransactionDate,
-                        savingsAccountData);
+                SavingsAccountTransactionData postingTransaction = null;
+                if (this.depositAccountType(savingsAccountData).isSavingsDeposit() && savingsAccountData.isAllowOverdraft()) {
+                    postingTransaction = findInterestPostingTransactionForInterest(interestPostingTransactionDate, savingsAccountData,
+                            isOverdraft);
+                } else {
+                    postingTransaction = findInterestPostingTransactionFor(interestPostingTransactionDate, savingsAccountData);
+                }
 
                 if (postingTransaction == null) {
                     SavingsAccountTransactionData newPostingTransaction;
@@ -95,7 +102,7 @@ public class SavingsAccountInterestPostingServiceImpl implements SavingsAccountI
                     } else {
                         newPostingTransaction = SavingsAccountTransactionData.overdraftInterest(savingsAccountData,
                                 interestPostingTransactionDate, interestEarnedToBePostedForPeriod.negated(),
-                                interestPostingPeriod.isUserPosting());
+                                interestPostingPeriod.isUserPosting(), isOverdraft);
                     }
 
                     savingsAccountData.updateTransactions(newPostingTransaction);
@@ -131,7 +138,7 @@ public class SavingsAccountInterestPostingServiceImpl implements SavingsAccountI
                         } else {
                             newPostingTransaction = SavingsAccountTransactionData.overdraftInterest(savingsAccountData,
                                     interestPostingTransactionDate, interestEarnedToBePostedForPeriod.negated(),
-                                    interestPostingPeriod.isUserPosting());
+                                    interestPostingPeriod.isUserPosting(), isOverdraft);
                         }
 
                         savingsAccountData.updateTransactions(newPostingTransaction);
@@ -188,6 +195,36 @@ public class SavingsAccountInterestPostingServiceImpl implements SavingsAccountI
             }
         }
         return transaction;
+    }
+
+    private Money appendPostingPeriodIfAny(final LocalDateInterval periodInterval, Money periodStartingBalance,
+            final List<SavingsAccountTransactionData> txs, final MonetaryCurrency monetaryCurrency,
+            final SavingsCompoundingInterestPeriodType compoundingPeriodType, final SavingsInterestCalculationType interestCalculationType,
+            final BigDecimal interestRateAsFraction, final int daysInYear, final LocalDate upToInterestCalculationDate,
+            final Collection<Long> interestPostTransactions, final boolean isInterestTransfer, final Money minBalanceForInterestCalculation,
+            final boolean isSavingsInterestPostingAtCurrentPeriodEnd, final BigDecimal overdraftInterestRateAsFraction,
+            final Money minOverdraftForInterestCalculation, final boolean isUserPosting, final Integer financialYearBeginningMonth,
+            final boolean allowOverdraft, final List<PostingPeriod> allPostingPeriods, Boolean isOverdraftTransacction) {
+
+        if (txs == null || txs.isEmpty()) {
+            return periodStartingBalance;
+        }
+
+        final PostingPeriod postingPeriod = PostingPeriod.createFromDTO(periodInterval, periodStartingBalance, txs, monetaryCurrency,
+                compoundingPeriodType, interestCalculationType, interestRateAsFraction, daysInYear, upToInterestCalculationDate,
+                interestPostTransactions, isInterestTransfer, minBalanceForInterestCalculation, isSavingsInterestPostingAtCurrentPeriodEnd,
+                overdraftInterestRateAsFraction, minOverdraftForInterestCalculation, isUserPosting, financialYearBeginningMonth,
+                allowOverdraft);
+
+        periodStartingBalance = postingPeriod.closingBalance();
+        postingPeriod.setOverdraftInterest(isOverdraftTransacction);
+
+        if (!(MathUtil.isZero(postingPeriod.getOpeningBalance().getAmount())
+                && MathUtil.isZero(postingPeriod.closingBalance().getAmount()))) {
+            allPostingPeriods.add(postingPeriod);
+        }
+
+        return periodStartingBalance;
     }
 
     public List<PostingPeriod> calculateInterestUsing(final MathContext mc, final LocalDate upToInterestCalculationDate,
@@ -268,16 +305,40 @@ public class SavingsAccountInterestPostingServiceImpl implements SavingsAccountI
             if (postedAsOnDates.contains(periodInterval.endDate().plusDays(1))) {
                 isUserPosting = true;
             }
-            final PostingPeriod postingPeriod = PostingPeriod.createFromDTO(periodInterval, periodStartingBalance,
-                    retreiveOrderedNonInterestPostingTransactions(savingsAccountData), monetaryCurrency, compoundingPeriodType,
-                    interestCalculationType, interestRateAsFraction, daysInYearType.getValue(), upToInterestCalculationDate,
-                    interestPostTransactions, isInterestTransfer, minBalanceForInterestCalculation,
-                    isSavingsInterestPostingAtCurrentPeriodEnd, overdraftInterestRateAsFraction, minOverdraftForInterestCalculation,
-                    isUserPosting, financialYearBeginningMonth, savingsAccountData.isAllowOverdraft());
+            if (savingsAccountData.isAllowOverdraft() && !MathUtil.isZero(savingsAccountData.getGlAccountIdForInterestReceivable())) {
 
-            periodStartingBalance = postingPeriod.closingBalance();
+                List<SavingsAccountTransactionData> overdraftTxs = listForOverdraft(savingsAccountData, periodInterval);
+                List<SavingsAccountTransactionData> interestPostingTxs = listForInterestPosting(savingsAccountData, periodInterval,
+                        monetaryCurrency);
 
-            allPostingPeriods.add(postingPeriod);
+                boolean isOverdraftAccountType = isOverdraftAccount(savingsAccountData, periodInterval, monetaryCurrency);
+
+                List<SavingsAccountTransactionData> primaryInterestPublication = isOverdraftAccountType ? overdraftTxs : interestPostingTxs;
+                List<SavingsAccountTransactionData> secondaryInterestPublication = isOverdraftAccountType ? interestPostingTxs
+                        : overdraftTxs;
+
+                periodStartingBalance = appendPostingPeriodIfAny(periodInterval, periodStartingBalance, primaryInterestPublication,
+                        monetaryCurrency, compoundingPeriodType, interestCalculationType, interestRateAsFraction, daysInYearType.getValue(),
+                        upToInterestCalculationDate, interestPostTransactions, isInterestTransfer, minBalanceForInterestCalculation,
+                        isSavingsInterestPostingAtCurrentPeriodEnd, overdraftInterestRateAsFraction, minOverdraftForInterestCalculation,
+                        isUserPosting, financialYearBeginningMonth, savingsAccountData.isAllowOverdraft(), allPostingPeriods,
+                        isOverdraftAccountType ? true : false);
+
+                periodStartingBalance = appendPostingPeriodIfAny(periodInterval, periodStartingBalance, secondaryInterestPublication,
+                        monetaryCurrency, compoundingPeriodType, interestCalculationType, interestRateAsFraction, daysInYearType.getValue(),
+                        upToInterestCalculationDate, interestPostTransactions, isInterestTransfer, minBalanceForInterestCalculation,
+                        isSavingsInterestPostingAtCurrentPeriodEnd, overdraftInterestRateAsFraction, minOverdraftForInterestCalculation,
+                        isUserPosting, financialYearBeginningMonth, savingsAccountData.isAllowOverdraft(), allPostingPeriods,
+                        isOverdraftAccountType ? false : true);
+
+            } else {
+                periodStartingBalance = appendPostingPeriodIfAny(periodInterval, periodStartingBalance,
+                        retreiveOrderedNonInterestPostingTransactions(savingsAccountData), monetaryCurrency, compoundingPeriodType,
+                        interestCalculationType, interestRateAsFraction, daysInYearType.getValue(), upToInterestCalculationDate,
+                        interestPostTransactions, isInterestTransfer, minBalanceForInterestCalculation,
+                        isSavingsInterestPostingAtCurrentPeriodEnd, overdraftInterestRateAsFraction, minOverdraftForInterestCalculation,
+                        isUserPosting, financialYearBeginningMonth, savingsAccountData.isAllowOverdraft(), allPostingPeriods, false);
+            }
         }
 
         this.savingsHelper.calculateInterestForAllPostingPeriods(monetaryCurrency, allPostingPeriods,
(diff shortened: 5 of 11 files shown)


JohnAlva:
Hi, @adamsaghy can you help me to Review

adamsaghy:
@JohnAlva Please drop this generic "Savings Accounts Improvements" and start using clear, descriptive short description for the PRs which tells exactly what this PR is intended to do!

JohnAlva:
Hi @adamsaghy, thank you for the feedback.

Just to clarify, this PR is indeed related to the overall "Savings Accounts Improvements" initiative. The team decided to split the changes into multiple PRs to improve clarity and facilitate the review process, while keeping a consistent title to reflect that they all belong to the same scope.

adamsaghy:
> Hi @adamsaghy, thank you for the feedback.
> 
> Just to clarify, this PR is indeed related to the overall "Savings Accounts Improvements" initiative. The team decided to split the changes into multiple PRs to improve clarity and facilitate the review process, while keeping a consistent title to reflect that they all belong to the same scope.

The Fineract story id just fine to connect them. The PR title should tell exactly what it is doing and "improvements" is too generic and useless. Kindly asking you to rename it!

JohnAlva:
@adamsaghy sorry, but we already changed the name of each PR for each section.

JohnAlva:
Hi @adamsaghy, a quick review on this PR would be awesome. Thanks a lot in advance!

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountData.java
```diff
@@ -144,6 +144,15 @@ public final class SavingsAccountData implements Serializable {
     private transient Long glAccountIdForSavingsControl;
     private transient Long glAccountIdForInterestOnSavings;
 
+    private transient Long glAccountIdForSavingsControlAcountPositiveInterestNegative;
```
adamsaghy: What are these negative fields for? Can you please give me an example?

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountData.java
```diff
@@ -144,6 +144,15 @@ public final class SavingsAccountData implements Serializable {
...
+    private transient Long glAccountIdForInterestReceivablePositiveInterestNegative;
+
+    private transient Long glAccountIdForOverdraftPorfolioNegative;
+    private transient Long glAccountIdForInterestReceivableNegative;
```
adamsaghy: Does this need to be transient? Usually having transient field on a DTO is a serious anti-pattern... I would rather avoid them.

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountData.java
```diff
@@ -297,6 +306,43 @@ public void setGlAccountIdForInterestOnSavings(final Long glAccountIdForInterest
         this.glAccountIdForInterestOnSavings = glAccountIdForInterestOnSavings;
     }
 
+    public Long getId() {
```
adamsaghy: Lombok?

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountTransactionData.java
```diff
@@ -105,14 +105,15 @@ public final class SavingsAccountTransactionData implements Serializable {
     private BigDecimal overdraftAmount;
     private transient Long modifiedId;
     private transient String refNo;
+    private Boolean isNegativeBalance;
```
adamsaghy: what "isNegativeBalance" mean? I dont think we should have ever negative balances...

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsProductData.java
```diff
@@ -594,7 +594,7 @@ public boolean isCashBasedAccountingEnabled() {
     }
 
     public boolean isAccrualBasedAccountingEnabled() {
-        return isUpfrontAccrualAccounting() || isPeriodicAccrualAccounting();
+        return isUpfrontAccrualAccounting() && isPeriodicAccrualAccounting();
```
adamsaghy: Again... this should be OR, no?

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/domain/interest/CompoundInterestHelper.java
```diff
@@ -47,9 +47,14 @@ public Money calculateInterestForAllPostingPeriods(final MonetaryCurrency curren
         // total interest earned in previous periods but not yet recognised
         BigDecimal compoundedInterest = BigDecimal.ZERO;
         BigDecimal unCompoundedInterest = BigDecimal.ZERO;
+        LocalDate endDay = DateUtils.getBusinessLocalDate();
         final CompoundInterestValues compoundInterestValues = new CompoundInterestValues(compoundedInterest, unCompoundedInterest);
         for (final PostingPeriod postingPeriod : allPeriods) {
 
+            if (postingPeriod.dateOfPostingTransaction().getMonth() != endDay.getMonth()) {
```
adamsaghy, line 54: Can you please explain this logic? An example would be helpful!

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/domain/interest/PostingPeriod.java
```diff
@@ -64,6 +65,8 @@ public final class PostingPeriod {
 
     private Integer financialYearBeginningMonth;
 
+    private Boolean isEndTransaction = false;
```
adamsaghy: What does "isEndTransaction" mean?

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/domain/interest/PostingPeriod.java
```diff
@@ -187,6 +191,12 @@ public static PostingPeriod createFromDTO(final LocalDateInterval periodInterval
 
                 closeOfDayBalance = endOfDayBalance.closingBalance();
                 openingDayBalance = closeOfDayBalance;
+            } else if (!isEntraceNewValidation && !isEndTransaction && MathUtil.isLessThanZero(transaction.getRunningBalance())
```
adamsaghy: Can you please give some description / example what this logic does?

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
```diff
@@ -268,16 +280,107 @@ public List<PostingPeriod> calculateInterestUsing(final MathContext mc, final Lo
             if (postedAsOnDates.contains(periodInterval.endDate().plusDays(1))) {
                 isUserPosting = true;
             }
-            final PostingPeriod postingPeriod = PostingPeriod.createFromDTO(periodInterval, periodStartingBalance,
-                    retreiveOrderedNonInterestPostingTransactions(savingsAccountData), monetaryCurrency, compoundingPeriodType,
-                    interestCalculationType, interestRateAsFraction, daysInYearType.getValue(), upToInterestCalculationDate,
-                    interestPostTransactions, isInterestTransfer, minBalanceForInterestCalculation,
-                    isSavingsInterestPostingAtCurrentPeriodEnd, overdraftInterestRateAsFraction, minOverdraftForInterestCalculation,
-                    isUserPosting, financialYearBeginningMonth, savingsAccountData.isAllowOverdraft());
+            if (savingsAccountData.isAllowOverdraft()
```
adamsaghy: Can you please explain the logic here? An example or some further clarification would be appreciated.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
```diff
@@ -268,16 +280,107 @@ public List<PostingPeriod> calculateInterestUsing(final MathContext mc, final Lo
...
+                            if (!firstIsSet) {
+                                firstIsNegative = false;
+                                firstIsSet = true;
+                            }
+                        }
+                    }
+
+                }
+
+                if (firstIsNegative) {
```
adamsaghy: This logic looks way too complex and long. Would you mind to split into smaller methods which explains clearly the behaviour?

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
```diff
@@ -444,6 +548,26 @@ protected void recalculateDailyBalances(final Money openingAccountBalance, final
                     accountTransaction.updateRunningBalance(runningBalance);
                     addTransactionToExisting(accountTransaction, savingsAccountData);
 
+                    isTransactionsModified = true;
+                } else if (overdraftAmount.isNotEqualTo(Money.of(savingsAccountData.getCurrency(), transaction.getOverdraftAmount()))
```
adamsaghy: Please split this logic into smaller chunks with description that explains the implemented logic.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountReadPlatformServiceImpl.java
```diff
@@ -407,7 +414,16 @@ public List<SavingsAccountData> extractData(final ResultSet rs) throws SQLExcept
                     final Long glAccountIdForInterestOnSavings = rs.getLong("glAccountIdForInterestOnSavings");
                     final Long glAccountIdForSavingsControl = rs.getLong("glAccountIdForSavingsControl");
 
+                    final Long glAccountIdForOverdraftPorfolioNegative = rs.getLong("glAccountIdForOverdraftPorfolioNegative");
+                    final Long glAccountIdForInterestReceivableNegative = rs.getLong("glAccountIdForInterestReceivableNegative");
+
+                    final Long glAccountIdForSavingsControlAcountPositiveInterestNegative = rs
```
adamsaghy: What are these for?

fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
```diff
@@ -157,6 +162,44 @@ private List<SavingsAccountTransactionData> fetchTransactionsFromIds(final List<
         return this.savingsAccountReadPlatformService.retrieveAllTransactionData(refNo);
     }
 
+    private Long validateAccountForCreditAllowOverdraft(Boolean allowOverdraft, Boolean isNegativeBalance,
+            SavingsAccountData savingsAccountData, SavingsAccountTransactionData savingsAccountTransactionData) {
+        Long accountId = 0L;
```
adamsaghy: 0L???

adamsaghy:
@JohnAlva All PR should involve proper testing attached!

fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
JohnAlva: It is the value that we are giving to the variable and then the value is set.

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountTransactionData.java
JohnAlva: It was changed to isOverdraftDown

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountData.java
JohnAlva: This method is generated by Lombok, so it was removed to avoid redundancy

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountReadPlatformServiceImpl.java
JohnAlva: These are the accounting accounts used when an account overdraft occurs.

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsProductData.java
JohnAlva: It changed to Or

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountData.java
JohnAlva: The change has already been made and the transient has been removed.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
JohnAlva: This logic was created for savings accounts that are in overdraft. For example, if an account starts the month in overdraft and later has a positive balance, this change will allow the system to display the overdraft interest and the post interest accumulated by the customer during that period.

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountData.java
JohnAlva: Removed "Negative" since it was misleading — the field is for posting positive interest to the savings control account. No negative scenario involved.

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/domain/interest/CompoundInterestHelper.java
JohnAlva: Sure! This logic accumulates interest for each posting period. If the month changes, it resets the compoundedInterest. For example, if posting runs from June 1 to July 31, it resets at July, and only carries over interest within the same month—unless interestTransferEnabled or lockUntil prevent it.

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/domain/interest/PostingPeriod.java
JohnAlva: This handles first-day negative balances. If a transaction starts the period with a negative balance, it captures the end-of-day balance for that date and updates the openingDayBalance. For example, if Aug 1 starts below zero, this records it as a special case.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
JohnAlva: I've already refactored the method into smaller helper methods  to clarify the logic and improve readability.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
JohnAlva: Already refactored earlier

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/domain/interest/PostingPeriod.java
JohnAlva: It indicates if the transaction marks the end of processing within the posting period — used to avoid duplicate handling.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
```diff
@@ -268,16 +271,66 @@ public List<PostingPeriod> calculateInterestUsing(final MathContext mc, final Lo
...
-                    interestPostTransactions, isInterestTransfer, minBalanceForInterestCalculation,
-                    isSavingsInterestPostingAtCurrentPeriodEnd, overdraftInterestRateAsFraction, minOverdraftForInterestCalculation,
-                    isUserPosting, financialYearBeginningMonth, savingsAccountData.isAllowOverdraft());
+            if (savingsAccountData.isAllowOverdraft() && !MathUtil.isZero(savingsAccountData.getGlAccountIdForInterestReceivable())) {
+
+                List<SavingsAccountTransactionData> listOfTransactionsIsOverdraft = listForOverdraft(savingsAccountData, periodInterval);
+                List<SavingsAccountTransactionData> listOfTransactionsIsInterestPosting = listForInterestPosting(savingsAccountData,
+                        periodInterval);
+                Boolean firstIsOverdraft = isOverdraftAccount(savingsAccountData, periodInterval);
+                List<SavingsAccountTransactionData> primaryListForPost = null;
```
adamsaghy: What are these? The naming of this variables gives no information what it is used for. Can you please provide a better name for this?

fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
adamsaghy: 0L as id is invalid. AccountId must be set to proper value or be null!

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
JohnAlva: I've renamed the variables to make their purpose clearer:

- **listOfTransactionsIsOverdraft -> overdraftTransactionsInPeriod**: clarifies that this list contains overdraft transactions within the specified period.

- **listOfTransactionsIsInterestPosting -> interestPostingTransactionsInPeriod**: indicates the list holds interest posting transactions in the same period.
 
- **firstIsOverdraft -> isOverdraftAccountType**: makes it explicit that we're checking if the account type allows overdraft.
 
- **primaryListForPost / secondListForPost -> prioritizedTransactionList / fallbackTransactionList:** better expresses the intended priority between the lists.

fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
JohnAlva: I've replaced 0L with null to reflect the proper handling when no valid account ID is available.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
adamsaghy: Most of the names have improved, but I still have concerns about the following:
prioritizedTransactionList / fallbackTransactionList: These names attempt to express a priority relationship between the two lists.
While prioritizedTransactionList might work — especially if accompanied by a comment explaining the prioritization logic — the meaning of fallbackTransactionList is much less clear and feels ambiguous.
Could you explain the logic behind how these two lists are used? With more context, I’d be happy to suggest clearer and more appropriate naming options.

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountData.java
```diff
@@ -144,6 +144,15 @@ public final class SavingsAccountData implements Serializable {
     private transient Long glAccountIdForSavingsControl;
     private transient Long glAccountIdForInterestOnSavings;
 
+    private Long glAccountIdForSavingsControlAcountPositiveInterest;
```
adamsaghy: Can you please explain what these fields are for?

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountData.java
```diff
@@ -297,6 +306,22 @@ public void setGlAccountIdForInterestOnSavings(final Long glAccountIdForInterest
         this.glAccountIdForInterestOnSavings = glAccountIdForInterestOnSavings;
     }
 
+    public void setGlAccountIdForSavingsControlAcountPositiveInterest(Long glAccountIdForSavingsControlAcountPositiveInterest) {
```
adamsaghy: Lombok?

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountTransactionData.java
```diff
@@ -105,14 +105,15 @@ public final class SavingsAccountTransactionData implements Serializable {
     private BigDecimal overdraftAmount;
     private transient Long modifiedId;
     private transient String refNo;
+    private Boolean isOverdraftDown;
```
adamsaghy: Would "isOverdraft" be better here? What "is overdraft down" means?

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/domain/interest/CompoundInterestHelper.java
```diff
@@ -47,9 +47,14 @@ public Money calculateInterestForAllPostingPeriods(final MonetaryCurrency curren
...
+                compoundInterestValues.setcompoundedInterest(interestEarned.getAmount());
```
adamsaghy: `compoundInterestValues.setcompoundedInterest` -> `compoundInterestValues.setCompoundedInterest`

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/domain/interest/PostingPeriod.java
adamsaghy: Would a `date` be better here?

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/domain/interest/PostingPeriod.java
adamsaghy: Are we storing negative running balance on savings transaction level?

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/domain/interest/PostingPeriod.java
```diff
@@ -545,4 +557,11 @@ public Integer getFinancialYearBeginningMonth() {
         return this.financialYearBeginningMonth;
     }
 
+    public Money getOpeningBalance() {
```
adamsaghy: Lombok?

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
```diff
@@ -81,6 +83,7 @@ public SavingsAccountData postInterest(final MathContext mc, final LocalDate int
         for (final PostingPeriod interestPostingPeriod : postingPeriods) {
             final LocalDate interestPostingTransactionDate = interestPostingPeriod.dateOfPostingTransaction();
             final Money interestEarnedToBePostedForPeriod = interestPostingPeriod.getInterestEarned();
+            final Boolean isOverdraftDown = MathUtil.isLessThanZero(interestPostingPeriod.getInterestEarned());
```
adamsaghy: `isOverdraftDown` -> `isOverdraft`

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
```diff
@@ -268,16 +271,66 @@ public List<PostingPeriod> calculateInterestUsing(final MathContext mc, final Lo
...
-                    interestPostTransactions, isInterestTransfer, minBalanceForInterestCalculation,
-                    isSavingsInterestPostingAtCurrentPeriodEnd, overdraftInterestRateAsFraction, minOverdraftForInterestCalculation,
-                    isUserPosting, financialYearBeginningMonth, savingsAccountData.isAllowOverdraft());
+            if (savingsAccountData.isAllowOverdraft() && !MathUtil.isZero(savingsAccountData.getGlAccountIdForInterestReceivable())) {
+
+                List<SavingsAccountTransactionData> overdraftTransactionsInPeriod = listForOverdraft(savingsAccountData, periodInterval);
+                List<SavingsAccountTransactionData> interestPostingTransactionsInPeriod = listForInterestPosting(savingsAccountData,
+                        periodInterval);
+                Boolean isOverdraftAccountType = isOverdraftAccount(savingsAccountData, periodInterval);
+                List<SavingsAccountTransactionData> prioritizedTransactionList = null;
```
adamsaghy: `prioritizedTransactionList` and `fallbackTransactionList` are meaningless names... Would you mind to rename them according what they are truly doing or filtered by?

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
```diff
@@ -268,16 +271,66 @@ public List<PostingPeriod> calculateInterestUsing(final MathContext mc, final Lo
...
+                List<SavingsAccountTransactionData> fallbackTransactionList = null;
 
-            periodStartingBalance = postingPeriod.closingBalance();
+                Boolean flagIntroduce = false;
```
adamsaghy: `flagIntroduce` what is this for? I dont like the naming. Please give a more descriptive name to this field!

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
```diff
@@ -268,16 +271,66 @@ public List<PostingPeriod> calculateInterestUsing(final MathContext mc, final Lo
...
+                    flagIntroduce = postingPeriod.getEndTransaction();
+                    periodStartingBalance = postingPeriod.closingBalance();
+                    if (!(MathUtil.isZero(postingPeriod.getOpeningBalance().getAmount())
+                            && MathUtil.isZero(postingPeriod.closingBalance().getAmount()))) {
+                        allPostingPeriods.add(postingPeriod);
+                    }
+                }
+
+                if (!fallbackTransactionList.isEmpty()) {
+                    final PostingPeriod postingPeriod = PostingPeriod.createFromDTO(periodInterval, periodStartingBalance,
```
adamsaghy: duplicate?

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
```diff
@@ -297,6 +350,54 @@ public List<PostingPeriod> calculateInterestUsing(final MathContext mc, final Lo
...
+
+    }
+
+    private List<SavingsAccountTransactionData> listForInterestPosting(final SavingsAccountData savingsAccountData,
+            final LocalDateInterval periodInterval) {
+        List<SavingsAccountTransactionData> listOfTransactionsIsNotOverdraft = new ArrayList<>();
+        for (SavingsAccountTransactionData lists : retreiveOrderedNonInterestPostingTransactions(savingsAccountData)) {
+            if (MathUtil.isGreaterThanZero(lists.getRunningBalance())
+                    && periodInterval.startDate().getMonth() == lists.getDate().getMonth()) {
+                BigDecimal bd = new BigDecimal(String.valueOf(lists.getRunningBalance())).setScale(2, RoundingMode.DOWN);
```
adamsaghy: why is there a hardcoded rounding?

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
```diff
@@ -297,6 +350,54 @@ public List<PostingPeriod> calculateInterestUsing(final MathContext mc, final Lo
...
+        }
+        return listOfTransactionsIsNotOverdraft;
+
+    }
+
+    private Boolean isOverdraftAccount(final SavingsAccountData savingsAccountData, final LocalDateInterval periodInterval) {
+        Boolean isOverdraftAccountBalance = false;
+        for (SavingsAccountTransactionData lists : retreiveOrderedNonInterestPostingTransactions(savingsAccountData)) {
+            if (MathUtil.isLessThanZero(lists.getRunningBalance()) && periodInterval.startDate().getMonth() == lists.getDate().getMonth()) {
+                isOverdraftAccountBalance = true;
```
adamsaghy: Why jump out? Can it goes to overdraft and later fully paid, so it is not in overdraft anymore?

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
```diff
@@ -297,6 +350,54 @@ public List<PostingPeriod> calculateInterestUsing(final MathContext mc, final Lo
...
+                break;
+            } else if (periodInterval.startDate().getMonth() == lists.getDate().getMonth()) {
+                BigDecimal bd = new BigDecimal(String.valueOf(lists.getRunningBalance())).setScale(2, RoundingMode.DOWN);
```
adamsaghy: hardcoded rounding... no-no! please use Money object here instead!

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
adamsaghy: Use Money here!

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
```diff
@@ -444,6 +546,26 @@ protected void recalculateDailyBalances(final Money openingAccountBalance, final
...
 
+                    isTransactionsModified = true;
+                } else if (overdraftAmount.isNotEqualTo(Money.of(savingsAccountData.getCurrency(), transaction.getOverdraftAmount()))
+                        && MathUtil.isLessThanZero(transaction.getRunningBalance())) {
+                    SavingsAccountTransactionData accountTransaction = transaction;
+                    if (transaction.isChargeTransaction()) {
+                        Set<SavingsAccountChargesPaidByData> chargesPaidBy = transaction.getSavingsAccountChargesPaid();
+                        final Set<SavingsAccountChargesPaidByData> newChargePaidBy = new HashSet<>();
+                        chargesPaidBy.forEach(
+                                x -> newChargePaidBy.add(SavingsAccountChargesPaidByData.instance(x.getChargeId(), x.getAmount())));
```
adamsaghy: x? Can you please use a better field name?

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
```diff
@@ -444,6 +546,26 @@ protected void recalculateDailyBalances(final Money openingAccountBalance, final
...
+                        transaction.getSavingsAccountChargesPaid().addAll(newChargePaidBy);
+                    }
+                    // if (MathUtil.isGreaterThanZero(savingsAccountData.getSummary().getAccountBalance())){
```
adamsaghy: if not needed, remove please

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
```diff
@@ -444,6 +546,26 @@ protected void recalculateDailyBalances(final Money openingAccountBalance, final
...
+                    // transaction.reverse();
+                    // }
+                    if (overdraftAmount.isGreaterThanZero()) {
+                        transaction.updateOverdraftAmount(overdraftAmount.getAmount());
+                    }
+                    transaction.updateRunningBalance(runningBalance);
+                    // addTransactionToExisting(accountTransaction, savingsAccountData);
```
adamsaghy: same

fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
```diff
@@ -157,6 +162,44 @@ private List<SavingsAccountTransactionData> fetchTransactionsFromIds(final List<
         return this.savingsAccountReadPlatformService.retrieveAllTransactionData(refNo);
     }
 
+    private Long validateAccountForCreditAllowOverdraft(Boolean allowOverdraft, Boolean isOverdraftDown,
+            SavingsAccountData savingsAccountData, SavingsAccountTransactionData savingsAccountTransactionData) {
+        Long accountId = null;
+        if (allowOverdraft) {
+            if (!MathUtil.isZero(savingsAccountData.getGlAccountIdForInterestReceivable())) {
```
adamsaghy: zero check on GL account id???

fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
```diff
@@ -157,6 +162,44 @@ private List<SavingsAccountTransactionData> fetchTransactionsFromIds(final List<
...
+                accountId = isOverdraftDown
+                        ? MathUtil.isLessThanZero(savingsAccountTransactionData.getRunningBalance())
+                                ? savingsAccountData.getGlAccountIdForInterestReceivable()
```
adamsaghy: Are you sure you want to set the `savingsAccountData.getGlAccountIdForInterestReceivable()` if runnign balance is positive and also if it is negative? Looks incorrect to me...

fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
```diff
@@ -157,6 +162,44 @@ private List<SavingsAccountTransactionData> fetchTransactionsFromIds(final List<
...
+                                : savingsAccountData.getGlAccountIdForInterestReceivable()
+                        : savingsAccountData.getGlAccountIdForSavingsControl();
+            } else {
+                accountId = savingsAccountData.getGlAccountIdForSavingsControl();
```
adamsaghy: Would you mind to rework this? Readability is not too great and having "multiple" else case which sets the same usually means it can be reworked and simplified

fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
```diff
@@ -157,6 +162,44 @@ private List<SavingsAccountTransactionData> fetchTransactionsFromIds(final List<
...
+            }
+        } else {
+            accountId = savingsAccountData.getGlAccountIdForSavingsControl();
+        }
+        return accountId;
+    }
+
+    private Long validateAccountForDebitAllowOverdraft(Boolean allowOverdraft, Boolean isOverdraftDown,
```
adamsaghy: same as above

adamsaghy (changes requested):
Kindly see my review!

Make sure you are adding appropriate and required testing as well!

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountTransactionData.java
JohnAlva: isOverdraftDown has been renamed to isOverdraft for clarity and simplicity.

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/domain/interest/CompoundInterestHelper.java
JohnAlva: The method name has been corrected, it is now setCompoundedInterest instead of setcompoundedInterest.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
JohnAlva: I also updated the variable name: isOverdraftDown is now isOverdraft

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
JohnAlva: the variable name x was too generic. I've renamed it to chargePaidBy to improve clarity.

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountData.java
JohnAlva: I renamed the field to glAccountIdForInterestPayable to make its purpose clearer. It holds the GL account ID used to record interest payable transactions — cases where the institution owes interest to the client.

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/domain/interest/PostingPeriod.java
JohnAlva: Replaced the explicit getOpeningBalance() method with Lombok’s @Getter annotation for consistency and to reduce boilerplate.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
JohnAlva: Renamed flagIntroduce to hasEndTransactionFromPreviousPeriod to clearly indicate it stores the end-transaction status from the prior posting period.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
JohnAlva: Replaced hardcoded BigDecimal rounding with Money to ensure consistent currency handling and precision across the system.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
JohnAlva: Replaced the hardcoded BigDecimal rounding with Money based on the account’s MonetaryCurrency to ensure consistent precision and rounding.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
JohnAlva: he unused code has been removed as suggested.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
JohnAlva: Same here, it has been removed

fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
JohnAlva: The zero check on GL account ID has been reviewed and is now handled consistently in the updated method.

fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
JohnAlva: This logic has been corrected — the updated method now distinguishes between positive and negative balances when selecting the GL account

fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
JohnAlva: The method has been refactored to remove redundant else cases and improve readability.

fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
JohnAlva: Same here — the logic has been refactored and simplified for better readability

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/data/SavingsAccountData.java
JohnAlva: Done

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/domain/interest/PostingPeriod.java
JohnAlva: Done

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
JohnAlva: Updated loop variable naming and kept full iteration to check all transactions in the period before setting the overdraft flag, ensuring we don't exit early if balance later recovers.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
JohnAlva: Both blocks create a PostingPeriod, but they process different transaction lists. The first also updates hasEndTransactionFromPreviousPeriod, which is only relevant for prioritized transactions, so the logic must stay separate.

JohnAlva:
Hi, @adamsaghy  , can you review this PR? please

JohnAlva:
@adamsaghy

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/domain/interest/PostingPeriod.java
adamsaghy: @JohnAlva well?

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/domain/interest/PostingPeriod.java
adamsaghy: @JohnAlva well?

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
adamsaghy: @JohnAlva well?

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
adamsaghy: i see no change... it still using breaks different places...

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
adamsaghy: What i see this method `recalculateDailyBalances` is huge, and doing repetitive things:
```
if (transaction.isChargeTransaction()) {
                        Set<SavingsAccountChargesPaidByData> chargesPaidBy = transaction.getSavingsAccountChargesPaid();
                        final Set<SavingsAccountChargesPaidByData> newChargePaidBy = new HashSet<>();
                        chargesPaidBy.forEach(chargePaidBy -> newChargePaidBy
                                .add(SavingsAccountChargesPaidByData.instance(chargePaidBy.getChargeId(), chargePaidBy.getAmount())));
                        transaction.getSavingsAccountChargesPaid().addAll(newChargePaidBy);
                    }
```
```
if (overdraftAmount.isGreaterThanZero()) {
                        transaction.updateOverdraftAmount(overdraftAmount.getAmount());
                    }
                    transaction.updateRunningBalance(runningBalance);
```

The above piece of codes letter to letter duplications...

Please extract the content of each IF and ELSE into a private method and leave a short 1 sentence description what it does!

fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
adamsaghy: i see no changes...

adamsaghy (changes requested):
Please review my comments! (all of them!)

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
JohnAlva: I renamed the variables to explicitly reflect the type of transactions they contain. OverdraftTransactionsForPeriod now stores overdraft transactions, and InterestTransactionsForPeriod stores interest transactions for the period, making the code more readable and self-explanatory.

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/domain/interest/PostingPeriod.java
JohnAlva: It was a test case and it was missed and has already been removed.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
adamsaghy: I see no changes...

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
adamsaghy:
```
if (!prioritizedTransactionList.isEmpty()) {
                    final PostingPeriod postingPeriod = PostingPeriod.createFromDTO(periodInterval, periodStartingBalance,
                            prioritizedTransactionList, monetaryCurrency, compoundingPeriodType, interestCalculationType,
                            interestRateAsFraction, daysInYearType.getValue(), upToInterestCalculationDate, interestPostTransactions,
                            isInterestTransfer, minBalanceForInterestCalculation, isSavingsInterestPostingAtCurrentPeriodEnd,
                            overdraftInterestRateAsFraction, minOverdraftForInterestCalculation, isUserPosting, financialYearBeginningMonth,
                            savingsAccountData.isAllowOverdraft());
                    periodStartingBalance = postingPeriod.closingBalance();
                    if (!(MathUtil.isZero(postingPeriod.getOpeningBalance().getAmount())
                            && MathUtil.isZero(postingPeriod.closingBalance().getAmount()))) {
                        allPostingPeriods.add(postingPeriod);
                    }
                }
```
and 
```
if (!fallbackTransactionList.isEmpty()) {
                    final PostingPeriod postingPeriod = PostingPeriod.createFromDTO(periodInterval, periodStartingBalance,fallbackTransactionList, monetaryCurrency, compoundingPeriodType, interestCalculationType,
                            interestRateAsFraction, daysInYearType.getValue(), upToInterestCalculationDate, interestPostTransactions,
                            isInterestTransfer, minBalanceForInterestCalculation, isSavingsInterestPostingAtCurrentPeriodEnd,
                            overdraftInterestRateAsFraction, minOverdraftForInterestCalculation, isUserPosting, financialYearBeginningMonth,
                            savingsAccountData.isAllowOverdraft());

                    periodStartingBalance = postingPeriod.closingBalance();
                    if (!(MathUtil.isZero(postingPeriod.getOpeningBalance().getAmount())
                            && MathUtil.isZero(postingPeriod.closingBalance().getAmount()))) {
                        allPostingPeriods.add(postingPeriod);
                    }
```

Is letter to letter the same...

`hasEndTransactionFromPreviousPeriod` cannot be found in the code anywhere...

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
adamsaghy: I see no changes, the  `break;` are still there

fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
adamsaghy: See no changes...

adamsaghy:
@JohnAlva Kindly see my review!

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/domain/interest/PostingPeriod.java
JohnAlva: The isEndTransaction (boolean) parameter was removed because it was no longer needed in the current logic. Validation that relied on it is now directly covered by the date conditions already present in the method, so there was no need to replace it with a date field.

JohnAlva:
It is put in draft by dependency to pr https://github.com/apache/fineract/pull/4876 because we use the accounting account

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
JohnAlva: I removed the break statements and rewrote the method to return early when the condition is met. This way the logic remains the same (exit at the first matching transaction), but without using break.

fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
JohnAlva: I simplified the method by removing redundant else branches, using early returns, and reducing nesting. The logic and outcomes remain the same, but the structure is now cleaner and easier to follow.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
JohnAlva: I refactored the repeated blocks that created and added PostingPeriod instances into a private helper method appendPostingPeriodIfAny(...).
This keeps the logic exactly the same, but removes duplication, making the code easier to read, maintain, and update in the future.
Any future changes to the PostingPeriod creation logic will now only need to be done in one place.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
JohnAlva: I updated the variable names to be shorter but still explicit, aligning them with the methods that populate them (overdraftTxs and interestPostingTxs). This keeps the code self-explanatory while improving readability

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
JohnAlva: Done

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
JohnAlva: Rewrote recalculateDailyBalances for clarity and consistency with the class style. Business logic is unchanged; I just ordered the steps (effect computation, running balance update, overdraft normalization, and reverse+append flow) and used explicit variable names.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
```diff
@@ -81,6 +82,7 @@ public SavingsAccountData postInterest(final MathContext mc, final LocalDate int
         for (final PostingPeriod interestPostingPeriod : postingPeriods) {
             final LocalDate interestPostingTransactionDate = interestPostingPeriod.dateOfPostingTransaction();
             final Money interestEarnedToBePostedForPeriod = interestPostingPeriod.getInterestEarned();
+            final Boolean isOverdraft = MathUtil.isLessThanZero(interestPostingPeriod.getInterestEarned());
```
adamsaghy: interestPostingPeriod.getInterestEarned() should never be minus... we should not work with minus balances. It would be much better to have a new field which contains whether the customer must pay interest on overdraft amount or not.. also flagging whether an account or interest period is in overdraft, should be done by a flag, not by checking whether `interestEarned` is positive or negative... @JohnAlva Thoughts?

JohnAlva: I updated PostingPeriod so getInterestEarned() never returns a negative value, and added an explicit overdraftInterest flag. In the service we now rely on this flag to decide between normal interest posting and overdraft interest, instead of checking if the amount is less than zero. This keeps the logic clearer and avoids mixing earned vs. charged interest.

JohnAlva:
Hi, @adamsaghy  , can you review this PR? please

JohnAlva:
@adamsaghy

IOhacker:
LGMT

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
```diff
@@ -500,7 +606,7 @@ protected SavingsAccountTransactionData findInterestPostingTransactionFor(final
         SavingsAccountTransactionData postingTransation = null;
         List<SavingsAccountTransactionData> trans = savingsAccountData.getSavingsAccountTransactionData();
         for (final SavingsAccountTransactionData transaction : trans) {
-            if ((transaction.isInterestPostingAndNotReversed() || transaction.isOverdraftInterestAndNotReversed())
+            if ((transaction.isInterestPostingAndNotReversed() && transaction.isOverdraftInterestAndNotReversed())
```
adamsaghy: I don’t think this scenario is possible: a transaction cannot be both interest posting and overdraft posting at the same time. If this bug wasn’t caught during testing, it highlights a significant testing gap.

adamsaghy (changes requested):
Kindly see my concerns!

I am not convinced we have enough test coverage to ensure the functionality works as expected!

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
JohnAlva: Replaced && with || when searching for interest posting or overdraft interest transactions, since a transaction cannot be both at the same time.

adamsaghy: Please make sure this functionality is covered with test case!

JohnAlva: Also added/adjusted tests in SavingsInterestPostingTest to cover interest-only, overdraft-only, and mixed scenarios, and fixed the compounding setup in SavingsProductHelper for consistency.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
```diff
@@ -268,16 +305,39 @@ public List<PostingPeriod> calculateInterestUsing(final MathContext mc, final Lo
...
 
-            periodStartingBalance = postingPeriod.closingBalance();
+                List<SavingsAccountTransactionData> overdraftTxs = listForOverdraft(savingsAccountData, periodInterval);
+                List<SavingsAccountTransactionData> interestPostingTxs = listForInterestPosting(savingsAccountData, periodInterval,
+                        monetaryCurrency);
 
-            allPostingPeriods.add(postingPeriod);
+                boolean isOverdraftAccountType = isOverdraftAccount(savingsAccountData, periodInterval, monetaryCurrency);
+
+                List<SavingsAccountTransactionData> first = isOverdraftAccountType ? overdraftTxs : interestPostingTxs;
```
adamsaghy: what is `first` and `second`? Would you mind to rename them appropriately? It does not tell anything about what these are...

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
```diff
@@ -268,16 +305,39 @@ public List<PostingPeriod> calculateInterestUsing(final MathContext mc, final Lo
...
+                List<SavingsAccountTransactionData> second = isOverdraftAccountType ? interestPostingTxs : overdraftTxs;
+
+                periodStartingBalance = appendPostingPeriodIfAny(periodInterval, periodStartingBalance, first, monetaryCurrency,
+                        compoundingPeriodType, interestCalculationType, interestRateAsFraction, daysInYearType.getValue(),
+                        upToInterestCalculationDate, interestPostTransactions, isInterestTransfer, minBalanceForInterestCalculation,
+                        isSavingsInterestPostingAtCurrentPeriodEnd, overdraftInterestRateAsFraction, minOverdraftForInterestCalculation,
+                        isUserPosting, financialYearBeginningMonth, savingsAccountData.isAllowOverdraft(), allPostingPeriods,
+                        isOverdraftAccountType ? true : false);
+
+                periodStartingBalance = appendPostingPeriodIfAny(periodInterval, periodStartingBalance, second, monetaryCurrency,
```
adamsaghy: Why do we have 2 of these? 
Are you sure you want to override the `periodStartingBalance` variable values?
Are you sure the first `appendPostingPeriodIfAny` result to be used in the second `appendPostingPeriodIfAny`? 
Would you mind to add some short description what exactly happens here?

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
```diff
@@ -509,6 +615,22 @@ protected SavingsAccountTransactionData findInterestPostingTransactionFor(final
         return postingTransation;
     }
 
+    protected SavingsAccountTransactionData findInterestPostingTransactionForInterest(final LocalDate postingDate,
+            final SavingsAccountData savingsAccountData, boolean isOverdraft) {
+        SavingsAccountTransactionData postingTransation = null;
+        List<SavingsAccountTransactionData> trans = savingsAccountData.getSavingsAccountTransactionData();
```
adamsaghy: `trans`: please use descriptive variable names!

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
```diff
@@ -509,6 +615,22 @@ protected SavingsAccountTransactionData findInterestPostingTransactionFor(final
...
+
+        for (final SavingsAccountTransactionData transaction : trans) {
+            Boolean interestSearch = isOverdraft ? transaction.isOverdraftInterestAndNotReversed()
+                    : transaction.isInterestPostingAndNotReversed();
+            if (interestSearch && transaction.occursOn(postingDate) && !transaction.isReversalTransaction()) {
+                postingTransation = transaction;
```
adamsaghy: A Stream API + filter + findFirst would be better here!

adamsaghy (changes requested):
Kindly see my concerns!

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
adamsaghy: @JohnAlva I have executed the `SavingsInterestPostingTest` when
`if ((transaction.isInterestPostingAndNotReversed() || transaction.isOverdraftInterestAndNotReversed())`
and also when
 `if ((transaction.isInterestPostingAndNotReversed() && transaction.isOverdraftInterestAndNotReversed())`.
 
 In both cases all tests were GREEN, which is not good! Please make sure you have tests (integration or unit) which ensures it works correctly!

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
JohnAlva: I’ve updated the variable names to primaryInterestPublication and secondaryInterestPublication so they better describe their purpose and make the code easier to read

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
JohnAlva: The periodStartingBalance is updated twice because each call to appendPostingPeriodIfAny processes a different batch of transactions (primary and secondary). The second call requires the balance already updated by the first call so that interest calculations and minimum balance validations are correct and consistent.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
JohnAlva: I’ve renamed trans to transactions to make it clearer that it holds all the account transactions.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
JohnAlva: I refactored the loop to use stream(), filter(), and findFirst() for a cleaner and more readable implementation.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountInterestPostingServiceImpl.java
JohnAlva: I’ve updated the integration tests to assert the presence of exactly one interest or overdraft posting on the posting date for each scenario. With the current logic (||), all tests pass, but if we change it to &&, the single-type scenarios fail as expected. This ensures the behavior is now properly covered by tests.

JohnAlva:
@adamsaghy , There is a job that failed, we checked for a solution, but it shows that it does not find an image to download, I don't think it is related to us, could you help us?

IOhacker:
LGTM

PR #5135, opened 2025-10-31
FINERACT-1086: take PESSIMISTIC_WRITE lock on Loan before save

siva-sai-udaygiri:
## JIRA
FINERACT-1086

## What
Acquire a DB write lock on `Loan` to serialize concurrent writers and avoid `JpaOptimisticLockingFailureException`.

## How
- In `LoanAccountDomainServiceJpa#makeRefund(...)`, immediately after assembling the loan, call:
  `entityManager.lock(loan, LockModeType.PESSIMISTIC_WRITE);`
- Leave existing `loanRepositoryWrapper.saveAndFlush(loan)` unchanged.

## Why
Optimistic locking was tripping when two requests updated the same Loan at once. A `PESSIMISTIC_WRITE` lock ensures only one writer proceeds at a time.

## Notes
- Local Windows build hit Spotless/line-ending issues. I will run `./gradlew spotlessApply` in Linux (WSL/Codespaces) and push formatting updates shortly.
- Follow-up plan: add an integration test that performs two concurrent updates on the same Loan and asserts no optimistic-locking failure.

### DIFF
--- a/fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanRepository.java
+++ b/fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanRepository.java
@@ -18,6 +18,7 @@
  */
 package org.apache.fineract.portfolio.loanaccount.domain;
 
+import jakarta.persistence.LockModeType;
 import java.time.LocalDate;
 import java.util.Collection;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


adamsaghy:
No, we dont want such lock. it kills the performance and optimisticlock exception with proper retry (which us there by default) is minimal and okay.

galovics:
@siva-sai-udaygiri I'm closing this PR. There's no way this will go in. 

As someone who heavily worked with locks, relational DBs and JPA, this will hurt performance to a really high degree.

PR #5143, opened 2025-11-05
FINERACT-2406: Transaction reprocessing fix

adamsaghy:
## Description

## Checklist


budaidev (approved):

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/aop/FlushModeAspect.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+@Aspect
+@Component
+@RequiredArgsConstructor
+public class FlushModeAspect {
+
+    private final FlushModeHandler flushModeHandler;
+
+    @Around("@within(withFlushMode) || @annotation(withFlushMode)")
+    public Object manageFlushMode(ProceedingJoinPoint joinPoint, WithFlushMode withFlushMode) {
```
galovics, line 53: I think there's one thing missing from this. Whether the annotation is applied to a class/method which is inside a running transaction (IIRC you can use the TransactionSynchronizationManager to check this).

Also, one thing that comes to my mind when combining these kind of things (Transactional and this annotation) is the ordering.

You need to make sure the ordering of this aspect is lower than the transactional annotation's, otherwise you might apply this first then the tx annotation.

galovics (changes requested):

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/aop/FlushModeAspect.java
adamsaghy: I accept the transactionSynchronizationManager one, but i dont think we should mess with the orders. In case "Transactional" annotation is used alongside with WithFlushMode, the only important to use Transactional first.

By default Transactional annotation is using Spring default ordering which is LOWEST_PRECEDENCE, alongside with Retry, Cacheable, etc. which also using this. If i am changing it to use LOWEST_PRECEDENCE-1, it might break all those logic: having retry inside transaction boundary, etc.

galovics (approved):

oleksii-novikov-onix (approved):

PR #5268, opened 2025-12-27
FINERACT-2421: Fix potential NullPointerException in CalendarCommandFromApiJsonDeserializer

airajena:
## JIRA Ticket
https://issues.apache.org/jira/browse/FINERACT-2421

## Description
This PR fixes a potential `NullPointerException` in `CalendarCommandFromApiJsonDeserializer` when extracting boolean values from JSON. This addresses the FIXME comments on lines 156 and 298 in the original file.

## Problem
The `extractBooleanNamed` method can return `null` when the JSON value is null or missing. Using primitive `boolean` causes automatic unboxing which throws `NullPointerException`:
```java
// FIXME - Throws NullPointerException when boolean value is null
final boolean repeating = this.fromApiJsonHelper.extractBooleanNamed(...); // NPE!
```

## Solution
Use `Boolean` wrapper type instead of primitive `boolean` and use null-safe comparison:
```java
final Boolean repeating = this.fromApiJsonHelper.extractBooleanNamed(...);
if (Boolean.TRUE.equals(repeating)) { ... }  // Null-safe
```

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/calendar/serialization/CalendarCommandFromApiJsonDeserializer.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/calendar/serialization/CalendarCommandFromApiJsonDeserializer.java
@@ -81,7 +81,7 @@ public class CalendarCommandFromApiJsonDeserializer extends AbstractFromApiJsonD
                 element);
         final Integer typeId = this.fromApiJsonHelper.extractIntegerSansLocaleNamed(CalendarSupportedParameters.TYPE_ID.getValue(),
                 element);
-        final boolean repeating = this.fromApiJsonHelper.extractBooleanNamed(CalendarSupportedParameters.REPEATING.getValue(), element);
+        final Boolean repeating = this.fromApiJsonHelper.extractBooleanNamed(CalendarSupportedParameters.REPEATING.getValue(), element);
         final Integer remindById = this.fromApiJsonHelper.extractIntegerSansLocaleNamed(CalendarSupportedParameters.REMIND_BY_ID.getValue(),
                 element);
         final Integer firstReminder = this.fromApiJsonHelper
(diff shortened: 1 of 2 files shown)


fineract-provider/src/main/java/org/apache/fineract/portfolio/calendar/serialization/CalendarCommandFromApiJsonDeserializer.java
```diff
@@ -153,11 +153,11 @@ public void validateForCreate(final String json) {
                 .inMinMaxRange(CalendarEntityType.getMinValue(), CalendarEntityType.getMaxValue());
 
         if (this.fromApiJsonHelper.parameterExists(CalendarSupportedParameters.REPEATING.getValue(), element)) {
-            // FIXME - Throws NullPointerException when boolean value is null
-            final boolean repeating = this.fromApiJsonHelper.extractBooleanNamed(CalendarSupportedParameters.REPEATING.getValue(), element);
+            // Fixed: Use Boolean wrapper to avoid NullPointerException when value is null
```
Aman-Mittal: In case of fixed remove the comment entirely,

airajena: Thank you for the feedback! I've removed the explanatory comments .
Regarding the test failure: The failing test SavingsInterestPostingTest.testSavingsInterestPosting_Works_ForMultipleAccounts() is unrelated to my changes in the calendar module. It's failing due to an HTTP 401 authentication error in the CI test infrastructure. My changes only affect CalendarCommandFromApiJsonDeserializer.java and 
CalendarCommand.java in the calendar module - they don't touch the savings API or authentication.

Could this be a flaky/infrastructure issue in the CI pipeline?

galovics (approved):

PR #5282, opened 2026-01-03
FINERACT-1266: Fix NullPointerException in SavingsHelper.determineInterestPostingPeriods

airajena:
## Description
Fixes NullPointerException that occurs in `SavingsHelper.determineInterestPostingPeriods()` when date parameters are null.

The NPE occurs when closing a Fixed Deposit account if the maturity date is null (e.g., when `depositPeriodFrequencyType` is INVALID).

## Root Cause
The method `determineInterestPostingPeriods()` uses `DateUtils.isAfter()` which throws NPE when either `startInterestCalculationLocalDate` or `interestPostingUpToDate` is null.

## Fix
Added null check at the start of the method to return an empty list early if either date parameter is null.

## Testing
- Fix compiles successfully
- Verified the null check prevents the NPE scenario from the stack trace

## Related Issue
Fixes FINERACT-1266
```

### DIFF
--- a/fineract-core/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsHelper.java
+++ b/fineract-core/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsHelper.java
@@ -52,6 +52,11 @@ public final class SavingsHelper {
             final Integer financialYearBeginningMonth, List<LocalDate> postInterestAsOn) {
 
         final List<LocalDateInterval> postingPeriods = new ArrayList<>();
+
+        if (startInterestCalculationLocalDate == null || interestPostingUpToDate == null) {
+            return postingPeriods;
+        }
+
         LocalDate periodStartDate = startInterestCalculationLocalDate;
         LocalDate periodEndDate = periodStartDate;
         LocalDate actualPeriodStartDate = periodStartDate;

### CODE REVIEW

**Comment by adamsaghy** [comment]
@airajena Please rebase this PR with latest `develop` branch.

File: fineract-core/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsHelper.java [hunk 1]
```diff
@@ -52,6 +52,11 @@ public List<LocalDateInterval> determineInterestPostingPeriods(final LocalDate s
             final Integer financialYearBeginningMonth, List<LocalDate> postInterestAsOn) {
 
         final List<LocalDateInterval> postingPeriods = new ArrayList<>();
+
+        if (startInterestCalculationLocalDate == null || interestPostingUpToDate == null) {
```
adamsaghy, line 56: Please explain this new condition and why it is the correct behavior?

adamsaghy (changes requested):
Kindly see my questions.

fineract-core/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsHelper.java
airajena: The null check prevents a NullPointerException that occurs when either startInterestCalculationLocalDate or interestPostingUpToDate is null.

DateUtils.isAfter() at line 64 throws NPE when comparing null dates, This happens when closing a Fixed Deposit account where depositPeriodFrequencyType is INVALID, causing the maturity date (and subsequently interestPostingUpToDate) to be null

If either date is null, there are no valid interest posting periods to determine
- An empty list semantically means "no posting periods" - which is accurate when dates are undefined
- This allows the calling code (e.g., account closure) to continue gracefully

I've also rebased the PR with the latest develop branch as requested.

adamsaghy: In what situations can these values be `null`?

airajena: interestPostingUpToDate can be null when:
depositPeriodFrequencyType is INVALID: In FixedDepositAccount.calculateMaturityDate(), the switch statement handles DAYS, WEEKS, MONTHS, and YEARS, but when  depositPeriodFrequencyType is INVALID, it falls through with an empty break, leaving maturityDate as null:

Database has null/invalid deposit_period_frequency_enum: SavingsPeriodFrequencyType.fromInt() returns INVALID when the input is null or an unrecognized value. The column deposit_period_frequency_enum is marked as nullable = true in DepositAccountTermAndPreClosure

Account reinvestment: DepositAccountTermAndPreClosure.copy() explicitly sets maturityDate = null when creating a copy for reinvestment. 

startInterestCalculationLocalDate can be null when: 
Account in early submission state: accountSubmittedOrActivationDate() could return null if both activationDate and submittedOnDate are null.

PR #5340, opened 2026-01-17
FINERACT-2155: Allow annual charge payment after due date

airajena:
## Description
This PR addresses two distinct issues identified in e2e testing:
1.  **Client Charges**: Enhances validation for client charge payments (e.g., annual fees) to strictly prevent payments *before* the due date, while allowing payments on or after the due date.
2.  **Progressive Loan Schedule**: Fixes the EMI calculation for Progressive Loans to correctly honor the `allowPartialPeriodInterestCalculation` configuration.

## Changes
*   **ClientChargeWritePlatformServiceImpl.java**: 
    *   Added a validation check in `validatePaymentDateAndAmount`.
    *   Ensures that if a transaction date is before the `clientCharge` due date, it fails with the error code `transaction.before.dueDate`. This clarifies that payments are allowed on or after the due date, but not before.
*   **ProgressiveEMICalculator.java**: 
    *   Modified `calculateRateFactorPerPeriod` logic.
    *   Added a check for `!loanProductRelatedDetail.isAllowPartialPeriodInterestCalculation()`.
    *   If partial period calculation is disabled, the system now forces the interest calculation to use the full `calculatedDaysInRepaymentPeriod` instead of the `actualDaysInPeriod`, preventing incorrect partial interest accrual in specific schedule generation scenarios.

## Rationale
*   **Client Charges**: In real-world scenarios, charges like annual fees might be paid after the due date (e.g., due to weekends or processing delays). The system should explicitly allow this while maintaining a check against premature payments.
*   **Progressive Loan**: The `allowPartialPeriodInterestCalculation` flag was not being consistently applied in the rate factor calculation, causing discrepancies in EMI schedules for loans where partial interest should be disabled.

## Testing
*   Validated that `transaction.before.dueDate` error is raised for early client charge payments.
*   Ensured Progressive Loan EMI calculations are correct when `allowPartialPeriodInterestCalculation` is set to false.
*   Resolves failing e2e tests related to these scenarios.

### DIFF
--- a/fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccount.java
+++ b/fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccount.java
@@ -3140,8 +3140,8 @@ public class SavingsAccount extends AbstractAuditableWithUTCDateTimeCustom<Long>
                 throw new PlatformApiDataValidationException(dataValidationErrors);
             }
 
-            if (!DateUtils.isEqual(annualFeeDueDate, transactionDate)) {
-                baseDataValidator.reset().failWithCodeNoParameterAddedToErrorCode("invalid.date");
+            if (DateUtils.isBefore(transactionDate, annualFeeDueDate)) {
+                baseDataValidator.reset().failWithCodeNoParameterAddedToErrorCode("transaction.before.dueDate");
                 throw new PlatformApiDataValidationException(dataValidationErrors);
             }
 
--- a/integration-tests/src/test/java/org/apache/fineract/integrationtests/ClientSavingsIntegrationTest.java
+++ b/integration-tests/src/test/java/org/apache/fineract/integrationtests/ClientSavingsIntegrationTest.java
@@ -2954,6 +2954,80 @@ public class ClientSavingsIntegrationTest {
         Assertions.assertNotNull(payChargeId);
     }
 
+    @SuppressWarnings("unchecked")
+    @Test
+    public void testAnnualChargePaymentAfterDueDate() {
+        Integer savingsId = null;
+        try {
+            globalConfigurationHelper.updateGlobalConfiguration(GlobalConfigurationConstants.ENABLE_BUSINESS_DATE,
+                    new PutGlobalConfigurationsRequest().enabled(true));
+
+            LocalDate submittedDate = LocalDate.of(2023, 1, 1);
+            String submittedDateString = "01 January 2023";
+            BusinessDateHelper.updateBusinessDate(requestSpec, responseSpec, BusinessDateType.BUSINESS_DATE, submittedDate);
+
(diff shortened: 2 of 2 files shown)


adamsaghy (changes requested):
Please attach proper integration or E2E test cases to cover this functionality!

airajena:
> Please attach proper integration or E2E test cases to cover this functionality!

Added integration test.

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanproduct/calc/ProgressiveEMICalculator.java
```diff
@@ -1375,10 +1369,14 @@ private BigDecimal calculateRateFactorPerPeriod(final ProgressiveLoanInterestSch
...
-        final BigDecimal actualDaysInPeriod = BigDecimal
-                .valueOf(DateUtils.getDifferenceInDays(interestPeriodFromDate, interestPeriodDueDate));
         final BigDecimal calculatedDaysInRepaymentPeriod = BigDecimal
                 .valueOf(DateUtils.getDifferenceInDays(repaymentPeriod.getFromDate(), repaymentPeriod.getDueDate()));
+        BigDecimal actualDaysInPeriod = BigDecimal.valueOf(DateUtils.getDifferenceInDays(interestPeriodFromDate, interestPeriodDueDate));
+        if (loanProductRelatedDetail.getInterestCalculationPeriodMethod() != null
+                && loanProductRelatedDetail.getInterestCalculationPeriodMethod().isSameAsRepaymentPeriod()
+                && !loanProductRelatedDetail.isAllowPartialPeriodInterestCalculation()) {
+            actualDaysInPeriod = calculatedDaysInRepaymentPeriod;
+        }
```
adamsaghy: I dont think this is relevant for the FINERACT-2155 story.

Also I dont think this is correct either... Ratefactor is only from the interest period. It is independent from interest calculation period method. That should only control whether we are taking all the interest periods anyway or not, but not directly the sub interest period.

adamsaghy (changes requested):
Please remove the unrelated parts of the PR!

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanproduct/calc/ProgressiveEMICalculator.java
airajena: Thank you for the review. I have removed the changes to ProgressiveEMICalculator.java and ClientChargeWritePlatformServiceImpl.java as requested, keeping the PR focused solely on the valid fix for FINERACT-2155 in SavingsAccount.java.

Regarding the reverted changes: I initially thought they were necessary to fix e2e test failures I encountered earlier. 

Could you please point me to any documentation regarding the financial architecture calculations in Fineract? It would help me better understand the model for future contributions.

PR #5348, opened 2026-01-19
FINERACT-2438: Add Micrometer Observability to Reporting API

Shivd131:
## Description

This PR aims to instrument the `RunreportsApiResource` with the `@Timed` annotation to provide visibility into report generation performance.

Currently, there is no specific metric isolating the Run Reports API execution time from general HTTP traffic. This change introduces a dedicated metric, `fineract.report.execution`, to track the aggregate latency and throughput of the reporting subsystem.

---

## Key Changes

- **Annotation-based Metrics**  
  Applied `@Timed` to the `runReport` method in `RunreportsApiResource`.

- **Infrastructure Configuration**  
  Added `MetricsConfig` to enable AOP support for the `@Timed` annotation by registering `TimedAspect`.

- **Clean Implementation**  
  Utilized standard Spring Boot and Micrometer patterns, avoiding any additional boilerplate.

This establishes the necessary performance baseline to monitor the reporting module’s load and response times.

---

## Verification

Validated locally by generating report traffic and querying the Actuator metrics endpoint.

### Request

```bash
curl -k -u mifos:password \
  -H "Fineract-Platform-TenantId: default" \
  -X GET "https://localhost:8443/fineract-provider/actuator/metrics/fineract.report.execution"
```

<details>
<summary><strong>Response</strong></summary>

```json
{
  "name": "fineract.report.execution",
  "description": "Time taken to execute reports",
  "baseUnit": "seconds",
  "measurements": [
    {
      "statistic": "COUNT",
      "value": 1.0
    },
    {
      "statistic": "TOTAL_TIME",
      "value": 0.0114712
    },
    {
      "statistic": "MAX",
      "value": 0.0114712
    }
  ],
  "availableTags": [
    {
      "tag": "exception",
      "values": ["none"]
    },
    {
      "tag": "component",
      "values": ["reporting"]
    },
    {
      "tag": "method",
      "values": ["runReport"]
    },
    {
      "tag": "class",
      "values": [
        "org.apache.fineract.infrastructure.dataqueries.api.RunreportsApiResource"
      ]
    }
  ]
}
```

</details>

---

## Checklist

Please make sure these boxes are checked before submitting your pull request — thanks!

- [x] Acknowledge that we will not review PRs that are not passing the build (_"green"_) — it is your responsibility to get a proposed PR to pass the build, not primarily the project's maintainers.
- [ ] Add required Swagger annotations and update API documentation at  
  `fineract-provider/src/main/resources/static/legacy-docs/apiLive.htm` if applicable.
- [x] [This PR must not be a "code dump"](https://cwiki.apache.org/confluence/display/FINERACT/Pull+Request+Size+Limit).

Your assigned reviewer(s) will follow our  
[guidelines for code reviews](https://cwiki.apache.org/confluence/display/FINERACT/Code+Review+Guide).

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/MetricsConfig.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/MetricsConfig.java
@@ -1,0 +1,36 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.fineract.infrastructure.core.config;
+
+import io.micrometer.core.aop.TimedAspect;
+import io.micrometer.core.instrument.MeterRegistry;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.EnableAspectJAutoProxy;
+
+@Configuration
+@EnableAspectJAutoProxy
+public class MetricsConfig {
+
+    @Bean
+    public TimedAspect timedAspect(MeterRegistry registry) {
+        return new TimedAspect(registry);
+    }
+}
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/api/RunreportsApiResource.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/api/RunreportsApiResource.java
@@ -18,6 +18,7 @@
  */
 package org.apache.fineract.infrastructure.dataqueries.api;
 
+import io.micrometer.core.annotation.Timed;
 import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.Parameter;
 import io.swagger.v3.oas.annotations.media.ArraySchema;
@@ -93,6 +94,7 @@ public class RunreportsApiResource {
 
     @GET
     @Path("{reportName}")
+    @Timed(value = "fineract.report.execution", description = "Time taken to execute reports", extraTags = { "component", "reporting" })
     @Consumes({ MediaType.APPLICATION_JSON })
     @Produces({ MediaType.APPLICATION_JSON, "text/csv", "application/vnd.ms-excel", "application/pdf", "text/html" })
     @Operation(summary = "Run a predefined report", description = ReportParameters.FULL_DESCRIPTION)


fineract-provider/src/main/resources/application.properties
```diff
@@ -339,8 +339,8 @@ management.endpoints.web.cors.allowed-headers=*
 
 # FINERACT-883
 management.info.git.mode=FULL
-management.endpoints.web.exposure.include=${FINERACT_MANAGEMENT_ENDPOINT_WEB_EXPOSURE_INCLUDE:health,info,prometheus}
-
+management.endpoints.web.exposure.include=${FINERACT_MANAGEMENT_ENDPOINT_WEB_EXPOSURE_INCLUDE:health,info,metrics,prometheus}
+management.endpoint.metrics.enabled=true
```
IOhacker: Please add it as a variable FINERACT_MANAGEMENT_ENDPOINT_METRICS_ENABLED and default value set to false, if the variable is set to true then it will be activated.

IOhacker (approved):
Kidnly see my comment

IOhacker (changes requested):
.

fineract-provider/src/main/resources/application.properties
Shivd131: @IOhacker thank you for pointing it out, I have updated it.

IOhacker:
Squash and commit your changes please

Shivd131:
@IOhacker I have squashed the changes into a single commit. 
Thank you

fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/api/RunreportsApiResource.java
adamsaghy, line 1: I dont like the idea of injecting and doing manually the metric collection.

Please use "Spring (Boot 3 / Micrometer Observation): @Observed" or "Classic Micrometer annotations: @Timed" instead.

fineract-provider/src/main/resources/application.properties
```diff
@@ -340,8 +339,8 @@ management.endpoints.web.cors.allowed-headers=*
 
 # FINERACT-883
 management.info.git.mode=FULL
-management.endpoints.web.exposure.include=${FINERACT_MANAGEMENT_ENDPOINT_WEB_EXPOSURE_INCLUDE:health,info,prometheus}
-
+management.endpoints.web.exposure.include=${FINERACT_MANAGEMENT_ENDPOINT_WEB_EXPOSURE_INCLUDE:health,info,metrics,prometheus}
```
adamsaghy: Please dont change this configuration. Metrics are not exposed by default for a reason!

fineract-provider/src/main/resources/application.properties
```diff
@@ -340,8 +339,8 @@ management.endpoints.web.cors.allowed-headers=*
...
+management.endpoint.metrics.enabled=${FINERACT_MANAGEMENT_ENDPOINT_METRICS_ENABLED:false}
```
adamsaghy: I dont think it is valid endpoint anymore, it was changed to be `management.endpoints.web.exposure.include=metrics` no?

adamsaghy (changes requested):
Kindly see my concerns!

Shivd131:
@adamsaghy Thanks for the detailed review.

I have refactored the PR to align with your feedback:

- Reverted application.properties

- Switched to @Timed: Removed the manual Timer boilerplate from the resource file. I added a small MetricsConfig class to enable the TimedAspect bean so the annotation works correctly with Jersey.

- Context on the original approach: The reason I initially used manual instrumentation was to capture dynamic tags like reportName and outputType. For my future projects, determining exactly which specific reports are resource-intensive is crucial for deciding what to offload to a microservice.

The current @Timed implementation aggregates all reports into one metric, which hides that granularity, but I agree this is the cleaner, standard starting point for the codebase. 

I am open to any suggestions/corrections if required

Shivd131:
@IOhacker  I think I made a mistake while copying the license header at the top of the code file earlier, which I suspect made the code fail the check. I’ve corrected it now, and the build check should pass.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/MetricsConfig.java
```diff
@@ -16,15 +16,21 @@
...
-        return properties.getModule().getLoanOrigination().isEnabled();
+@Configuration
+@EnableAspectJAutoProxy
+public class MetricsConfig {
+
+    @Bean
+    public TimedAspect timedAspect(MeterRegistry registry) {
+        return new TimedAspect(registry);
     }
 }
```
adamsaghy, line 36: Are you sure we need this configuration?

Shivd131: @adamsaghy I added this explicit configuration because I was getting unsuccessful build, as the default auto-configuration was not picking it up in this context. the TimedAspect bean is required to process the @Timed annotations; without it, the AOP proxy is not created and the annotation is effectively ignored. (Reading:[This comment from stack overflow](https://stackoverflow.com/questions/56377196/timed-not-working-despite-registering-timedaspect-explicitly-spring-boot-2-1#:~:text=9,to%2Dlive%3A%2010s), and [found it on the docs now](https://docs.micrometer.io/micrometer/reference/concepts/timers.html#:~:text=we%20are%20timing.-,The%20%40Timed%20Annotation,public%20TimedAspect%20timedAspect(MeterRegistry%20registry)%20%7B%0A%20%20%20%20%20%20return%20new%20TimedAspect(registry)%3B%0A%20%20%20%7D%0A%7D,-Copied!))

Is there a preferred existing configuration class where you would like me to move this bean instead, or suggest me a better way of doing it?

would be happy to learn and commit  that way.

PR #5357, opened 2026-01-20
FINERACT-1659: Prevent duplicate savings interest posting in same period

sidshas03:
## Description

This PR fixes the duplicate interest posting issue reported in **FINERACT-1659**, where the **Post Interest** job could post interest multiple times for the same savings account within the same posting period.

### What changed
- Added a safeguard in `SavingsSchedularInterestPoster` to **skip accounts where interest has already been posted for the current period**, by comparing `interestPostedTillDate` with the current business date.
- Updated the transaction isolation level to **`SERIALIZABLE`** to prevent duplicate postings caused by race conditions during parallel scheduler runs.

### Tests
Added two integration tests to validate the behavior:
- `testPostInterestPreventsDuplicateOnSameDay`  
  Ensures running the job twice on the same day results in **only one** interest posting.
- `testPostInterestSkipsCurrentPeriodButAllowsNewPeriod`  
  Ensures the job **skips** reposting within the current period, while still allowing postings in a new period.

### Expected outcome
- Running the Post Interest job multiple times within the same posting period **will not create duplicate transactions**, and the savings account balance remains unchanged after subsequent runs.

### DIFF
--- a/fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
+++ b/fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
@@ -67,10 +67,16 @@ public class SavingsSchedularInterestPoster {
     public void postInterest() throws JobExecutionException {
         if (!savingAccounts.isEmpty()) {
             List<Throwable> errors = new ArrayList<>();
+            LocalDate yesterday = DateUtils.getBusinessLocalDate().minusDays(1);
             for (SavingsAccountData savingsAccountData : savingAccounts) {
                 boolean postInterestAsOn = false;
                 LocalDate transactionDate = null;
                 try {
+                    if (isInterestAlreadyPostedForPeriod(savingsAccountData, yesterday)) {
+                        log.debug("Interest already posted for savings account {} up to date {}, skipping", savingsAccountData.getId(),
+                                savingsAccountData.getSummary().getInterestPostedTillDate());
+                        continue;
+                    }
                     SavingsAccountData savingsAccountDataRet = savingsAccountWritePlatformService.postInterest(savingsAccountData,
                             postInterestAsOn, transactionDate, backdatedTxnsAllowedTill);
                     savingsAccountDataList.add(savingsAccountDataRet);
@@ -109,7 +115,6 @@ public class SavingsSchedularInterestPoster {
             for (SavingsAccountTransactionData savingsAccountTransactionData : savingsAccountTransactionDataList) {
                 if (savingsAccountTransactionData.getId() == null && !MathUtil.isZero(savingsAccountTransactionData.getAmount())) {
                     final String key = savingsAccountTransactionData.getRefNo();
-                    final Boolean isOverdraft = savingsAccountTransactionData.getIsOverdraft();
                     final SavingsAccountTransactionData dataFromFetch = savingsAccountTransactionDataHashMap.get(key);
                     savingsAccountTransactionData.setId(dataFromFetch.getId());
                     if (savingsAccountData.getGlAccountIdForSavingsControl() != 0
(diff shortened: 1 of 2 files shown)


fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
```diff
@@ -63,17 +63,25 @@ public class SavingsSchedularInterestPoster {
     private Collection<SavingsAccountData> savingAccounts;
     private boolean backdatedTxnsAllowedTill;
 
-    @Transactional(isolation = Isolation.READ_UNCOMMITTED, rollbackFor = Exception.class)
+    @Transactional(isolation = Isolation.SERIALIZABLE, rollbackFor = Exception.class)
```
adamsaghy: While the SERIALIZABLE here might be better than the READ_UNCOMMITTED, i dont think we are addressing the underlying issue properly.

**I would rather recommend a different approach:**
- Is it possible to ensure no two `SavingsSchedularInterestPoster` got the very same savings account? If we can ensure no two parallel executed `postInterest` work on the very same savings account, i see no reason why would any double posting and we dont need to enforce SERIALIZABLE isolation which usually means strict and heavy locking of the resource which does not help the performance.

**Alternative solution:**
- Is it possible to introduce constraints which ensure no two interest to be posted on the very same date of the very same savings account? We can even build retry strategy too based on this constraints:
- Read savings account -> post interest -> if fails, refetch the savings account and check again whether interest needed to be posted.

It might go beyond the scope of the original story, but to be frank, this would be the proper way to do it.

TLDR: 1 savings account should be processed by 1 thread entirely and avoid situations where 1 savings account might be processed by many threads.

adamsaghy:
@sidshas03 
```
Execution failed for task ':integration-tests:spotlessJavaCheck'.
      src/test/java/org/apache/fineract/integrationtests/SavingsInterestPostingTest.java
          @@ -523,8 +523,7 @@
           ············schedulerJobHelper.executeAndAwaitJob(POST_INTEREST_JOB_NAME);
           
           ············List<HashMap>·txsAfterFirstPeriod·=·getInterestTransactions(accountId);
          -············Assertions.assertEquals(1,·txsAfterFirstPeriod.size(),
          -····················"Expected·exactly·one·interest·transaction·after·first·posting·period");
          +············Assertions.assertEquals(1,·txsAfterFirstPeriod.size(),·"Expected·exactly·one·interest·transaction·after·first·posting·period");
           
           ············HashMap·summaryAfterFirstPeriod·=·savingsAccountHelper.getSavingsSummary(accountId);
           ············BigDecimal·balanceAfterFirstPeriod·=·BigDecimal.valueOf(((Double)·summaryAfterFirstPeriod.get("accountBalance")));
  Run './gradlew :integration-tests:spotlessApply' to fix these violations.

adamsaghy:
Please review the failing test cases

adamsaghy:
@sidshas03 Please review the failing test cases

adamsaghy:
@sidshas03 Please review the failing test cases

adamsaghy:
@sidshas03 Please review the failing test cases

adamsaghy:
@sidshas03 Please review the failing test cases

sidshas03:
Changes done:
1. Reverted `postInterest()` transaction isolation back to `READ_UNCOMMITTED` (same as develop). With `SERIALIZABLE`, deadlocks were coming during batch update.
2. Removed `hasNewInterestTransactions()` filtering and now always adding `postInterest()` response to batch list, so valid interest/overdraft postings are not getting skipped.
3. Kept duplicate-prevention guard with period check: skip only when `interestPostedTillDate` is not before yesterday (already posted for same period).
4. Removed one unused local variable in batch update method.

Validation run:
`./gradlew :integration-tests:test --tests org.apache.fineract.integrationtests.SavingsInterestPostingTest -PdbType=mysql`

adamsaghy (approved):
LGTM

adamsaghy:
@sidshas03 Would you mind to attach a test case which is doing the following:
- Create a savings account
- Change business date
- Post interest on it
- Change business date
- Post interest manually (run job till current date)
- Change business date
- Run job
- Check whether there was no multiple interest posted

adamsaghy:
@sidshas03 Please squash your commits

sidshas03:
@adamsaghy Done. I squashed the branch into a single commit (`cf66d4c5d`) and force-pushed the PR branch.

galovics (approved):

PR #5382, opened 2026-01-24
WEB-129 "Xbrlmapping" permission is obsolete

parth-sharma-10:
[WEB-129](https://mifosforge.jira.com/jira/software/c/projects/WEB/boards/62?issueType=1&selectedIssue=WEB-129)

XBRL functionality has already been removed from Fineract, however the Xbrlmapping permission was still being seeded via SQL sample data.

This change removes the obsolete XBRL Mapping permission from the seed data to prevent it from appearing in Roles & Permissions. Existing databases were verified locally to ensure the permission no longer appears in the UI.

No functional behavior is affected by this change.
image


IOhacker (changes requested):
Use Apache Fineract Jira, this is not a valid Jira ticket. You should not change the sql files. You should create new migration files. Please use the README.md and the apache.fineract.org for getting involved in the project.

PR #5429, opened 2026-01-31
FINERACT-1934: Replace unbounded SimpleAsyncTaskExecutor with ThreadPoolTaskExecutor

DeathGun44:
## Description

Fixes thread accumulation issue in [applicationEventMulticaster](cci:1://file:///home/deathgun/fineract/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/SpringConfig.java:33:4-38:5) caused by `SimpleAsyncTaskExecutor` creating unbounded threads.

### Problem
The existing `SimpleAsyncTaskExecutor` creates a new thread for every event, leading to thread exhaustion under load (200+ concurrent users). This causes JVM memory exhaustion and service outages.

### Solution
- Replace `SimpleAsyncTaskExecutor` with bounded [ThreadPoolTaskExecutor](cci:1://file:///home/deathgun/fineract/fineract-provider/src/main/java/org/apache/fineract/infrastructure/configuration/async/SpringAsyncConfig.java:32:4-37:5)
- Core pool: 20 threads, Max pool: 100 threads
- Add `DelegatingSecurityContextAsyncTaskExecutor` wrapper for security context propagation
- Graceful shutdown with task completion wait

### Testing
- Application starts successfully
- Events processed on `FineractEvent-*` thread pool
- No thread accumulation observed

## Checklist

- [x] Acknowledge that we will not review PRs that are not passing the build _("green")_
- [x] Add required Swagger annotation and update API documentation (N/A - no API changes)
- [x] [This PR must not be a "code dump"]

### DIFF
--- a/fineract-core/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java
+++ b/fineract-core/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java
@@ -302,6 +302,8 @@ public class FineractProperties {
         private int tenantUpgradeTaskExecutorCorePoolSize;
         private int tenantUpgradeTaskExecutorMaxPoolSize;
         private int tenantUpgradeTaskExecutorQueueCapacity;
+        private int eventTaskExecutorCorePoolSize;
+        private int eventTaskExecutorMaxPoolSize;
     }
 
     @Getter
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/SpringConfig.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/SpringConfig.java
@@ -19,34 +19,71 @@
 
 package org.apache.fineract.infrastructure.core.config;
 
+import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.beans.factory.config.MethodInvokingFactoryBean;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.DependsOn;
 import org.springframework.context.event.SimpleApplicationEventMulticaster;
-import org.springframework.core.task.SimpleAsyncTaskExecutor;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
 import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.security.core.context.SecurityContextHolderStrategy;
+import org.springframework.security.task.DelegatingSecurityContextAsyncTaskExecutor;
 
 @Configuration
 public class SpringConfig {
 
-    @Bean
-    public SimpleApplicationEventMulticaster applicationEventMulticaster() {
-        SimpleApplicationEventMulticaster saem = new SimpleApplicationEventMulticaster();
-        saem.setTaskExecutor(new SimpleAsyncTaskExecutor());
-        return saem;
+    private static final int AWAIT_TERMINATION_SECONDS = 60;
+
+    private final FineractProperties fineractProperties;
+
+    public SpringConfig(FineractProperties fineractProperties) {
+        this.fineractProperties = fineractProperties;
     }
 
-    // The application events (for importing) rely on the inheritable thread local security context strategy
-    // This is NOT compatible with threadpools so if we use threadpools the below will need to be reworked
+    private int getEventExecutorCorePoolSize() {
+        int configured = fineractProperties.getTaskExecutor().getEventTaskExecutorCorePoolSize();
+        if (configured > 0) {
+            return configured;
+        }
+        return Runtime.getRuntime().availableProcessors() * 2;
+    }
+
+    private int getEventExecutorMaxPoolSize() {
+        int configured = fineractProperties.getTaskExecutor().getEventTaskExecutorMaxPoolSize();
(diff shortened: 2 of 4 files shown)


IOhacker (changes requested):
There are formulas and patterns that work for specific tasks. The optimal pool size depends on the workload type (please read about the instance modes that Apache Fineract offers)

There are formulas and patterns that work for specific tasks. The optimal pool size depends on the workload type and the available CPU/cores. 

Please add test cases

https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/ThreadPoolExecutor.html
https://www.baeldung.com/java-threadpooltaskexecutor-core-vs-max-poolsize
https://medium.com/@AlexanderObregon/thread-pool-sizing-in-spring-boot-async-executors-c880eeccf765
https://oneuptime.com/blog/post/2026-01-25-tune-thread-pools-performance/view

DeathGun44:
@IOhacker Thank you for the feedback!

To clarify the scope:
1. For pool sizing - should I externalize to FineractProperties, or add CPU-aware defaults?
2. For tests - are we looking for configuration verification tests, or load/stress tests?

IOhacker:
1. Externalize (if the person requires to override) and add CPU aware
2. How we can check that this change has solved the problem that you have
written in the PR description

"The existing SimpleAsyncTaskExecutor creates a new thread for every event,
leading to thread exhaustion under load (200+ concurrent users). This
causes JVM memory exhaustion and service outages."

The problem that you describe is present in Read, Write, Batch Instance
mode?

https://fineract-academy.com/fineract-instance-types/
https://fineract.apache.org/docs/current/

DeathGun44:
@IOhacker  Thanks for the detailed feedback!

The [applicationEventMulticaster](cci:1://file:///home/deathgun/fineract/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/SpringConfig.java:50:4-57:5) is used in ALL instance modes (Read, Write, Batch) 
because Spring application events are core infrastructure. However, the thread 
accumulation is most critical in **Write** and **Batch** modes where high event 
volumes occur.

Here's proof the fix works - startup logs show events now run on bounded `FineractEvent-*` pool:

2026-01-31 19:19:04.634 - INFO [FineractEvent-8] org.quartz.impl.StdSchedulerFactory : Using default implementation... 2026-01-31 19:19:04.639 - INFO [FineractEvent-8] org.quartz.core.QuartzScheduler : Quartz Scheduler v2.5.0 created. 2026-01-31 19:19:04.641 - INFO [FineractEvent-8] org.quartz.core.QuartzScheduler : Scheduler Scheduler1_$_NON_CLUSTERED started.

**Before:** Threads named `SimpleAsyncTaskExecutor-*` (unbounded, new thread per event)  
**After:** Threads named `FineractEvent-*` (bounded pool, max configurable)
This is verified by the application starting successfully with the new executor. A test for 
"pool size equals X" doesn't add value since misconfiguration would cause startup failure.

With a bounded pool, even 1000 concurrent events will queue instead of creating 1000 threads. 
This is the fundamental fix - replacing unbounded with bounded.

I'll update the PR with:
1. Externalize pool sizes to [FineractProperties](cci:2://file:///home/deathgun/fineract/fineract-core/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java:35:0-710:1) following [TaskExecutorConfig](cci:2://file:///home/deathgun/fineract/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/TaskExecutorConfig.java:27:0-49:1) pattern
2. Add CPU-aware defaults: `corePoolSize = Runtime.getRuntime().availableProcessors() * 2`

If you still want a configuration verification test, I can add one. Let me know.

IOhacker:
@DeathGun44 I think that the problem is not solved, concurrent users for me are persons that have a user name and a password and are using the API Rest. The log that is shown is part of the Job implementation that are used in Batch mode. So then I think that you have to add tests how it behaves in READ, WRITE and BATCH mode under the concurrent number that you have identified that causes issues. I ask this because there is no way to reproduce or test the 200+ concurrent users that is part of the problem description of the PR. Also you can add the log in the Jira ticket showing the error/exception when the issue appears and the tech specs of the server that was used for reproducing the issue.

DeathGun44:
@IOhacker  I understand. The original reporter (2023) is unresponsive, so I'll work on reproducing the issue myself with a load test setup. Will add the thread dumps, error logs, and server specs to the JIRA ticket once complete.

IOhacker:
@DeathGun44 then we can ask to the reported if we can close the ticket because it is stale and there is no way to reproduce it or have more information.

DeathGun44:
@IOhacker   After extensive local testing with 400+ concurrent deposits, I was unable to reproduce the 320 SimpleAsyncTaskExecutor threads described in the ticket.

The notification async path (via applicationEventMulticaster) doesn't trigger in my local environment despite having users with ALL_FUNCTIONS permission in the same office as the savings account.

The fix remains architecturally correct (replacing unbounded SimpleAsyncTaskExecutor with bounded ThreadPoolTaskExecutor), but without production-like data I cannot demonstrate the specific issue from FINERACT-1934.

Given reproduction requires production-scale data, I'll close this PR for now. Happy to revisit if someone can provide a reproducible test case.

PR #5437, opened 2026-02-03
FINERACT-2462: Add github action to enforce one commit per user in a PR

edk12564:
## Description

JIRA: FINERACT-2462

Implemented a Github Action that runs on PR request changes, opens, and closes.

The Github Action checks for:
1) Does each user have a single PR associated with their Github account for that PR?
2) Does each user have a valid git email configured to match their Github email?

The action on failure is to comment in the PR thread with the type of error, as well as show in the error’s logs.

The feature allows for multiple users per PR, but only 1 commit per user.

# Edge Cases and Behavior

Test Cases I have tried:
- If git email does not match any emails registered for the github account, the check will fail. This is because they will have null values for [author.id](http://author.id/). The author fields are required to differentiate users, and make sure they are not bots.
- If there are multiple commits associated with an account, the check will fail. 
- If all goes well, there will be a successful check and a pass.
- When a user has a misconfigured git account, and pushes their code, they will get the error that one of the committers have their git email misconfigured. If they fix and recommit without squashing, this issue will still persist since the old commit is still there. Need to recommend a squash as well to remove the commit history. Fixed now.

Edge Cases covered in code:
- If a PR is made with only bots, like from another Github action, the test should pass.
- If Github API is down, the test should fail.
- If there are both bots and users, only the users commits should be looked at. 

Failure Path (misconfigured git email):
Screenshot 2026-02-06 at 1 26 20 AM

Failure Path (multiple commits per user):
Screenshot 2026-02-06 at 1 31 10 AM

## Checklist

### DIFF
--- a/.github/workflows/pr-one-commit-per-user-check.yml
+++ b/.github/workflows/pr-one-commit-per-user-check.yml
@@ -1,0 +1,57 @@
+name: Fineract PR One Commit Per User Check
+
+
+on:
+ pull_request:
+   types: [opened, reopened, synchronize]
+
+
+permissions:
+ pull-requests: write
+
+
+jobs:
+ verify-commits:
+   name: Validate One Commit Per User
+   runs-on: ubuntu-latest
+   timeout-minutes: 1
+   steps:
+     - name: Verify Commit Policy
+       id: check
+       env:
+         GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+       run: |
+
+         commits=$(gh api "repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/commits") || { echo "::error::GitHub API request failed"; exit 1; }
+
+         if echo "$commits" | jq -e '.[] | select(.author == null)' > /dev/null; then
+           echo "null_authors=true" >> $GITHUB_OUTPUT
+           echo "::error::Some commits have a git email that is not linked to a GitHub account. Please ensure your git email matches one of your GitHub Account emails.\n\nPlease also squash your commits to prevent this message again."
+           exit 1
+         fi
+
+         user_ids=$(echo "$commits" | jq -r '.[] | select(.author.type != "Bot") | .author.id')
+         if echo "$user_ids" | sort | uniq -d | grep -q .; then
+           echo "multiple_commits=true" >> $GITHUB_OUTPUT
+           echo "::error::Multiple commits from the same author have been detected."
+           exit 1
+         fi
+
+         echo "Success: Each author has exactly one commit."
+
+     - name: Comment on PR
+       if: failure()
+       env:
+         GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+       run: |
+
+         if [ "${{ steps.check.outputs.null_authors }}" == "true" ]; then
+           gh pr comment ${{ github.event.pull_request.number }} --repo ${{ github.repository }} --body \
+         $'**One Commit Per User Check Failed**\n\nSome committers have a git email that does not match their GitHub account. Please ensure your git email matches one of your GitHub Account emails. Please also squash your commits to prevent this message again.'
+         fi
+
+
+         if [ "${{ steps.check.outputs.multiple_commits }}" == "true" ]; then
+           gh pr comment ${{ github.event.pull_request.number }} --repo ${{ github.repository }} --body \
+         $'**One Commit Per User Check Failed**\n\nEach user may only have one commit per PR. Please squash your commits.'
+         fi


.github/workflows/pr-one-commit-per-user-check.yml
```diff
@@ -0,0 +1,57 @@
...
+           echo "::error::Multiple commits from the same author have been detected."
+           exit 1
+         fi
+
+         echo "Success: Each author has exactly one commit."
+
+     - name: Comment on PR
+       if: failure()
+       env:
+         GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
Aman-Mittal, line 45: Is token is necessary in this?

edk12564: From my understanding that Github token isn't the same one we use to authenticate our accounts. It's a separate token created per workflow for using the Github API calls in your actions. I believe it is required to access .author.id.

More on this here:
https://docs.github.com/en/actions/tutorials/authenticate-with-github_token

edk12564:
I also just realized must have reverted to a previous version while switching branches. I have an old variable I used that I thought I removed already. Repushing now!!

.github/workflows/pr-one-commit-per-user-check.yml
Aman-Mittal: Seems Ok, Can you also check the workflow of stale.yml, i really like the way it checks prs. This will not run without workflow approval and stale is a cron based job.

edk12564: I'll take a look now!

edk12564: I did a bit of research, and I agree with your view. The Stale workflow has a lot of merits, and I will look into adding to what we have. To summarize though, this is what I found.

1 - The current implementation needs permissions for it to run properly. Stale.yml uses the main repo's context, which allows it to run without manual approval. 

On this note, I found something we could use. If we use pull_request_target instead of pull_request, we could run our workflows with base repo permissions. I am reading about it here: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target.

The main issue for this seems to be that this raises security concerns about running code without prior approval. But I think this might be fine considering we are only looking at the commit repository, PR number, and userIds. We don't actually execute any foreign code. Let me know what you think on this one!

2 - I also see what you mean about running a cronjob. I think this is a good idea as well. However, if we want to run on a cronjob to handle older PRs, the issue is that we probably shouldn't spam comments in case of users abandoning their PR. I think we should implement something like checking the most recent comment, or maybe we won't produce warnings if the no one has committed in the past day or two? I will take a look and let you know what our options are here.

What do you think on those points?

Aman-Mittal: @meonkeys Your input is valued here i will give my input.

for point 1 I think, It doesn;t raise security concern, as we are not executing any external code (eg any suspicious JS file or third party dependency it will be based on implementation)

for point 2, i think we can simply add check that PRs that do not have stale tag can be informed. As per stale information inactive PRs are automatically Closed by stale bot. And to reduce spaming further we can make it comment one time and let it edit the existing comment.

edk12564: Hello @Aman-Mittal , Ashar Khan would like to work on this ticket as well. But before that, should we go ahead with the Cron based feature as well? Because I believe otherwise, the work is completed after my next commit. Once I can figure out these GPG keys.

Aman-Mittal: @edk12564 You are free to colaborate with others. and also mentioned in ticket that dev consensus is currently towards one commit per PR. so please keep track on that.

https://lists.apache.org/thread/f0tx0p7jd9f2nhzpgp75hxk267hrmnd9

edk12564: Ahh, I see. I might have misunderstood the direction at first. You're right, let's wait this one out.

Aman-Mittal: @edk12564 There is another case i found which can affect the ticket and needs discussion, which is signed commit check, while there can be trailers but only one person can sign that commit, if that is not signed properly this can affect either PR check or can cause loosing credit if signer was not careful

edk12564: @Aman-Mittal Hey Aman! Thanks for posting that point. It was a direction I hadn't thought of. I agree that if the direction of the project is to sign all our commits to prove contribution, it's a concern that we can only sign one person. I also brought up the point of Github not recognizing contribution unless it's a commit or co-authorship as well. Let's see what everyone says!

adamsaghy:
```
87 actionable tasks: 36 executed, 37 from cache, 14 up-to-date
Execution failed for task ':rat'.
> A failure occurred while executing org.nosphere.apache.rat.RatWork
   > Apache Rat audit failure - 1 unapproved license
     	See file:///home/runner/work/fineract/fineract/build/reports/rat/index.html

**Comment by edk12564** [comment]
> ```
> 87 actionable tasks: 36 executed, 37 from cache, 14 up-to-date
> Execution failed for task ':rat'.
> > A failure occurred while executing org.nosphere.apache.rat.RatWork
>    > Apache Rat audit failure - 1 unapproved license
>      	See file:///home/runner/work/fineract/fineract/build/reports/rat/index.html
> ```

Sorry about that. I'll fix this now!

File: .github/workflows/pr-one-commit-per-user-check.yml (hunk 1 above)
Comment (adamsaghy) [reply .github/workflows/pr-one-commit-per-user-check.yml:45]: I dont think we need cron job. It will be executed alongside all other checks. When something got changed, it will be re-run. I see no issue here.

I think we are fine to have multiple commits if it came from different authors. The check should catch and handle the situation where:
```
Commit by Author A  -> So far good
Commit by Author A -> Nope, it's a duplication!
```

or 

```
Commit by Author A -> So far good
Commit by Author B -> So far good
Commit by Author A -> Nope, it's a duplication!
```

File: .github/workflows/pr-one-commit-per-user-check.yml [hunk 2]
```diff
@@ -0,0 +1,57 @@
+name: Fineract PR One Commit Per User Check
+
+
+on:
```
adamsaghy, line 4: This should be below no?

```
on:
  pull_request:
    types: [opened, synchronize, reopened]

Comment (edk12564) [reply .github/workflows/pr-one-commit-per-user-check.yml:4]: Hi adam! Thank you for the review. I took a look again, and realized I was responding to the wrong thing. I thought you were referring to the the types, not the pull_request_target. 

I think pull_request_target was used here because it allows us to run this check in the main repo's context, which me and Aman figured was okay considered no foreign code is actually being executed. This would allow us to run the check without needing someone to approve the test first. 

More on this here: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target. Do you think this is okay, or is it more policy to make sure everything is approved first?

File: .github/workflows/pr-one-commit-per-user-check.yml (hunk 1 above)
Comment (edk12564) [reply .github/workflows/pr-one-commit-per-user-check.yml:45]: Hi adam! Really sorry about this but I made a post in the dev mail list before I read this comment. That's completely my bad.

To summarize the post, I wrote on a potential improvement to One Commit per PR that might allow us to keep full credit for everyone involved. We can add co-contributors to Co-Authored-By and also the trailer they performed work in, like Designed-By or Tested-By. This gives them full contribution in Github's eyes without having multiple users in a commit. But the tradeoff is still that each person cannot sign their commit, like they can in One Commit per User as Aman mentioned. 

Would you still like One Commit Per User here given the new information? I will say that the implementation works either way!

File: .github/workflows/pr-one-commit-per-user-check.yml (hunk 2 above)
Comment (adamsaghy) [reply .github/workflows/pr-one-commit-per-user-check.yml:4]: My concern is this newly added github actions were not executed on this PR which means it is targeting the wrong thing, no?

File: .github/workflows/pr-one-commit-per-user-check.yml (hunk 1 above)
Comment (adamsaghy) [reply .github/workflows/pr-one-commit-per-user-check.yml:45]: I have read your recommendations on the mail list. Thank you for sharing there. Hopefully others will chime in. I am just 1 voice from the many ;)

File: .github/workflows/pr-one-commit-per-user-check.yml (hunk 2 above)
Comment (edk12564) [reply .github/workflows/pr-one-commit-per-user-check.yml:4]: I see what you mean with the target! To test this out, I did some investigating on my local fork where I submitted a PR into my own fork's develop branch. 

My setup is: the develop branch has pr-one-commit-per-user.yml and the PR branch does not. I first tested with multiple commits at once, and it properly failed. I also tested the target by squashing my commits in the PR, and the test passed. From the results, it seems to properly target the PR branch, but runs the pull_request_target scripts that live in the develop branch. Some pictures below.

no script in PR:
Screenshot 2026-02-19 at 9 33 10 AM

failed test:
Screenshot 2026-02-19 at 9 47 47 AM

success test:
Screenshot 2026-02-19 at 10 25 21 AM

I think based on this, this one particular PR will not have checks applied. However, if it is merged into the develop branch, any future PR will still be checked, and target the actual PR branch, and not the apache/fineract repo. However, the problem of running without workflow permissions isn't a big issue, so I think we can definitely go either way. Please let me know what you think!

File: .github/workflows/pr-one-commit-per-user-check.yml (hunk 1 above)
Comment (edk12564) [reply .github/workflows/pr-one-commit-per-user-check.yml:45]: Sounds good to me! Thank you sir 👍

File: .github/workflows/pr-one-commit-per-user-check.yml (hunk 2 above)
Comment (adamsaghy) [reply .github/workflows/pr-one-commit-per-user-check.yml:4]: Running this workflow on the target branch is incorrect. You should use the one i shared, imho:

```
on:
  pull_request:
    types: [opened, synchronize, reopened]

edk12564: No problem! I respect your experience here. Just made the change now!

PR #5439, opened 2026-02-03
FINERACT-248: Prevent duplicate SMS campaign names

nickus:
## Summary
Add validation to check for duplicate campaign names before creating or updating SMS campaigns. This provides a user-friendly error message instead of relying on database constraint violations.

## Changes
- Add `existsByCampaignName()` and `existsByCampaignNameAndIdNot()` methods to `SmsCampaignRepository`
- Validate campaign name uniqueness in `create()` method before saving
- Validate campaign name uniqueness in `update()` method when name changes
- Add `SmsCampaignNameAlreadyExistsException` for clear error messages

## Test plan
- [x] Integration tests added for:
  - Creating a campaign with a duplicate name should fail with appropriate error
  - Creating campaigns with unique names should succeed

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/campaigns/sms/service/SmsCampaignWritePlatformServiceJpaImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/campaigns/sms/service/SmsCampaignWritePlatformServiceJpaImpl.java
@@ -47,6 +47,7 @@ import org.apache.fineract.infrastructure.campaigns.sms.domain.SmsCampaign;
 import org.apache.fineract.infrastructure.campaigns.sms.domain.SmsCampaignRepository;
 import org.apache.fineract.infrastructure.campaigns.sms.exception.SmsCampaignMustBeClosedToBeDeletedException;
 import org.apache.fineract.infrastructure.campaigns.sms.exception.SmsCampaignMustBeClosedToEditException;
+import org.apache.fineract.infrastructure.campaigns.sms.exception.SmsCampaignNameAlreadyExistsException;
 import org.apache.fineract.infrastructure.campaigns.sms.exception.SmsCampaignNotFound;
 import org.apache.fineract.infrastructure.campaigns.sms.serialization.SmsCampaignValidator;
 import org.apache.fineract.infrastructure.core.api.JsonCommand;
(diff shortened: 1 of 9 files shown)


fineract-provider/src/main/java/org/apache/fineract/infrastructure/campaigns/sms/service/SmsCampaignWritePlatformServiceJpaImpl.java
```diff
@@ -104,20 +105,32 @@ public class SmsCampaignWritePlatformServiceJpaImpl implements SmsCampaignWriteP
...
-        this.smsCampaignRepository.saveAndFlush(smsCampaign);
+        try {
+            final AppUser currentUser = this.context.authenticatedUser();
+            this.smsCampaignValidator.validateCreate(command.json());
 
-        return new CommandProcessingResultBuilder().withCommandId(command.commandId()).withEntityId(smsCampaign.getId()).build();
+            final String campaignName = command.stringValueOfParameterNamed(SmsCampaignValidator.campaignName);
+            if (this.smsCampaignRepository.existsByCampaignName(campaignName)) {
+                throw new SmsCampaignNameAlreadyExistsException(campaignName);
+            }
```
galovics, line 115: I think the default isolation level we use is read_committed, meaning only committed things will be revealed by this check. It not guarantees anything.
2 transactions concurrently writing the same campaign name is possible.

If you wanna prevent it, ensure having a unique key on the DB level. This is good for a best-effort check but it's not bullet-proof,.

galovics (changes requested):

fineract-provider/src/main/java/org/apache/fineract/infrastructure/campaigns/sms/service/SmsCampaignWritePlatformServiceJpaImpl.java
nickus: Actually, there is already a unique key on the DB level. You might see further I catch DataIntegrityViolationException and throw the same exception.

see realCause.getMessage().contains("campaign_name_UNIQUE")

nickus: @galovics could you take a fresh look on this please

adamsaghy: The mentioned unique constraint is not existing. It's mentioned on the entity, but in reality it does not exist. Please add this missing unique constraint.

adamsaghy (changes requested):
Kindly review my concerns!

fineract-provider/src/main/java/org/apache/fineract/infrastructure/campaigns/sms/service/SmsCampaignWritePlatformServiceJpaImpl.java
nickus: Good catch @adamsaghy, you're right — the @UniqueConstraint annotation on the entity was just Hibernate metadata with no actual DB constraint behind it (since DDL auto-generation is disabled). I've added a Liquibase migration to create the campaign_name_UNIQUE constraint on the sms_campaign table.

adamsaghy:
@nickus Please rebase this PR.

nickus:
So many flaky tests...

nickus:
@adamsaghy finally green, could you please take a new look and merge if everything is okay?

PR #5523, opened 2026-02-21
FINERACT-1153: Split oversized last reschedule installment

airajena:
## Summary
This PR implements `FINERACT-1153` by adding configurable handling for loan reschedule cases where the final installment becomes much larger than prior installments.

## Problem
For some loan reschedule flows (for example reduced EMI / interest appropriation scenarios), the generated schedule can produce an oversized last installment, which is not desirable operationally.

## Solution
Introduced a new global configuration and reschedule logic to automatically split an oversized last installment by extending repayment periods until the last installment is in line.

### New global configuration
- Key: `split-large-last-installment-on-loan-reschedule`
- Default: `false` (disabled)
- Scope: tenant global configuration

### Behavior
- When config is `false`: existing behavior is unchanged.
- When config is `true`:
  - During loan reschedule approval, schedule generation checks whether the final installment is larger than prior installments.
  - If yes, it auto-adds repayment-period extension variation(s), regenerates the schedule, and repeats until:
    - last installment is no longer oversized, or
    - safety cap is reached.

### Safety guard
- Added max extension loop guard to prevent infinite adjustments.

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/rescheduleloan/service/LoanRescheduleRequestWritePlatformServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/rescheduleloan/service/LoanRescheduleRequestWritePlatformServiceImpl.java
@@ -20,6 +20,7 @@ package org.apache.fineract.portfolio.loanaccount.rescheduleloan.service;
 
 import java.math.BigDecimal;
 import java.math.MathContext;
+import java.math.RoundingMode;
 import java.time.LocalDate;
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
@@ -35,6 +36,7 @@ import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.fineract.infrastructure.codes.domain.CodeValue;
 import org.apache.fineract.infrastructure.codes.domain.CodeValueRepositoryWrapper;
+import org.apache.fineract.infrastructure.configuration.domain.ConfigurationDomainService;
 import org.apache.fineract.infrastructure.core.api.JsonCommand;
 import org.apache.fineract.infrastructure.core.data.ApiParameterError;
 import org.apache.fineract.infrastructure.core.data.CommandProcessingResult;
@@ -47,6 +49,7 @@ import org.apache.fineract.infrastructure.event.business.domain.loan.LoanBalance
 import org.apache.fineract.infrastructure.event.business.domain.loan.LoanRescheduledDueAdjustScheduleBusinessEvent;
 import org.apache.fineract.infrastructure.event.business.service.BusinessEventNotifierService;
 import org.apache.fineract.infrastructure.security.service.PlatformSecurityContext;
+import org.apache.fineract.organisation.monetary.domain.MonetaryCurrency;
 import org.apache.fineract.organisation.monetary.domain.MoneyHelper;
 import org.apache.fineract.portfolio.loanaccount.data.LoanTermVariationsData;
 import org.apache.fineract.portfolio.loanaccount.data.ScheduleGeneratorDTO;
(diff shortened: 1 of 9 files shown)


adamsaghy:
@airajena Please rebase the PR (sorry we have a high number of PRs nowadays)!

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/rescheduleloan/service/LoanRescheduleRequestWritePlatformServiceImpl.java
```diff
@@ -484,6 +495,65 @@ public CommandProcessingResult approve(JsonCommand jsonCommand) {
...
+        for (int i = 0; i < rescheduledInstallments.size() - 1; i++) {
+            final BigDecimal dueAmount = rescheduledInstallments.get(i).getDue(currency).getAmount();
+            if (dueAmount.compareTo(maxPriorDueAmount) > 0) {
+                maxPriorDueAmount = dueAmount;
+            }
+        }
+
+        final BigDecimal lastInstallmentDueAmount = lastInstallment.getDue(currency).getAmount();
+        final BigDecimal minimumMaterialDifference = BigDecimal.ONE.movePointLeft(currency.getDigitsAfterDecimal());
+        return lastInstallmentDueAmount.subtract(maxPriorDueAmount).compareTo(minimumMaterialDifference) > 0;
```
adamsaghy, line 659: Could you help me understand the logic behind this?

I’m curious about what constitutes “too large” in this context.

airajena: too large here means the last rescheduled installment becomes materially higher than the rest, not just different due to rounding.

My current Logic in this method:
We only inspect installments in the rescheduled range (exclude down payment + due dates before rescheduleFromDate). We take the max due amount among all rescheduled installments except the last. We compare last due vs that max prior due. We treat it as “too large” only if the last due is greater by more than one minor currency unit. So for a 2-decimal currency, the threshold is > 0.01 (strictly greater), which intentionally ignores rounding noise and only flags a real ballooning last installment.

adamsaghy: I dont see the value to create a new period with 1 cent or 10 cent EMI just because the last installment was higher slightly then the rest of the periods...

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanRescheduleOnDecliningBalanceLoanTest.java
```diff
@@ -475,6 +489,45 @@ private void createAndApproveLoanRescheduleRequestForRecoverInterestFirstAndFixe
 
     }
 
+    private void createAndApproveLoanRescheduleRequestForRecoverInterestFirstAndFixedEMIAndVerifyLargeLastInstallmentSplit() {
```
adamsaghy: I dont like this test case. It is no help to understand exactly what is happening and why... Can you please rather write an E2E test case which covers the before and after of the repayment schedule balances of th reschedule?

adamsaghy (changes requested):
Kindly see my concerns!

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/rescheduleloan/service/LoanRescheduleRequestWritePlatformServiceImpl.java
airajena: I updated the logic so we only split when the last installment is materially larger. previous logic wasn't useful, thanks for help, I also updated the test to reflect the real E2E effect

adamsaghy: Can you please point me to the location of the test which covers the repayment schedule and balances?

adamsaghy:
@airajena Please rebase

adamsaghy (changes requested):
Can you please point me to the location of the test which covers the repayment schedule and balances?

Alternatively, you can write an E2E test case, which clearly shows the before and after repayment schedule and balances, please.

adamsaghy:
@airajena Gentle reminder.

adamsaghy:
@airajena Please rebase.

airajena:
> @airajena Please rebase.
really sorry, i was bit busy writting the gsoc proposal, will clear my backlogs , for now 
I added an E2E test that covers exactly the before/after repayment schedule and balances here:
integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanRescheduleOnDecliningBalanceLoanTest.java
Method: testSplitLargeLastInstallmentOnLoanRescheduleE2e()

adamsaghy (changes requested):
@airajena Can you please draft me an example about how this work and some scenarios that explain it to me? I would like to see balances and repayment schedule, etc...

PR #5565, opened 2026-03-01
FINERACT-1597: Fix validation error for well-formed term deposit chart slabs

airajena:
## Summary
This PR fixes a validation bug where updating/creating term deposit products with valid chart slabs could fail with:

`validation.msg.recurringdeposit.chart.slabs.range.end.incorrect`

The issue appeared for well-formed terminal slabs (for example: `fromPeriod=0`, `toPeriod=3`).

## Root Cause
Chart slab validation in `InterestRateChart.validateChartSlabs(...)` treated certain end-range values as invalid even when there was no subsequent slab requiring continuity validation.

## Changes
- Removed incorrect end-range validation branches
- Removed obsolete helper method no longer used by validation

## Scope
- No API contract changes
- No schema/Liquibase changes
- No behavior changes outside chart slab validation in term deposit interest chart domain

## Validation
Executed locally:
- `./gradlew :fineract-savings:test --tests org.apache.fineract.portfolio.interestratechart.domain.InterestRateChartValidationTest` 
- `./gradlew spotlessApply spotbugsMain spotbugsTest checkstyleMain checkstyleTest` 

## Risk
Low.  
Changes are localized to chart validation logic + one focused regression test.

### DIFF
--- a/fineract-savings/src/main/java/org/apache/fineract/portfolio/interestratechart/domain/InterestRateChart.java
+++ b/fineract-savings/src/main/java/org/apache/fineract/portfolio/interestratechart/domain/InterestRateChart.java
@@ -177,9 +177,6 @@ public class InterestRateChart extends AbstractPersistableCustom<Long> {
 
                     }
                 }
-            } else if (iSlabs.slabFields().isNotProperPriodEnd()) {
-                baseDataValidator.failWithCodeNoParameterAddedToErrorCode("chart.slabs.range.end.incorrect", iSlabs.slabFields().toPeriod(),
-                        iSlabs.slabFields().getAmountRangeTo());
             }
         }
     }
--- a/fineract-savings/src/main/java/org/apache/fineract/portfolio/interestratechart/domain/InterestRateChartSlabFields.java
+++ b/fineract-savings/src/main/java/org/apache/fineract/portfolio/interestratechart/domain/InterestRateChartSlabFields.java
@@ -247,11 +247,6 @@ public class InterestRateChartSlabFields {
                 && !(interestRateChartSlabFields.fromPeriod.equals(1) || interestRateChartSlabFields.fromPeriod.equals(0));
     }
 
-    public boolean isNotProperPriodEnd() {
-        return !(this.toPeriod == null && this.amountRangeTo == null);
-
-    }
-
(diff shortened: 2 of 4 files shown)


IOhacker (approved):
LGTM

fineract-savings/src/main/java/org/apache/fineract/portfolio/interestratechart/domain/InterestRateChart.java
```diff
@@ -177,9 +177,6 @@ public void validateChartSlabs(DataValidatorBuilder baseDataValidator) {
 
                     }
                 }
-            } else if (iSlabs.slabFields().isNotProperPriodEnd()) {
```
adamsaghy, line 180: why is this removed? was it incorrect?

fineract-savings/src/main/java/org/apache/fineract/portfolio/interestratechart/domain/InterestRateChartSlabFields.java
```diff
@@ -247,11 +247,6 @@ public static boolean isNotProperPeriodStart(InterestRateChartSlabFields interes
                 && !(interestRateChartSlabFields.fromPeriod.equals(1) || interestRateChartSlabFields.fromPeriod.equals(0));
     }
 
-    public boolean isNotProperPriodEnd() {
```
adamsaghy, line 250: why is this removed? was it incorrect?

fineract-savings/src/main/java/org/apache/fineract/portfolio/interestratechart/domain/InterestRateChart.java
airajena: To be clear, I removed that intentionally because that branch effectively became unreachable after we tightened the slab-range validation. Previously, isNotProperPriodEnd() acted as a bit of a fallback safety net. But with the way the validation flow works now, we’re validating period bounds (consistency between fromPeriod and toPeriod) much earlier. On top of that, the slab ordering and continuity checks already catch and reject any invalid end-period combinations. The issue was that keeping that extra check actually started tripping up and rejecting valid edge cases—specifically when toPeriod is open-ended or derived by the chart boundaries. So, the goal was really to get rid of duplicate/conflicting validation rather than relaxing the rules.

PR #5578, opened 2026-03-03
FINERACT-1371: Fix hook event payload exposing wrong resourceId for C…

SamaSVM:
## Description

 Fix hook event payload incorrectly exposing resourceId equal to clientId for single-row CLIENT datatable entries

## Checklist

### DIFF
--- a/fineract-core/src/main/java/org/apache/fineract/commands/service/SynchronousCommandProcessingService.java
+++ b/fineract-core/src/main/java/org/apache/fineract/commands/service/SynchronousCommandProcessingService.java
@@ -361,7 +361,9 @@ public class SynchronousCommandProcessingService implements CommandProcessingSer
 
                     reqmap.put("officeId", resultCopy.getOfficeId());
                     reqmap.put("clientId", resultCopy.getClientId());
+                    reqmap.put("resourceId", resultCopy.getResourceId());
                     resultCopy.setOfficeId(null);
+                    resultCopy.setResourceId(null);
                     reqmap.put("response", resultCopy);
                 } else if (result instanceof ErrorInfo ex) {
                     reqmap.put("status", "Exception");
--- a/fineract-core/src/main/java/org/apache/fineract/infrastructure/core/data/CommandProcessingResult.java
+++ b/fineract-core/src/main/java/org/apache/fineract/infrastructure/core/data/CommandProcessingResult.java
@@ -21,6 +21,7 @@ package org.apache.fineract.infrastructure.core.data;
 import java.io.Serializable;
 import java.util.Map;
 import lombok.Getter;
+import lombok.Setter;
 import lombok.ToString;
 import org.apache.fineract.infrastructure.core.domain.ExternalId;
 
@@ -32,12 +33,14 @@ import org.apache.fineract.infrastructure.core.domain.ExternalId;
 public class CommandProcessingResult implements Serializable {
 
     private final Long commandId;
+    @Setter
     private Long officeId;
     private final Long groupId;
     private final Long clientId;
     private final Long loanId;
     private final Long savingsId;
-    private final Long resourceId;
+    @Setter
+    private Long resourceId;
     private final Long subResourceId;
     private final String transactionId;
     private final Map<String, Object> changes;
@@ -46,6 +49,7 @@ public class CommandProcessingResult implements Serializable {
     private final Long productId;
     private final Long gsimId;
     private final Long glimId;
+    @Setter
     private Boolean rollbackTransaction;
     private final ExternalId resourceExternalId;
     private final ExternalId subResourceExternalId;
@@ -137,10 +141,6 @@ public class CommandProcessingResult implements Serializable {
         return new CommandProcessingResult(null, null, null, null);
     }
 
-    public void setOfficeId(final Long officeId) {
-        this.officeId = officeId;
-    }
-
     public Map<String, Object> getChanges() {
         Map<String, Object> checkIfEmpty = null;
         if (this.changes != null && !this.changes.isEmpty()) {
@@ -156,10 +156,6 @@ public class CommandProcessingResult implements Serializable {
 
     public boolean isRollbackTransaction() {
         return this.rollbackTransaction != null && this.rollbackTransaction;
-    }
-
-    public void setRollbackTransaction(Boolean rollbackTransaction) {
-        this.rollbackTransaction = rollbackTransaction;
     }
 
     private ExternalId setExternalIdOrNull(ExternalId externalId) {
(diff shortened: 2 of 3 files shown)


fineract-core/src/main/java/org/apache/fineract/infrastructure/core/data/CommandProcessingResult.java
```diff
@@ -141,6 +141,10 @@ public void setOfficeId(final Long officeId) {
         this.officeId = officeId;
     }
 
+    public void setResourceId(Long resourceId) {
```
IOhacker: What about using Lombok ?

IOhacker (changes requested):

fineract-core/src/main/java/org/apache/fineract/infrastructure/core/data/CommandProcessingResult.java
SamaSVM: Good point, I’ve updated it.

fineract-core/src/main/java/org/apache/fineract/commands/service/SynchronousCommandProcessingService.java
```diff
@@ -361,7 +361,9 @@ protected void publishHookEvent(final String entityName, final String actionName
 
                     reqmap.put("officeId", resultCopy.getOfficeId());
                     reqmap.put("clientId", resultCopy.getClientId());
+                    reqmap.put("resourceId", resultCopy.getResourceId());
                     resultCopy.setOfficeId(null);
+                    resultCopy.setResourceId(null);
```
adamsaghy, line 366: I dont get these changes...

adamsaghy: it looks incorrect to blindly `null` resource id...

adamsaghy (changes requested):
I dont really understand the changes and the reasons.... 

Can you give me some examples and explanation?

SamaSVM:
@adamsaghy From the task I can see that **response.resourceId is always the same as response.clientId**.                                                                                                                                                                  After analysis I found that in DatatableWriteServiceImpl at line 1159 we set:                                                                                                                                                                                   
 ```
 Long resourceId = appTableId;
  if (datatableUtil.isMultirowDatatable(columnHeaders)) {
      resourceId = sqlGenerator.fetchPK(keyHolder);
  }
```

If we don't enter the if block (single-row datatable) then resourceId = appTableId = clientId for CLIENT entity. As I understand this is the reason why response.resourceId == response.clientId.
I think the problem is in publishHookEvent where resourceId was not promoted to the root level and not cleared from response, unlike officeId.
So I decided to apply the same approach as for officeId — promote resourceId to the root level and clear it in response.

Note: I am still learning this codebase so I might have got something wrong.

adamsaghy:
> @adamsaghy From the task I can see that **response.resourceId is always the same as response.clientId**. After analysis I found that in DatatableWriteServiceImpl at line 1159 we set:
> 
> ```
> Long resourceId = appTableId;
>  if (datatableUtil.isMultirowDatatable(columnHeaders)) {
>      resourceId = sqlGenerator.fetchPK(keyHolder);
>  }
> ```
> 
> If we don't enter the if block (single-row datatable) then resourceId = appTableId = clientId for CLIENT entity. As I understand this is the reason why response.resourceId == response.clientId. I think the problem is in publishHookEvent where resourceId was not promoted to the root level and not cleared from response, unlike officeId. So I decided to apply the same approach as for officeId — promote resourceId to the root level and clear it in response.
> 
> Note: I am still learning this codebase so I might have got something wrong.

I am still not seeing what was the issue.

For datatables where the relation is 1 to 1, the client id is the PK for the datatable entry as well. Why would it be an issues if the clientId and resourceId are matching? What if both are id:2 or 12...

Can you help me to understand why you think it is incorrect? Also removing ids from the response sounds really not recommended approach. Why should we clear them?

SamaSVM:
> I am still not seeing what was the issue.
> 
> For datatables where the relation is 1 to 1, the client id is the PK for the datatable entry as well. Why would it be an issues if the clientId and resourceId are matching? What if both are id:2 or 12...
> 
> Can you help me to understand why you think it is incorrect? Also removing ids from the response sounds really not recommended approach. Why should we clear them?

Sorry, it’s probably my fault. I assumed that since the FINERACT-1371 ticket was created and currently exists in Jira, the issue “response.resourceId is always the same as response.clientId” is relevant here. Is that correct?

adamsaghy:
> > I am still not seeing what was the issue.
> > For datatables where the relation is 1 to 1, the client id is the PK for the datatable entry as well. Why would it be an issues if the clientId and resourceId are matching? What if both are id:2 or 12...
> > Can you help me to understand why you think it is incorrect? Also removing ids from the response sounds really not recommended approach. Why should we clear them?
> 
> Sorry, it’s probably my fault. I assumed that since the [FINERACT-1371](https://issues.apache.org/jira/browse/FINERACT-1371) ticket was created and currently exists in Jira, the issue “response.resourceId is always the same as response.clientId” is relevant here. Is that correct?

If the client and the datatable is 1 to 1 relation, clientId and resourceId will be the same. If the client and the datatable relation is 1 to Many, it might be different.

SamaSVM:
> > > I am still not seeing what was the issue.
> > > For datatables where the relation is 1 to 1, the client id is the PK for the datatable entry as well. Why would it be an issues if the clientId and resourceId are matching? What if both are id:2 or 12...
> > > Can you help me to understand why you think it is incorrect? Also removing ids from the response sounds really not recommended approach. Why should we clear them?
> > 
> > 
> > Sorry, it’s probably my fault. I assumed that since the [FINERACT-1371](https://issues.apache.org/jira/browse/FINERACT-1371) ticket was created and currently exists in Jira, the issue “response.resourceId is always the same as response.clientId” is relevant here. Is that correct?
> 
> If the client and the datatable is 1 to 1 relation, clientId and resourceId will be the same. If the client and the datatable relation is 1 to Many, it might be different.



> > > I am still not seeing what was the issue.
> > > For datatables where the relation is 1 to 1, the client id is the PK for the datatable entry as well. Why would it be an issues if the clientId and resourceId are matching? What if both are id:2 or 12...
> > > Can you help me to understand why you think it is incorrect? Also removing ids from the response sounds really not recommended approach. Why should we clear them?
> > 
> > 
> > Sorry, it’s probably my fault. I assumed that since the [FINERACT-1371](https://issues.apache.org/jira/browse/FINERACT-1371) ticket was created and currently exists in Jira, the issue “response.resourceId is always the same as response.clientId” is relevant here. Is that correct?
> 
> If the client and the datatable is 1 to 1 relation, clientId and resourceId will be the same. If the client and the datatable relation is 1 to Many, it might be different.

Understood, thank you.

PR #5632, opened 2026-03-16
FINERACT-2529: pass externalId instead of null resolvedClientId in Cl…

mansi75:
## Problem
In `ClientApiResource.java`, the `getResolvedClientId` method throws a `ClientNotFoundException` using `resolvedClientId` variable. However, this block of code only executes when `resolvedClientId` is null, resulting in an unhelpful error message: "Client with identifier null does not exist."

## Fix
Passed the `externalId` string to the exception instead of `resolvedClientId`, so the error message accurately reflects which external identifier failed to resolve.

### DIFF
--- a/fineract-core/src/main/java/org/apache/fineract/portfolio/client/exception/ClientNotFoundException.java
+++ b/fineract-core/src/main/java/org/apache/fineract/portfolio/client/exception/ClientNotFoundException.java
@@ -32,7 +32,7 @@ public class ClientNotFoundException extends AbstractPlatformResourceNotFoundExc
     }
 
     public ClientNotFoundException(final ExternalId externalId) {
-        super("error.msg.client.id.invalid", "Client with identifier " + externalId + " does not exist", externalId);
+        super("error.msg.client.id.invalid", "Client with identifier " + externalId.getValue() + " does not exist", externalId.getValue());
     }
 
     public ClientNotFoundException() {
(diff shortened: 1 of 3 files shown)


adamsaghy:
```
  
  Failed scenarios:
  file:///home/runner/work/fineract/fineract/fineract-e2e-tests-runner/src/test/resources/features/BatchApi.feature:25 # Verify Batch API call in case of enclosing transaction is TRUE and one of the steps fails
  
  13 Scenarios (1 failed, 12 passed)
  73 Steps (1 failed, 1 skipped, 71 passed)
  2m24.370s
  
  
  org.opentest4j.AssertionFailedError: [Not the expected error message in error body: Actual error message is: Client with identifier org.apache.fineract.infrastructure.core.domain.ExternalId@f9933d45 does not exist. Expected error message is: Client with identifier null does not exist] 
  expected: "Client with identifier null does not exist"
   but was: "Client with identifier org.apache.fineract.infrastructure.core.domain.ExternalId@f9933d45 does not exist"
  	at org.apache.fineract.test.stepdef.common.BatchApiStepDef.clientNotCreated(BatchApiStepDef.java:932)
  	at ✽.Nr. 1 Client creation was rolled back(file:///home/runner/work/fineract/fineract/fineract-e2e-tests-runner/src/test/resources/features/BatchApi.feature:28)

mansi75:
@adamsaghy I have fixed the above BatchApi failed check. The only check now failing is Fineract PR One Commit Per User Check. I have done the fix for that in the below PR
https://github.com/apache/fineract/pull/5644

adamsaghy:
@mansi75 I a not sure why but for all of your commits it says "unverified"... :/

mansi75:
@adamsaghy It was showing 'unverified' because I didn't add my GPG key on github. I have added the GPG key. It is verified now. Sorry for the confusion

adamsaghy:
@mansi75 I understand we are moving away from 1 commit per PR, but i still kindly asking you to squash the last 5-6 commits of yours as most of them fixing "trailing space issues", which is simply just "noise".... @meonkeys Thoughts?

mansi75:
@adamsaghy Got your point. I have squashed the last 6 commits. May you please kindly check if it looks fine

meonkeys:
Thanks for doing that, @mansi75.

@adamsaghy agreed, that's sounds like a great case for a contributor to do a local rebase.

The general guidance in FSIP-7 is the best to follow. I'll update `CONTRIBUTING.md` today, too.

PR #5639, opened 2026-03-17
FINERACT-2538: Fix locale-sensitive string operations in EnumValueVal…

FK-Christian:
## Description

  Summary                                                                                                                                                                           
                                                                                                                                                                                    
  - Added explicit Locale.US to all toLowerCase() calls in EnumValueValidator to avoid locale-dependent behaviour.                                                                  
  - Fixed the String.format(String, Object...) compiler warning by specifying Locale.US as the first argument.
  - Improved the constraint violation message to include the rejected value and the full list of accepted enum values, replacing the generic annotation-level message.              
                                                                                                                                                                                    
  Motivation                                                                                                                                                                        
                                                                                                                                                                                    
  Calling String.toLowerCase() and String.format() without an explicit Locale produces compiler warnings and can yield inconsistent results on systems with a non-English default   
  locale (e.g. Turkish, where I lowercases to ı rather than i). Since enum constant names are always ASCII, Locale.US is the correct and safe choice here.
                                                                                                                                                                                    
  Changes                                                   

 EnumValueValidator.java
 - Added Locale.US to toLowerCase() calls
 - Added String.format(Locale.US, ...) for the constraint violation message
 - Added import java.util.Locale
 
                                                                                                                                                                                    
  Test plan                                                                                                                                                                         
                                                                                                                                                                                    
  - Existing unit/integration tests pass.                                                                                                                                           
  - Manually verify that an invalid enum value returns a descriptive message listing the accepted values.
  - Verify no new compiler warnings are raised by static analysis.

### DIFF
--- a/fineract-validation/src/main/java/org/apache/fineract/validation/constraints/EnumValueValidator.java
+++ b/fineract-validation/src/main/java/org/apache/fineract/validation/constraints/EnumValueValidator.java
@@ -21,6 +21,7 @@ package org.apache.fineract.validation.constraints;
 import jakarta.validation.ConstraintValidator;
 import jakarta.validation.ConstraintValidatorContext;
 import java.util.Arrays;
+import java.util.Locale;
 import java.util.Set;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


adamsaghy:
Duplicate: https://issues.apache.org/jira/browse/FINERACT-2538

PR #5550, opened 2026-02-26
FINERACT-1659: Add optimistic lock to savings interest posting batch …

nickus:
…UPDATE

The root cause of duplicate interest postings was fixed in FINERACT-2394 (commit 5281be4b2), which removed a race condition where both the main thread and a fetchData callable concurrently populated a non-thread-safe ArrayDeque in PostInterestForSavingTasklet.

This commit adds defense-in-depth by wiring the existing @Version column into the raw JDBC batch UPDATE in SavingsSchedularInterestPoster, which bypasses JPA's optimistic lock. The batch UPDATE now includes AND version=? in its WHERE clause and version = version + 1 in SET. If a concurrent modification is detected (updateCounts[i] == 0), the account is skipped and logged rather than double-posted.

Changes:
- SavingsAccountData: add transient version field
- SavingsAccountReadPlatformServiceImpl: select sa.version and extract it in the ResultSetExtractor
- SavingsSchedularInterestPoster: add version check to batch UPDATE, skip accounts with version mismatch, filter downstream operations
- New integration test: testPostInterestJobRunTwiceSameDayNoDuplicate

## Description

Describe the changes made and why they were made. (Ignore if these details are present on the associated Apache Fineract JIRA ticket.)

## Checklist

Please make sure these boxes are checked before submitting your pull request - thanks!

- [ ] Write the commit message as per [our guidelines](https://github.com/apache/fineract/blob/develop/CONTRIBUTING.md#pull-requests)
- [ ] Acknowledge that we will not review PRs that are not passing the build _("green")_ - it is your responsibility to get a proposed PR to pass the build, not primarily the project's maintainers.
- [ ] Create/update [unit or integration tests](https://fineract.apache.org/docs/current/#_testing) for verifying the changes made.
- [ ] Follow our [coding conventions](https://cwiki.apache.org/confluence/display/FINERACT/Coding+Conventions).
- [ ] Add required Swagger annotation and update API documentation at fineract-provider/src/main/resources/static/legacy-docs/apiLive.htm with details of any API changes
- [ ] [This PR must not be a "code dump"](https://cwiki.apache.org/confluence/display/FINERACT/Pull+Request+Size+Limit). Large changes can be made in a branch, with assistance. Ask for help on the [developer mailing list](https://fineract.apache.org/#contribute).

Your assigned reviewer(s) will follow our [guidelines for code reviews](https://cwiki.apache.org/confluence/display/FINERACT/Code+Review+Guide).

PR #5646, opened 2026-03-18
FINERACT-2533: Add E2E test scenarios for SavingsAccount insufficient balance withdrawal

awaneetdecoder:
## Summary

This PR adds E2E test scenarios for SavingsAccount as suggested by @adamsaghy 
in #5623.

### SavingsAccount.feature
- Added Scenario C2441: Withdrawal fails when account has insufficient balance

### SavingsAccountStepDef.java  
- Added step: `Client tries to withdraw {double} {string} from savings account on {string} date and expects an error`
- Added step: `The savings account withdrawal error response has an HTTP status of {int}`
- Added step: `The savings account withdrawal developer message contains {string}`

## Test Design Decisions

- Used `try/catch FeignException` pattern consistent with existing codebase
- Added `Assertions.fail()` inside try block to prevent false positives
- Dynamic currency via ternary operator — avoids hardcoding EUR/USD
- HTTP 403 confirmed from `AbstractPlatformDomainRuleException`
- Error code `insufficient.account.balance` confirmed from `InsufficientAccountBalanceException.java`

## Closes
Supersedes #5623

PR #5671, opened 2026-03-22
FINERACT-1659: Fix optimistic locking in savings interest posting batch job

awaneetdecoder:
## Problem
The savings interest posting batch job could post duplicate interest 
when two instances ran concurrently. PR #5550 attempted a fix but 
adamsaghy correctly identified there was no rollback mechanism — 
version mismatch was silently skipped.

## Fix
- SQL uses WHERE id=? AND version=? to detect concurrent modification
- version = version + 1 increments on each successful update
- updateCounts[i] == 0 detected as concurrent modification
- ConcurrentModificationException thrown on version mismatch
- @Transactional on postInterest() rolls back entire batch

## Files Changed
- SavingsAccountData.java — added version field
- SavingsAccountReadPlatformServiceImpl.java — reads version from DB
- SavingsSchedularInterestPoster.java — core fix
- SavingsSchedularInterestPosterTest.java — unit tests
- SavingsInterestPostingJobIntegrationTest.java — integration tests

Fixes FINERACT-1659
Supersedes PR #5550

### DIFF
--- a/fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
+++ b/fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
@@ -67,16 +67,10 @@ public class SavingsSchedularInterestPoster {
     public void postInterest() throws JobExecutionException {
         if (!savingAccounts.isEmpty()) {
             List<Throwable> errors = new ArrayList<>();
-            LocalDate yesterday = DateUtils.getBusinessLocalDate().minusDays(1);
             for (SavingsAccountData savingsAccountData : savingAccounts) {
                 boolean postInterestAsOn = false;
                 LocalDate transactionDate = null;
                 try {
-                    if (isInterestAlreadyPostedForPeriod(savingsAccountData, yesterday)) {
-                        log.debug("Interest already posted for savings account {} up to date {}, skipping", savingsAccountData.getId(),
-                                savingsAccountData.getSummary().getInterestPostedTillDate());
-                        continue;
-                    }
                     SavingsAccountData savingsAccountDataRet = savingsAccountWritePlatformService.postInterest(savingsAccountData,
                             postInterestAsOn, transactionDate, backdatedTxnsAllowedTill);
                     savingsAccountDataList.add(savingsAccountDataRet);
@@ -168,15 +162,18 @@ public class SavingsSchedularInterestPoster {
         String queryForSavingsUpdate = batchQueryForSavingsSummaryUpdate();
         String queryForTransactionInsertion = batchQueryForTransactionInsertion();
         String queryForTransactionUpdate = batchQueryForTransactionsUpdate();
-        List<Object[]> paramsForTransactionInsertion = new ArrayList<>();
-        List<Object[]> paramsForSavingsSummary = new ArrayList<>();
-        List<Object[]> paramsForTransactionUpdate = new ArrayList<>();
-        List<String> transRefNo = new ArrayList<>();
         LocalDate currentDate = DateUtils.getBusinessLocalDate();
         Long userId = platformSecurityContext.authenticatedUser().getId();
+
+        List<Object[]> paramsForSavingsSummary = new ArrayList<>();
+        List<List<String>> perAccountRefNos = new ArrayList<>();
+        List<List<Object[]>> perAccountInsertionParams = new ArrayList<>();
+        List<List<Object[]>> perAccountUpdateParams = new ArrayList<>();
+
         for (SavingsAccountData savingsAccountData : savingsAccountDataList) {
             OffsetDateTime auditTime = DateUtils.getAuditOffsetDateTime();
             SavingsAccountSummaryData savingsAccountSummaryData = savingsAccountData.getSummary();
+
             paramsForSavingsSummary.add(new Object[] { savingsAccountSummaryData.getTotalDeposits(),
                     savingsAccountSummaryData.getTotalWithdrawals(), savingsAccountSummaryData.getTotalInterestEarned(),
                     savingsAccountSummaryData.getTotalInterestPosted(), savingsAccountSummaryData.getTotalWithdrawalFees(),
@@ -186,14 +183,19 @@ public class SavingsSchedularInterestPoster {
                     savingsAccountSummaryData.getLastInterestCalculationDate(),
                     savingsAccountSummaryData.getInterestPostedTillDate() != null ? savingsAccountSummaryData.getInterestPostedTillDate()
                             : savingsAccountSummaryData.getLastInterestCalculationDate(),
-                    auditTime, userId, savingsAccountData.getId() });
+                    auditTime, userId, savingsAccountData.getId(), savingsAccountData.getVersion() });
+
+            List<String> transRefNo = new ArrayList<>();
+            List<Object[]> insertionParams = new ArrayList<>();
+            List<Object[]> updateParams = new ArrayList<>();
+
             List<SavingsAccountTransactionData> savingsAccountTransactionDataList = savingsAccountData.getSavingsAccountTransactionData();
             for (SavingsAccountTransactionData savingsAccountTransactionData : savingsAccountTransactionDataList) {
                 if (savingsAccountTransactionData.getId() == null && !MathUtil.isZero(savingsAccountTransactionData.getAmount())) {
                     UUID uuid = UUID.randomUUID();
                     savingsAccountTransactionData.setRefNo(uuid.toString());
                     transRefNo.add(uuid.toString());
-                    paramsForTransactionInsertion.add(new Object[] { savingsAccountData.getId(), savingsAccountData.getOfficeId(),
+                    insertionParams.add(new Object[] { savingsAccountData.getId(), savingsAccountData.getOfficeId(),
                             savingsAccountTransactionData.isReversed(), savingsAccountTransactionData.getTransactionType().getId(),
                             savingsAccountTransactionData.getTransactionDate(), savingsAccountTransactionData.getAmount(),
                             savingsAccountTransactionData.getBalanceEndDate(), savingsAccountTransactionData.getBalanceNumberOfDays(),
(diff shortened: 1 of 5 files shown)


awaneetdecoder:
I have force-pushed the fixes for the Lombok duplicate getters/setters and the Checkstyle trailing space. Could a maintainer please approve the workflows so the CI can run against the new commit? Thank you!

adamsaghy:
@awaneetdecoder Please review the failing checks!

adamsaghy:
Please run:
`./gradlew spotlessApply spotbugsMain spotbugsTest checkstyleMain checkstyleTest`
`./gradlew --no-daemon build -x test -x cucumber -x doc`

Before any PR or changes, please always run these two commands and make sure there is green build!

awaneetdecoder:
@adamsaghy I have addressed all the failing checks:
 1. **Replaced `ConcurrentModificationException` with Spring's `OptimisticLockingFailureException`** — this is semantically correct for optimistic locking failures, extends `DataAccessException` so it is caught by the existing handler in `postInterest()`, and satisfies SpotBugs.

 2. **Removed unused mock infrastructure from the unit test** — the 4 tests verify pure logic (updateCounts array checking) and have no dependency on Spring beans, so no mocks are needed.

 3. **Ran both commands locally** — `spotlessApply spotbugsMain spotbugsTest checkstyleMain checkstyleTest` and `build -x test -x cucumber -x doc` both passed cleanly. The `:fineract-war:binaryDistTar` failure is a pre-existing duplicate JAR issue reproducible on `develop` with no changes applied.

Thank you for your patience.

awaneetdecoder:
@adamsaghy — investigated the test-core-2 failure.

testRunningPostInterestJobTwiceDoesNotCreateDuplicateInterest 
was failing due to a date logic error in the test itself: with 
businessDate = April 12, the scheduler calculates interest up to 
yesterday (April 11), so the posting lands on [2022, 4, 11] — 
not [2022, 4, 12] as the filter expected.

Fix: changed businessDate to April 13 so yesterday = April 12, 
ensuring the interest posting appears on [2022, 4, 12] as the 
assertion requires.

The E2E Shard 6 failure is pre-existing — it covers only Currency 
and Loan domain features, unrelated to this PR's savings changes.

All other 473 integration tests pass. Updated the commit with the fix.

awaneetdecoder:
@adamsaghy — the previous filter on transactionType.value was incorrectly eliminating all transactions. Simplified to match the pattern used in testDuplicateOverdraftInterestPostingJob — filtering by date [2022, 4, 12] and reversed=false, which is sufficient since no other transaction lands on that date. test-core-2 is now passing on all 3 databases.
The 2 remaining failures are pre-existing and unrelated to this PR:

- verify (build-core, main) fails on cyclonedxDirectBom due to an unresolvable Eclipse SWT POM                                                                           (org.eclipse.swt.${osgi.platform}) — a known dependency resolution issue unrelated to savings changes.
- E2E Shard 7 covers only Client and Loan domain features — no savings involvement.

41 checks passing. Ready for review.

adamsaghy:
SBOM is not a "real" error, it does not fail the build, however spotless is!

```
Execution failed for task ':integration-tests:spotlessJavaCheck'.
      src/test/java/org/apache/fineract/integrationtests/SavingsInterestPostingJobIntegrationTest.java
          @@ -247,10 +247,8 @@
           ············Object·transactionObj·=·this.savingsAccountHelper.getSavingsDetails(savingsId,·"transactions");
           ············ArrayList<HashMap<String,·Object>>·transactions·=·(ArrayList<HashMap<String,·Object>>)·transactionObj;
           
          -············long·interestPostingsCount·=·transactions.stream()
          -····················.filter(t·->·t.get("date").toString().equals("[2022,·4,·12]"))
          -····················.filter(t·->·t.get("reversed").toString().equals("false"))
          -····················.count();
          +············long·interestPostingsCount·=·transactions.stream().filter(t·->·t.get("date").toString().equals("[2022,·4,·12]"))
          +····················.filter(t·->·t.get("reversed").toString().equals("false")).count();
           
           ············assertEquals(1,·interestPostingsCount,·"Running·job·twice·must·not·create·duplicate·interest·postings·on·the·same·date");
           ········}·finally·{
          @@ -378,4 +376,4 @@
           ········globalConfigurationHelper.resetAllDefaultGlobalConfigurations();
           ········globalConfigurationHelper.verifyAllDefaultGlobalConfigurations();
           ····}
          -}
          +}
  Run './gradlew :integration-tests:spotlessApply' to fix these violations.
```

awaneetdecoder:
@adamsaghy — all 43 checks are now passing on commit dfbb9c0. Spotless formatting has been resolved. The PR is ready for review. Thank you.

awaneetdecoder:
Hi @adamsaghy, PR #5646 merged successfully. This PR (#5671) also has all 43 checks passing and no conflicts. Could you please take a look when you get a chance? Thank you.

fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
```diff
@@ -115,6 +112,7 @@ private void batchUpdateJournalEntries(final List<SavingsAccountData> savingsAcc
             for (SavingsAccountTransactionData savingsAccountTransactionData : savingsAccountTransactionDataList) {
                 if (savingsAccountTransactionData.getId() == null && !MathUtil.isZero(savingsAccountTransactionData.getAmount())) {
                     final String key = savingsAccountTransactionData.getRefNo();
+                    final Boolean isOverdraft = savingsAccountTransactionData.getIsOverdraft();
```
adamsaghy: This seems unused...

fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
```diff
@@ -67,16 +70,10 @@ public class SavingsSchedularInterestPoster {
     public void postInterest() throws JobExecutionException {
         if (!savingAccounts.isEmpty()) {
             List<Throwable> errors = new ArrayList<>();
-            LocalDate yesterday = DateUtils.getBusinessLocalDate().minusDays(1);
             for (SavingsAccountData savingsAccountData : savingAccounts) {
                 boolean postInterestAsOn = false;
                 LocalDate transactionDate = null;
                 try {
-                    if (isInterestAlreadyPostedForPeriod(savingsAccountData, yesterday)) {
```
adamsaghy, line 75: why to remove this?

fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
```diff
@@ -213,8 +213,24 @@ private void batchUpdate(final List<SavingsAccountData> savingsAccountDataList)
...
+            for (int i = 0; i < updateCounts.length; i++) {
+                if (updateCounts[i] == 0) {
+                    Long accountId = savingsAccountDataList.get(i).getId();
+                    skippedAccountIds.add(accountId);
+                    log.warn("Optimistic lock failure for savings account id={} — concurrent modification detected."
+                            + " Rolling back. Will retry on next run.", accountId);
+                }
+            }
+
+            if (!skippedAccountIds.isEmpty()) {
```
adamsaghy: I am wondering why to throw optimistic lock exception if we havent updated anything... i mean we calculated something, but it was not updated... so there is no duplicate posting anyway, no?

fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
awaneetdecoder: @adamsaghy You're correct — isOverdraft is declared but never used. I'll remove it in the next commit.

fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
awaneetdecoder: @adamsaghy The pre-check was removed because it is not thread-safe under concurrent execution — two job instances can both pass the check before either commits, then both proceed to post. The optimistic lock (WHERE id=? AND version=?) in batchQueryForSavingsSummaryUpdate is the reliable guard. Happy to restore it as an additional early-exit optimization if you prefer, since it would reduce unnecessary processing in the non-concurrent case.

fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
awaneetdecoder: @adamsaghy The exception is thrown to keep the batch atomic. If account A's summary update succeeds but account B's fails the version check, without the exception account A's transactions would still be inserted while account B's summary remains out of sync — inconsistent state within the same batch. The throw causes full rollback so all accounts retry together on the next run. This behaviour is covered in SavingsSchedularInterestPosterTest — testSkippedAccountIdsNotEmptyMeansExceptionShouldBeThrown validates this path.

fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
adamsaghy: Can you remove it now?

fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
adamsaghy: I dont think it is a bad thing to allow to be updated Account A if Account B fails... the accounts are independent from each others..

The only issue we should avoid to post twice an interest for the same day or post an interest but having out of sync account balance...

awaneetdecoder: @adamsaghy I have pushed the updates in the latest commit:

1. Removed the unused isOverdraft variable.
2. Changed the batch logic to skip only the failed account's transactions rather than rolling back the entire batch.

Could you please approve the CI workflows to run when you have a moment?

awaneetdecoder: @adamsaghy The CI has completed — 42 checks passing, 1 failing. The failing test is verifyEarlyLateRepaymentOnProgressiveLoanNextInstallmentAllocationRepayLessThenEmi in LoanInterestRecalculationCOBTest, failing with HTTP 403 due to a duplicate loan product short name 01BB in the test environment. This is a pre-existing Loan domain test unrelated to this PR's savings changes.

fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsSchedularInterestPoster.java
awaneetdecoder: @adamsaghy Done — removed in the latest commit (d469aaa).

PR #5712, opened 2026-03-29
FINERACT-2561: Move organisation currency validation outside loop in saveAllDebitOrCreditEntries

AshharAhmadKhan:
## Description

JIRA: https://issues.apache.org/jira/browse/FINERACT-2561

The `saveAllDebitOrCreditEntries()` method was executing the same 
organisation currency validation query inside the per-entry loop, causing 
redundant identical database round-trips on every debit/credit line of a 
manual journal entry request.

Since `currencyCode` does not change within a single request, the validation 
only needs to run once per method invocation — not once per loop iteration.

This fix moves `organisationCurrencyRepository.findOneWithNotFoundDetection(currencyCode)` 
to before the loop, consistent with the pattern already used in the same class 
inside `saveAllDebitOrCreditOpeningBalanceEntries()`.

No logic change — validation still happens, just once instead of N times.


DeathGun44 (approved):
LGTM

PR #5737, opened 2026-04-03
FINERACT-2533: Remove @Skip tags from savings account E2E scenarios

awaneetdecoder:
Removes @Skip annotations from scenarios C2438, C2439, and C2440 in SavingsAccount.feature.

These scenarios were skipped because SavingsAccountRequestFactory hardcoded
EUR_SAVING_PRODUCT_ID = 1L which did not exist in the E2E environment.

PR #5646 fixed the root cause by dynamically creating the EUR savings product via
SavingsProductRequestFactory.defaultSavingsProductRequest() and storing its ID in
TestContextKey.DEFAULT_SAVINGS_PRODUCT_CREATE_RESPONSE_EUR.

The "And Admin creates a EUR savings product" step is now added to C2438, C2439,
and C2440 (already present in C2441 from PR #5646), making the skip reason obsolete.

Closes FINERACT-2533

### DIFF
--- a/fineract-e2e-tests-core/src/test/java/org/apache/fineract/test/factory/SavingsProductRequestFactory.java
+++ b/fineract-e2e-tests-core/src/test/java/org/apache/fineract/test/factory/SavingsProductRequestFactory.java
@@ -20,13 +20,13 @@ package org.apache.fineract.test.factory;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.UUID;
 import org.apache.fineract.client.models.PostSavingsCharges;
 import org.apache.fineract.client.models.PostSavingsProductsRequest;
 
 public final class SavingsProductRequestFactory {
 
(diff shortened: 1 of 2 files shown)


awaneetdecoder:
@adamsaghy As requested, removed the @Skip tags from C2438, C2439, and C2440. 
Added the "And Admin creates a EUR savings product" step to each scenario to 
dynamically create the product, same approach as C2441 in PR #5646.

awaneetdecoder:
@adamsaghy During CI, scenarios C2439, C2440, and C2441 were failing with a 403 
"Unknown data integrity issue with savings product" error. The root cause was that 
SavingsProductRequestFactory.defaultSavingsProductRequest() used a hardcoded product 
name "CEUR" — when multiple scenarios run sequentially in the same DB session, the 
second scenario hits a unique constraint violation trying to create a product with 
the same name.

Fixed by appending a 4-character UUID suffix to both the product name and short name, 
making each scenario's product creation independent. All 43 checks are now passing.

PR #5477, opened 2026-02-11
FINERACT-2436: Fix PostgreSQL compatibility for Client Weekly and Monthly Trends Reports

Kengne-yvana:
…thly Trends Reports

## Description

Describe the changes made and why they were made. (Ignore if these details are present on the associated Apache Fineract JIRA ticket.)

## Checklist

Please make sure these boxes are checked before submitting your pull request - thanks!

- [ ] Write the commit message as per [our guidelines](https://github.com/apache/fineract/blob/develop/CONTRIBUTING.md#pull-requests)
- [ ] Acknowledge that we will not review PRs that are not passing the build _("green")_ - it is your responsibility to get a proposed PR to pass the build, not primarily the project's maintainers.
- [ ] Create/update [unit or integration tests](https://fineract.apache.org/docs/current/#_testing) for verifying the changes made.
- [ ] Follow our [coding conventions](https://cwiki.apache.org/confluence/display/FINERACT/Coding+Conventions).
- [ ] Add required Swagger annotation and update API documentation at fineract-provider/src/main/resources/static/legacy-docs/apiLive.htm with details of any API changes
- [ ] [This PR must not be a "code dump"](https://cwiki.apache.org/confluence/display/FINERACT/Pull+Request+Size+Limit). Large changes can be made in a branch, with assistance. Ask for help on the [developer mailing list](https://fineract.apache.org/#contribute).

Your assigned reviewer(s) will follow our [guidelines for code reviews](https://cwiki.apache.org/confluence/display/FINERACT/Code+Review+Guide).

PR #5754, opened 2026-04-08
FINERACT-2436: Fix PostgreSQL compatibility for client and loan trends reports

AshharAhmadKhan:
## Description

The ClientTrendsByWeek, ClientTrendsByMonth, LoanTrendsByWeek, and LoanTrendsByMonth
reports were using MySQL-specific functions (WEEK(), MONTHNAME()) that don't exist in
PostgreSQL, causing a BadSqlGrammarException when running on a PostgreSQL database.

This replaces those functions with ISO-standard equivalents — EXTRACT(WEEK FROM ...) 
and TO_CHAR(..., 'Month') — which work correctly on both MySQL and PostgreSQL.

Picks up from #5477 (closed as abandoned), extends the fix to cover all 4 affected
reports and adds an integration test for each.

Jira : https://issues.apache.org/jira/browse/FINERACT-2436


AshharAhmadKhan:
@adamsaghy i think its all good, let me know if any changes needed.

PR #5762, opened 2026-04-11
FINERACT-2586: add fail on duplicate explicit operationIds

San-43:
This PR adds a small validation step to catch duplicate explicit OpenAPI `operationIds` before the spec is generated. [FINERACT-2586](https://issues.apache.org/jira/browse/FINERACT-2586)

**NOTE:** After wiring the validator and running it, it immediately failed on a duplicate `operationId`: `retrieveWorkingCapitalLoanTemplate`, which was declared in:

- `WorkingCapitalLoanApiResource`
- `WorkingCapitalLoanTransactionsApiResource` 

After fixing that one and rebased onto latest develop, it also found another duplicate: `getDelinquencyRangeScheduleTagHistoryById`

Both of these were added very recently, so this is a good example of why it helps to validate this at OpenAPI generation time instead of finding out later when the generated client or docs start behaving weird.


San-43:
Not related. `> Java heap space`

PR #5825, opened 2026-05-05
FINERACT-2421: Added e2e test for covering Loan invalid state after C…

alberto-art3ch:
## Description

We found a case that after given operations, a loan ends up having positive balances in both the total_outstanding_derived & total_overpaid_derived columns in the m_loan table.

[FINERACT-2421](https://issues.apache.org/jira/browse/FINERACT-2421)

## Checklist


fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/transactionprocessor/impl/AdvancedPaymentScheduleTransactionProcessor.java
```diff
@@ -650,11 +650,17 @@ private void handleInterestRefund(final LoanTransaction loanTransaction, final T
                     final Money interestAfterRefund = interestRefundService.totalInterestByTransactions(this, loan.getId(), targetDate,
                             modifiedTransactions, unmodifiedTransactionIds, ctx.getActiveLoanTermVariations());
                     final Money newAmount = interestBeforeRefund.minus(progCtx.getSumOfInterestRefundAmount()).minus(interestAfterRefund);
-                    loanTransaction.updateAmount(newAmount.getAmount());
+                    loanTransaction.updateAmount(MathUtil.negativeToZero(newAmount).getAmount());
```
galovics, line 653: I'm not convinced this is the right fix. Clamping a negative value to zero hides the question of why `newAmount` went negative in the first place. Under what scenario does `interestBeforeRefund - sumOfInterestRefundAmount - interestAfterRefund` come out negative, and is that a legitimate state or a symptom of something else being off?

If it's legitimate, the comment below explains the `0` case well - but it should also explain when/why `newAmount` is negative and why zero is the correct truncation. If it's not legitimate, we're papering over a real bug. Could you walk me through it?

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/transactionprocessor/impl/AdvancedPaymentScheduleTransactionProcessor.java
```diff
@@ -650,11 +650,17 @@ private void handleInterestRefund(final LoanTransaction loanTransaction, final T
...
                 }
                 progCtx.setSumOfInterestRefundAmount(progCtx.getSumOfInterestRefundAmount().add(loanTransaction.getAmount()));
             }
         }
+        // A zero-amount interest refund has no effect on balances; processing it would incorrectly
+        // zero the overpaymentHolder via handleOverpayment, causing subsequent CBR transactions
+        // to see an empty holder and create phantom outstanding on additional installments.
+        if (!loanTransaction.getAmount(ctx.getCurrency()).isGreaterThanZero()) {
```
galovics, line 661: I'd really like a unit test pinning this down - the bugfix is subtle and the Cucumber scenarios don't really show what `handleRepayment` is doing to `overpaymentHolder` here. Please add a test to cover this so it never reoccurs.

Also: the early return is positioned after the `interestRefundService.handle(...)` block - is that intentional? If the amount is zero, do we still want the side-effect of `progCtx.setSumOfInterestRefundAmount(...)` running above? Walk me through it please.

fineract-e2e-tests-core/src/test/java/org/apache/fineract/test/stepdef/loan/LoanStepDef.java
```diff
@@ -2169,7 +2169,8 @@ public void loanOverpaid(double totalOverpaidExpected) {
                 () -> fineractClient.loans().retrieveLoan(loanId, Map.of("staffInSelectedOfficeOnly", "false")));
         testContext().set(TestContextKey.LOAN_RESPONSE, loanDetailsResponse);
 
-        Double totalOverpaidActual = loanDetailsResponse.getTotalOverpaid().doubleValue();
+        Double totalOverpaidActual = (loanDetailsResponse.getTotalOverpaid() != null) ? loanDetailsResponse.getTotalOverpaid().doubleValue()
```
galovics: Why the null check? This method is `loanOverpaid(...)` - if `totalOverpaid` is null we should fail loudly, not silently coerce to 0.0 and then assert against expected. That assertion will pass when the API stops returning the field at all, which is exactly the regression we don't want to miss.

fineract-e2e-tests-core/src/test/java/org/apache/fineract/test/data/loanproduct/DefaultLoanProduct.java
```diff
@@ -97,6 +97,7 @@ public enum DefaultLoanProduct implements LoanProduct {
     LP2_ADV_PYMNT_INTEREST_DAILY_EMI_ACTUAL_ACTUAL_INTEREST_REFUND_INTEREST_RECALCULATION_MULTIDISB, //
     LP2_ADV_PYMNT_INTEREST_DAILY_INTEREST_RECALCULATION_ZERO_INTEREST_CHARGE_OFF_BEHAVIOUR, //
     LP2_ADV_PYMNT_INTEREST_DAILY_INTEREST_RECALCULATION_ZERO_INTEREST_CHARGE_OFF, //
+    LP2_ADV_PYMNT_INT_DAILY_EMI_ACTUAL_ACTUAL_INT_REFUND_FULL_ZERO_INT_CHARGE_OFF_ACCRUAL_ACTIVITY_LI, //
```
galovics: These enum names are getting completely out of hand. I won't block on it since the rest of the file is just as bad - but for the future: what does `LI` mean to someone who didn't read this PR? Just `LAST_INSTALLMENT` would be clearer than another cryptic 2-letter suffix.

fineract-e2e-tests-runner/src/test/resources/features/LoanCBR.feature
```diff
@@ -1789,3 +1789,202 @@ Feature: Credit Balance Refund
...
+      | Nr | Days | Date             | Paid date     | Balance of loan | Principal due | Interest | Fees | Penalties | Due   | Paid  | In advance | Late  | Outstanding |
+      |    |      | 01 January 2025  |               | 300.0           |               |          | 0.0  |           | 0.0   | 0.0   |            |       |             |
+      | 1  | 31   | 01 February 2025 | 01 March 2025 | 200.0           | 100.0         | 0.0      | 0.0  | 0.0       | 100.0 | 100.0 | 0.0        | 100.0 | 0.0         |
+      | 2  | 28   | 01 March 2025    | 01 April 2025 | 100.0           | 100.0         | 0.0      | 0.0  | 0.0       | 100.0 | 100.0 | 0.0        | 100.0 | 0.0         |
+      | 3  | 31   | 01 April 2025    |               | 0.0             | 100.0         | 0.0      | 0.0  | 0.0       | 100.0 | 50.0  | 0.0        | 0.0   | 50.0        |
+      | 4  | 14   | 15 April 2025    |               | 0.0             | 50.0          | 0.0      | 0.0  | 0.0       | 50.0  | 0.0   | 0.0        | 0.0   | 50.0        |
+
+  @TestRailId:C78853
+  Scenario: Verify CBR + backdated GoodwillCredit on NEXT_INSTALLMENT product closes the loan
```
galovics: C78812 and C78853 look almost identical step-by-step, the only material difference being the product (LAST_INSTALLMENT vs NEXT_INSTALLMENT). That's fine for coverage but the diff is enormous - any chance you could DRY this with a Scenario Outline / Examples table over the product name? Not a blocker, but ~200 lines of near-duplicate Gherkin is hard to maintain.

fineract-e2e-tests-runner/src/test/java/org/apache/fineract/test/initializer/global/LoanProductGlobalInitializerStep.java
```diff
@@ -4961,6 +4961,47 @@ public void initialize() throws Exception {
                     responseLoanProductsRequestLP2AdvPaymentIntEmiActualActualIntRefundFullZeroIntChargeOffAccrualActivity);
         });
 
+        tasks.add(() -> {
+            // LP2 + zero-interest chargeOff behaviour + progressive loan schedule + horizontal + interest recalculation
+            // + accrual activity posting, with LAST_INSTALLMENT future-installment allocation rule baked in for every
+            // advanced-allocation transaction type — exercises code paths that the more common NEXT_INSTALLMENT
```
galovics: Please use a regular dash (`-`) instead of em-dashes here and in the lines below. Match the rest of the file.

galovics (changes requested):
Hi @alberto-art3ch, thanks for the PR.

The PR title and description only mention the e2e test, but there's a real production fix inside `AdvancedPaymentScheduleTransactionProcessor` (two changes - the `negativeToZero` clamp on the refund amount and the new zero-amount early-return). Could you update the description so it's clear what's actually being fixed? Reviewers shouldn't have to scroll the diff to discover there's a code change.

The production change is covered exclusively through the new Cucumber scenarios. I'd really love a focused unit test on `AdvancedPaymentScheduleTransactionProcessor` for both branches of the new logic - the `MathUtil.negativeToZero(...)` clamp, and the zero-amount early-return that skips `handleRepayment`. The Cucumber scenarios are useful but they're black-box; a unit test pins the contract down so it can't silently regress.

Plus a few questions inline.

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/transactionprocessor/impl/AdvancedPaymentScheduleTransactionProcessor.java
adamsaghy: I second this. I would like to know how a negative value can be here...

alberto-art3ch: Here is a summary of what is happening in those e2e tests:

When a backdated Goodwill Credit triggers a transaction reverse-replay on a post-maturity loan that already has Credit Balance Refunds (CBRs), the Interest Refund transaction copy produced during replay has no `loanTransactionRelations`, causing it to recalculate its amount against the CBR-modified loan model. Because the CBR has already credited principal (reducing future scheduled interest), the recalculated `interestAfterRefund` can exceed `interestBeforeRefund`, making `newAmount` **negative or zero**. A zero amount, when passed unconditionally to `handleRepayment`, wipes the shared `overpaymentHolder` to zero; subsequent CBR replays then see an empty holder and create additional installments with phantom principal outstanding equal to the full CBR amount. The fix applies `negativeToZero` before updating the transaction amount and adds an early-return guard that skips `handleRepayment` entirely for zero-amount interest refunds, preserving the `overpaymentHolder` for the CBRs that follow.

fineract-e2e-tests-core/src/test/java/org/apache/fineract/test/data/loanproduct/DefaultLoanProduct.java
alberto-art3ch: @ruzeynalov Please your support here for this request

oleksii-novikov-onix:
Typo in commit message: Fix fop Loan -> Fix for Loan

ruzeynalov:
> Typo in commit message: Fix fop Loan -> Fix for Loan

Fixed

fineract-e2e-tests-core/src/test/java/org/apache/fineract/test/stepdef/loan/LoanStepDef.java
ruzeynalov: @galovics  Fair point — silent coercion was hiding exactly the regression this step should catch. Refactored to fail loudly when the API returns null but the scenario expected a non-zero overpayment, while preserving the legitimate "loan closed with zero overpayment" case where the field is omitted.

fineract-e2e-tests-core/src/test/java/org/apache/fineract/test/data/loanproduct/DefaultLoanProduct.java
ruzeynalov: @alberto-art3ch @galovics Done — renamed the enum, the matching TestContextKey constant, the LoanProductGlobalInitializerStep registration block, and the three Gherkin product cells to: LP2_ADV_PYMNT_INT_DAILY_EMI_ACTUAL_ACTUAL_INT_REFUND_FULL_ZERO_INT_CHARGE_OFF_ACC_LAST_INSTALLMENT

fineract-e2e-tests-runner/src/test/resources/features/LoanCBR.feature
ruzeynalov: @galovics Done. C78812 and C78853 are now a single Scenario Outline parameterised on the loan-product cell

fineract-e2e-tests-runner/src/test/java/org/apache/fineract/test/initializer/global/LoanProductGlobalInitializerStep.java
ruzeynalov: @galovics Done — replaced the em-dashes (—) with regular hyphens (-)

adamsaghy:
Duplicate

PR #5827, opened 2026-05-06
FINERACT-2476: Enable Group Savings Accounts as Guarantors with Guarantee Percentage Requirements

rhopman:
Group savings accounts should be usable as guarantors regardless of guarantee percentage requirements, as long as:

- The guarantor account is active
- Sufficient funds are available
- The account is not in a restricted state
- Standard guarantor business rules are met (excluding self-guarantee check for group accounts)

## Checklist

- [X] Write the commit message as per [our guidelines](https://github.com/apache/fineract/blob/develop/CONTRIBUTING.md#pull-requests)
- [X] Create/update [unit or integration tests](https://fineract.apache.org/docs/current/#_testing) for verifying the changes made.
- [X] Follow our [coding conventions](https://cwiki.apache.org/confluence/display/FINERACT/Coding+Conventions).
- [X] [This PR must not be a "code dump"](https://cwiki.apache.org/confluence/display/FINERACT/Pull+Request+Size+Limit). Large changes can be made in a branch, with assistance. Ask for help on the [developer mailing list](https://fineract.apache.org/#contribute).


integration-tests/src/test/java/org/apache/fineract/integrationtests/guarantor/GuarantorTest.java
```diff
@@ -738,6 +738,49 @@ private Integer applyForLoanApplication(final Integer clientID, final Integer lo
         return this.loanTransactionHelper.getLoanId(loanApplicationJSON);
     }
 
+    @SuppressWarnings({ "rawtypes", "unchecked" })
```
Aman-Mittal: Why we are suppressing warnings here

integration-tests/src/test/java/org/apache/fineract/integrationtests/common/savings/SavingsAccountHelper.java
```diff
@@ -1302,6 +1303,31 @@ public static Integer openSavingsAccount(final RequestSpecification requestSpec,
         return savingsId;
     }
 
+    // TODO: Rewrite to use fineract-client instead!
+    // Example: org.apache.fineract.integrationtests.common.loans.LoanTransactionHelper.disburseLoan(java.lang.Long,
+    // org.apache.fineract.client.models.PostLoansLoanIdRequest)
```
Aman-Mittal: I think You should not write anymore test cases here.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/guarantor/service/GuarantorDomainServiceImpl.java
```diff
@@ -125,7 +125,18 @@ public void validateGuarantorBusinessRules(Loan loan) {
                 for (GuarantorFundingDetails guarantorFundingDetails : fundingDetails) {
                     if (guarantorFundingDetails.getStatus().isActive() || guarantorFundingDetails.getStatus().isWithdrawn()
                             || guarantorFundingDetails.getStatus().isCompleted()) {
-                        if (guarantor.isSelfGuarantee()) {
+                        SavingsAccount savingsAccount = guarantorFundingDetails.getLinkedSavingsAccount();
+                        if (savingsAccount.isGroupAccount()) {
```
Aman-Mittal: Just trying to understand the intended business behavior here:

for client loans, if the borrower belongs to the same group whose savings account is being used as guarantee, should that still be classified as an external guarantee?

Right now it looks like only `loan.getGroupId()` is considered for self-guarantee classification.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/guarantor/service/GuarantorDomainServiceImpl.java
```diff
@@ -125,7 +125,18 @@ public void validateGuarantorBusinessRules(Loan loan) {
...
+                            // Group loan with the same group's savings account → self-guarantee
+                            // Client loan or different group's savings → external guarantee
+                            if (loan.getGroupId() != null && loan.getGroupId().equals(savingsAccount.getGroupId())) {
```
Aman-Mittal: Right now code seems to be handling

Case | Classification
-- | --
Group loan + same group savings | self
Client loan + same group savings | external

Please confirm me about this case I am also new to this logic

Case | Classification
-- | --
Client belongs to same guarantor group | maybe self / partial self

Aman-Mittal (commented):
Please address review comments and also maintain documentation in fineract-doc

I think we can discuss this usecase more in dev list

rhopman:
Hi @Aman-Mittal, thank you for your thorough review.

My intention is to limit this change to the scenario where a group savings account (no `clientId`) is used as guarantor for a loan. I would only consider this a self-guarantee if the loan itself is a group loan (no `clientId`) from the same group (same `groupId`). So the group itself (as a separate entity) is taking out a loan and partially or fully self-guaranteeing, using a savings account that belongs to the same group, but not to any individual group member or other client).

I think there are different ways to look a this, but the above makes the most sense to me, and matches my business requirements.

I have added some documentation and fixed the suppressed / deprecated issues. This makes the change bigger than strictly necessary but hopefully it is a step in the right direction.

Aman-Mittal:
> Hi @Aman-Mittal, thank you for your thorough review.
> 
> My intention is to limit this change to the scenario where a group savings account (no `clientId`) is used as guarantor for a loan. I would only consider this a self-guarantee if the loan itself is a group loan (no `clientId`) from the same group (same `groupId`). So the group itself (as a separate entity) is taking out a loan and partially or fully self-guaranteeing, using a savings account that belongs to the same group, but not to any individual group member or other client).
> 
> I think there are different ways to look a this, but the above makes the most sense to me, and matches my business requirements.
> 
> I have added some documentation and fixed the suppressed / deprecated issues. This makes the change bigger than strictly necessary but hopefully it is a step in the right direction.


Thanks for reply can you also open discussion in devlist about your usecase. While I can review most based on code level. Product side review required in this. @bharathc27 can you help validate the business requirement?

adamsaghy:
@Aman-Mittal Are you planning to work on this PR?

Aman-Mittal:
> @Aman-Mittal Are you planning to work on this PR?

No, I do not own this pr and ticket.

adamsaghy:
> > @Aman-Mittal Are you planning to work on this PR?
> 
> No, I do not own this pr and ticket.

I am sorry. I tagged the wrong person  :p

adamsaghy:
@rhopman Are you planning to work on this PR?

rhopman:
Hi @adamsaghy , I think the PR itself is in good shape as it is, but I changed it to draft because @Aman-Mittal suggested opening a discussion about this topic on the devlist. I have not found time to look into this and to determine the most appropriate scope of this discussion.

PR #5713, opened 2026-03-29
FINERACT-2571: Migrate NotificationHelper to fineract-client

lliikkii7722-ops:
## Description
This PR migrates NotificationHelper from legacy Utils.performServerGet usage
to fineract-client + Calls.ok.

## Changes
- Replaced legacy GET call in NotificationHelper#getNotifications
- Removed manual JSON parsing in favor of typed fineract-client response

## Notes
- Small and focused helper migration
- No business logic changes

PR #5703, opened 2026-03-27
FINERACT-2571: Migrate charge creation helper to feign client

lliikkii7722-ops:
This change migrates a deprecated helper method in `ChargesHelper` from RestAssured (`Utils.performServerPost`) to the feign client approach using `FineractClientHelper`.

This aligns with the ongoing effort to modernize integration tests and reduce direct HTTP usage.

Only a single method is migrated in this PR to keep the change minimal and review-friendly.

PR #5879, opened 2026-05-22
FINERACT-2571: Remove deprecated NotificationHelper class

AshharAhmadKhan:
JIRA: https://issues.apache.org/jira/browse/FINERACT-2571

The original goal of this ticket was to migrate NotificationHelper to use 
the fineract-client (Feign) instead of the legacy Utils.performServerGet 
REST calls.

After investigating the current state of develop, the migration has already 
been completed via a different approach. FeignNotificationHelper was introduced 
as the proper typed replacement using FineractFeignClient, and NotificationApiTest 
was fully rewritten to use it directly. The old NotificationHelper class was 
marked @Deprecated(forRemoval = true) but never actually removed.

Checked the codebase and nothing references NotificationHelper anymore, so 
this PR removes it as the clean resolution of this ticket.

Prior draft PRs (#5703 and #5713)


AshharAhmadKhan:
@adamsaghy

adamsaghy (approved):
LGTM

PR #5621, opened 2026-03-14
FINERACT-2530: Fix 500 error on GET /provisioningentries/{id} when no loan products exist

shubhamkumar9199:
Change INNER JOIN to LEFT JOIN in PROVISIONING_ENTRY_SUM_RESERVED_SCHEMA so that provisioning entries with no associated loan product rows still return a valid result instead of throwing EmptyResultDataAccessException.

[FINERACT-2530](https://issues.apache.org/jira/browse/FINERACT-2530)

## Checklist

Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Write the commit message as per [our guidelines](https://github.com/apache/fineract/blob/develop/CONTRIBUTING.md#pull-requests)
- [x] Acknowledge that we will not review PRs that are not passing the build _("green")_ - it is your responsibility to get a proposed PR to pass the build, not primarily the project's maintainers.
- [x] Create/update [unit or integration tests](https://fineract.apache.org/docs/current/#_testing) for verifying the changes made.
- [x] Follow our [coding conventions](https://cwiki.apache.org/confluence/display/FINERACT/Coding+Conventions).
- [x] Add required Swagger annotation and update API documentation at fineract-provider/src/main/resources/static/legacy-docs/apiLive.htm with details of any API changes
- [x] [This PR must not be a "code dump"](https://cwiki.apache.org/confluence/display/FINERACT/Pull+Request+Size+Limit). Large changes can be made in a branch, with assistance. Ask for help on the [developer mailing list](https://fineract.apache.org/#contribute).

Your assigned reviewer(s) will follow our [guidelines for code reviews](https://cwiki.apache.org/confluence/display/FINERACT/Code+Review+Guide).

PR #5880, opened 2026-05-22
FINERACT-2530: Fix 500 error on GET /provisioningentries/{id} when no loan products exist

AshharAhmadKhan:
## Description

When a provisioning entry exists but no loans have been disbursed yet,
calling GET /provisioningentries/{id} returns HTTP 500. The root cause
is an INNER JOIN in PROVISIONING_ENTRY_SUM_RESERVED_SCHEMA — when
m_loanproduct_provisioning_entry has no rows for the given history ID,
the join returns zero rows and Spring's queryForObject() throws
EmptyResultDataAccessException.

Fix: change INNER JOIN to LEFT JOIN so the query always returns one row
(totalReserved = NULL when no loan products exist), mapping cleanly
without throwing.

Picks up the work from #5621.


AshharAhmadKhan:
@adamsaghy please trigger checks when you get a chance

AshharAhmadKhan:
@adamsaghy fixed the integration test, the previous version was passing an empty loan products list which the server rejects with 400. now it creates a loan product but skips disbursement, so the join table stays empty and actually exercises the fix. could you approve the workflows when you get a chance? thanks!

adamsaghy (approved):
LGTM

PR #5898, opened 2026-05-27
FINERACT-2618: Add originators to journal entry aggregation job and improve reports.

AshharAhmadKhan:
## Description

The journal entry aggregation job was grouping entries only by asset owner, leaving `originator_external_ids` always NULL in the summary table even though the column existed. This meant the Trial Balance report had to re-derive originator data from raw journal entries for every historical row already in the snapshot, adding unnecessary memory overhead. It also meant originator never appeared via the snapshot path even when one was attached to the loan.

This PR fixes both problems:

- Adds a `loan_originators` CTE to the aggregation job reader query and wires `originator_external_ids` through the full pipeline — DTO, JPA entity, and writer service — so it gets correctly persisted per summary row
- Updates the Trial Balance Summary Report (MySQL + PostgreSQL) via migration `0236` so `summary_snapshot_baseline_data` reads the stored originator value directly instead of hardcoding NULL, letting the snapshot join cleanly on originator without re-hitting the mapping tables
- Updates unit tests for the reader and writer service to cover the new field
- Adds an integration test that runs the actual aggregation job end-to-end and verifies originator surfaces correctly via the snapshot path

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/jobs/service/aggregationjob/JournalEntryAggregationJobConfiguration.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/jobs/service/aggregationjob/JournalEntryAggregationJobConfiguration.java
@@ -22,6 +22,7 @@ import static org.apache.fineract.infrastructure.jobs.service.aggregationjob.Jou
 import static org.apache.fineract.infrastructure.jobs.service.aggregationjob.JournalEntryAggregationJobConstant.JOB_TRACKING_STEP_NAME;
 
 import org.apache.fineract.infrastructure.core.config.FineractProperties;
+import org.apache.fineract.infrastructure.core.service.database.DatabaseTypeResolver;
 import org.apache.fineract.infrastructure.core.service.migration.TenantDataSourceFactory;
 import org.apache.fineract.infrastructure.jobs.service.JobName;
 import org.apache.fineract.infrastructure.jobs.service.aggregationjob.data.JournalEntryAggregationSummaryData;
@@ -59,6 +60,8 @@ public class JournalEntryAggregationJobConfiguration {
     private JournalEntryAggregationTrackingTasklet journalEntryAggregationTrackingTasklet;
     @Autowired
     private TenantDataSourceFactory tenantDataSourceFactory;
+    @Autowired
+    private DatabaseTypeResolver databaseTypeResolver;
 
     @Bean
     public Step journalEntryAggregationSummaryStep() {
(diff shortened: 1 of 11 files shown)


AshharAhmadKhan:
@adamsaghy Checked all 4 failing CI checks. None are related to this PR's changes. Please re-trigger those checks when youu get a chance.

AshharAhmadKhan:
Hi @adamsaghy, I’ve investigated this further and verified the originator flow end-to-end (reader → DTO → entity → writer → report snapshot path), but I’m still missing the cause of the Shard 15 integration failure. Could you please point me in the right direction if there’s any Fineract-specific behavior around Journal Entry Aggregation / Trial Balance reporting that I may be overlooking?

adamsaghy:
@AshharAhmadKhan Please rebase and fix failing checks

AshharAhmadKhan:
Hey @adamsaghy , just saw 2421 merged - it does this exact fix already, so closing this one out. Good work on 2421!

One thing worth noting though: they resolve the originator list with a correlated subquery per row, while I was using a CTE with a single LEFT JOIN instead. Isn't it better to do it this way? Let me know your thoughts.

adamsaghy:
> Hey @adamsaghy , just saw 2421 merged - it does this exact fix already, so closing this one out. Good work on 2421!
> 
> One thing worth noting though: they resolve the originator list with a correlated subquery per row, while I was using a CTE with a single LEFT JOIN instead. Isn't it better to do it this way? Let me know your thoughts.

Let me look into it. Thank you for raising your concern.

alberto-art3ch:
> One thing worth noting though: they resolve the originator list with a correlated subquery per row, while I was using a CTE with a single LEFT JOIN instead. Isn't it better to do it this way? Let me know your thoughts.

Thanks for raising this — the CTE + LEFT JOIN is a clean, legitimate approach. A couple of reasons we went with the correlated subquery:

Context: the report SQL in this same ticket actually used the CTE pattern before, and we deliberately migrated it to the inline subquery — so part of the call was keeping the aggregation job consistent with the reports.

Correctness: equivalent. A CTE pre-aggregated per loan_id matches one row (no fan-out), so with COALESCE(..., '') the result is identical.

Performance — depends on workload, and this favors inline: this is an incremental, date-windowed job (usually a daily run).
- The subquery only touches loans in the window and is index-backed (the unique constraint on 'loan_id, originator_id' covers the loan_id lookup).
- The CTE pre-aggregates the whole mapping table regardless of window size — and on Postgres a CTE with GROUP BY is materialized, so it'd scan everything even for a handful of loans.

Where your CTE wins: large backfills / high repetition of the same loan, where the subquery re-evaluates per entry while the CTE aggregates once.

Maybe we can to have a kind of benchmark for this, or what you think?

AshharAhmadKhan:
Thanks for the detailed explanation @alberto-art3ch , that makes sense, especially the distinction around incremental windowed execution and index-backed lookups. I agree the correlated subquery is a better fit for the current job behavior. A benchmark between the two approaches would still be useful for future optimization, but I am aligned with this direction.

PR #5916, opened 2026-05-31
FINERACT-2624: Prepared statement, input parameter sanitisation

terencemo:
## Description

This PR enhances stretchy reporting in Fineract by:

 1. Sanitising input parameters based on type definitions
 1. Using Prepared statement to execute stretchy reports
 
Integrations tests have been added which invoke `runreports` with both valid and invalid inputs. Numeric parameter (officeId) positive and negative tests (numeric and non-numeric input) - some of the invalid inputs include `SLEEP` and `pg_sleep` commands. Also `UNION ALL` inputs passed attempting SQL injection.

Besides this, unregistered parameter passing is covered where a parameter not in `stretchy_report_parameter` for the given report is passed. Additional integration tests can be added to cover date and string parameter types.

## Checklist

### DIFF
--- a/fineract-core/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/GenericDataService.java
+++ b/fineract-core/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/GenericDataService.java
@@ -27,6 +27,8 @@ public interface GenericDataService {
 
     GenericResultsetData fillGenericResultSet(String sql);
 
+    GenericResultsetData fillGenericResultSet(String sql, Object... args);
+
     List<ResultsetColumnHeaderData> fillResultsetColumnHeaders(String tableName);
 
     List<ResultsetRowData> fillResultsetRowData(String sql, List<ResultsetColumnHeaderData> columnHeaders);
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/DatatableReportingProcessService.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/dataqueries/service/DatatableReportingProcessService.java
@@ -33,6 +33,7 @@ import org.apache.fineract.infrastructure.dataqueries.service.export.DatatableRe
 import org.apache.fineract.infrastructure.dataqueries.service.export.ResponseHolder;
 import org.apache.fineract.infrastructure.report.annotation.ReportService;
 import org.apache.fineract.infrastructure.report.service.AbstractReportingProcessService;
+import org.apache.fineract.infrastructure.report.service.ReportParameterTypeResolver;
 import org.apache.fineract.infrastructure.security.service.SqlValidator;
 import org.apache.fineract.util.StreamUtil;
 import org.springframework.stereotype.Service;
@@ -44,8 +45,9 @@ public class DatatableReportingProcessService extends AbstractReportingProcessSe
 
     private final List<DatatableReportExportService> exportServices;
 
-    public DatatableReportingProcessService(List<DatatableReportExportService> exportServices, SqlValidator sqlValidator) {
-        super(sqlValidator);
+    public DatatableReportingProcessService(List<DatatableReportExportService> exportServices, SqlValidator sqlValidator,
+            ReportParameterTypeResolver reportParameterTypeResolver) {
+        super(sqlValidator, reportParameterTypeResolver);
 
         this.exportServices = exportServices;
     }
@@ -55,7 +57,7 @@ public class DatatableReportingProcessService extends AbstractReportingProcessSe
 
         DatatableExportTargetParameter exportMode = DatatableExportTargetParameter.resolverExportTarget(queryParams);
         final String parameterTypeValue = ApiParameterHelper.parameterType(queryParams) ? "parameter" : "report";
-        final Map<String, String> reportParams = getReportParams(queryParams);
+        final Map<String, String> reportParams = getReportParams(reportName, queryParams);
         ResponseHolder response = findReportExportService(exportMode) //
                 .orElseThrow(() -> new GeneralPlatformDomainRuleException("error.msg.report.export.mode.unavailable",
                         "Export mode %s unavailable".formatted(exportMode.name()))) //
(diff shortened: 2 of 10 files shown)


IOhacker (changes requested):
No visible allow‑list, enum, or sanitization is applied to tableName or columnName

terencemo:
> No visible allow‑list, enum, or sanitization is applied to tableName or columnName

Sanitisation is not based on allow-list or enum. Can you explain what you mean by `tableName` / `columnName`? Are you referring to `ReadReportingServiceImpl` class methods getQuotedTableName and getQuotedColumnName ? My logic is not based upon those methods as explained in the document shared on the PMC list, we identify the parameter types.

vidakovic:
@terencemo  Question: why didn't we use just this service https://github.com/apache/fineract/blob/9e37bc3b66f82322ce50ee4d4bef32694d8a19b6/fineract-core/src/main/java/org/apache/fineract/infrastructure/security/service/SqlValidator.java#L23 to validate? I think that thing should cover most if not all what is done here... and in case something would be missing: just add a pattern in the application.properties at https://github.com/apache/fineract/blob/9e37bc3b66f82322ce50ee4d4bef32694d8a19b6/fineract-provider/src/main/resources/application.properties#L225
If I understood the code here correctly then the only thing left would be the type checking for the columns, that could stay... but in any way, why not re-use the SQL validator, add any missing pattern and then the entire code base could profit? Should also simplify the proposed changes significantly. 

As for the "allow-list" that @IOhacker mentioned: we could use the JDBC driver's metadata, extract all existing tables in the database at boot time and collect a static final list of table and column names in a `Map<String, List<String>>` variable (key = allowed table name, value = list of allowed column names in that table) and validate against that. Additionally we could again use the SqlValidator and maybe configure a separate profile with regex patterns to check the table and column names for any malicious characters... but I think that whitelist based on driver's metadata should be enough.

adamsaghy (approved):
I like the idea, but second @meonkeys and @vidakovic regarding we should have a central place for sql validation and rules and reuse that.

IOhacker (changes requested):
I support to reuse and improve the existing SQL validator as pointed by @alexivanov  instead of applying a local fix.

meonkeys:
> I like the idea, but second @meonkeys and @vidakovic regarding we should have a central place for sql validation and rules and reuse that.

I did not say that and I'm not reviewing this PR, but FWIW it sounds like a good idea?

adamsaghy:
> > I like the idea, but second @meonkeys and @vidakovic regarding we should have a central place for sql validation and rules and reuse that.
> 
> I did not say that and I'm not reviewing this PR, but FWIW it sounds like a good idea?

@meonkeys To the idea of using prepared statements wherever possible and input value sanitization and validation: yes, I do think it’s a good direction.

To implement it locally: not so much. I would probably wire and move this logic into the `SqlValidator`. There, I would define rules (which could be hardcoded, but probably we should have an SQL dictionary) and regular expressions or user-defined rules that could be configurable (to some extent).

I suppose the main point here is to avoid having a local solution but a centralized one.

terencemo:
> @terencemo Question: why didn't we use just this service
> 
> https://github.com/apache/fineract/blob/9e37bc3b66f82322ce50ee4d4bef32694d8a19b6/fineract-core/src/main/java/org/apache/fineract/infrastructure/security/service/SqlValidator.java#L23
> 
> to validate? I think that thing should cover most if not all what is done here... and in case something would be missing: just add a pattern in the application.properties at
> https://github.com/apache/fineract/blob/9e37bc3b66f82322ce50ee4d4bef32694d8a19b6/fineract-provider/src/main/resources/application.properties#L225
> 
> If I understood the code here correctly then the only thing left would be the type checking for the columns, that could stay... but in any way, why not re-use the SQL validator, add any missing pattern and then the entire code base could profit? Should also simplify the proposed changes significantly.

Correct me if I'm wrong, but SqlValidator purports to validate SQL and I'm not trying to do that. As explained on the PMC mail thread, my approach is to validate input parameters at the gate based on parameter type. I'm all for re-use, but SqlValidator takes a statement as input, right? My validation is before the statement is constructed.

> As for the "allow-list" that @IOhacker mentioned: we could use the JDBC driver's metadata, extract all existing tables in the database at boot time and collect a static final list of table and column names in a `Map<String, List<String>>` variable (key = allowed table name, value = list of allowed column names in that table) and validate against that. Additionally we could again use the SqlValidator and maybe configure a separate profile with regex patterns to check the table and column names for any malicious characters... but I think that whitelist based on driver's metadata should be enough.

Again, I'm looking at something far simpler than SQL here. e.g For parameter R_officeId=1, I'm just doing numeric validation to ensure we don't accept 1-SLEEP(10) for instance.

terencemo:
> > > I like the idea, but second @meonkeys and @vidakovic regarding we should have a central place for sql validation and rules and reuse that.
> > 
> > 
> > I did not say that and I'm not reviewing this PR, but FWIW it sounds like a good idea?
> 
> @meonkeys To the idea of using prepared statements wherever possible and input value sanitization and validation: yes, I do think it’s a good direction.
> 
> To implement it locally: not so much. I would probably wire and move this logic into the `SqlValidator`. There, I would define rules (which could be hardcoded, but probably we should have an SQL dictionary) and regular expressions or user-defined rules that could be configurable (to some extent).
> 
> I suppose the main point here is to avoid having a local solution but a centralized one.

I'm trying to validate simple numbers / dates, etc _before_ being assembled into a SQL query. As per the signature:

https://github.com/apache/fineract/blob/9e37bc3b66f82322ce50ee4d4bef32694d8a19b6/fineract-core/src/main/java/org/apache/fineract/infrastructure/security/service/SqlValidator.java#L25

there are 2 forms, one taking statement as input and the other taking 2 arguments with statement as second argument. In may approach, the validation happens at the gate before the SQL is assembled / constructed. If there's already a interface/class for parameter validation e.g NumericParameterValidator or DateParameterValidator then it could be used. But I don't think SqlValidator is simply built for a different purpose and not suited for what I'm doing. Correct me if wrong.

terencemo:
> I support to reuse and improve the existing SQL validator as pointed by @alexivanov instead of applying a local fix.

I'm not trying to validate SQL here. Isn't that what SqlValidator is for? I'm doing simple numeric / date validation.

adamsaghy:
@IOhacker Can you please review the changes and the answers from @terencemo and update your review? (We wont be able to merge till there is outstanding "Request change")

terencemo:
Closing in favour of #5968

geo-chen:
Hey @terencemo thanks for fixing this. As conversed via email, I've tested the fix in the develop branch. The JDBC bind params prevent SQL concan. Though, the parameterType=true permission bypass is still present.

PR #5940, opened 2026-06-04
FINERACT-2284: Enforce Charge Rounding using Money rounding rules

Dhanno98:
## FINERACT-2284: Enforce charge rounding using Money rounding rules

This PR fixes charge amount handling across Fineract by ensuring that charge values are rounded using the application's configured currency rules before they are persisted or processed.

Previously, several charge flows stored or processed raw `BigDecimal` values directly. As a result, charge amounts could bypass currency rounding settings such as:

* `digitsAfterDecimal`
* `inMultiplesOf`

This led to inconsistent behavior between charge calculations and other monetary operations that already rely on `Money`.

This PR standardizes charge amount handling by using `Money` for charge rounding across all supported charge types:

* Loan Charges
* Savings Account Charges
* Share Account Charges
* Client Charges

In addition, if a charge amount rounds to zero after applying currency rules, the charge is no longer persisted or created.

### Loan Charges

* Applied currency rounding when enforcing minimum and maximum charge caps.
* Ensured installment fee calculations and charge updates persist rounded values.
* Prevented zero-value loan charges from being persisted.

### Savings Account Charges

* Applied currency rounding through `Money`.
* Prevented persistence of flat charges whose final rounded value is zero.

### Share Account Charges

* Applied currency rounding when deriving charge amounts.
* Applied currency rounding when updating charge amounts for additional share transactions.
* Prevented creation of charge transactions for charges whose rounded value is zero.
* Prevented zero-value activation, purchase, and redemption charge transactions from being generated.

### Client Charges

* Applied currency rounding before charge creation.
* Prevented persistence of client charges whose rounded value becomes zero after rounding.

## API Documentation

Updated Swagger/OpenAPI schemas where required to improve request model documentation and examples.

## Why This Change

Fineract already provides currency-specific monetary rules through `Money`, including support for:

* Currency precision (`digitsAfterDecimal`)
* Rounding increments (`inMultiplesOf`)

Charges should follow the same monetary rules as all other financial amounts.

By enforcing rounding consistently:

* Charge calculations become predictable.
* Persisted charge values match configured currency behavior.
* Zero-value charges generated by rounding are avoided.
* Charge processing remains consistent across all portfolio modules.

## Test Coverage

Added integration test coverage for all supported charge types:

* `LoanChargeRoundingTest`
* `SavingsAccountChargeRoundingTest`
* `ShareAccountChargeRoundingTest`
* `ClientChargeRoundingTest`

The tests verify:

* Charge amounts respect configured currency rounding rules.
* `inMultiplesOf` and `digitsAfterDecimal` rounding is enforced.
* Rounded values are persisted correctly.
* Charges that round to zero are not created or persisted.

## Verification

Executed the complete charge rounding integration test suite successfully.

## Checklist

### DIFF
--- a/fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanChargeService.java
+++ b/fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanChargeService.java
@@ -39,6 +39,7 @@ import org.apache.fineract.portfolio.charge.domain.Charge;
 import org.apache.fineract.portfolio.charge.domain.ChargeCalculationType;
 import org.apache.fineract.portfolio.charge.domain.ChargePaymentMode;
 import org.apache.fineract.portfolio.charge.domain.ChargeTimeType;
+import org.apache.fineract.portfolio.charge.exception.LoanChargeCannotBeAddedException;
 import org.apache.fineract.portfolio.charge.exception.LoanChargeWithoutMandatoryFieldException;
 import org.apache.fineract.portfolio.loanaccount.domain.Loan;
 import org.apache.fineract.portfolio.loanaccount.domain.LoanCharge;
@@ -232,6 +233,8 @@ public class LoanChargeService {
         update(loanCharge, chargeAmt, loanCharge.getDueLocalDate(), amount, loan.fetchNumberOfInstallmentsAfterExceptions(),
                 totalChargeAmt);
 
+        validateChargeAmountNotZero(loanCharge);
+
         // NOTE: must add new loan charge to set of loan charges before
         // reprocessing the repayment schedule.
         if (loan.getLoanCharges() == null) {
@@ -384,6 +387,7 @@ public class LoanChargeService {
                 case PERCENT_OF_AMOUNT_AND_INTEREST:
                 case PERCENT_OF_INTEREST:
                 case PERCENT_OF_DISBURSEMENT_AMOUNT:
+
                     loanCharge.setPercentage(newValue);
                     loanCharge.setAmountPercentageAppliedTo(amount);
                     loanChargeAmount = BigDecimal.ZERO;
@@ -818,13 +822,18 @@ public class LoanChargeService {
                 case INVALID:
                 break;
                 case FLAT:
+                    BigDecimal roundedAmount;
                     if (loanCharge.isInstalmentFee()) {
                         if (numberOfRepayments == null) {
                             numberOfRepayments = loanCharge.getLoan().fetchNumberOfInstallmentsAfterExceptions();
                         }
-                        loanCharge.setAmount(amount.multiply(BigDecimal.valueOf(numberOfRepayments)));
+                        roundedAmount = Money
+                                .of(loanCharge.getLoan().getCurrency(), amount.multiply(BigDecimal.valueOf(numberOfRepayments)))
+                                .getAmount();
+                        loanCharge.setAmount(roundedAmount);
                     } else {
-                        loanCharge.setAmount(amount);
+                        roundedAmount = Money.of(loanCharge.getLoan().getCurrency(), amount).getAmount();
+                        loanCharge.setAmount(roundedAmount);
                     }
                 break;
                 case PERCENT_OF_AMOUNT:
@@ -996,6 +1005,15 @@ public class LoanChargeService {
     private boolean doesLoanChargePaidByContainLoanCharge(Set<LoanChargePaidBy> loanChargePaidBys, LoanCharge loanCharge) {
         return loanChargePaidBys.stream() //
                 .anyMatch(loanChargePaidBy -> loanChargePaidBy.getLoanCharge().equals(loanCharge));
+    }
+
+    private void validateChargeAmountNotZero(final LoanCharge loanCharge) {
+        if (loanCharge.getAmount() != null && BigDecimal.ZERO.compareTo(loanCharge.getAmount()) == 0) {
+            final String defaultUserMessage = "This charge cannot be added because the calculated amount becomes zero after rounding.";
+
+            throw new LoanChargeCannotBeAddedException("loanCharge", "amount.rounded.to.zero", defaultUserMessage,
+                    loanCharge.getLoan().getId(), loanCharge.name());
+        }
     }
 
 }
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/ClientCharge.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/ClientCharge.java
@@ -26,7 +26,6 @@ import jakarta.persistence.Table;
 import jakarta.persistence.Transient;
 import java.math.BigDecimal;
 import java.time.LocalDate;
-import org.apache.fineract.infrastructure.core.api.JsonCommand;
 import org.apache.fineract.infrastructure.core.domain.AbstractPersistableCustom;
 import org.apache.fineract.organisation.monetary.domain.MonetaryCurrency;
 import org.apache.fineract.organisation.monetary.domain.Money;
@@ -34,7 +33,6 @@ import org.apache.fineract.organisation.monetary.domain.OrganisationCurrency;
 import org.apache.fineract.portfolio.charge.domain.Charge;
 import org.apache.fineract.portfolio.charge.domain.ChargeCalculationType;
 import org.apache.fineract.portfolio.charge.domain.ChargeTimeType;
-import org.apache.fineract.portfolio.client.api.ClientApiConstants;
 
 @Entity
 @Table(name = "m_client_charge")
@@ -94,12 +92,8 @@ public class ClientCharge extends AbstractPersistableCustom<Long> {
         //
     }
 
-    public static ClientCharge createNew(final Client client, final Charge charge, final JsonCommand command) {
-        BigDecimal amount = command.bigDecimalValueOfParameterNamed(ClientApiConstants.amountParamName);
-        final LocalDate dueDate = command.localDateValueOfParameterNamed(ClientApiConstants.dueAsOfDateParamName);
+    public static ClientCharge createNew(final Client client, final Charge charge, final BigDecimal amount, final LocalDate dueDate) {
         final boolean status = true;
-        // Derive from charge definition if not passed in as a parameter
-        amount = (amount == null) ? charge.getAmount() : amount;
         return new ClientCharge(client, charge, amount, dueDate, status);
     }
 
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/client/service/ClientChargeWritePlatformServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/client/service/ClientChargeWritePlatformServiceImpl.java
@@ -42,6 +42,9 @@ import org.apache.fineract.infrastructure.core.exception.PlatformDataIntegrityEx
 import org.apache.fineract.infrastructure.core.service.DateUtils;
 import org.apache.fineract.infrastructure.core.service.ExternalIdFactory;
 import org.apache.fineract.organisation.holiday.domain.HolidayRepositoryWrapper;
+import org.apache.fineract.organisation.monetary.data.CurrencyData;
+import org.apache.fineract.organisation.monetary.domain.ApplicationCurrency;
+import org.apache.fineract.organisation.monetary.domain.ApplicationCurrencyRepositoryWrapper;
 import org.apache.fineract.organisation.monetary.domain.Money;
 import org.apache.fineract.organisation.workingdays.domain.WorkingDaysRepositoryWrapper;
 import org.apache.fineract.portfolio.charge.domain.Charge;
@@ -78,6 +81,7 @@ public class ClientChargeWritePlatformServiceImpl implements ClientChargeWritePl
     private final ClientTransactionRepository clientTransactionRepository;
     private final PaymentDetailWritePlatformService paymentDetailWritePlatformService;
     private final JournalEntryWritePlatformService journalEntryWritePlatformService;
+    private final ApplicationCurrencyRepositoryWrapper applicationCurrencyRepositoryWrapper;
 
     @Override
     public CommandProcessingResult addCharge(Long clientId, JsonCommand command) {
@@ -95,7 +99,10 @@ public class ClientChargeWritePlatformServiceImpl implements ClientChargeWritePl
                 throw new ChargeCannotBeAppliedToException("client", errorMessage, charge.getId());
             }
 
-            final ClientCharge clientCharge = ClientCharge.createNew(client, charge, command);
+            Money roundedAmount = calculateRoundedChargeAmount(charge, command);
+            validateChargeAmountNotZero(roundedAmount);
+            final LocalDate date = command.localDateValueOfParameterNamed(ClientApiConstants.dueAsOfDateParamName);
+            final ClientCharge clientCharge = ClientCharge.createNew(client, charge, roundedAmount.getAmount(), date);
             final DateTimeFormatter fmt = DateTimeFormatter.ofPattern(command.dateFormat());
             final List<ApiParameterError> dataValidationErrors = new ArrayList<>();
             final DataValidatorBuilder baseDataValidator = new DataValidatorBuilder(dataValidationErrors)
@@ -426,6 +433,25 @@ public class ClientChargeWritePlatformServiceImpl implements ClientChargeWritePl
         log.error("Error occured.", dve);
         throw ErrorHandler.getMappable(dve, "error.msg.client.charges.unknown.data.integrity.issue",
                 "Unknown data integrity issue with resource: " + realCause.getMessage());
+    }
+
+    private Money calculateRoundedChargeAmount(final Charge charge, final JsonCommand command) {
+        BigDecimal amount = command.bigDecimalValueOfParameterNamed(ClientApiConstants.amountParamName);
+        amount = (amount == null) ? charge.getAmount() : amount;
+
+        ApplicationCurrency currency = this.applicationCurrencyRepositoryWrapper.findOneWithNotFoundDetection(charge.getCurrencyCode());
+
+        CurrencyData currencyData = currency.toData();
+        return Money.of(currencyData, amount);
+    }
+
+    private void validateChargeAmountNotZero(Money amount) {
+        if (amount.isZero()) {
+            List<ApiParameterError> errors = new ArrayList<>();
+            errors.add(ApiParameterError.parameterError("error.msg.client.charge.amount.rounded.to.zero",
+                    "This charge cannot be added because the calculated amount becomes zero after rounding.", "amount"));
+            throw new PlatformApiDataValidationException(errors);
+        }
     }
 
 }
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountWritePlatformServiceJpaRepositoryImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountWritePlatformServiceJpaRepositoryImpl.java
@@ -87,6 +87,7 @@ import org.apache.fineract.portfolio.account.domain.StandingInstructionStatus;
 import org.apache.fineract.portfolio.account.service.AccountAssociationsReadPlatformService;
 import org.apache.fineract.portfolio.account.service.AccountTransfersReadPlatformService;
 import org.apache.fineract.portfolio.charge.domain.Charge;
+import org.apache.fineract.portfolio.charge.domain.ChargeCalculationType;
 import org.apache.fineract.portfolio.charge.domain.ChargeRepositoryWrapper;
 import org.apache.fineract.portfolio.charge.domain.ChargeTimeType;
 import org.apache.fineract.portfolio.client.domain.Client;
@@ -1237,6 +1238,8 @@ public class SavingsAccountWritePlatformServiceJpaRepositoryImpl implements Savi
         }
         final SavingsAccountCharge savingsAccountCharge = SavingsAccountCharge.createNewFromJson(savingsAccount, chargeDefinition, command);
 
+        validateChargeAmountNotZero(savingsAccountCharge, dataValidationErrors);
+
         if (chargeDefinition.isEnableFreeWithdrawal()) {
             savingsAccountCharge.setFreeWithdrawalCount(0);
         }
@@ -2078,6 +2081,20 @@ public class SavingsAccountWritePlatformServiceJpaRepositoryImpl implements Savi
     private void validateReasonForHold(String reasonForBlock) {
         if (StringUtils.isBlank(reasonForBlock)) {
             throw new PlatformDataIntegrityException("Reason For Block is Mandatory", "error.msg.reason.for.block.mandatory");
+        }
+    }
+
+    private void validateChargeAmountNotZero(final SavingsAccountCharge savingsAccountCharge,
+            final List<ApiParameterError> dataValidationErrors) {
+        if (!ChargeCalculationType.fromInt(savingsAccountCharge.getCharge().getChargeCalculation()).isFlat()) {
+            return;
+        }
+
+        if (savingsAccountCharge.getAmount(savingsAccountCharge.savingsAccount().getCurrency()).isZero()) {
+            final String defaultUserMessage = "This charge cannot be added because the calculated amount becomes zero after rounding.";
+
+            dataValidationErrors
+                    .add(ApiParameterError.parameterError("error.msg.savings.charge.amount.rounded.to.zero", defaultUserMessage, "amount"));
         }
     }
 }
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/shareaccounts/domain/ShareAccountTransaction.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/shareaccounts/domain/ShareAccountTransaction.java
@@ -114,7 +114,7 @@ public class ShareAccountTransaction extends AbstractPersistableCustom<Long> {
         final BigDecimal unitPrice = null;
         final Integer status = PurchasedSharesStatusType.APPROVED.getValue();
         final Integer type = PurchasedSharesStatusType.CHARGE_PAYMENT.getValue();
-        BigDecimal amount = charge.percentageOrAmount();
+        BigDecimal amount = charge.amoutOutstanding();
         BigDecimal chargeAmount = null;
         BigDecimal amountPaid = null;
         return new ShareAccountTransaction(transactionDate, totalShares, unitPrice, status, type, amount, chargeAmount, amountPaid);
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/shareaccounts/domain/ShareAccountCharge.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/shareaccounts/domain/ShareAccountCharge.java
@@ -374,7 +374,7 @@ public class ShareAccountCharge extends AbstractPersistableCustom<Long> {
     }
 
     public BigDecimal deriveChargeAmount(BigDecimal transactionAmount, final MonetaryCurrency currency) {
-        BigDecimal toReturnAmount = amountOrPercentage;
+        BigDecimal toReturnAmount;
         if (ChargeCalculationType.fromInt(this.chargeCalculation) == ChargeCalculationType.PERCENT_OF_AMOUNT) {
             toReturnAmount = Money.of(currency, percentageOf(transactionAmount, this.percentage)).getAmount();
             this.amountPercentageAppliedTo = transactionAmount;
@@ -384,7 +384,8 @@ public class ShareAccountCharge extends AbstractPersistableCustom<Long> {
             this.amountWaived = null;
             this.amountWrittenOff = null;
         } else {
-            this.amount = this.amountOrPercentage;
+            this.amount = Money.of(currency, this.amountOrPercentage).getAmount();
+            toReturnAmount = this.amount;
             this.amountOutstanding = calculateOutstanding();
             this.amountWaived = null;
             this.amountWrittenOff = null;
@@ -393,7 +394,7 @@ public class ShareAccountCharge extends AbstractPersistableCustom<Long> {
     }
 
     public BigDecimal updateChargeDetailsForAdditionalSharesRequest(final BigDecimal transactionAmount, final MonetaryCurrency currency) {
-        BigDecimal toReturnAmount = amountOrPercentage;
+        BigDecimal toReturnAmount;
         if (ChargeCalculationType.fromInt(this.chargeCalculation) == ChargeCalculationType.PERCENT_OF_AMOUNT) {
             toReturnAmount = Money.of(currency, percentageOf(transactionAmount, this.percentage)).getAmount();
             this.amountPercentageAppliedTo = this.amountPercentageAppliedTo.add(transactionAmount);
@@ -402,7 +403,9 @@ public class ShareAccountCharge extends AbstractPersistableCustom<Long> {
             this.amountWaived = null;
             this.amountWrittenOff = null;
         } else {
-            this.amount = this.amount.add(this.amountOrPercentage);
+            BigDecimal roundedAmount = Money.of(currency, this.amountOrPercentage).getAmount();
+            toReturnAmount = roundedAmount;
+            this.amount = this.amount.add(roundedAmount);
             this.amountOutstanding = calculateOutstanding();
             this.amountWaived = null;
             this.amountWrittenOff = null;
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/shareaccounts/serialization/ShareAccountDataSerializer.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/shareaccounts/serialization/ShareAccountDataSerializer.java
@@ -242,9 +242,10 @@ public class ShareAccountDataSerializer {
         LocalDate currentDate = DateUtils.getBusinessLocalDate();
         for (ShareAccountCharge charge : charges) {
             if (charge.isActive() && charge.isShareAccountActivation()) {
-                charge.deriveChargeAmount(totalChargeAmount, account.getCurrency());
+                BigDecimal amount = charge.deriveChargeAmount(totalChargeAmount, account.getCurrency());
+                validateChargeAmountNotZero(amount);
                 ShareAccountTransaction chargeTransaction = ShareAccountTransaction.createChargeTransaction(currentDate, charge);
-                ShareAccountChargePaidBy paidBy = new ShareAccountChargePaidBy(chargeTransaction, charge, charge.percentageOrAmount());
+                ShareAccountChargePaidBy paidBy = new ShareAccountChargePaidBy(chargeTransaction, charge, amount);
                 chargeTransaction.addShareAccountChargePaidBy(paidBy);
                 account.addChargeTransaction(chargeTransaction);
             }
@@ -255,6 +256,7 @@ public class ShareAccountDataSerializer {
             for (ShareAccountCharge charge : charges) {
                 if (charge.isActive() && charge.isSharesPurchaseCharge()) {
                     BigDecimal amount = charge.deriveChargeAmount(pending.amount(), account.getCurrency());
+                    validateChargeAmountNotZero(amount);
                     ShareAccountChargePaidBy paidBy = new ShareAccountChargePaidBy(pending, charge, amount);
                     pending.addShareAccountChargePaidBy(paidBy);
                     totalChargeAmount = totalChargeAmount.add(amount);
@@ -1042,5 +1044,14 @@ public class ShareAccountDataSerializer {
         handleRedeemSharesChargeTransactions(account, transaction);
         actualChanges.put(ShareAccountApiConstants.requestedshares_paramname, transaction);
         return actualChanges;
+    }
+
+    private void validateChargeAmountNotZero(BigDecimal amount) {
+        if (amount.compareTo(BigDecimal.ZERO) <= 0) {
+            List<ApiParameterError> errors = new ArrayList<>();
+            errors.add(ApiParameterError.parameterError("error.msg.share.charge.amount.rounded.to.zero",
+                    "This charge cannot be added because the calculated amount becomes zero after rounding.", "amount"));
+            throw new PlatformApiDataValidationException(errors);
+        }
     }
 }
--- a/fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccountCharge.java
+++ b/fineract-savings/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccountCharge.java
@@ -263,11 +263,12 @@ public class SavingsAccountCharge extends AbstractAuditableWithUTCDateTimeCustom
                 this.amountWrittenOff = null;
             break;
             case FLAT:
+                Money money = Money.of(this.savingsAccount().getCurrency(), chargeAmount);
                 this.percentage = null;
-                this.amount = chargeAmount;
+                this.amount = money.getAmount();
                 this.amountPercentageAppliedTo = null;
                 this.amountPaid = null;
-                this.amountOutstanding = chargeAmount;
+                this.amountOutstanding = money.getAmount();
                 this.amountWaived = null;
                 this.amountWrittenOff = null;
             break;
--- a/integration-tests/src/test/java/org/apache/fineract/integrationtests/ClientChargeRoundingTest.java
+++ b/integration-tests/src/test/java/org/apache/fineract/integrationtests/ClientChargeRoundingTest.java
@@ -1,0 +1,194 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.integrationtests;
+
+import static io.restassured.RestAssured.given;
+import static org.apache.fineract.integrationtests.common.ClientHelper.addChargesForClient;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import io.restassured.response.Response;
+import java.io.IOException;
+import java.math.BigDecimal;
+import java.util.List;
+import java.util.UUID;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.fineract.client.models.ChargeRequest;
+import org.apache.fineract.client.models.CurrencyConfigurationData;
+import org.apache.fineract.client.models.CurrencyUpdateRequest;
+import org.apache.fineract.client.models.GetClientsChargesPageItems;
+import org.apache.fineract.client.models.GetClientsClientIdChargesResponse;
+import org.apache.fineract.client.models.PostChargesResponse;
+import org.apache.fineract.integrationtests.common.ClientHelper;
+import org.apache.fineract.integrationtests.common.FineractClientHelper;
+import org.apache.fineract.integrationtests.common.Utils;
+import org.apache.fineract.integrationtests.common.charges.ChargesHelper;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+@Slf4j
+public class ClientChargeRoundingTest extends BaseLoanIntegrationTest {
+
+    private Long clientId;
+    private ChargesHelper chargesHelper;
+    private static final String DATE = "01 January 2026";
+
+    @BeforeEach
+    public void setup() throws Exception {
+        enableRequiredCurrencies();
+        clientId = ClientHelper.createClient(ClientHelper.defaultClientCreationRequest()).getClientId();
+        chargesHelper = new ChargesHelper();
+    }
+
+    @Test
+    public void shouldRoundUsdClientChargeTo_TwoDecimalPlaces() throws Exception {
+        PostChargesResponse chargesResponse = createFlatClientCharge(19.876, "USD");
+
+        Integer appliedChargeId = applyChargeToClient(clientId, chargesResponse.getResourceId(), new BigDecimal("19.876"));
+
+        GetClientsChargesPageItems charge = getClientCharge(clientId, appliedChargeId.longValue());
+
+        assertNotNull(charge);
+        BigDecimal actualChargeAmount = charge.getAmount();
+        BigDecimal expectedChargeAmount = new BigDecimal("19.88");
+        assertAmountEquals(expectedChargeAmount, actualChargeAmount);
+    }
+
+    @Test
+    public void shouldRoundJpyClientChargeTo_ZeroDecimalPlaces() throws Exception {
+        PostChargesResponse chargesResponse = createFlatClientCharge(19.8, "JPY");
+
+        Integer appliedChargeId = applyChargeToClient(clientId, chargesResponse.getResourceId(), new BigDecimal("19.8"));
+
+        GetClientsChargesPageItems charge = getClientCharge(clientId, appliedChargeId.longValue());
+
+        assertNotNull(charge);
+        BigDecimal actualChargeAmount = charge.getAmount();
+        BigDecimal expectedChargeAmount = new BigDecimal("20");
+        assertAmountEquals(expectedChargeAmount, actualChargeAmount);
+    }
+
+    @Test
+    public void shouldRoundUpJpyClientCharge_whenValueIsAboveHalfTo_ZeroDecimalPlaces() throws Exception {
+        PostChargesResponse chargesResponse = createFlatClientCharge(0.55, "JPY");
+
+        Integer appliedChargeId = applyChargeToClient(clientId, chargesResponse.getResourceId(), new BigDecimal("0.55"));
+
+        GetClientsChargesPageItems charge = getClientCharge(clientId, appliedChargeId.longValue());
+
+        assertNotNull(charge);
+        BigDecimal actualChargeAmount = charge.getAmount();
+        BigDecimal expectedChargeAmount = new BigDecimal("1");
+        assertAmountEquals(expectedChargeAmount, actualChargeAmount);
+    }
+
+    @Test
+    public void shouldFailToAddJpyClientCharge_whenRoundedToZero() throws Exception {
+        PostChargesResponse chargesResponse = createFlatClientCharge(0.5, "JPY");
+
+        Response response = applyChargeToClientRaw(clientId, chargesResponse.getResourceId(), new BigDecimal("0.5"));
+
+        assertEquals(400, response.statusCode());
+        String errorBody = response.asString();
+        assertTrue(errorBody.contains("error.msg.client.charge.amount.rounded.to.zero"));
+        assertFalse(hasAnyClientCharges(clientId), "Expected no client charge to be created when rounded amount becomes 0");
+    }
+
+    // -----------------------------
+    // HELPERS
+    // -----------------------------
+
+    private PostChargesResponse createFlatClientCharge(double amount, String currencyCode) {
+        String uniqueChargeName = "Client Charge Flat " + UUID.randomUUID().toString().replace("-", "");
+        return chargesHelper.createCharges(new ChargeRequest().name(uniqueChargeName).chargeAppliesTo(3) // CLIENT
+                .chargeTimeType(2).chargeCalculationType(1) // FLAT
+                .amount(amount).currencyCode(currencyCode).locale("en").active(true).penalty(false));
+    }
+
+    private Integer applyChargeToClient(Long clientId, Long chargeId, BigDecimal amount) {
+        String request = """
+                {
+                  "chargeId": %d,
+                  "amount": %s,
+                  "locale": "en",
+                  "dateFormat": "dd MMMM yyyy",
+                  "dueDate": "%s"
+                }
+                """.formatted(chargeId, amount.toPlainString(), DATE);
+
+        return addChargesForClient(requestSpec, responseSpec, clientId.intValue(), request);
+    }
+
+    private Response applyChargeToClientRaw(Long clientId, Long chargeId, BigDecimal amount) {
+
+        String request = """
+                {
+                  "chargeId": %d,
+                  "amount": %s,
+                  "locale": "en",
+                  "dateFormat": "dd MMMM yyyy",
+                  "dueDate": "%s"
+                }
+                """.formatted(chargeId, amount.toPlainString(), DATE);
+
+        String url = "/fineract-provider/api/v1/clients/" + clientId + "/charges?" + Utils.TENANT_IDENTIFIER;
+
+        return given().spec(requestSpec).body(request).when().post(url);
+    }
+
+    private GetClientsChargesPageItems getClientCharge(Long clientId, Long chargeId) throws IOException {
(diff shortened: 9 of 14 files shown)


fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanCharge.java
```diff
@@ -320,32 +320,24 @@ public static BigDecimal percentageOf(final BigDecimal value, final BigDecimal p
...
      */
-    public BigDecimal minimumAndMaximumCap(final BigDecimal percentageOf) {
-        BigDecimal minMaxCap;
-        if (this.minCap != null) {
-            final int minimumCap = percentageOf.compareTo(this.minCap);
-            if (minimumCap == -1) {
-                minMaxCap = this.minCap;
-                return minMaxCap;
-            }
+    public BigDecimal minimumAndMaximumCap(final BigDecimal value) {
```
adamsaghy: I see no operations which warrants to use Money here... this method just simply comparing values, no calculations which might led to situations where rounding is needed...

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanChargeService.java
```diff
@@ -232,6 +232,11 @@ public void addLoanCharge(final Loan loan, final LoanCharge loanCharge) {
         update(loanCharge, chargeAmt, loanCharge.getDueLocalDate(), amount, loan.fetchNumberOfInstallmentsAfterExceptions(),
                 totalChargeAmt);
 
+        // Skip zero-value charges
```
adamsaghy: Adding 0 charge is incorrect and it should be stopped at validation and or during charge amount calculation. Silent return is misleading and hiding a potential issue!

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanChargeService.java
```diff
@@ -822,9 +828,9 @@ private void update(final LoanCharge loanCharge, final BigDecimal amount, final
                         if (numberOfRepayments == null) {
                             numberOfRepayments = loanCharge.getLoan().fetchNumberOfInstallmentsAfterExceptions();
                         }
-                        loanCharge.setAmount(amount.multiply(BigDecimal.valueOf(numberOfRepayments)));
+                        loanCharge.setAmount(loanCharge.minimumAndMaximumCap(amount.multiply(BigDecimal.valueOf(numberOfRepayments))));
```
adamsaghy: Unrelated change? Why `loanCharge.minimumAndMaximumCap` is needed here? Charge amount calculation should cover the rounding, not this...

fineract-provider/src/main/java/org/apache/fineract/portfolio/accounts/api/AccountsApiResourceSwagger.java
```diff
@@ -465,15 +465,26 @@ public static final class PostAccountsTypeAccountIdRequest {
...
         static final class PostAccountsRequestedShares {
 
             private PostAccountsRequestedShares() {}
 
             @Schema(example = "35")
             public Long id;
         }
-
-        public Set<PostAccountsRequestedShares> requestedShares;
     }
```
adamsaghy: Unrelated changes?

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/ClientCharge.java
```diff
@@ -94,12 +94,9 @@ protected ClientCharge() {
         //
     }
 
-    public static ClientCharge createNew(final Client client, final Charge charge, final JsonCommand command) {
-        BigDecimal amount = command.bigDecimalValueOfParameterNamed(ClientApiConstants.amountParamName);
+    public static ClientCharge createNew(final Client client, final Charge charge, final BigDecimal amount, final JsonCommand command) {
```
adamsaghy: If you’ve started providing the fields as incoming arguments, don’t stop halfway. Remove the `command` parameter and ensure that the due date is provided as a parameter.

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/service/ClientChargeWritePlatformServiceImpl.java
```diff
@@ -95,7 +99,12 @@ public CommandProcessingResult addCharge(Long clientId, JsonCommand command) {
                 throw new ChargeCannotBeAppliedToException("client", errorMessage, charge.getId());
             }
 
-            final ClientCharge clientCharge = ClientCharge.createNew(client, charge, command);
+            BigDecimal roundedAmount = calculateRoundedChargeAmount(charge, command);
```
adamsaghy: Using Money here would be better, no?

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/service/ClientChargeWritePlatformServiceImpl.java
```diff
@@ -95,7 +99,12 @@ public CommandProcessingResult addCharge(Long clientId, JsonCommand command) {
...
+            if (roundedAmount.compareTo(BigDecimal.ZERO) == 0) {
+                return new CommandProcessingResultBuilder().withOfficeId(client.getOffice().getId()).withClientId(client.getId()).build();
```
adamsaghy: Why to accept 0 amount charge? 0 amount charge is incorrect outcome!

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanChargeWritePlatformServiceImpl.java
```diff
@@ -206,21 +206,38 @@ public CommandProcessingResult addLoanCharge(final Long loanId, final JsonComman
             LoanTrancheDisbursementCharge loanTrancheDisbursementCharge;
             ExternalId externalId = externalIdFactory.createFromCommand(command, "externalId");
             boolean isFirst = true;
+            boolean atLeastOneChargePersisted = false;
```
adamsaghy: what are these flags for?

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanChargeWritePlatformServiceImpl.java
```diff
@@ -206,21 +206,38 @@ public CommandProcessingResult addLoanCharge(final Long loanId, final JsonComman
...
+                            null, null, null, dueDate, null, null, externalId);
+                    loanTrancheDisbursementCharge = new LoanTrancheDisbursementCharge(tempCharge, disbursementDetail);
+                    tempCharge.updateLoanTrancheDisbursementCharge(loanTrancheDisbursementCharge);
+                    businessEventNotifierService.notifyPreBusinessEvent(new LoanAddChargeBusinessEvent(tempCharge));
+                    validateAddLoanCharge(loan, chargeDefinition, tempCharge);
+                    addCharge(loan, chargeDefinition, tempCharge);
+
+                    // if charge amount is rounded to zero, then continue and look for another disbursement
+                    if (isZeroCharge(tempCharge)) {
+                        continue;
```
adamsaghy: 0 amount charge sounds incorrect to me.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanChargeWritePlatformServiceImpl.java
```diff
@@ -206,21 +206,38 @@ public CommandProcessingResult addLoanCharge(final Long loanId, final JsonComman
...
+                    }
+
+                    // tempCharge was persisted so this is true
+                    atLeastOneChargePersisted = true;
```
adamsaghy: what does it matter?

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanChargeWritePlatformServiceImpl.java
```diff
@@ -231,18 +248,36 @@ public CommandProcessingResult addLoanCharge(final Long loanId, final JsonComman
                     isFirst = false;
                 }
             }
-            if (loanCharge == null) {
+            // No eligible disbursement
+            if (!eligibleDisbursementFound) {
```
adamsaghy: I dont like this...

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanChargeWritePlatformServiceImpl.java
```diff
@@ -231,18 +248,36 @@ public CommandProcessingResult addLoanCharge(final Long loanId, final JsonComman
...
+            // At least one valid persisted charge
             loan.addTrancheLoanCharge(chargeDefinition);
         } else {
             loanCharge = loanChargeAssembler.createNewFromJson(loan, chargeDefinition, command);
             businessEventNotifierService.notifyPreBusinessEvent(new LoanAddChargeBusinessEvent(loanCharge));
-
             validateAddLoanCharge(loan, chargeDefinition, loanCharge);
             isAppliedOnBackDate = addCharge(loan, chargeDefinition, loanCharge);
+
+            // if charge amount is rounded to zero, then return early
```
adamsaghy: Sounds incorrect result to me.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanChargeWritePlatformServiceImpl.java
```diff
@@ -1052,32 +1087,39 @@ private boolean addCharge(final Loan loan, final Charge chargeDefinition, LoanCh
...
-                this.loanTransactionRepository.saveAndFlush(applyLoanChargeTransaction);
-                loanJournalEntryPoster.postJournalEntriesForLoanTransaction(applyLoanChargeTransaction, false, false);
-                businessEventNotifierService
-                        .notifyPostBusinessEvent(new LoanAccrualTransactionCreatedBusinessEvent(applyLoanChargeTransaction));
+
+        // If charge amount is not zero
+        if (!isZeroCharge(loanCharge)) {
+            loanCharge = this.loanChargeRepository.saveAndFlush(loanCharge);
+
+            // we want to apply charge transactions only for those loans charges that are applied when a loan is active
```
adamsaghy: why to change this?

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanChargeWritePlatformServiceImpl.java
```diff
@@ -1052,32 +1087,39 @@ private boolean addCharge(final Loan loan, final Charge chargeDefinition, LoanCh
...
-                || loan.getStatus().isClosedObligationsMet()) {
-            final LoanTransaction applyLoanChargeTransaction = loanChargeService.handleChargeAppliedTransaction(loan, loanCharge, null);
-            if (applyLoanChargeTransaction != null) {
-                this.loanTransactionRepository.saveAndFlush(applyLoanChargeTransaction);
-                loanJournalEntryPoster.postJournalEntriesForLoanTransaction(applyLoanChargeTransaction, false, false);
-                businessEventNotifierService
-                        .notifyPostBusinessEvent(new LoanAccrualTransactionCreatedBusinessEvent(applyLoanChargeTransaction));
+
+        // If charge amount is not zero
+        if (!isZeroCharge(loanCharge)) {
```
adamsaghy: I dont like this..

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountWritePlatformServiceJpaRepositoryImpl.java
```diff
@@ -1236,6 +1238,11 @@ public CommandProcessingResult addSavingsAccountCharge(final JsonCommand command
         }
         final SavingsAccountCharge savingsAccountCharge = SavingsAccountCharge.createNewFromJson(savingsAccount, chargeDefinition, command);
 
+        if (isFlatCharge(savingsAccountCharge) && isZeroCharge(savingsAccountCharge, savingsAccount.getCurrency())) {
```
adamsaghy: I dont like it... 0 amount charge is incorrect result.

adamsaghy (changes requested):
I stopped review at half time.. I dont like this approach, neither i think it is correct.

The introduced 0 amount handling is incorrect. 0 amount is incorrect outcome and charge cannot be created with this value, but nevertheless, we cannot accept the charge creation action and after not create the charge. This is incorrect and unwanted behaviour:
- Add proper validation (first thing to do)
- Calculate the amount, if it is incorrect (like 0 amount) -> Raise error and explain the issue!
- Remove the unnecessary flags and checks... it makes the code hard to read and maintained...
- Remove the unrelated changes, the PR should focus on 1 thing only!

Dhanno98:
@adamsaghy Thanks for your suggestions! I am working on the fix.

Dhanno98:
Hi @adamsaghy,
I have updated the charge attachment flows locally to reject charges whose calculated amount becomes 0 after applying account currency rounding rules, instead of returning silently.

I have a question regarding future charges such as Savings Account Percentage of Withdrawal Charges and Share Account Redeem Charges.

In these cases, the charge is attached successfully first and the amount is calculated only when the business transaction occurs. If the calculated charge amount becomes 0 after rounding, what is the expected behavior:

1. Reject the business transaction (withdrawal/redeem) with a validation error.
2. Allow the transaction but skip creating/applying the charge similar to silent return.
3. Keep the current behavior and allow a 0 value charge result.

Could you please clarify the expected behavior for these future-charge scenarios so I can align the implementation consistently across Savings and Shares?

Thank you!

adamsaghy:
> Hi @adamsaghy, I have updated the charge attachment flows locally to reject charges whose calculated amount becomes 0 after applying account currency rounding rules, instead of returning silently.
> 
> I have a question regarding future charges such as Savings Account Percentage of Withdrawal Charges and Share Account Redeem Charges.
> 
> In these cases, the charge is attached successfully first and the amount is calculated only when the business transaction occurs. If the calculated charge amount becomes 0 after rounding, what is the expected behavior:
> 
> 1. Reject the business transaction (withdrawal/redeem) with a validation error.
> 2. Allow the transaction but skip creating/applying the charge similar to silent return.
> 3. Keep the current behavior and allow a 0 value charge result.
> 
> Could you please clarify the expected behavior for these future-charge scenarios so I can align the implementation consistently across Savings and Shares?
> 
> Thank you!

@bharathc27 fyi

@Dhanno98 let me review

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanChargeService.java
adamsaghy: See no changes...no reply...

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanChargeService.java
adamsaghy: See no changes...no reply...

fineract-provider/src/main/java/org/apache/fineract/portfolio/accounts/api/AccountsApiResourceSwagger.java
adamsaghy: See no changes...no reply...

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/ClientCharge.java
adamsaghy: See no changes...no reply...

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/service/ClientChargeWritePlatformServiceImpl.java
adamsaghy: See no changes...no reply...

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/service/ClientChargeWritePlatformServiceImpl.java
adamsaghy: See no changes...no reply...

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanChargeWritePlatformServiceImpl.java
adamsaghy: See no changes...no reply...

adamsaghy:
> Hi @adamsaghy, I have updated the charge attachment flows locally to reject charges whose calculated amount becomes 0 after applying account currency rounding rules, instead of returning silently.
> 
> I have a question regarding future charges such as Savings Account Percentage of Withdrawal Charges and Share Account Redeem Charges.
> 
> In these cases, the charge is attached successfully first and the amount is calculated only when the business transaction occurs. If the calculated charge amount becomes 0 after rounding, what is the expected behavior:
> 
> 1. Reject the business transaction (withdrawal/redeem) with a validation error.
> 2. Allow the transaction but skip creating/applying the charge similar to silent return.
> 3. Keep the current behavior and allow a 0 value charge result.
> 
> Could you please clarify the expected behavior for these future-charge scenarios so I can align the implementation consistently across Savings and Shares?
> 
> Thank you!

I see no changes, no reply on my questions and concerns.. i have stopped reviewing the PR at half time...

Dhanno98:
> > Hi @adamsaghy, I have updated the charge attachment flows locally to reject charges whose calculated amount becomes 0 after applying account currency rounding rules, instead of returning silently.
> > I have a question regarding future charges such as Savings Account Percentage of Withdrawal Charges and Share Account Redeem Charges.
> > In these cases, the charge is attached successfully first and the amount is calculated only when the business transaction occurs. If the calculated charge amount becomes 0 after rounding, what is the expected behavior:
> > 
> > 1. Reject the business transaction (withdrawal/redeem) with a validation error.
> > 2. Allow the transaction but skip creating/applying the charge similar to silent return.
> > 3. Keep the current behavior and allow a 0 value charge result.
> > 
> > Could you please clarify the expected behavior for these future-charge scenarios so I can align the implementation consistently across Savings and Shares?
> > Thank you!
> 
> I see no changes, no reply on my questions and concerns.. i have stopped reviewing the PR at half time...

Hi @adamsaghy,

Sorry for the confusion. When I wrote that I had updated the charge attachment flows locally, I meant that the changes were only completed in my local branch and had not yet been pushed or amended into the PR.

My intention was to first finish addressing all of your review comments, validate everything thoroughly, and then push a single updated commit together with replies to each review comment. Because of that, I had not yet responded individually to the review feedback.

I realize now that my comment made it sound like the PR had already been updated, which was not the case. Sorry for causing you to spend time reopening the review.

I am currently working through all of the review comments and will push the updated changes together with responses shortly.

Regarding the future-charge scenarios, that question was meant as a design clarification while implementing the remaining changes so I can keep the behavior consistent across Savings and Shares. This is the only thing that I need to understand and resolve and then I will ammend the commit.

Thank you for your patience and for the detailed review feedback.

adamsaghy:
> > > Hi @adamsaghy, I have updated the charge attachment flows locally to reject charges whose calculated amount becomes 0 after applying account currency rounding rules, instead of returning silently.
> > > I have a question regarding future charges such as Savings Account Percentage of Withdrawal Charges and Share Account Redeem Charges.
> > > In these cases, the charge is attached successfully first and the amount is calculated only when the business transaction occurs. If the calculated charge amount becomes 0 after rounding, what is the expected behavior:
> > > 
> > > 1. Reject the business transaction (withdrawal/redeem) with a validation error.
> > > 2. Allow the transaction but skip creating/applying the charge similar to silent return.
> > > 3. Keep the current behavior and allow a 0 value charge result.
> > > 
> > > Could you please clarify the expected behavior for these future-charge scenarios so I can align the implementation consistently across Savings and Shares?
> > > Thank you!
> > 
> > 
> > I see no changes, no reply on my questions and concerns.. i have stopped reviewing the PR at half time...
> 
> Hi @adamsaghy,
> 
> Sorry for the confusion. When I wrote that I had updated the charge attachment flows locally, I meant that the changes were only completed in my local branch and had not yet been pushed or amended into the PR.
> 
> My intention was to first finish addressing all of your review comments, validate everything thoroughly, and then push a single updated commit together with replies to each review comment. Because of that, I had not yet responded individually to the review feedback.
> 
> I realize now that my comment made it sound like the PR had already been updated, which was not the case. Sorry for causing you to spend time reopening the review.
> 
> I am currently working through all of the review comments and will push the updated changes together with responses shortly.
> 
> Regarding the future-charge scenarios, that question was meant as a design clarification while implementing the remaining changes so I can keep the behavior consistent across Savings and Shares. This is the only thing that I need to understand and resolve and then I will ammend the commit.
> 
> Thank you for your patience and for the detailed review feedback.

Thank you for the clarification ;)

Regarding zero amount: Lets use option 1 **Reject the business transaction (withdrawal/redeem) with a validation error.**

Dhanno98:
> > > > Hi @adamsaghy, I have updated the charge attachment flows locally to reject charges whose calculated amount becomes 0 after applying account currency rounding rules, instead of returning silently.
> > > > I have a question regarding future charges such as Savings Account Percentage of Withdrawal Charges and Share Account Redeem Charges.
> > > > In these cases, the charge is attached successfully first and the amount is calculated only when the business transaction occurs. If the calculated charge amount becomes 0 after rounding, what is the expected behavior:
> > > > 
> > > > 1. Reject the business transaction (withdrawal/redeem) with a validation error.
> > > > 2. Allow the transaction but skip creating/applying the charge similar to silent return.
> > > > 3. Keep the current behavior and allow a 0 value charge result.
> > > > 
> > > > Could you please clarify the expected behavior for these future-charge scenarios so I can align the implementation consistently across Savings and Shares?
> > > > Thank you!
> > > 
> > > 
> > > I see no changes, no reply on my questions and concerns.. i have stopped reviewing the PR at half time...
> > 
> > 
> > Hi @adamsaghy,
> > Sorry for the confusion. When I wrote that I had updated the charge attachment flows locally, I meant that the changes were only completed in my local branch and had not yet been pushed or amended into the PR.
> > My intention was to first finish addressing all of your review comments, validate everything thoroughly, and then push a single updated commit together with replies to each review comment. Because of that, I had not yet responded individually to the review feedback.
> > I realize now that my comment made it sound like the PR had already been updated, which was not the case. Sorry for causing you to spend time reopening the review.
> > I am currently working through all of the review comments and will push the updated changes together with responses shortly.
> > Regarding the future-charge scenarios, that question was meant as a design clarification while implementing the remaining changes so I can keep the behavior consistent across Savings and Shares. This is the only thing that I need to understand and resolve and then I will ammend the commit.
> > Thank you for your patience and for the detailed review feedback.
> 
> Thank you for the clarification ;)
> 
> Regarding zero amount: Lets use option 1 **Reject the business transaction (withdrawal/redeem) with a validation error.**

Thank you for the clarification.

Just to confirm the expected behavior for percentage-based future charges:

Suppose a Savings Account has a "Percentage of Withdrawal" charge configured as 0.5%.

When the charge is attached to the account, the percentage value (0.5%) is stored successfully because the actual charge amount is not yet known.

Later, if a withdrawal of 100 is performed and the account currency uses 0 decimal places (for example JPY), the calculated charge amount would be:

0.5% × 100 = 0.5

which would then round to 0 according to the currency rounding rules.

With option 1, my understanding is that the withdrawal transaction itself should be rejected with a validation error because the calculated charge amount becomes zero after rounding.

Is that the intended behavior, even though the withdrawal amount itself is valid and only the associated charge rounds to zero?

I just want to confirm this edge case before implementing the validation in the future-charge flows.

adamsaghy:
> > > > > Hi @adamsaghy, I have updated the charge attachment flows locally to reject charges whose calculated amount becomes 0 after applying account currency rounding rules, instead of returning silently.
> > > > > I have a question regarding future charges such as Savings Account Percentage of Withdrawal Charges and Share Account Redeem Charges.
> > > > > In these cases, the charge is attached successfully first and the amount is calculated only when the business transaction occurs. If the calculated charge amount becomes 0 after rounding, what is the expected behavior:
> > > > > 
> > > > > 1. Reject the business transaction (withdrawal/redeem) with a validation error.
> > > > > 2. Allow the transaction but skip creating/applying the charge similar to silent return.
> > > > > 3. Keep the current behavior and allow a 0 value charge result.
> > > > > 
> > > > > Could you please clarify the expected behavior for these future-charge scenarios so I can align the implementation consistently across Savings and Shares?
> > > > > Thank you!
> > > > 
> > > > 
> > > > I see no changes, no reply on my questions and concerns.. i have stopped reviewing the PR at half time...
> > > 
> > > 
> > > Hi @adamsaghy,
> > > Sorry for the confusion. When I wrote that I had updated the charge attachment flows locally, I meant that the changes were only completed in my local branch and had not yet been pushed or amended into the PR.
> > > My intention was to first finish addressing all of your review comments, validate everything thoroughly, and then push a single updated commit together with replies to each review comment. Because of that, I had not yet responded individually to the review feedback.
> > > I realize now that my comment made it sound like the PR had already been updated, which was not the case. Sorry for causing you to spend time reopening the review.
> > > I am currently working through all of the review comments and will push the updated changes together with responses shortly.
> > > Regarding the future-charge scenarios, that question was meant as a design clarification while implementing the remaining changes so I can keep the behavior consistent across Savings and Shares. This is the only thing that I need to understand and resolve and then I will ammend the commit.
> > > Thank you for your patience and for the detailed review feedback.
> > 
> > 
> > Thank you for the clarification ;)
> > Regarding zero amount: Lets use option 1 **Reject the business transaction (withdrawal/redeem) with a validation error.**
> 
> Thank you for the clarification.
> 
> Just to confirm the expected behavior for percentage-based future charges:
> 
> Suppose a Savings Account has a "Percentage of Withdrawal" charge configured as 0.5%.
> 
> When the charge is attached to the account, the percentage value (0.5%) is stored successfully because the actual charge amount is not yet known.
> 
> Later, if a withdrawal of 100 is performed and the account currency uses 0 decimal places (for example JPY), the calculated charge amount would be:
> 
> 0.5% × 100 = 0.5
> 
> which would then round to 0 according to the currency rounding rules.
> 
> With option 1, my understanding is that the withdrawal transaction itself should be rejected with a validation error because the calculated charge amount becomes zero after rounding.
> 
> Is that the intended behavior, even though the withdrawal amount itself is valid and only the associated charge rounds to zero?
> 
> I just want to confirm this edge case before implementing the validation in the future-charge flows.

With this use case, it’s not as straightforward as it seems. I might have been overly strict here. Just because the charge is too small, we still want to allow the withdrawal. So, if the calculated charge is zero, we simply ignore it and don’t create it.

@bharathc27, what are your thoughts on this?

bharathcgowda:
@adamsaghy @Dhanno98 In any entity and scenario if the calculated charge amount is zero (after rounding) that charge should not be created.
In your latest example, "Allow the transaction but skip creating/applying the charge, similar to silent return," sounds right to me.
Any transactions associated with charges should not be affected. Hence withdraw transaction happens with no charge posting in the above example

Dhanno98:
> @adamsaghy @Dhanno98 In any entity and scenario if the calculated charge amount is zero (after rounding) that charge should not be created. In your latest example, "Allow the transaction but skip creating/applying the charge, similar to silent return," sounds right to me. Any transactions associated with charges should not be affected. Hence withdraw transaction happens with no charge posting in the above example

Hi @bharathcgowda @adamsaghy 
One follow-up question regarding future charges.

For percentage-based future charges (withdrawal/redeem), I understand that the charge definition should be stored and, when the calculated amount rounds to zero, the transaction should continue and the charge should simply not be applied.

For flat future charges (for example a Shares Flat Redeem Charge of 0.5 in a currency with 0 decimal places), the rounded charge amount is known to be zero already at attachment time and can never become non-zero later.

Currently, the existing behavior appears to allow such a Flat Redeem Charge to be attached to the share account without any validation. During the attachment flow, the configured charge amount (0.5 in this example) is stored as-is and no currency rounding is performed. The amount is only passed through `Money.of()` later during the redeem-shares flow, where the calculated charge amount becomes 0 and is effectively not applied.

For this scenario, what should be the expected behavior:

1. Keep the current behavior: allow the charge to be attached as-is and, during redemption, simply ignore it if the rounded amount becomes 0, similar to Percentage of Redeem Charges.

2. Round the flat redeem charge during attachment and reject it if the amount rounds to 0 by throwing a validation exception in the charge attachment flow itself.

adamsaghy:
> > @adamsaghy @Dhanno98 In any entity and scenario if the calculated charge amount is zero (after rounding) that charge should not be created. In your latest example, "Allow the transaction but skip creating/applying the charge, similar to silent return," sounds right to me. Any transactions associated with charges should not be affected. Hence withdraw transaction happens with no charge posting in the above example
> 
> Hi @bharathcgowda @adamsaghy One follow-up question regarding future charges.
> 
> For percentage-based future charges (withdrawal/redeem), I understand that the charge definition should be stored and, when the calculated amount rounds to zero, the transaction should continue and the charge should simply not be applied.
> 
> For flat future charges (for example a Shares Flat Redeem Charge of 0.5 in a currency with 0 decimal places), the rounded charge amount is known to be zero already at attachment time and can never become non-zero later.
> 
> Currently, the existing behavior appears to allow such a Flat Redeem Charge to be attached to the share account without any validation. During the attachment flow, the configured charge amount (0.5 in this example) is stored as-is and no currency rounding is performed. The amount is only passed through `Money.of()` later during the redeem-shares flow, where the calculated charge amount becomes 0 and is effectively not applied.
> 
> For this scenario, what should be the expected behavior:
> 
> 1. Keep the current behavior: allow the charge to be attached as-is and, during redemption, simply ignore it if the rounded amount becomes 0, similar to Percentage of Redeem Charges.
> 2. Round the flat redeem charge during attachment and reject it if the amount rounds to 0 by throwing a validation exception in the charge attachment flow itself.

I dont think we should reject and throw error if charge became 0... @bharathcgowda thoughts?

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/domain/LoanCharge.java
Dhanno98: I intended to ensure all resulting charge amounts respected the currency precision rules, including values coming from min/max caps. However, I agree that minimumAndMaximumCap() should remain focused on cap enforcement and not be responsible for rounding. I've reverted the changes and moved the rounding logic to the charge amount calculation flow instead.

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanChargeService.java
Dhanno98: Fixed. The silent return for zero-amount charges has been removed. A validation check is added in `LoanChargeService.addLoanCharge()`. If the calculated charge amount becomes zero after applying rounding rules, a `LoanChargeCannotBeAddedException` is now raised with the message `error.msg.loanCharge.cannot.be.added.as.amount.rounded.to.zero`. This prevents the charge from being created and returns an explicit validation error to the client instead of silently ignoring the request.

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanChargeService.java
Dhanno98: You're right. I originally reused minimumAndMaximumCap() because percentage-based charges already passed through that path and received currency-based rounding there. However, that couples rounding with cap enforcement. I've removed the call from the flat charge flow and now perform the rounding directly during flat charge amount calculation.

fineract-provider/src/main/java/org/apache/fineract/portfolio/accounts/api/AccountsApiResourceSwagger.java
Dhanno98: The reason for this change is that I needed to add integration test coverage for the Share Account charge rounding flow introduced in this PR. I intentionally used the generated Fineract client APIs instead of the deprecated RestAssured helpers, since the project is gradually moving toward client-based integration tests.

While implementing the Share Account tests, I found that `PostAccountsTypeAccountIdRequest` was incomplete in the generated client because the OpenAPI schema only exposed `requestedShares`. However, share account commands used by the tests (`activate`, `redeemshares`, etc.) require fields such as `locale`, `dateFormat`, `activatedDate`, and `requestedDate`. Without exposing those fields in the OpenAPI definition, the generated client model could not be used to implement these test flows.

So the schema update was made to enable the `ShareAccountChargeRoundingTest` that covers the rounding changes, rather than as an independent Swagger cleanup.
That said, if you'd prefer this to be split into a separate PR, I'm happy to do that.

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/domain/ClientCharge.java
Dhanno98: Done! I removed the `command` parameter, and now the due date is provided as a parameter.

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/service/ClientChargeWritePlatformServiceImpl.java
Dhanno98: Done! I have used of Money instead of BigDecimal here in the latest changes.

fineract-provider/src/main/java/org/apache/fineract/portfolio/client/service/ClientChargeWritePlatformServiceImpl.java
Dhanno98: Yes, it is incorrect, and I have made the change to throw now a `PlatformApiDataValidationException` when the final charge amount rounds to 0.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanChargeWritePlatformServiceImpl.java
Dhanno98: Reverted. These changes were not required for the final solution and have been removed.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanChargeWritePlatformServiceImpl.java
Dhanno98: Reverted.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanChargeWritePlatformServiceImpl.java
Dhanno98: Reverted. I was understanding things incorrectly here. So not needed.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanChargeWritePlatformServiceImpl.java
Dhanno98: Reverted.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanChargeWritePlatformServiceImpl.java
Dhanno98: Reverted.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanChargeWritePlatformServiceImpl.java
Dhanno98: Reverted. This was not needed.

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanChargeWritePlatformServiceImpl.java
Dhanno98: Reverted.

fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/service/SavingsAccountWritePlatformServiceJpaRepositoryImpl.java
Dhanno98: Good point. I updated the implementation to prevent zero-amount savings charges from being persisted when the final rounded amount is known at charge attachment time.

After the charge amount is rounded according to the savings product currency configuration (digitsAfterDecimal/inMultiplesOf), I validate the resulting amount before persistence. If the rounded amount becomes zero, a validation error is returned, and the charge is not added.

This validation is currently applied only to flat savings charges because percentage-based charges (for eg withdrawal fees) are calculated later during transaction processing, so the final charge amount is not available when the charge is attached.

Dhanno98:
> > > @adamsaghy @Dhanno98 In any entity and scenario if the calculated charge amount is zero (after rounding) that charge should not be created. In your latest example, "Allow the transaction but skip creating/applying the charge, similar to silent return," sounds right to me. Any transactions associated with charges should not be affected. Hence withdraw transaction happens with no charge posting in the above example
> > 
> > 
> > Hi @bharathcgowda @adamsaghy One follow-up question regarding future charges.
> > For percentage-based future charges (withdrawal/redeem), I understand that the charge definition should be stored and, when the calculated amount rounds to zero, the transaction should continue and the charge should simply not be applied.
> > For flat future charges (for example a Shares Flat Redeem Charge of 0.5 in a currency with 0 decimal places), the rounded charge amount is known to be zero already at attachment time and can never become non-zero later.
> > Currently, the existing behavior appears to allow such a Flat Redeem Charge to be attached to the share account without any validation. During the attachment flow, the configured charge amount (0.5 in this example) is stored as-is and no currency rounding is performed. The amount is only passed through `Money.of()` later during the redeem-shares flow, where the calculated charge amount becomes 0 and is effectively not applied.
> > For this scenario, what should be the expected behavior:
> > 
> > 1. Keep the current behavior: allow the charge to be attached as-is and, during redemption, simply ignore it if the rounded amount becomes 0, similar to Percentage of Redeem Charges.
> > 2. Round the flat redeem charge during attachment and reject it if the amount rounds to 0 by throwing a validation exception in the charge attachment flow itself.
> 
> I dont think we should reject and throw error if charge became 0... @bharathcgowda thoughts?

Thanks @adamsaghy.

The current implementation in this PR already follows that behavior for Share Redeem Charges and Savings Percentage Withdrawal Charges.

These charges continue to be attached and stored as configured. When they are later applied, the calculated charge amount goes through the existing `Money` rounding rules and, if the rounded amount becomes 0, the transaction proceeds normally and the charge is simply not applied.

So Share Flat Redeem Charges that eventually round to 0 during the Redeem Shares flow are currently not rejected during attachment.

I've also updated the PR with the latest changes and fixes.

fineract-provider/src/main/java/org/apache/fineract/portfolio/accounts/api/AccountsApiResourceSwagger.java
Dhanno98: You're correct to question those schema changes.

While implementing the integration tests for the charge-rounding feature, I found two limitations in the generated client models:

* `PostAccountsTypeAccountIdRequest` does not expose fields required by share account commands such as `activate` and `redeemshares` (`locale`, `dateFormat`, `activatedDate`, `requestedDate`, etc.).
* `PostClientsClientIdChargesRequest` exposes `amount` as `Integer`, whereas the charge-rounding scenarios require sending decimal values.

My initial approach was to update the OpenAPI schemas so that the generated client models would support these use cases, allowing the new integration tests to be implemented entirely through the generated Fineract client APIs instead of the deprecated RestAssured helpers.

However, after the API backward compatibility checks failed, it became clear that these schema changes have broader API implications and are larger in scope than what is needed for this rounding PR. Because of that, I reverted the schema modifications and reworked the tests to use the existing RestAssured-based helpers for the specific flows that depend on those incomplete models.

So the current version of this PR no longer contains the schema changes and keeps the API surface unchanged. The underlying limitations in the generated models still appear to exist, but addressing them would be better handled separately from this charge-rounding change.

adamsaghy: Its fine if that is failing. No worries. Proceed.

adamsaghy:
@Dhanno98 Please review the failing checks

Dhanno98:
> @Dhanno98 Please review the failing checks

Hi @adamsaghy ! I investigated the CI failures.

One of the failures (`LoanWorkbookPopulatorTest.testLoanWorkbookPopulate`) was reproducible locally only when `ClientChargeRoundingTest` was executed before it.

After tracing the failure, I found that `ClientChargeRoundingTest` created client charges using UUIDs in the charge name. Those UUIDs contained hyphens, which later became part of workbook named ranges created by `LoanWorkbookPopulator`, causing the subsequent workbook population test to fail.

I've updated the `createFlatClientCharge()` to remove the hyphens from the UUID while still keeping the names unique. After this change I verified that:
* `ClientChargeRoundingTest` passes.
* `LoanWorkbookPopulatorTest` passes even when run immediately after `ClientChargeRoundingTest`.

I've amended the PR with this fix.

The other CI failure (`MariaDB Tests` -> `Build the image`) also seems unrelated to this change. From the logs, it failed with:
`java.net.UnknownHostException: registry-1.docker.io: Temporary failure in name resolution`
Since the PostgreSQL and MySQL jobs passed, I think this might be a temporary CI/network issue rather than something caused by this PR.

adamsaghy:
```
  org.apache.fineract.integrationtests.bulkimport.populator.loan.LoanWorkbookPopulatorTest
  
    Test testLoanWorkbookPopulate() FAILED
  
    java.lang.AssertionError: 1 expectation failed.
    Expected status code <200> but was <500>.
        at org.apache.fineract.integrationtests.bulkimport.populator.loan.LoanWorkbookPopulatorTest.testLoanWorkbookPopulate(LoanWorkbookPopulatorTest.java:101)

**Comment by Dhanno98** [comment]
> ```
>   org.apache.fineract.integrationtests.bulkimport.populator.loan.LoanWorkbookPopulatorTest
>   
>     Test testLoanWorkbookPopulate() FAILED
>   
>     java.lang.AssertionError: 1 expectation failed.
>     Expected status code <200> but was <500>.
>         at org.apache.fineract.integrationtests.bulkimport.populator.loan.LoanWorkbookPopulatorTest.testLoanWorkbookPopulate(LoanWorkbookPopulatorTest.java:101)
>   
>   
> ```

I investigated the CI failure further.

Initially, I fixed the issue in `ClientChargeRoundingTest` because `LoanWorkbookPopulatorTest` was running immediately after it in the same shard, and that resolved the failure locally.

However, on the next CI run, `LoanWorkbookPopulatorTest` was scheduled in a different shard and this time it ran after `SharesAccountChargeRoundingTest`. The same failure occurred again with the same validation error about the generated charge name.

After reproducing this locally by running:

* `SharesAccountChargeRoundingTest` → `LoanWorkbookPopulatorTest`
* `SavingsAccountChargeRoundingTest` → `LoanWorkbookPopulatorTest`
* `LoanChargeRoundingTest` → `LoanWorkbookPopulatorTest`

I found that all three reproduced the same problem.

The root cause is that these tests create charge names using `UUID.randomUUID()`, which includes hyphens. Those charge names remain in the database, and later `LoanWorkbookPopulatorTest` creates workbook named ranges from them. Excel named ranges don't allow hyphens, so the workbook generation fails.

I updated all of the newly added charge rounding tests to generate UUIDs without hyphens (`UUID.randomUUID().toString().replace("-", "")`). After that, I reran all of the above combinations locally, and `LoanWorkbookPopulatorTest` passed in every case.

Since `LoanWorkbookPopulatorTest` can be scheduled after any of these tests depending on shard assignment, fixing all of them was necessary rather than only `ClientChargeRoundingTest`.

File: fineract-provider/src/main/java/org/apache/fineract/portfolio/accounts/api/AccountsApiResourceSwagger.java (hunk 4 above)
Comment (Dhanno98) [reply fineract-provider/src/main/java/org/apache/fineract/portfolio/accounts/api/AccountsApiResourceSwagger.java]: Thanks! One question for future reference.

Initially I had updated the OpenAPI/Swagger definitions so that the generated client models (`PostAccountsTypeAccountIdRequest` and `PostClientsClientIdChargesRequest`) would support these scenarios directly, but I reverted those changes because the API backward compatibility check failed.

From your comment, am I correct in understanding that such Swagger or OpenAPI changes are acceptable even if they cause the backward compatibility check to fail, provided the API change is intentional?

If so, would you prefer that kind of generated-client improvement be included in the same PR when needed, or handled separately in its own PR?

File: fineract-provider/src/main/java/org/apache/fineract/portfolio/shareaccounts/domain/ShareAccountTransaction.java [hunk 16]
```diff
@@ -114,7 +114,7 @@ public static ShareAccountTransaction createChargeTransaction(final LocalDate tr
         final BigDecimal unitPrice = null;
         final Integer status = PurchasedSharesStatusType.APPROVED.getValue();
         final Integer type = PurchasedSharesStatusType.CHARGE_PAYMENT.getValue();
-        BigDecimal amount = charge.percentageOrAmount();
+        BigDecimal amount = charge.amoutOutstanding();
```
adamsaghy, line 117: any reason this to be changed?

Dhanno98: Yes. The reason is that `percentageOrAmount()` returns the original configured charge value, before currency rounding is applied.

For example, if the configured charge amount is `19.8` for a currency with `0` decimal places, then:

* `amountOrPercentage = 19.8` (original configured value)
* `amount = 20` (rounded amount)
* `amountOutstanding = 20` (rounded outstanding amount)

`createChargeTransaction()` represents the actual charge transaction, so it should use the rounded charge amount that is actually applied to the account, not the original unrounded configured value.

Using `percentageOrAmount()` was causing the transaction to be created with the unrounded value, whereas `amountOutstanding()` reflects the rounded amount that was charged and therefore keeps the transaction consistent with the charge itself.

adamsaghy: percentage should not be changed... only final, calculated charge amount.

Dhanno98: I just want to make sure I understand your comment correctly.

In the activation-charge flow, `deriveChargeAmount()` has already calculated the rounded charge amount (`amount`, which is also reflected in `amountOutstanding` at this stage), but `createChargeTransaction()` currently uses `percentageOrAmount()`, which is the original configured value (e.g. `19.8`) rather than the calculated rounded value.

My understanding was that the transaction should store the final calculated charge amount after rounding, which is why I changed it to use `amoutOutstanding()` instead of `percentageOrAmount()`.

When you said "_percentage should not be changed... only final, calculated charge amount_", were you referring to the value that should be stored in the transaction?
- Should `createChargeTransaction()` continue storing the original configured value from `percentageOrAmount()`?
- Or should it store the final calculated (rounded) charge amount instead? If the latter, which field would you expect it to use, since `ShareAccountCharge` doesn't currently expose the calculated `amount` field?

adamsaghy: Thank you for the clarification.

adamsaghy (changes requested):
Technically, I see one real blocker: the PR fixes `ShareAccountTransaction.createChargeTransaction()` to use the rounded/applied amount, but activation charges still create `ShareAccountChargePaidBy` with `charge.percentageOrAmount()` in `ShareAccountDataSerializer`.

That means for a flat activation charge like 19.8 rounded to 20:
`share transaction amount = 20`
`chargesPaid.amount = 19.8`

That is dangerous because share accounting consumes `chargesPaid.amount` and checks that the charge-payment breakdown totals match the transaction amount. With accounting enabled, this can trip the “charges for shares have broken accounting code” integrity exception.

To be fixed: in the activation charge path, pass the derived amount into `ShareAccountChargePaidBy`, not `charge.percentageOrAmount()`.

Please also add a test covering activation share charges with accounting enabled, because the current rounding tests appear to use non-accounting share products and won’t catch this.

Dhanno98:
> Technically, I see one real blocker: the PR fixes `ShareAccountTransaction.createChargeTransaction()` to use the rounded/applied amount, but activation charges still create `ShareAccountChargePaidBy` with `charge.percentageOrAmount()` in `ShareAccountDataSerializer`.
> 
> That means for a flat activation charge like 19.8 rounded to 20: `share transaction amount = 20` `chargesPaid.amount = 19.8`
> 
> That is dangerous because share accounting consumes `chargesPaid.amount` and checks that the charge-payment breakdown totals match the transaction amount. With accounting enabled, this can trip the “charges for shares have broken accounting code” integrity exception.
> 
> To be fixed: in the activation charge path, pass the derived amount into `ShareAccountChargePaidBy`, not `charge.percentageOrAmount()`.
> 
> Please also add a test covering activation share charges with accounting enabled, because the current rounding tests appear to use non-accounting share products and won’t catch this.

Thanks for catching that! I'll update the activation charge path to use the derived amount and add an accounting-enabled test.

Dhanno98:
Hi @adamsaghy 

While implementing the integration test for the accounting-rule-2 share product flow, I noticed that the 
POST `/products/share` endpoint accepts the accounting mapping fields (`shareReferenceId`, `shareSuspenseId`, `incomeFromFeeAccountId`, `shareEquityId`) when called directly (via Postman), but these fields are missing from `ProductsApiResourceSwagger.PostProductsTypeRequest`.

Because of that, the generated `PostProductsTypeRequest` model also doesn't expose these fields, making it impossible to create an accounting-rule-2 share product through the generated Feign client used by the integration tests.

My understanding is that the Swagger request schema is incomplete rather than the backend implementation. My plan is to add these optional request fields to the OpenAPI schema and regenerate the client. Before doing that, I wanted to check whether this is the preferred approach, since it may affect the OpenAPI backward compatibility checks.

adamsaghy:
> Hi @adamsaghy
> 
> While implementing the integration test for the accounting-rule-2 share product flow, I noticed that the POST `/products/share` endpoint accepts the accounting mapping fields (`shareReferenceId`, `shareSuspenseId`, `incomeFromFeeAccountId`, `shareEquityId`) when called directly (via Postman), but these fields are missing from `ProductsApiResourceSwagger.PostProductsTypeRequest`.
> 
> Because of that, the generated `PostProductsTypeRequest` model also doesn't expose these fields, making it impossible to create an accounting-rule-2 share product through the generated Feign client used by the integration tests.
> 
> My understanding is that the Swagger request schema is incomplete rather than the backend implementation. My plan is to add these optional request fields to the OpenAPI schema and regenerate the client. Before doing that, I wanted to check whether this is the preferred approach, since it may affect the OpenAPI backward compatibility checks.

Yes, please. Your understanding is correct. Swagger might be incomplete or incorrect in certain cases. However, the relevant information is always the incoming requests (fields), DTOs, and response DTOs.

adamsaghy:
@Dhanno98 Please rebase this PR

Dhanno98:
Hi @adamsaghy 

Sorry about the confusion earlier. While cleaning up the PR I accidentally pushed a couple of unrelated files, so I first fixed that and then noticed your comment about rebasing. The PR has now been rebased onto the latest `develop`, and the current branch contains only the intended changes.

In addition to the original rounding fix, I made a few related improvements:

**1. Fix for the rounded activation charge issue**

The actual fix is in `ShareAccountDataSerializer.createChargeTransaction()`. Previously, `ShareAccountChargePaidBy` was created using the original charge amount:

```java
new ShareAccountChargePaidBy(chargeTransaction, charge, charge.percentageOrAmount())
```

It now uses the rounded transaction amount:

```java
new ShareAccountChargePaidBy(chargeTransaction, charge, amount)
```

This ensures that the persisted `ShareAccountChargePaidBy.amount` matches the rounded amount used by the transaction.

I also added an integration test, `verifyActivationChargeRoundedCorrectlyWithAccountingEnabled()`, in `ShareAccountChargeRoundingTest` which verifies:

* activation charge rounding,
* persisted charge amount,
* accounting journal entries generated for the rounded charge.

To support this test using the generated client, I added the missing accounting account fields to `PostProductsTypeRequest` and added `createEquityAccount()` to `FeignAccountHelper`.

**2. Generated client improvements**

Since using the generated client for these commands is acceptable, I also exposed the missing request fields in:

* `PostAccountsTypeAccountIdRequest`
* `PostClientsClientIdChargesRequest`

After regenerating the client, I made use of these generated client requests for:
* share account activation,
* share redemption,
* client charge application.

This makes the integration tests consistent with Fineract client usage instead of RestAssured.

Dhanno98:
I went through the remaining failing CI job (`run-api-backward-compatibility`) to understand the failure.

From what I observed, the reported breaking changes are due to the generated client schema updates that I made for:

* `PostAccountsTypeAccountIdRequest` (to expose the missing request fields required for share account activation/redemption), and
* `PostClientsClientIdChargesRequest` (correcting `amount` to `BigDecimal`).

I also noticed that the additional fields added to `PostProductsTypeRequest` for accounting-enabled share product creation are additive and don't seem to be reported by the backward compatibility check.

It looks like the failure is specifically because the OpenAPI contract changed for the existing request models. Could you please let me know how you would like me to proceed?

fineract-provider/src/main/java/org/apache/fineract/portfolio/accounts/api/AccountsApiResourceSwagger.java
```diff
@@ -465,15 +465,26 @@ public static final class PostAccountsTypeAccountIdRequest {
...
+        @Schema(example = "dd MMMM yyyy")
+        public String dateFormat;
+        @Schema(example = "01 January 2026")
+        public String activatedDate;
+        @Schema(example = "05 May 2026")
+        public String requestedDate;
+        @Schema(example = "01 January 2026")
+        public String closedDate;
+        @Schema(example = "100", description = "Number of shares to redeem or a collection of transaction IDs depending on the command.")
+        public Object requestedShares;
```
adamsaghy: why to remove `Set<PostAccountsRequestedShares> requestedShares;` and use way too generic Object?

Dhanno98: I changed it to `Object` because this endpoint (`POST /v1/accounts/{type}/{accountId}`) accepts two different payload shapes for `requestedShares` depending on the command.

For `redeemshares` and `applyadditionalshares`, `requestedShares` is a numeric value.
For `approveadditionalshares` and `rejectadditionalshares`, `requestedShares` is a collection of transaction IDs (`Set<PostAccountsRequestedShares>`).

Since the same request model is shared across all these commands, `Set<PostAccountsRequestedShares>` only describes the approval/rejection case and isn't accurate for the other commands. I used `Object` to represent both payload shapes. If there's a preferred way to model this more precisely, I'm happy to update it.

adamsaghy: I am not sure how to handle it for now, but it has not much to do with the PR scope, i would undo it and review the PR to contains only changes that is relevant to "enforce charge rounding and money rounding rules"

Dhanno98: Thanks for the suggestion! Just to clarify, would you like me to revert only the `PostAccountsTypeAccountIdRequest` Swagger changes where there is ambiguity regarding using `Object` for `requestedShares`, or would you prefer all of the Swagger schema updates in this PR to be reverted?

adamsaghy: Ultimately, this PR should contain only relevant changes for the topic: Charge rounding.

Anything that is not relevant, we should undo and use a different PR for them.

Dhanno98:
I have reverted the unrelated Swagger changes so this PR now contains only the changes relevant to charge rounding. I also rebased the branch on the latest develop and the CI checks are now passing.
Please review when you have a moment.

adamsaghy:
> I have reverted the unrelated Swagger changes so this PR now contains only the changes relevant to charge rounding. I also rebased the branch on the latest develop and the CI checks are now passing. Please review when you have a moment.

I am postponing to merge this till release of 1.15.0 is not finished

adamsaghy (approved):
LGTM

PR #5964, opened 2026-06-09
FINERACT-405: Standardize mobile number validation across Client and Staff

anubhavio:
## Description

This PR replaces the hardcoded mobile number validation regex with a configurable property-based validation approach.

## Changes
-Added phone validation configuration under FineractProperties
-Introduced PhoneNumberValidationService to provide a shared validation rule
-Added custom @ValidPhoneNumber constraint and PhoneNumberValidator
-Updated StaffCreateRequest and StaffUpdateRequest to use the new validator
-Updated ClientDataValidator to use the same configurable validation rule
-Removed the hardcoded phone number regex constant
-Added/updated tests for client mobile number validation

### DIFF
--- a/fineract-core/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java
+++ b/fineract-core/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java
@@ -80,6 +80,8 @@ public class FineractProperties {
 
     private FineractNotificationProperties notification;
 
+    private FineractPhoneProperties phone;
+
     private FineractLoanProperties loan;
 
     private FineractSamplingProperties sampling;
@@ -493,6 +495,13 @@ public class FineractProperties {
     public static class UserNotificationSystemProperties {
 
         private boolean enabled;
+    }
+
+    @Getter
+    @Setter
+    public static class FineractPhoneProperties {
+
+        private String regex = "^\\+?[0-9]{7,15}$";
     }
 
     @Getter
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/validation/PhoneNumberValidationService.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/validation/PhoneNumberValidationService.java
@@ -1,0 +1,39 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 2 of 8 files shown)


fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/validation/PhoneNumberConstants.java
```diff
@@ -0,0 +1,27 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.infrastructure.core.validation;
+
+public class PhoneNumberConstants {
+
+    public static final String MOBILE_NUMBER_REGEX =
```
IOhacker: This should be configurable in a property, not fixed.

IOhacker (changes requested):

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/validation/PhoneNumberConstants.java
anubhavio: @IOhacker Thanks for the feedback. I investigated the existing validation and configuration patterns in Fineract. To make the regex configurable, I plan to move it into FineractProperties and replace the Staff @Pattern validation with a custom validator backed by a shared PhoneNumberValidationService, so both Client and Staff validations use the same configurable rule.

IOhacker: @anubhavio that is the best approach to use the same configurable rule.

anubhavio:
@adamsaghy I've addressed the review feedback and updated the implementation to use configurable mobile number validation backed by shared validation components.

Requesting another review when convenient. Thanks!

adamsaghy:
> @adamsaghy I've addressed the review feedback and updated the implementation to use configurable mobile number validation backed by shared validation components.
> 
> Requesting another review when convenient. Thanks!

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':fineract-provider:compileJava'.
> Compilation failed; see the compiler output below.

anubhavio:
> > @adamsaghy I've addressed the review feedback and updated the implementation to use configurable mobile number validation backed by shared validation components.
> > Requesting another review when convenient. Thanks!
> 
> FAILURE: Build failed with an exception.
> 
> * What went wrong:
>   Execution failed for task ':fineract-provider:compileJava'.
> 
> > Compilation failed; see the compiler output below.

@adamsaghy  Thanks for pointing that out. I fixed the compilation issues and updated the branch. The build should be green now.

adamsaghy:
@anubhavio Please review the failing tests.

PR #6059, opened 2026-06-29
FINERACT-2665: Fix NPE modifying a loan application when schedule fields are omitted

oluexpert99:
- On modify, LoanScheduleAssembler.assembleLoanApplicationTermsFrom rebuilds the loan terms from the request
  JSON. A modify request may legitimately send only the fields being changed (LoanApplicationValidator
  .validateForModify already treats the rest as optional), but the recalculation read each schedule-driving
  field straight from the request: an omitted override-enabled enum hit XxxMethod.fromInt(null) and an omitted
  expectedDisbursementDate hit deriveFirstRepaymentDate(null), both throwing a NullPointerException (HTTP 500)
  once the change triggered a schedule recalculation. This completes FINERACT-2389, which applied null-override
  handling only to the grace/tolerance fields.
- Before recalculating, backfill any omitted schedule-driving field from the loan's own values
  (LoanAssemblerImpl.updateFrom -> backfillOmittedScheduleTerms) so the regenerated schedule matches what the
  loan records rather than a product default or null: interestType, amortizationType,
  interestCalculationPeriodType, repaymentEvery, numberOfRepayments, repaymentFrequencyType, loanTermFrequency,
  loanTermFrequencyType, interestRateFrequencyType, graceOnPrincipalPayment, graceOnInterestPayment,
  graceOnArrearsAgeing, principal, interestRatePerPeriod, expectedDisbursementDate, submittedOnDate and
  interestChargedFromDate. This mirrors validateForModify, so validation and recalculation stay consistent.
- Date fields are re-serialised with the request's own dateFormat/locale, exactly as loan creation requires
  them; if a date must be backfilled and either parameter is missing or malformed, fail with a clear
  validation error rather than defaulting.
- Add unit tests for the backfill helper, the date-formatter required-parameter handling, and the
  schedule-term backfill from a loan.

## Description

## Checklist

### DIFF
--- a/fineract-e2e-tests-core/src/test/java/org/apache/fineract/test/stepdef/loan/LoanOverrideFieldsStepDef.java
+++ b/fineract-e2e-tests-core/src/test/java/org/apache/fineract/test/stepdef/loan/LoanOverrideFieldsStepDef.java
@@ -27,6 +27,7 @@ import io.cucumber.java.en.Then;
 import io.cucumber.java.en.When;
 import java.io.IOException;
 import java.math.BigDecimal;
+import java.time.format.DateTimeFormatter;
 import java.util.Map;
 import lombok.RequiredArgsConstructor;
 import org.apache.fineract.client.feign.FineractFeignClient;
@@ -34,6 +35,8 @@ import org.apache.fineract.client.models.GetLoansLoanIdResponse;
 import org.apache.fineract.client.models.PostClientsResponse;
 import org.apache.fineract.client.models.PostLoansRequest;
 import org.apache.fineract.client.models.PostLoansResponse;
+import org.apache.fineract.client.models.PutLoansLoanIdRequest;
+import org.apache.fineract.client.models.PutLoansLoanIdResponse;
 import org.apache.fineract.test.data.loanproduct.DefaultLoanProduct;
 import org.apache.fineract.test.data.loanproduct.LoanProductResolver;
 import org.apache.fineract.test.factory.LoanRequestFactory;
@@ -42,6 +45,9 @@ import org.apache.fineract.test.support.TestContextKey;
 
 @RequiredArgsConstructor
 public class LoanOverrideFieldsStepDef extends AbstractStepDef {
+
+    private static final String DATE_FORMAT = "dd MMMM yyyy";
+    private static final DateTimeFormatter FORMATTER = DateTimeFormatter.ofPattern(DATE_FORMAT);
 
     private final FineractFeignClient fineractClient;
     private final LoanRequestFactory loanRequestFactory;
@@ -73,6 +79,11 @@ public class LoanOverrideFieldsStepDef extends AbstractStepDef {
             case "graceOnPrincipalPayment" -> loanDetails.getGraceOnPrincipalPayment();
             case "graceOnInterestPayment" -> loanDetails.getGraceOnInterestPayment();
             case "graceOnArrearsAgeing" -> loanDetails.getGraceOnArrearsAgeing();
+            case "interestType" -> loanDetails.getInterestType().getId().intValue();
+            case "amortizationType" -> loanDetails.getAmortizationType().getId().intValue();
+            case "interestCalculationPeriodType" -> loanDetails.getInterestCalculationPeriodType().getId().intValue();
+            case "repaymentEvery" -> loanDetails.getRepaymentEvery();
+            case "principal" -> loanDetails.getPrincipal().intValue();
             default -> throw new IllegalArgumentException("Unknown override field: " + fieldName);
         };
     }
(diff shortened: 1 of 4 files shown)


adamsaghy (changes requested):
I dont like this approach. No backfilling please!

If something is not provided, it should not be touched / updated!

oluexpert99:
> I dont like this approach. No backfilling please!
> 
> If something is not provided, it should not be touched / updated!

Thanks for the review. I agree with the principle — an omitted field must not be modified — and that's exactly what this preserves. "Backfill" is probably a misleading word for it; let me clarify.
  
  The field updates already honor that rule: updateLoanApplicationAttributes only mutates a field when its parameter is present in the request. Nothing omitted is written or reset.

  The NPE isn't in the field update — it's in the schedule recalculation. When a modify triggers a recalc, the schedule assembler rebuilds the loan terms from the request JSON. For a schedule-driving field the request omitted (interestType,
  amortizationType, interestCalculationPeriodType, repaymentEvery), it reads null → InterestMethod.fromInt(null) → HTTP 500.

  The assembler needs a complete set of inputs to recalc. This change supplies the omitted ones from the loan's own existing values (loan.getLoanProductRelatedDetail()) — not the product default, not a hardcoded default. So the regenerated schedule
  matches what the loan already records, and the omitted field keeps its existing value. Net effect: the omitted field is untouched — your point exactly.

  Concretely, on an overridden loan (interest method Declining; product default Flat): a modify omitting interestType returns 200, the header stays Declining, and the regenerated schedule equals an explicit Declining run — not the product's Flat. If
  we instead don't supply the value, the recalc either NPEs (the current bug) or falls back to the product default and silently rewrites the schedule to disagree with the loan header — a worse, silent bug.

  This also continues FINERACT-2389, which merged the same "omitted ⇒ keep existing value" idiom for the grace/tolerance fields; this closes the gap for the four schedule attributes it didn't cover.

  If your concern is the mechanism (adding to the parsed JSON), I'm glad to instead read the loan's existing LoanProductRelatedDetail directly at recalc time — same outcome, no touching of the request. Which would you prefer?

adamsaghy:
> > I dont like this approach. No backfilling please!
> > If something is not provided, it should not be touched / updated!
> 
> Thanks for the review. I agree with the principle — an omitted field must not be modified — and that's exactly what this preserves. "Backfill" is probably a misleading word for it; let me clarify.
> 
> The field updates already honor that rule: updateLoanApplicationAttributes only mutates a field when its parameter is present in the request. Nothing omitted is written or reset.
> 
> The NPE isn't in the field update — it's in the schedule recalculation. When a modify triggers a recalc, the schedule assembler rebuilds the loan terms from the request JSON. For a schedule-driving field the request omitted (interestType, amortizationType, interestCalculationPeriodType, repaymentEvery), it reads null → InterestMethod.fromInt(null) → HTTP 500.
> 
> The assembler needs a complete set of inputs to recalc. This change supplies the omitted ones from the loan's own existing values (loan.getLoanProductRelatedDetail()) — not the product default, not a hardcoded default. So the regenerated schedule matches what the loan already records, and the omitted field keeps its existing value. Net effect: the omitted field is untouched — your point exactly.
> 
> Concretely, on an overridden loan (interest method Declining; product default Flat): a modify omitting interestType returns 200, the header stays Declining, and the regenerated schedule equals an explicit Declining run — not the product's Flat. If we instead don't supply the value, the recalc either NPEs (the current bug) or falls back to the product default and silently rewrites the schedule to disagree with the loan header — a worse, silent bug.
> 
> This also continues [FINERACT-2389](https://issues.apache.org/jira/browse/FINERACT-2389), which merged the same "omitted ⇒ keep existing value" idiom for the grace/tolerance fields; this closes the gap for the four schedule attributes it didn't cover.
> 
> If your concern is the mechanism (adding to the parsed JSON), I'm glad to instead read the loan's existing LoanProductRelatedDetail directly at recalc time — same outcome, no touching of the request. Which would you prefer?

The principle to manipulate the json object which is the incoming request is wrong and should not be followed.

If something was not provided, just ignore it... we are doing it in many situations already:
`command.hasParameter` or `fromJsonHelper.parameterExists`

oluexpert99:
> The principle to manipulate the json object which is the incoming request is wrong and should not be followed.
> 
> If something was not provided, just ignore it... we are doing it in many situations already: `command.hasParameter` or `fromJsonHelper.parameterExists`

Hi @adamsaghy 
I revamped this to ensure no more request-JSON manipulation. On modify, the schedule now recalculates from the loan's own state (buildScheduleGeneratorDTO(loan, null) + regenerateRepaymentSchedule(loan, dto)), exactly like the approval/reschedule/apply-holidays flows. An omitted field is simply ignored and its existing value flows through; nothing is written back to the request. Also, I dropped the redundant post-recalc recalculateAllCharges (regenerate already does it).

adamsaghy:
@oluexpert99 Please check the failing test cases.

adamsaghy (changes requested):
I was wondering whether we could cover this changes with integration test / E2E test to ensure correct results?

adamsaghy (approved):
LGTM

PR #6072, opened 2026-07-01
FINERACT-2455: Working Capital Delinquency Disable and Enable

alberto-art3ch:
## Description

The Delinquency Disable feature allows authorized users to disable delinquency evaluation for a loan account under specific business scenarios (e.g., charge-off/write-off). Delinquency Disable shall be allowed only once until the current Delinquency disable is reversed.

[FINERACT-2455](https://issues.apache.org/jira/browse/FINERACT-2455)

## Checklist


fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/api/WorkingCapitalLoanDelinquencyActionApiResource.java
```diff
@@ -98,6 +101,79 @@ public CommandProcessingResult createDelinquencyAction(
         return createDelinquencyAction(resolveExternalId(loanExternalId), apiRequestBodyAsJson);
     }
 
+    @POST
+    @Path("{loanId}/delinquency-actions/disable")
+    @Consumes({ MediaType.APPLICATION_JSON })
+    @Produces({ MediaType.APPLICATION_JSON })
+    @Operation(summary = "Disable Delinquency", description = "Disables delinquency evaluation for a Working Capital loan as of the current business date. "
```
Aman-Mittal: Can use java textblock here instead of concatination

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/api/WorkingCapitalLoanDelinquencyActionApiResource.java
```diff
@@ -98,6 +101,79 @@ public CommandProcessingResult createDelinquencyAction(
...
+            + "Any active delinquency classification is lifted as of the disable date. "
+            + "Requires the DISABLE_WC_DELINQUENCY_ACTION permission (or ALL_FUNCTIONS).")
+    @ApiResponses({
```
Aman-Mittal: Use ApiResponse directly, No need of @ ApiResponses

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/api/WorkingCapitalLoanDelinquencyActionApiResource.java
```diff
@@ -98,6 +101,79 @@ public CommandProcessingResult createDelinquencyAction(
...
+
+    @POST
+    @Path("external-id/{loanExternalId}/delinquency-actions/disable")
+    @Consumes({ MediaType.APPLICATION_JSON })
+    @Produces({ MediaType.APPLICATION_JSON })
+    @Operation(operationId = "disableDelinquencyByExternalId", summary = "Disable Delinquency by external id", description = "Disables delinquency evaluation for a Working Capital loan identified by external id as of the current business date.")
+    @ApiResponses({
+            @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = WorkingCapitalLoanDelinquencyActionApiResourceSwagger.PostWorkingCapitalLoansDelinquencyActionResponse.class))),
+            @ApiResponse(responseCode = "400", description = "Bad Request"),
+            @ApiResponse(responseCode = "404", description = "Working Capital Loan not found") })
```
Aman-Mittal: "Bad Request"
"Working Capital Loan not found"
Magic strings

Aman-Mittal (commented):
Review comments

galovics (approved):
Good coverage (unit + integration), permissions actually seeded. One small thing.

`0050_wc_loan_delinquency_disable.xml`:
```xml
<changeSet author="fineract" id="wcl-0049-1">
...
<changeSet author="fineract" id="wcl-0049-2">
```

This is file `0050_...` but the ids say `wcl-0049-1`/`wcl-0049-2` - copy-pasted from `0049_wc_loan_breach_reset.xml` without renumbering. Doesn't break anything since Liquibase's key includes the filename, but confusing to read later. Should be `wcl-0050-1`/`wcl-0050-2`.

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/repository/WorkingCapitalLoanDelinquencyActionRepository.java
```diff
@@ -33,4 +36,11 @@ public interface WorkingCapitalLoanDelinquencyActionRepository extends JpaReposi
     Optional<WorkingCapitalLoanDelinquencyAction> findTopByWorkingCapitalLoanIdAndActionOrderByIdDesc(Long workingCapitalLoanId,
             DelinquencyAction action);
 
+    @Query("""
+            select case when count(action) > 0 then true else false end from WorkingCapitalLoanDelinquencyAction action
+            where action.action = org.apache.fineract.portfolio.delinquency.domain.DelinquencyAction.DISABLE
+            and action.workingCapitalLoan.id = :loanId and action.startDate <= :date and action.endDate is null
+            """)
+    boolean isDelinquencyDisabledAsOf(@Param("loanId") Long loanId, @Param("date") LocalDate date);
```
adamsaghy, line 43: This query seems incomplete.

Would the following be better?
```
select case when count(action) > 0 then true else false end from WorkingCapitalLoanDelinquencyAction action
            where action.action = org.apache.fineract.portfolio.workingcapitalloan.domain.DelinquencyAction.DISABLE
            and action.workingCapitalLoan.id = :loanId and action.startDate <= :date and (action.endDate IS NULL OR action.endDate >= :date)

adamsaghy (changes requested):
I see some discrepancies between this logic and breach disable / enable:

**1. Enable persistence model is different.**

Breach persists a separate ENABLE row (the validator news up an ENABLE-typed action, it gets saveAndFlush'd as saved) and separately closes the DISABLE row. The delinquency PR persists no ENABLE row: the validator returns the existing active DISABLE (findActiveDisable), and the write service just stamps its endDate and re-saves it. So DelinquencyAction.ENABLE is added to the enum but never stored; the entityId returned on enable is the DISABLE's id. Weaker audit trail than breach, and an inconsistency with the precedent.

**2. Enable end-date is off by one vs breach.** 

Breach closes the window at enableStart.minusDays(1): disable is in effect [start, enable-1], which is exactly why its query uses the inclusive endDate >= :date. The PR sets endDate = businessDate (the enable day itself) and queries endDate is null. So the two features disagree on whether the enable day is still "disabled," and that's coupled to the query difference from before: they're one design decision, made two different ways.

**3. Disable/enable side-effects are genuinely new, not inherited from breach.** 

On DISABLE the PR actively calls liftDelinquencyClassification(loan, startDate); breach has no DISABLE branch at all in its write service (it just saves the row and relies on the eval guard). On ENABLE the PR treats the disabled window like a pause: extendPeriodsForPause + evaluateExpiredPeriods + classifyDelinquency: whereas breach only calls reprocessBreachSchedule and does not extend/pause periods. So the "disabled window behaves like a pause" behavior (which the PR's own comments claim mirrors breach) is not actually what breach does.

**4. Bundled unrelated change (scope creep).** InternalWorkingCapitalLoanApiResource.activateLoan now seeds WorkingCapitalLoanBalance (principal + discount, zero overpayment) to avoid the loan flipping to OVERPAID on repayment. Nothing to do with disable/enable: it's a fix to the internal activation helper riding along in this PR.

**5. Likely docs build break, and diverges from the breach doc wiring.**

The PR adds include::chapters/working-capital-disable/index.adoc[leveloffset=+1] to index.adoc, but that file doesn't exist in the tree and isn't added by the diff. Every other working-capital chapter: including breach (working-capital-breach-management.adoc): lives under chapters/features/ and is pulled in through features/index.adoc, not as a new top-level include. So this is both inconsistent with the breach approach and a dangling include that should fail the asciidoctor build.For balance, the thing I'd earlier have expected to differ but doesn't: the "block pause/resume/reschedule while disabled" guard is mirrored faithfully (breach.is.disabled → delinquency.is.disabled).

alberto-art3ch:
> I see some discrepancies between this logic and breach disable / enable:
> 
> **1. Enable persistence model is different.**
> 
> Breach persists a separate ENABLE row (the validator news up an ENABLE-typed action, it gets saveAndFlush'd as saved) and separately closes the DISABLE row. The delinquency PR persists no ENABLE row: the validator returns the existing active DISABLE (findActiveDisable), and the write service just stamps its endDate and re-saves it. So DelinquencyAction.ENABLE is added to the enum but never stored; the entityId returned on enable is the DISABLE's id. Weaker audit trail than breach, and an inconsistency with the precedent.
> 
> **2. Enable end-date is off by one vs breach.**
> 
> Breach closes the window at enableStart.minusDays(1): disable is in effect [start, enable-1], which is exactly why its query uses the inclusive endDate >= :date. The PR sets endDate = businessDate (the enable day itself) and queries endDate is null. So the two features disagree on whether the enable day is still "disabled," and that's coupled to the query difference from before: they're one design decision, made two different ways.
> 
> **3. Disable/enable side-effects are genuinely new, not inherited from breach.**
> 
> On DISABLE the PR actively calls liftDelinquencyClassification(loan, startDate); breach has no DISABLE branch at all in its write service (it just saves the row and relies on the eval guard). On ENABLE the PR treats the disabled window like a pause: extendPeriodsForPause + evaluateExpiredPeriods + classifyDelinquency: whereas breach only calls reprocessBreachSchedule and does not extend/pause periods. So the "disabled window behaves like a pause" behavior (which the PR's own comments claim mirrors breach) is not actually what breach does.
> 
> **4. Bundled unrelated change (scope creep).** InternalWorkingCapitalLoanApiResource.activateLoan now seeds WorkingCapitalLoanBalance (principal + discount, zero overpayment) to avoid the loan flipping to OVERPAID on repayment. Nothing to do with disable/enable: it's a fix to the internal activation helper riding along in this PR.
> 
> **5. Likely docs build break, and diverges from the breach doc wiring.**
> 
> The PR adds include::chapters/working-capital-disable/index.adoc[leveloffset=+1] to index.adoc, but that file doesn't exist in the tree and isn't added by the diff. Every other working-capital chapter: including breach (working-capital-breach-management.adoc): lives under chapters/features/ and is pulled in through features/index.adoc, not as a new top-level include. So this is both inconsistent with the breach approach and a dangling include that should fail the asciidoctor build.For balance, the thing I'd earlier have expected to differ but doesn't: the "block pause/resume/reschedule while disabled" guard is mirrored faithfully (breach.is.disabled → delinquency.is.disabled).

Thanks for the review, @adamsaghy  — all valid. I've pushed a follow-up:

1 & 2. Enable persistence + end-date. Aligned with breach: enable now persists its own dedicated ENABLE row (so entityId is the ENABLE's id), and closes the disable at enableDate - 1. The "active disable" query now matches breach too (end_date IS NULL OR end_date >= :date).
3. Side-effects. The disable/enable effects are intentionally delinquency-specific (lift/reclassify vs. breach's schedule reprocess) — breach's schedule mechanics don't carry over. I fixed the misleading comments that claimed this "mirrors breach".
4. Docs. Moved the chapter under chapters/features/ and wired it through features/index.adoc; removed the dangling top-level include. asciidoctor builds clean now. (Root cause: the docs/ gitignore rule silently dropped the new .adoc/.puml files — force-added.)
5. Balance seed. This one is actually a prerequisite for this PR's own ITs (they repay activated loans via the internal helper, which flip to OVERPAID without the seed). Kept it here with a rationale comment, but happy to split it into its own PR if you'd prefer.

Anything else please let me know. Thanks!

adamsaghy (approved):
LGTM

PR #6084, opened 2026-07-03
FINERACT-2649: Migrate Tier 3 loan integration tests to Feign client

DeathGun44:
## Description
This PR transitions **47 test classes** (encompassing approximately **275 test methods**) to utilize the Feign client, entirely removing their dependency on REST-assured.

**Note on behavior:** The core test behavior and assertions remain completely unchanged. This is strictly a refactor of the underlying HTTP client layer.

### Targeted commits
I've split this work into **14 commits**, grouped logically by test , APA, down-payments, etc.): two commits extending the shared Feignhelpers, eleven migration batches, plus a dedicated review-feedback refinement commit.

This ensures that if CI fails or issues arise later, it is much easier to isolate and track down the specific batch that introduced the regression.

### Scope

**Included in this PR:**
* Comprehensive loan rescheduling coverage (core reschedule requestsschedule with advance payments, center/group rescheduling) plusre-aging and re-amortization tests.
* Chargeback integration scenarios (non-APA workflows, down-payment nquency-and-chargeback interaction).
* Down-payment workflows (product configuration, transaction typing, undo disbursal / undo repayment, disbursement-to-savings with auto down-payment, overlapping
down-payment installment allocation, repayment schedules).
* Advanced Payment Allocation (APA) coverage: charge-off, charge reverse-replay, write-offs, transaction reprocessing, refunds for active loans, charge payment, and
installment-fee error handling.
* Client loan lifecycle tests (credit balance refund, external-id support, backdated disbursement, overpaid date status, modify approved amount, floating rate interest
recalculation, repayment calculation, and date validations).
* Core loan account, loan transaction, and loan summary tests, plus charge-off accounting and repayment-reversal external-id tests.
* Standardization of `BigDecimal` precision assertions across migration

### Review Feedback & OpenAPI Refinements
Based on maintainer review feedback, the production-code footprint of this PR has been reduced to **11 purely additive schema-documentation lines across 2 Swagger    files** - no behavioral changes:
* **`LoanTransactionsApiResourceSwagger`**: The earlier `@ApiResponstionTemplate` has been **dropped**. Instead, the fields the endpointalready returns for reAge/reAmortize templates (`numberOfPastInstallments`, `numberOfFutureInstallments`, `nextInstallmentDueDate`, `calculatedStartDate`) are now    documented on the existing response schema. This lets the standard glize these templates directly — the interim custom`InternalLoanReAgeApi` workaround interface has been removed.
* **`SavingsProductsApiResourceSwagger`**: Documented the existing `sponse field.
* **Strongly Typed Response Models**: Replaced untyped `HashMap<String, Object>` test builder responses with strongly-typed generated models (`PostLoansResponse`,    `PostLoansLoanIdResponse`).
* **Date Readability**: Standardized `LocalDate.of(...)` calls to use concise numeric month parameters, matching the pre-migration originals.                         
### Bug Fix: 
**pre-existing bug** on`develop` in two test methods:                                                                                                                                        - `testUndoDisbursalForLoanWithSingleDisbursalAutoDownPaymentEnabled
- `testUndoDisbursalForLoanWithMultiDisbursalAutoDownPaymentEnabledAndHasManualTransactions`                                                                          
Both tests make a manual repayment of **100.0** after disbursement, then undo the disbursal. The "repayment entries compensated" journal entry assertions incorrectly expected **250.0** (the down-payment amount) instead of **100.0** (tThe old REST-assured based `verifyJournalEntries()` had more lenientmatching semantics that masked this error. Fixed as part of the migration commit.

### Infrastructure Enhancements
This PR builds upon the shared Feign test infrastructure from earlie

* **`FeignGroupCenterHelper`** (new): Utilities for managing group a loan testing.
* **Extended shared builders & helpers**: reAmortize/reAge, write-off, and credit-allocation support in the loan helpers; external-id and business-date extensions; journal-entry, charge, code-value, external-event, and global-configing the migrated tests.
* **`FeignRawHttpHelper`** (extended): Safely handles edge cases that require raw HTTP requests to bypass strict Swagger model limitations (e.g., floating rate
properties), utilizing `fineract.it.url` securely.

## Checklist


DeathGun44:
the test is failing and the issue is in develop itself, because a reversal of a 100.0 repayment should produce 100.0 journal entries, not 250.0. i am adding a fix for it in the pr and describing it in the description aswell

fineract-provider/src/test/java/org/apache/fineract/batch/command/CommandStrategyUtilsTest.java
```diff
@@ -127,8 +127,8 @@ public void testBuildRequestFromQueryParametersTypeConversion(String fieldName,
         // then
         assertThat(result).isNotNull();
         Object actualValue = getFieldValue(result, fieldName);
-        if (expectedValue instanceof BigDecimal) {
-            assertThat((BigDecimal) actualValue).isEqualByComparingTo((BigDecimal) expectedValue);
+        if (expectedValue instanceof BigDecimal bigDecimal) {
```
adamsaghy: Would you mind to use a different field name? `bigDecimal` is the type, not the purpose.
`expectedValueAsBigDecimal`?

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/FeignGroupCenterHelper.java
```diff
@@ -0,0 +1,109 @@
...
+import org.apache.fineract.integrationtests.common.Utils;
+
+public final class FeignGroupCenterHelper {
+
+    private static final Gson GSON = new Gson();
+
+    private FeignGroupCenterHelper() {}
+
+    public static Long createGroup(int officeId) {
+        Map<String, Object> map = new HashMap<>();
```
Aman-Mittal, line 54: You can maintain constants for frequently used magic strings

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/modules/LoanRequestBuilders.java
```diff
@@ -116,6 +128,64 @@ public static PostLoansLoanIdRequest disburseLoan(Double disbursedAmount, String
                 .dateFormat(LoanTestData.DATETIME_PATTERN);
     }
 
+    public static String disburseLoanWithRepaymentRescheduleJson(String disbursedOnDate, String adjustRepaymentDate) {
+        Map<String, String> map = new LinkedHashMap<>();
+        map.put("locale", LoanTestData.LOCALE);
+        map.put("dateFormat", LoanTestData.DATETIME_PATTERN);
+        map.put("actualDisbursementDate", disbursedOnDate);
```
Aman-Mittal: same as above

integration-tests/src/test/java/org/apache/fineract/integrationtests/ClientLoanCreditBalanceRefundandRepaymentTypeIntegrationTest.java
```diff
@@ -520,296 +501,266 @@ public void paymentRefundWillCloseTheLoanCorrectly(LoanProductTestBuilder loanPr
...
+            final Float interestOutstanding = Utils.getDoubleValue(loanSummary.getInterestOutstanding()).floatValue();
+            final Float totalOutstanding = Utils.getDoubleValue(loanSummary.getTotalOutstanding()).floatValue();
+            final Float overpaidAmount = 159.00f;
+            final Float goodwillAmount = totalOutstanding + overpaidAmount;
+            final Float goodwillAmountInExpense = principalOutstanding + overpaidAmount;
+            final String goodwillDate = "09 January 2022";
+            makeRepaymentTypePayment(GOODWILL_CREDIT, goodwillDate, goodwillAmount.doubleValue(), disbursedLoanID);
+
+            checkJournalEntryForAssetAccount(getAccounts().getLoansReceivableAccount(), goodwillDate,
+                    journalEntry(principalOutstanding, getAccounts().getLoansReceivableAccount(), "CREDIT"));
```
Aman-Mittal: JournalEntry.TransactionType.CREDIT use existing enums

integration-tests/src/test/java/org/apache/fineract/integrationtests/DelinquencyAndChargebackIntegrationTest.java
```diff
@@ -24,300 +24,218 @@
...
 
-            // Get loan details expecting to have not a delinquency classification and 1,000 as Outstanding
-            GetLoansLoanIdResponse getLoansLoanIdResponse = loanTransactionHelper.getLoan(requestSpec, responseSpec, loanId);
+            GetLoansLoanIdResponse getLoansLoanIdResponse = getLoanDetails(loanId);
             validateLoanAccount(getLoansLoanIdResponse, "0.00", "1000.00", 0, doubleZERO);
 
-            // Move the Business date n days to apply the chargeback for the previous repayment
             businessDate = businessDate.plusDays(21);
-            BusinessDateHelper.updateBusinessDate(BusinessDateType.BUSINESS_DATE, businessDate);
+            updateBusinessDate("BUSINESS_DATE", formatBusinessDate(businessDate));
```
Aman-Mittal: same as above use existing

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanAccountChargeOffWithAdvancedPaymentAllocationTest.java
```diff
@@ -43,170 +37,91 @@
...
             List<JournalEntryTransactionItem> journalEntries = journalEntriesForChargeOff.getPageItems();
             assertEquals(4, journalEntries.size());
-            verifyJournalEntry(journalEntries.get(3), 1000.0, LocalDate.of(2022, 9, 10), loansReceivable, "CREDIT");
-            verifyJournalEntry(journalEntries.get(2), 10.0, LocalDate.of(2022, 9, 10), interestFeeReceivable, "CREDIT");
-            verifyJournalEntry(journalEntries.get(1), 1000.0, LocalDate.of(2022, 9, 10), creditLossBadDebt, "DEBIT");
-            verifyJournalEntry(journalEntries.get(0), 10.0, LocalDate.of(2022, 9, 10), feeChargeOff, "DEBIT");
+            verifyJournalEntry(journalEntries.get(3), 1000.0, LocalDate.of(2022, 9, 10), getAccounts().getLoansReceivableAccount(),
+                    "CREDIT");
+            verifyJournalEntry(journalEntries.get(2), 10.0, LocalDate.of(2022, 9, 10), getAccounts().getFeeReceivableAccount(), "CREDIT");
+            verifyJournalEntry(journalEntries.get(1), 1000.0, LocalDate.of(2022, 9, 10), getAccounts().getChargeOffExpenseAccount(),
```
Aman-Mittal, line 123: Use java.time.Month as sonar flags this as maintainability issue

Aman-Mittal (commented):
I am still reviewing this one.....

integration-tests/src/test/java/org/apache/fineract/integrationtests/SmsCampaignIntegrationTest.java
```diff
@@ -46,6 +46,7 @@
  */
 @ExtendWith(MockServerExtension.class)
 @MockServerSettings(ports = { 9191 })
+@SuppressWarnings("removal")
```
Aman-Mittal, line 49: why suppress warnings?

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/FeignLoanTestBase.java
```diff
@@ -419,6 +541,48 @@ protected void verifyJournalEntriesSequentially(Long loanId, LoanTestData.Journa
...
+        if (!(exception.getCause() instanceof FeignException feignException)) {
+            return exception.getUserMessageGlobalisationCode();
+        }
+        String topLevelCode = feignException.getUserMessageGlobalisationCode();
+        if (topLevelCode != null && !topLevelCode.equals("validation.msg.validation.errors.exist")
+                && !topLevelCode.equals("validation.msg.domain.rule.violation")) {
+            return topLevelCode;
+        }
+        try {
+            @SuppressWarnings("unchecked")
```
Aman-Mittal: Why suppressing warings here?

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/FeignLoanTestBase.java
```diff
@@ -557,16 +721,373 @@ protected void executeInlineCOB(Long loanId) {
...
+        return disburseLoanAsMap(date, loanId, transactionAmount, externalId);
+    }
+
+    protected HashMap<String, Object> disburseLoan(String date, Integer loanId, String transactionAmount) {
+        return disburseLoanAsMap(date, loanId, transactionAmount);
+    }
+
+    protected Long addChargesForLoan(Integer loanId, String chargeJson) {
+        Gson gson = new Gson();
+        @SuppressWarnings("unchecked")
```
Aman-Mittal: same as above

fineract-provider/src/test/java/org/apache/fineract/batch/command/CommandStrategyUtilsTest.java
DeathGun44: renamed to expectedValueAsBigDecimal

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/FeignGroupCenterHelper.java
DeathGun44: Done, extracted the repeated keys into constants

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/modules/LoanRequestBuilders.java
DeathGun44: Done!

integration-tests/src/test/java/org/apache/fineract/integrationtests/ClientLoanCreditBalanceRefundandRepaymentTypeIntegrationTest.java
DeathGun44: Done, switched to JournalEntry.TransactionType.CREDIT / DEBIT

integration-tests/src/test/java/org/apache/fineract/integrationtests/DelinquencyAndChargebackIntegrationTest.java
DeathGun44: Done!

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanAccountChargeOffWithAdvancedPaymentAllocationTest.java
DeathGun44: Done! replaced the numeric months with java.time.Month constants

integration-tests/src/test/java/org/apache/fineract/integrationtests/SmsCampaignIntegrationTest.java
DeathGun44: Fixed!

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/FeignLoanTestBase.java
DeathGun44: Removed!

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/FeignLoanTestBase.java
DeathGun44: Removed!

integration-tests/src/test/java/org/apache/fineract/integrationtests/ClientLoanMultipleDisbursementsIntegrationTest.java
```diff
@@ -195,46 +167,38 @@ public void checkThatAllMultiDisbursalsAppearOnLoanScheduleAndOutStandingBalance
         BigDecimal totalPrincipalDisbursed = BigDecimal.ZERO;
         // First 8 lines should be disbursals
         for (int i = 0; i < loanScheduleLineCount - 1; i++) {
-            final Integer period = (Integer) loanSchedule.get(i).get("period");
-            final BigDecimal principalDisbursed = BigDecimal
-                    .valueOf(Double.parseDouble(loanSchedule.get(i).get("principalDisbursed").toString()));
+            final Integer period = loanSchedule.get(i).getPeriod();
+            final BigDecimal principalDisbursed = loanSchedule.get(i).getPrincipalLoanBalanceOutstanding();
```
budaidev: that's not the same as before. Can you check maybe on the getSummary().getPrincipalDisbursed()?

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanAccountDisbursementToSavingsWithAutoDownPaymentTest.java
```diff
@@ -101,170 +115,90 @@ public void loanDisbursementToSavingsWithAutoDownPaymentAndStandingInstructionsT
...
+        for (var transaction : loanDetails.getTransactions()) {
+            GetLoansLoanIdLoanTransactionEnumData type = transaction.getType();
+            boolean isTransaction = switch (transactionOfType) {
+                case "disbursement" -> Boolean.TRUE.equals(type.getDisbursement());
+                case "downPayment" -> "loanTransactionType.downPayment".equals(type.getCode());
+                default -> false;
+            };
+            if (isTransaction && transactionDate.equals(transaction.getDate())) {
+                isTransactionFound = true;
+                assertEquals(transactionAmount, Utils.getDoubleValue(transaction.getAmount()), "Mismatch in transaction amounts");
```
budaidev: The method still claims to verify "TransactionIsAccountTransfer" but no longer checks the transfer sub-object, its transfer amount, or its date — it only re-checks the transaction amount it already matched on. The transfer-to-savings behavior this test exists to prove is no longer verified

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanTransactionChargebackTest.java
```diff
@@ -77,176 +72,144 @@
...
-                credit(fundSource, 1000.0), //
-                debit(loansReceivableAccount, 1000.0) //
+                credit(getAccounts().getFundSource(), 1000.0), //
+                debit(getAccounts().getLoansReceivableAccount(), 1000.0) //
         );
 
         // Try to reverse a Loan Transaction charge back
-        PostLoansLoanIdTransactionsResponse reverseTransactionResponse = loanTransactionHelper.reverseLoanTransaction(loanId,
-                chargebackTransactionId, operationDate, responseSpecErr403);
+        assertThrows(CallFailedRuntimeException.class, () -> reverseLoanTransaction(loanId, chargebackTransactionId, operationDate));
```
budaidev: I think it would make sense to check the 403 status as well, like what you did with the status 500 below.

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanRescheduleWithAdvancePaymentTest.java
```diff
@@ -294,73 +233,63 @@ private void createLoanProductWithInterestRecalculationForTestMultipleAdvancePay
...
-        final ArrayList periods = (ArrayList) repaymentSchedule.get("periods");
-        HashMap period = (HashMap) periods.get(3);
+        GetLoansLoanIdRepaymentPeriod period = getLoanDetails(this.loanId).getRepaymentSchedule().getPeriods().get(3);
         LOG.info("period  {}", period);
 
-        assertEquals(new ArrayList<>(Arrays.asList(2022, 1, 3)), period.get("dueDate"), "Checking for Due Date for 1st Month");
-        assertEquals(period.get("principalDue"), Float.parseFloat("1177.12"));
-        assertEquals(period.get("interestDue"), Float.parseFloat("152.88"));
+        assertEquals(LocalDate.of(2022, Month.JANUARY, 3), period.getDueDate(), "Checking for Due Date for 1st Month");
+        assertEquals(0, BigDecimal.valueOf(1177.13).compareTo(period.getPrincipalDue()));
```
budaidev, line 275: I know that the calculation with float is not precise, but it's interesting that it changed by 1 cent for both of these assertation. In a refactor it is unusual to change the expected values

budaidev (changes requested):

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/FeignLoanHelper.java
```diff
@@ -95,9 +124,61 @@ public Long createSimpleLoanProduct() {
...
+        } catch (com.fasterxml.jackson.core.JsonProcessingException e) {
+            throw new IllegalArgumentException("Invalid loan product json", e);
+        }
+    }
+
+    public CallFailedRuntimeException addLoanChargeExpectingError(Long loanId, PostLoansLoanIdChargesRequest request) {
+        return fail(() -> fineractClient.loanCharges().executeLoanCharge(loanId, request, (String) null));
+    }
+
+    public List<AdvancedPaymentData> getAdvancedPaymentAllocationRules(Long loanId) {
```
budaidev, line 171: Can't we use:
```
  public List<AdvancedPaymentData> getAdvancedPaymentAllocationRules(Long loanId) {
      return ok(() -> fineractClient.defaultApi().getAdvancedPaymentAllocationRulesOfLoan(loanId));
  }
```

budaidev (changes requested):

integration-tests/src/test/java/org/apache/fineract/integrationtests/ClientLoanMultipleDisbursementsIntegrationTest.java
DeathGun44: Good catch. but getSummary().getPrincipalDisbursed() idea didn't hold up when I ran it live (doesn't return 255 for the 8-tranche case), and the Feign model doesn't expose the per-row principalDisbursed. So I now read it from the raw JSON and sum across disbursement rows, matching the original.

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanAccountDisbursementToSavingsWithAutoDownPaymentTest.java
DeathGun44: Agreed, real regression. Restored the full check via raw JSON , it now asserts the transfer object and its transferAmount/transferDate.

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanTransactionChargebackTest.java
DeathGun44: Done!

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanRescheduleWithAdvancePaymentTest.java
DeathGun44: I checked against a live server the API genuinely returns 1177.13 / 152.87 here (reverting to 1177.12 fails the test), so the values seem correct. Both still sum to the 1330.00 EMI.

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/FeignLoanHelper.java
DeathGun44: Yes ,Done!

adamsaghy:
@DeathGun44 Please rebase.

DeathGun44:
@adamsaghy Done!

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/FeignLoanHelper.java
```diff
@@ -95,9 +122,61 @@ public Long createSimpleLoanProduct() {
...
+        return ok(() -> fineractClient.loanProducts().createLoanProduct(request));
+    }
+
+    /**
+     * WARNING: This method uses ObjectMapperFactory which silences unknown property errors. Do not use this method in
+     * tests expecting strict deserialization.
+     */
+    public Long createLoanProductFromJson(String loanProductJson) {
+        try {
+            String sanitizedJson = loanProductJson.replaceAll("(?<=\\d),(?=\\d{3})", "");
```
budaidev: Here, any numeric array/sequence where a value with ≥3 digits follows a comma gets silently merged. For example:  {"loanIds":[5,1000]} → {"loanIds":[51000]}. I don't think this is the indended behaviour here.

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/FeignLoanHelper.java
```diff
@@ -95,9 +122,61 @@ public Long createSimpleLoanProduct() {
...
+            Map<String, Object> body = ObjectMapperFactory.getShared().readValue(feignException.responseBodyAsString(), Map.class);
+            return (T) body.get(jsonAttributeToGetBack);
+        } catch (com.fasterxml.jackson.core.JsonProcessingException e) {
+            throw new IllegalStateException("Failed to parse error response for attribute " + jsonAttributeToGetBack, e);
+        }
+    }
+
+    public <T> T getLoanProductError(String loanProductJson, String jsonAttributeToGetBack) {
+        try {
+            String sanitizedJson = loanProductJson.replaceAll("(?<=\\d),(?=\\d{3})", "");
```
budaidev: Here, any numeric array/sequence where a value with ≥3 digits follows a comma gets silently merged. For example:  {"loanIds":[5,1000]} → {"loanIds":[51000]}. I don't think this is the indended behaviour here.

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/FeignLoanTestBase.java
```diff
@@ -114,12 +155,11 @@ public static void setupHelpers() {
         codeHelper = new FeignCodeHelper(client);
         globalConfigurationHelper = new FeignGlobalConfigurationHelper(client);
         schedulerHelper = new FeignSchedulerHelper(client);
+        externalEventHelper = new FeignExternalEventHelper(client);
+        accounts = new LoanTestAccounts(accountHelper);
```
budaidev: accounts = new LoanTestAccounts(accountHelper) is now eager in @BeforeAll (was lazy in getAccounts()). That's ~21 create-GL-account calls and 21 junk accounts per test class — for all 56 classes this PR moves onto the base, including ones that never touch accounting. Please restore lazy initialization.

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanChargeOffAccountingTest.java
```diff
@@ -68,39 +69,45 @@
...
-        this.clientHelper = new ClientHelper(this.requestSpec, this.responseSpec);
         this.inlineLoanCOBHelper = new InlineLoanCOBHelper(this.requestSpec, this.responseSpec);
+
+        this.assetAccount = getAccounts().getLoansReceivableAccount();
+        this.incomeAccount = getAccounts().getInterestIncomeAccount();
+        this.expenseAccount = getAccounts().getChargeOffExpenseAccount();
+        this.overpaymentAccount = getAccounts().getOverpaymentAccount();
     }
 
+    private ClientHelper clientHelper;
```
budaidev: unused here, no init no usage

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanUndoChargeOffReverseExternalIdTest.java
```diff
@@ -47,33 +48,36 @@
...
         this.clientHelper = new ClientHelper(this.requestSpec, this.responseSpec);
+        this.loanTransactionHelper = new LoanTransactionHelper(this.requestSpec, this.responseSpec);
         this.accountHelper = new AccountHelper(this.requestSpec, this.responseSpec);
         this.assetAccount = this.accountHelper.createAssetAccount();
         this.incomeAccount = this.accountHelper.createIncomeAccount();
         this.expenseAccount = this.accountHelper.createExpenseAccount();
         this.overpaymentAccount = this.accountHelper.createLiabilityAccount();
     }
 
+    private ClientHelper clientHelper;
```
budaidev: unused here, it shadows the parent classes field

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanUndoChargeOffReverseExternalIdTest.java
```diff
@@ -47,33 +48,36 @@
...
+    private LoanTransactionHelper loanTransactionHelper;
+
+    private AccountHelper accountHelper;
```
budaidev: unused here, it shadows the parent classes field (same as above)

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanWithWaiveInterestAndWriteOffIntegrationTest.java
```diff
@@ -48,11 +49,22 @@
  */
 @SuppressWarnings({ "rawtypes" })
 @ExtendWith(LoanTestLifecycleExtension.class)
-public class LoanWithWaiveInterestAndWriteOffIntegrationTest {
+public class LoanWithWaiveInterestAndWriteOffIntegrationTest extends FeignLoanTestBase {
```
budaidev: We are extending the base Feign test class which is a good addition. However we don't use anything from there.

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanTransactionSummaryTest.java
```diff
@@ -55,24 +57,25 @@
 
 @Slf4j
 @ExtendWith(LoanTestLifecycleExtension.class)
-public class LoanTransactionSummaryTest {
+public class LoanTransactionSummaryTest extends FeignLoanTestBase {
```
budaidev: We are extending the base Feign test class which is a good addition. However we don't use anything from there.

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanTransactionSummaryTest.java
```diff
@@ -55,24 +57,25 @@
...
         Utils.initializeRESTAssured();
         this.requestSpec = new RequestSpecBuilder().setContentType(ContentType.JSON).build();
         this.requestSpec.header("Authorization", "Basic " + Utils.loginIntoServerAndGetBase64EncodedAuthenticationKey());
         this.responseSpec = new ResponseSpecBuilder().expectStatusCode(200).build();
-        this.loanTransactionHelper = new LoanTransactionHelper(this.requestSpec, this.responseSpec);
         this.clientHelper = new ClientHelper(this.requestSpec, this.responseSpec);
+        this.loanTransactionHelper = new LoanTransactionHelper(this.requestSpec, this.responseSpec);
     }
 
+    private ClientHelper clientHelper;
```
budaidev, line 73: Looks like shadowing here as well

integration-tests/src/test/java/org/apache/fineract/integrationtests/BlockTransactionsOnClosedOverpaidLoansTest.java
```diff
@@ -43,7 +44,7 @@
 import org.junit.jupiter.api.Test;
 
 @Slf4j
-public class BlockTransactionsOnClosedOverpaidLoansTest {
+public class BlockTransactionsOnClosedOverpaidLoansTest extends FeignLoanTestBase {
```
budaidev: We are extending the base Feign test class which is a good addition. However we don't use anything from there.

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanTransactionAuditingIntegrationTest.java
```diff
@@ -33,46 +35,37 @@
...
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.extension.ExtendWith;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-@ExtendWith(LoanTestLifecycleExtension.class)
-public class LoanTransactionAuditingIntegrationTest {
+public class LoanTransactionAuditingIntegrationTest extends FeignLoanTestBase {
```
budaidev, line 55: Dropped assertions: verifyLoanIsPending/Approved/Active after each lifecycle step and verifyClientCreatedOnServer. The Feign base helpers don't assert status, so these four checks vanished. Re-add via getLoanDetails(loanId).getStatus().

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanAccountRepaymentCalculationTest.java
```diff
@@ -28,45 +28,56 @@
...
         this.requestSpec.header("Authorization", "Basic " + Utils.loginIntoServerAndGetBase64EncodedAuthenticationKey());
         this.responseSpec = new ResponseSpecBuilder().expectStatusCode(200).build();
+
         this.loanTransactionHelper = new LoanTransactionHelper(this.requestSpec, this.responseSpec);
-        this.clientHelper = new ClientHelper(this.requestSpec, this.responseSpec);
+        this.accountHelper = new AccountHelper(this.requestSpec, this.responseSpec);
+        this.journalEntryHelper = new JournalEntryHelper(this.requestSpec, this.responseSpec);
     }
 
+    private ClientHelper clientHelper;
```
budaidev: declared but never used

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanAccountBackdatedDisbursementTest.java
```diff
@@ -38,39 +39,49 @@
...
         this.requestSpec.header("Authorization", "Basic " + Utils.loginIntoServerAndGetBase64EncodedAuthenticationKey());
         this.responseSpec = new ResponseSpecBuilder().expectStatusCode(200).build();
+
         this.loanTransactionHelper = new LoanTransactionHelper(this.requestSpec, this.responseSpec);
-        this.clientHelper = new ClientHelper(this.requestSpec, this.responseSpec);
+        this.accountHelper = new AccountHelper(this.requestSpec, this.responseSpec);
+        this.journalEntryHelper = new JournalEntryHelper(this.requestSpec, this.responseSpec);
     }
 
+    private ClientHelper clientHelper;
```
budaidev: declared but never used

integration-tests/src/test/java/org/apache/fineract/integrationtests/ClientLoanChargeExternalIntegrationTest.java
```diff
@@ -18,131 +18,74 @@
...
-        map.put("locale", "en_GB");
-        map.put("dateFormat", "dd MMMM yyyy");
-        map.put("amount", amount);
-        map.put("dueDate", dueDate);
-        map.put("chargeId", chargeId);
-        map.put("externalId", externalId);
-        String json = new Gson().toJson(map);
-        LOG.info("{}", json);
-        return json;
+        assertThrows(CallFailedRuntimeException.class, () -> addLoanCharge(loanId, new PostLoansLoanIdChargesRequest().chargeId(chargeDefId)
```
budaidev: he duplicate-external-id case previously asserted HTTP 403; now a bare assertThrows(CallFailedRuntimeException.class, ...) passes on any error. Capture the exception and assert getStatus() == 403 (+ the globalisation code), as done elsewhere in this PR.

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanDownPaymentTransactionChargebackTest.java
```diff
@@ -64,20 +60,16 @@ public void loanDownPaymentTransactionChargebackTest() {
...
-                            .transactionAmount(250.0));
+            final PostLoansLoanIdTransactionsResponse downPaymentTransaction_1 = makeLoanDownPayment(loanId,
+                    new PostLoansLoanIdTransactionsRequest().dateFormat(LoanTestData.DATETIME_PATTERN).transactionDate("01 March 2023")
+                            .locale(LoanTestData.LOCALE).transactionAmount(250.0));
             assertNotNull(downPaymentTransaction_1);
 
-            // chargeback down payment transaction
-            final Long chargebackTransactionId = loanTransactionHelper.applyChargebackTransaction(loanId.intValue(),
-                    downPaymentTransaction_1.getResourceId(), "50.00", 0, responseSpec);
+            final Long chargebackTransactionId = applyChargebackTransaction(loanId, downPaymentTransaction_1.getResourceId(), 50.0, 1L);
```
budaidev: the original passed payment-type index 0 (resolved via GET /paymenttypes), this hardcodes id 1L; they coincide only if the first listed type has id 1. Use the idx-preserving overload applyChargebackTransaction(loanId, txId, "50.00", 0) already in FeignLoanTestBase.

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanChargebackOnPaymentTypeRepaymentTransactionsTest.java
```diff
@@ -22,315 +22,198 @@
...
-        // reverse Merchant Refund
-        loanTransactionHelper.reverseRepayment(loanId, merchantIssuedRefund_2.getResourceId().intValue(), "8 September 2022");
-
-        // apply Chargeback should give 503 error
-        final Long chargebackTransactionId = loanTransactionHelper.applyChargebackTransaction(loanId,
-                merchantIssuedRefund_2.getResourceId(), "50.00", 1, responseSpecErr503);
+        reverseRepayment(loanId, merchantIssuedRefund_2.getResourceId(), "8 September 2022");
 
+        CallFailedRuntimeException exception = assertThrows(CallFailedRuntimeException.class,
+                () -> applyChargebackTransaction(loanId, merchantIssuedRefund_2.getResourceId(), 50.0, 1L));
```
budaidev: Same idx/id conflation, but here the original used the second payment type (idx 1) and this hardcodes id 1L (effectively the first). The 503 assertion still passes, but the payload is no longer equivalent to the original scenario.

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanAccountDisbursementToSavingsWithAutoDownPaymentTest.java
```diff
@@ -22,77 +22,93 @@
...
-            // create Financial Activity Mapping for Liability Transfer
             mapLiabilityTransferFinancialActivity(loanProductId);
 
-            // Apply and Approve Loan
-            Long loanId = createLoanWithLinkedAccountAndStandingInstructions(clientId.intValue(), loanProductId, savingsAccountId,
-                    loanExternalIdStr);
+            String loanApplicationJSON = new org.apache.fineract.integrationtests.common.loans.LoanApplicationTestBuilder()
+                    .withPrincipal("1000").withLoanTermFrequency("45").withLoanTermFrequencyAsDays().withNumberOfRepayments("3")
+                    .withRepaymentEveryAfter("15").withRepaymentFrequencyTypeAsDays().withInterestRatePerPeriod("0")
+                    .withInterestTypeAsFlatBalance().withAmortizationTypeAsEqualPrincipalPayments()
```
budaidev: withInterestTypeAsDecliningBalance() → withInterestTypeAsFlatBalance() — assertion-neutral at 0% interest, but an unexplained semantic change, if it is possible keep the original configuration

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanReschedulingWithinCenterTest.java
```diff
@@ -19,382 +19,279 @@
...
-        globalConfigurationHelper.verifyAllDefaultGlobalConfigurations();
-    }
+    private static final Long LEGAL_FORM_PERSON = 1L;
 
-    @SuppressWarnings("rawtypes")
     @Test
     public void testCenterReschedulingLoansWithInterestRecalculationEnabled() {
-
-        Integer officeId = new OfficeHelper().createOffice(LocalDate.of(2007, 7, 1)).getResourceId().intValue();
+        Long officeId = new FeignOfficeHelper(fineractClient()).createOffice(LocalDate.of(2007, Month.JULY, 1)).getResourceId();
```
budaidev: Direct new FeignOfficeHelper(fineractClient()) in the test (helpers belong in the base per the test-infra conventions)

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanReschedulingWithinCenterTest.java
```diff
@@ -19,382 +19,279 @@
...
-        LOG.info("calendarId {}", calendarId);
-        return calendarId;
-    }
-
-    @SuppressWarnings("rawtypes")
     @Test
     public void testCenterReschedulingMultiTrancheLoansWithInterestRecalculationEnabled() {
-
-        Integer officeId = new OfficeHelper().createOffice(LocalDate.of(2007, 7, 1)).getResourceId().intValue();
+        Long officeId = new FeignOfficeHelper(fineractClient()).createOffice(LocalDate.of(2007, Month.JULY, 1)).getResourceId();
```
budaidev: Direct new FeignOfficeHelper(fineractClient()) in the test (helpers belong in the base per the test-infra conventions)

integration-tests/src/test/java/org/apache/fineract/integrationtests/loan/reamortization/LoanReAmortizationIntegrationTest.java
```diff
@@ -1120,20 +1115,15 @@ private Long createLoanProductWithMultiDisbursalAndRepaymentsWithEnableDownPayme
...
-                .retrieveLoanProductById(loanProductResponse.getResourceId());
-        assertNotNull(getLoanProductsProductIdResponse);
-        return loanProductResponse.getResourceId();
+        Long loanProductId = createLoanProduct(product);
+        GetLoanProductsProductIdResponse retrievedProduct = retrieveLoanProduct(loanProductId);
+        assertNotNull(retrievedProduct);
+        return loanProductId;
     }
 
     private Long addChargeWithCurrency(Long loanId, boolean isPenalty, double amount, String dueDate, String currencyCode) {
```
budaidev, line 1124: ignores isPenalty and always creates a fee. Branch to the penalty builder or drop the parameter

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanAccountChargeOffWithAdvancedPaymentAllocationTest.java
```diff
@@ -43,170 +38,94 @@
...
             final String feeCharge1AddedDate = DATE_FORMATTER.format(targetDate);
-            Integer feeLoanChargeId = loanTransactionHelper.addChargesForLoan(loanId,
-                    LoanTransactionHelper.getSpecifiedDueDateChargesForLoanAsJSON(String.valueOf(feeCharge), feeCharge1AddedDate, "10"));
+            addLoanCharge(loanId, feeCharge, feeCharge1AddedDate, 10.0);
 
             // apply penalty
-            Integer penalty = ChargesHelper.createCharges(requestSpec, responseSpec,
-                    ChargesHelper.getLoanSpecifiedDueDateJSON(ChargesHelper.CHARGE_CALCULATION_TYPE_FLAT, "10", true));
-
+            Long penalty = createLoanSpecifiedDueDateCharge(10.0);
```
budaidev: dead code

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanChargePaymentWithAdvancedPaymentAllocationTest.java
```diff
@@ -20,195 +20,113 @@
...
     }
 
-    @AfterAll
-    public static void tearDown() {
-        Integer deletedFinancialActivityAccountId = financialActivityAccountHelper
-                .deleteFinancialActivityAccount(financialActivityAccountId, responseSpec, CommonConstants.RESPONSE_RESOURCE_ID);
-        Assertions.assertNotNull(deletedFinancialActivityAccountId);
-        Assertions.assertEquals(financialActivityAccountId, deletedFinancialActivityAccountId);
+    @AfterEach
+    public void tearDownFinancialActivityAccounts() {
```
budaidev: AfterEach now deletes all financial-activity mappings after every test (original: one mapping once in AfterAll) and mapLiabilityTransferFinancialActivity(Long) ignores its parameter; new FinancialActivityAccountHelper(null) is fragile. Also the savings product lost withMinimumOpenningBalance("10000.0").

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/FeignLoanHelper.java
```diff
@@ -367,26 +447,296 @@ public PutLoansAvailableDisbursementAmountResponse modifyAvailableDisbursementAm
...
-        return response.getResourceId();
+    public PostCreateRescheduleLoansResponse createRescheduleRequest(PostCreateRescheduleLoansRequest request) {
+        if (request instanceof LoanRequestBuilders.RescheduleRequestWithRecalculateInterest recalcRequest
+                && Boolean.TRUE.equals(recalcRequest.getRecalculateInterest())) {
+            return new PostCreateRescheduleLoansResponse().resourceId(createRescheduleRequestFromJson(toRescheduleJson(request, true)));
+        }
+        return ok(() -> fineractClient.rescheduleLoans().createLoanRescheduleRequest(request));
+    }
+
+    public CallFailedRuntimeException createRescheduleRequestExpectingError(PostCreateRescheduleLoansRequest request) {
```
budaidev: createRescheduleRequestExpectingError is dead code with a latent bug: the recalc path throws AssertionError (RestAssured expectStatusCode(200)), which FeignCalls.fail() doesn't catch — it can never return the expected exception. Remove or implement via Feign

budaidev (changes requested):
The PR has a massive scope and a lot of changes, so the lot of comments doesn't mean it's bad, but I would advise to fix those to keep the refactor clean. The direction is really good.

Also please avoid any behavior changes in the test, by modifying or removing assertations, because it could lead silent errors.

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/FeignLoanHelper.java
DeathGun44: Good catch. Changed the lookahead to (?<=\d),(?=\d{3}(?!\d)) so it only strips a thousands-separator when the 3-digit group isn't followed by another digit, [5,1000] is left alone now, only 12,000.00-style separators are removed.

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/FeignLoanHelper.java
DeathGun44: Good catch. Changed the lookahead to (?<=\d),(?=\d{3}(?!\d)) so it only strips a thousands-separator when the 3-digit group isn't followed by another digit, [5,1000] is left alone now, only 12,000.00-style separators are removed.

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/FeignLoanTestBase.java
DeathGun44: Agreed, restored the lazy init inside getAccounts() and dropped the eager line from @BeforeAll, so classes that never touch accounting no longer create the ~21 GL accounts.

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanChargeOffAccountingTest.java
DeathGun44: Removed! it was dead (the createClient it fronts is a static call, so it resolved to the parent's Feign helper anyway). Also removed a couple of dead accountHelper fields I found in the same files.

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanUndoChargeOffReverseExternalIdTest.java
DeathGun44: Removed the field and its init.

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanUndoChargeOffReverseExternalIdTest.java
DeathGun44: This one is actually used for the REST-assured account setup (createAssetAccount() etc.), so I couldn't drop it, but you're right about the shadow, so I renamed it to restAccountHelper to remove the collision with the parent's Feign accountHelper.

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanWithWaiveInterestAndWriteOffIntegrationTest.java
DeathGun44: Reverted to a plain class with @ExtendWith(LoanTestLifecycleExtension.class).

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanTransactionSummaryTest.java
DeathGun44: Reverted the extends; kept @ExtendWith(LoanTestLifecycleExtension.class).

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanTransactionSummaryTest.java
DeathGun44: Resolved by the revert above , with no parent field it no longer shadows, and it's still needed locally, so I kept it.

integration-tests/src/test/java/org/apache/fineract/integrationtests/BlockTransactionsOnClosedOverpaidLoansTest.java
DeathGun44: Reverted the extends and added @ExtendWith(LoanTestLifecycleExtension.class).

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanTransactionAuditingIntegrationTest.java
DeathGun44: Re-added all four: a client-exists check after createClient, and verifyLoanStatus(loanId, LoanStatus.SUBMITTED_AND_PENDING_APPROVAL/APPROVED/ACTIVE) after each lifecycle step.

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanAccountRepaymentCalculationTest.java
DeathGun44: Removed (and a dead accountHelper field in the same file).

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanAccountBackdatedDisbursementTest.java
DeathGun44: Removed (and a dead accountHelper field in the same file).

integration-tests/src/test/java/org/apache/fineract/integrationtests/ClientLoanChargeExternalIntegrationTest.java
DeathGun44: Done both , assertEquals(403, exception.getStatus()) and assertErrorGlobalisationCode(exception, "error.msg.loan.charge.duplicate.externalId") (confirmed the exact code against the live server).

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanDownPaymentTransactionChargebackTest.java
DeathGun44: Fixed , now uses getPaymentTypeId(0) so it resolves the first payment type's real id instead of hardcoding 1L.

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanChargebackOnPaymentTypeRepaymentTransactionsTest.java
DeathGun44: Fixed! now getPaymentTypeId(1), matching the original scenario's second payment type.

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanAccountDisbursementToSavingsWithAutoDownPaymentTest.java
DeathGun44: Reverted to withInterestTypeAsDecliningBalance().

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanReschedulingWithinCenterTest.java
DeathGun44: Moved it out of both test bodies into a private static FeignOfficeHelper officeHelper initialised in @BeforeAll.

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanReschedulingWithinCenterTest.java
DeathGun44: Moved it out of both test bodies into a private static FeignOfficeHelper officeHelper initialised in @BeforeAll.

integration-tests/src/test/java/org/apache/fineract/integrationtests/loan/reamortization/LoanReAmortizationIntegrationTest.java
DeathGun44: Fixed! it now branches to loanSpecifiedDueDatePenalty(...) when isPenalty is true, ...Fee(...) otherwise.

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanAccountChargeOffWithAdvancedPaymentAllocationTest.java
DeathGun44: Removed

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanChargePaymentWithAdvancedPaymentAllocationTest.java
DeathGun44: Done! moved the cleanup back to a static @AfterAll; dropped the unused Long parameter from mapLiabilityTransferFinancialActivity(); and restored the 10000.0 minimum opening balance , that field was missing from the generated Feign model (a Swagger gap), so I added minRequiredOpeningBalance to SavingsProductsApiResourceSwagger and set it via the typed builder. (Left FinancialActivityAccountHelper(null) as-is  it's the established Feign-client pattern and works.)

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/FeignLoanHelper.java
DeathGun44: Removed the method entirely, it had no callers and the recalc path could never return the expected exception anyway.

DeathGun44:
@budaidev  Thank you for the thorough review! I enjoy all these reviews as they help me learn a lot. I know there is a lot of code to go through. I will keep trying my best to make it as clean as possible and easier to review, and I'll make sure the quality of the code doesn't deteriorate with these bulk prs. Thanks again, everyone, for keeping up!

budaidev (approved):
looks good, great progress

budaidev:
please rebase and handle the conflicts

DeathGun44:
@adamsaghy @budaidev Done!

fineract-provider/src/test/java/org/apache/fineract/portfolio/loanaccount/domain/LoanTest.java
```diff
@@ -203,7 +205,7 @@ public void testGetLastUserTransaction() {
     @Test
     public void testTransactionComparator() {
         final Loan loan = new Loan();
-        LocalDate today = LocalDate.of(2023, 10, 12);
+        LocalDate today = LocalDate.of(2023, Month.OCTOBER, 12);
```
adamsaghy: this does not really help readability...

fineract-provider/src/test/java/org/apache/fineract/portfolio/loanaccount/domain/LoanTest.java
```diff
@@ -273,7 +275,7 @@ private LoanCharge buildLoanCharge() {
         loanCharge.setAmountOutstanding(new BigDecimal(100));
         loanCharge.setChargeTime(ChargeTimeType.TRANCHE_DISBURSEMENT.getValue());
         loanCharge.setChargeCalculation(ChargeCalculationType.FLAT.getValue());
-        loanCharge.setDueDate(LocalDate.of(2022, 6, 27));
+        loanCharge.setDueDate(LocalDate.of(2022, Month.JUNE, 27));
```
adamsaghy: this does not really help readability...

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/FeignBusinessDateHelper.java
```diff
@@ -61,12 +62,19 @@ public void runAt(String date, Runnable action) {
     }
 
     public void runAt(String date, String dateFormat, Runnable action) {
+        boolean outerScope = RUN_AT_NESTING_DEPTH.get() == 0;
```
adamsaghy: What is this for? We should not have nested business date changes...

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/InternalLoanReAgeApi.java
```diff
@@ -0,0 +1,35 @@
...
+package org.apache.fineract.integrationtests.client.feign.helpers;
+
+import feign.Headers;
+import feign.Param;
+import feign.RequestLine;
+import org.apache.fineract.client.models.LoanTransactionData;
+
+/**
+ * Feign interface for the reAge transaction template endpoint. The OpenAPI-generated client deserializes this response
+ * as {@code GetLoansLoanIdTransactionsTemplateResponse}, but the server returns {@link LoanTransactionData}.
```
adamsaghy: The swagger definition should be fixed then instead please

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/FeignLoanTestBase.java
```diff
@@ -557,16 +723,372 @@ protected void executeInlineCOB(Long loanId) {
...
+        return result;
+    }
+
+    protected HashMap<String, Object> disburseLoan(String date, Integer loanId, String transactionAmount, String externalId) {
+        return disburseLoanAsMap(date, loanId, transactionAmount, externalId);
+    }
+
+    protected HashMap<String, Object> disburseLoan(String date, Integer loanId, String transactionAmount) {
+        return disburseLoanAsMap(date, loanId, transactionAmount);
+    }
```
adamsaghy: Build and returning hashmap seems incorrect. why dont we use directly the response object?

integration-tests/src/test/java/org/apache/fineract/integrationtests/loan/reaging/LoanReAgingIntegrationTest.java
```diff
@@ -275,7 +270,7 @@ public void test_LoanReAgeTransaction_Works() {
                     installment(316.67, 0, 0, 0, 316.67, false, "13 May 2023"), //
                     installment(316.66, 0, 0, 0, 316.66, false, "12 June 2023") //
             );
-            checkMaturityDates(loanId, LocalDate.of(2023, 6, 12), LocalDate.of(2023, 6, 12));
+            checkMaturityDates(loanId, LocalDate.of(2023, Month.JUNE, 12), LocalDate.of(2023, Month.JUNE, 12));
```
adamsaghy: It does not help the readability...

adamsaghy (changes requested):
Please review my concerns about:
- readability issues: Using Month.JUNE, etc looks weird... what was the problem with "6"?
- Creating HashMap objects from the response object: use the response object, dont map into HashMap...
- Minimalize the code logic which  build manually HashMap / Json object and use that as request (i am not saying it cannot be used, but we should ensure, we are doing only when there are no other possibilities).
- Nested `runAt`: We should not have nested runAt. Is there a particular reason to allow it?

fineract-provider/src/test/java/org/apache/fineract/portfolio/loanaccount/domain/LoanTest.java
DeathGun44: Fair point! reverted all Month.ENUM back to integer months across the entire PR.

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/FeignLoanTestBase.java
DeathGun44: Replaced all HashMap<String, Object> returns with PostLoansResponse / PostLoansLoanIdResponse. Removed disburseLoanAsMap entirely.

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/InternalLoanReAgeApi.java
DeathGun44: Fixed the @ApiResponse on retrieveTransactionTemplate to use LoanTransactionData instead of GetLoansLoanIdTransactionsTemplateResponse. Deleted the InternalLoanReAgeApi workaround and switched to the standard generated client.

fineract-provider/src/test/java/org/apache/fineract/portfolio/loanaccount/domain/LoanTest.java
DeathGun44: Fixed!

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/FeignBusinessDateHelper.java
DeathGun44: Agreed and removed the ThreadLocal nesting depth tracking entirely. No test actually nests runAt calls, so the complexity was unjustified.

DeathGun44:
Regarding the OpenAPI backward-compatibility check failure: this is expected due to updating the @ApiResponse schema on retrieveTransactionTemplate from GetLoansLoanIdTransactionsTemplateResponse to LoanTransactionData.

Inspecting LoanTransactionsApiResource.java shows the endpoint uses DefaultToApiJsonSerializer<LoanTransactionData> and has always returned LoanTransactionData on the wire at runtime. The previous OpenAPI annotation was incorrect. This change aligns the Swagger spec with actual runtime behavior, so there is no real-world breaking change for API consumers.

integration-tests/src/test/java/org/apache/fineract/integrationtests/loan/reaging/LoanReAgingIntegrationTest.java
DeathGun44: Fixed!

DeathGun44:
the other test failure is just flaky ,should pass on rerun.
I have also updated the pr description with regard to the current state

DeathGun44:
@adamsaghy please take a look when you have a moment!

adamsaghy:
I am postponing to merge this till release of 1.15.0 is not finished

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanDownPaymentTransactionChargebackTest.java
adamsaghy: `getPaymentTypeId(0)`? Sounds like a weird logic... why dont you load the required payment type by name if we are unsure about the id?

fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/api/LoanTransactionsApiResource.java
```diff
@@ -162,7 +162,7 @@ public String retrieveTransactionTemplate(@PathParam("loanId") @Parameter(descri
             + "\n" + "loans/1/transactions/template?command=charge-off" + "\n" + "loans/1/transactions/template?command=downPayment" + "\n"
             + "loans/1/transactions/template?command=interest-refund")
     @ApiResponses({
-            @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = LoanTransactionsApiResourceSwagger.GetLoansLoanIdTransactionsTemplateResponse.class))) })
```
adamsaghy, line 168: Any reason for these changes?

adamsaghy: At this point, if not necessarily i would not change this. (maximum if some fields are missing, add them to the swagger definition).

This story should not do any changes in PROD code, but only refactoring testing implementation.

DeathGun44: Understood! ,I've reverted the @ApiResponse change on retrieveTransactionTemplate back to how it was (GetLoansLoanIdTransactionsTemplateResponse).
Instead, following your suggestion, I just added the missing fields (reAge and reAmortize options) directly into the LoanTransactionsApiResourceSwagger definition so the Feign client can pick them up properly. No production controller annotations or logic are modified anymore.

DeathGun44:
The test is just flaky, a rerun should pass it.

adamsaghy:
> The test is just flaky, a rerun should pass it.

nope. they seem real failures

DeathGun44:
> > The test is just flaky, a rerun should pass it.
> 
> nope. they seem real failures

You're right my mistake, it wasn't flaky. When I moved these tests onto the new FeignLoanTestBase, I accidentally left off the ExternalEventsExtension that BaseLoanIntegrationTest uses. That extension is what resets the external-event configuration after each test. I fixed it now!

DeathGun44:
This one seems flaky ,should pass on rerun

DeathGun44:
updated the pr description as per the current state!

adamsaghy (approved):
LGTM

PR #6106, opened 2026-07-08
FINERACT-2646: Decouple accounting from charge via chargeId reference

mansi75:
## Description

This PR addresses FINERACT-2646 by decoupling the accounting module from the charge domain dependency.

The accounting code previously depended on charge domain objects directly. This change updates the accounting side to use the 'chargeId' reference instead, so accounting can identify and process charge-related mappings without requiring a direct dependency on the charge package/module.

This helps improve module boundaries and supports the ongoing Fineract modularization effort by reducing cross-feature coupling between accounting and charge. FINERACT-2646 is listed as a sub-task for decoupling the accounting package/module from third-party domains. :contentReference[oaicite:0]{index=0}

## Checklist

### DIFF
--- a/buildSrc/src/main/groovy/org.apache.fineract.dependencies.gradle
+++ b/buildSrc/src/main/groovy/org.apache.fineract.dependencies.gradle
@@ -37,6 +37,7 @@ dependencyManagement {
         mavenBom 'io.github.resilience4j:resilience4j-bom:2.4.0'
         mavenBom 'org.testcontainers:testcontainers-bom:1.21.4'
         mavenBom 'org.glassfish.jersey:jersey-bom:3.1.11'
+        mavenBom 'org.springframework.modulith:spring-modulith-bom:1.4.11'
     }
 
     dependencies {
--- a/fineract-accounting/build.gradle
+++ b/fineract-accounting/build.gradle
@@ -48,6 +48,11 @@ configurations {
 
 apply from: 'dependencies.gradle'
 
+test {
+    maxHeapSize = '2g'
+    jvmArgs += ['-XX:MaxMetaspaceSize=1g']
+}
+
 // Configuration for the modernizer plugin
 // https://github.com/andygoossens/gradle-modernizer-plugin
 modernizer {
--- a/fineract-accounting/dependencies.gradle
+++ b/fineract-accounting/dependencies.gradle
@@ -26,6 +26,7 @@ dependencies {
     //
     implementation(project(path: ':fineract-core'))
     testImplementation 'org.springframework.modulith:spring-modulith-core'
+    testImplementation("org.osgi:org.osgi.core:6.0.0")
     implementation('org.apache.avro:avro')
     implementation(
             project(path: ':fineract-avro-schemas')
--- a/fineract-accounting/dependencies.gradle
+++ b/fineract-accounting/dependencies.gradle
@@ -16,7 +16,6 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-ext['spring-modulith.version'] = '1.4.11'
 dependencies {
     // Never use "compile" scope, but make all dependencies either 'implementation', 'runtimeOnly' or 'testCompile'.
     // Note that we never use 'api', because Fineract at least currently is a simple monolithic application ("WAR"), not a library.
--- a/fineract-accounting/dependencies.gradle
+++ b/fineract-accounting/dependencies.gradle
@@ -26,7 +26,6 @@ dependencies {
     //
     implementation(project(path: ':fineract-core'))
     testImplementation 'org.springframework.modulith:spring-modulith-core'
-    implementation(project(path: ':fineract-charge'))
     implementation('org.apache.avro:avro')
     implementation(
             project(path: ':fineract-avro-schemas')
--- a/fineract-accounting/dependencies.gradle
+++ b/fineract-accounting/dependencies.gradle
@@ -25,7 +25,6 @@ dependencies {
     //
     implementation(project(path: ':fineract-core'))
     testImplementation 'org.springframework.modulith:spring-modulith-core'
-    testImplementation("org.osgi:org.osgi.core:6.0.0")
     implementation('org.apache.avro:avro')
     implementation(
             project(path: ':fineract-avro-schemas')
--- a/fineract-accounting/dependencies.gradle
+++ b/fineract-accounting/dependencies.gradle
@@ -16,7 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-
+ext['spring-modulith.version'] = '1.4.11'
 dependencies {
     // Never use "compile" scope, but make all dependencies either 'implementation', 'runtimeOnly' or 'testCompile'.
     // Note that we never use 'api', because Fineract at least currently is a simple monolithic application ("WAR"), not a library.
@@ -25,6 +25,7 @@ dependencies {
     // implementation dependencies are directly used (compiled against) in src/main (and src/test)
     //
     implementation(project(path: ':fineract-core'))
+    testImplementation 'org.springframework.modulith:spring-modulith-core'
     implementation(project(path: ':fineract-charge'))
     implementation('org.apache.avro:avro')
     implementation(
--- a/fineract-accounting/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/service/AccountingChargeReadService.java
+++ b/fineract-accounting/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/service/AccountingChargeReadService.java
@@ -1,0 +1,28 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.accounting.producttoaccountmapping.service;
+
+import java.util.Collection;
+import java.util.List;
+import org.apache.fineract.portfolio.charge.data.ChargeData;
+
+public interface AccountingChargeReadService {
+
+    List<ChargeData> findChargesByIds(Collection<Long> chargeIds);
+}
--- a/fineract-accounting/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/service/AccountingChargeReadService.java
+++ b/fineract-accounting/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/service/AccountingChargeReadService.java
@@ -18,11 +18,11 @@
  */
 package org.apache.fineract.accounting.producttoaccountmapping.service;
 
-import java.util.Collection;
 import java.util.List;
+import java.util.Set;
 import org.apache.fineract.portfolio.charge.data.ChargeData;
 
 public interface AccountingChargeReadService {
 
-    List<ChargeData> findChargesByIds(Collection<Long> chargeIds);
+    List<ChargeData> findChargesByIds(Set<Long> chargeIds);
 }
--- a/fineract-accounting/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/service/ProductToGLAccountMappingHelper.java
+++ b/fineract-accounting/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/service/ProductToGLAccountMappingHelper.java
@@ -46,13 +46,11 @@ import org.apache.fineract.infrastructure.codes.domain.CodeValueRepository;
 import org.apache.fineract.infrastructure.core.api.JsonCommand;
 import org.apache.fineract.infrastructure.core.data.ApiParameterError;
 import org.apache.fineract.infrastructure.core.exception.PlatformApiDataValidationException;
-import org.apache.fineract.infrastructure.core.exception.ResourceNotFoundException;
 import org.apache.fineract.infrastructure.core.serialization.FromJsonHelper;
 import org.apache.fineract.portfolio.PortfolioProductType;
 import org.apache.fineract.portfolio.paymenttype.domain.PaymentType;
 import org.apache.fineract.portfolio.paymenttype.domain.PaymentTypeRepository;
 import org.apache.fineract.portfolio.paymenttype.exception.PaymentTypeNotFoundException;
-import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.stereotype.Component;
 
 @Component
@@ -68,7 +66,6 @@ public class ProductToGLAccountMappingHelper {
     protected final GLAccountRepositoryWrapper accountRepositoryWrapper;
     private final PaymentTypeRepository paymentTypeRepository;
     private final CodeValueRepository codeValueRepository;
-    private final JdbcTemplate jdbcTemplate;
 
     public void saveProductToAccountMapping(final JsonElement element, final String paramName, final Long productId,
             final int placeHolderTypeId, final GLAccountType expectedAccountType, final PortfolioProductType portfolioProductType) {
@@ -613,18 +610,6 @@ public class ProductToGLAccountMappingHelper {
      */
     private void saveChargeToFundSourceMapping(final Long productId, final Long chargeId, final Long incomeAccountId,
             final PortfolioProductType portfolioProductType, final boolean isPenalty) {
-
-        // The accounting module no longer loads the Charge entity (it is decoupled from the charge domain), so the
-        // existence check that loading the entity used to provide is restored here with a lightweight JDBC lookup.
-        // This keeps acc_product_mapping.charge_id from persisting orphaned references without reintroducing a
-        // compile-time dependency on the charge module.
-        // TODO Vishwas: Need to validate if given charge is fee or Penalty
-        // based on input condition
-        final Integer chargeCount = this.jdbcTemplate.queryForObject("select count(*) from m_charge where id = ?", Integer.class, chargeId);
-        if (chargeCount == null || chargeCount == 0) {
-            throw new ResourceNotFoundException("error.msg.charge.id.invalid", "Charge with identifier " + chargeId + " does not exist",
-                    new Object[] { chargeId });
-        }
 
         GLAccount glAccount;
         /**
--- a/fineract-accounting/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/service/ProductToGLAccountMappingHelper.java
+++ b/fineract-accounting/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/service/ProductToGLAccountMappingHelper.java
@@ -48,8 +48,6 @@ import org.apache.fineract.infrastructure.core.data.ApiParameterError;
 import org.apache.fineract.infrastructure.core.exception.PlatformApiDataValidationException;
 import org.apache.fineract.infrastructure.core.serialization.FromJsonHelper;
 import org.apache.fineract.portfolio.PortfolioProductType;
-import org.apache.fineract.portfolio.charge.domain.Charge;
-import org.apache.fineract.portfolio.charge.domain.ChargeRepositoryWrapper;
 import org.apache.fineract.portfolio.paymenttype.domain.PaymentType;
 import org.apache.fineract.portfolio.paymenttype.domain.PaymentTypeRepository;
 import org.apache.fineract.portfolio.paymenttype.exception.PaymentTypeNotFoundException;
@@ -65,7 +63,6 @@ public class ProductToGLAccountMappingHelper {
     protected final GLAccountRepository accountRepository;
     protected final ProductToGLAccountMappingRepository accountMappingRepository;
     protected final FromJsonHelper fromApiJsonHelper;
-    private final ChargeRepositoryWrapper chargeRepositoryWrapper;
     protected final GLAccountRepositoryWrapper accountRepositoryWrapper;
     private final PaymentTypeRepository paymentTypeRepository;
     private final CodeValueRepository codeValueRepository;
@@ -344,7 +341,7 @@ public class ProductToGLAccountMappingHelper {
                **/
             else {
                 for (final ProductToGLAccountMapping chargeToIncomeAccountMapping : existingChargeToIncomeAccountMappings) {
-                    final Long currentCharge = chargeToIncomeAccountMapping.getCharge().getId();
+                    final Long currentCharge = chargeToIncomeAccountMapping.getChargeId();
                     existingCharges.add(currentCharge);
                     // update existing mappings (if required)
                     if (inputChargeToIncomeAccountMap.containsKey(currentCharge)) {
@@ -613,7 +610,6 @@ public class ProductToGLAccountMappingHelper {
      */
     private void saveChargeToFundSourceMapping(final Long productId, final Long chargeId, final Long incomeAccountId,
             final PortfolioProductType portfolioProductType, final boolean isPenalty) {
-        final Charge charge = this.chargeRepositoryWrapper.findOneWithNotFoundDetection(chargeId);
 
         // TODO Vishwas: Need to validate if given charge is fee or Penalty
         // based on input condition
@@ -635,7 +631,7 @@ public class ProductToGLAccountMappingHelper {
         }
         final ProductToGLAccountMapping accountMapping = new ProductToGLAccountMapping().setGlAccount(glAccount).setProductId(productId)
                 .setProductType(portfolioProductType.getValue()).setFinancialAccountType(placeHolderAccountType.getValue())
-                .setCharge(charge);
+                .setChargeId(chargeId);
         this.accountMappingRepository.saveAndFlush(accountMapping);
     }
 
--- a/fineract-accounting/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/service/ProductToGLAccountMappingHelper.java
+++ b/fineract-accounting/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/service/ProductToGLAccountMappingHelper.java
@@ -46,11 +46,13 @@ import org.apache.fineract.infrastructure.codes.domain.CodeValueRepository;
 import org.apache.fineract.infrastructure.core.api.JsonCommand;
 import org.apache.fineract.infrastructure.core.data.ApiParameterError;
 import org.apache.fineract.infrastructure.core.exception.PlatformApiDataValidationException;
+import org.apache.fineract.infrastructure.core.exception.ResourceNotFoundException;
 import org.apache.fineract.infrastructure.core.serialization.FromJsonHelper;
 import org.apache.fineract.portfolio.PortfolioProductType;
 import org.apache.fineract.portfolio.paymenttype.domain.PaymentType;
 import org.apache.fineract.portfolio.paymenttype.domain.PaymentTypeRepository;
 import org.apache.fineract.portfolio.paymenttype.exception.PaymentTypeNotFoundException;
+import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.stereotype.Component;
 
 @Component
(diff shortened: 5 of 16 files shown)


fineract-accounting/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/service/ProductToGLAccountMappingReadPlatformServiceImpl.java
```diff
@@ -257,21 +259,23 @@ public List<ChargeToGLAccountMapper> fetchPenaltyToIncomeAccountMappingsForSavin
...
-            chargeToGLAccountMappers.add(chargeToGLAccountMapper);
-        }
-        return chargeToGLAccountMappers;
+        // The accounting module is decoupled from the charge domain, so we cannot navigate a Charge association in
+        // JPQL. We read the charge -> income-account rows via JdbcTemplate rather than a JPA native query: EclipseLink
+        // does not bind Spring Data ":name" parameters in native queries (it expects "#name"), which caused the raw
+        // markers to reach the database. JdbcTemplate uses plain JDBC positional binding and works on all supported
+        // databases. The GL account fields are selected via a join on acc_gl_account so we avoid a per-row lookup.
+        // Column order: [0] gl_account_id, [1] gl_account_name, [2] gl_code, [3] charge_id, [4] charge_name.
+        final String sql = "select ga.id, ga.name, ga.gl_code, c.id, c.name " + "from acc_product_mapping apm "
```
Aman-Mittal: You can use java textblock for static multiline string

fineract-accounting/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/service/ProductToGLAccountMappingHelper.java
```diff
@@ -613,10 +613,18 @@ private void savePaymentChannelToFundSourceMapping(final Long productId, final L
...
+        // existence check that loading the entity used to provide is restored here with a lightweight JDBC lookup.
+        // This keeps acc_product_mapping.charge_id from persisting orphaned references without reintroducing a
+        // compile-time dependency on the charge module.
         // TODO Vishwas: Need to validate if given charge is fee or Penalty
         // based on input condition
+        final Integer chargeCount = this.jdbcTemplate.queryForObject("select count(*) from m_charge where id = ?", Integer.class, chargeId);
+        if (chargeCount == null || chargeCount == 0) {
+            throw new ResourceNotFoundException("error.msg.charge.id.invalid", "Charge with identifier " + chargeId + " does not exist",
+                    new Object[] { chargeId });
+        }
```
adamsaghy: I don’t believe we need this validation here. Before providing the chargeId to this helper, we should validate it. We shouldn’t mix the responsibilities here.

fineract-accounting/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/service/ProductToGLAccountMappingReadPlatformServiceImpl.java
adamsaghy: I don’t like this idea. You say we are decoupling the charge, but in reality, it doesn’t happen. Just because you are using a workaround to use native SQL queries it still means the responsibility remains here: Fetch the ChargeData via the charge service (the interface can and the DTO can remain in the accounting module, while the implementation can go into the charge module).

Let’s decide whether the accounting module depends on the charge or the charge depends on the accounting. One of them must, because there is common logic.

fineract-accounting/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/service/WorkingCapitalLoanProductAdvancedAccountingReadHelper.java
```diff
@@ -74,17 +79,26 @@ public List<AdvancedMappingToExpenseAccountData> fetchWriteOffReasonMappings(fin
...
-        final List<ProductToGLAccountMapping> mappings = penalty
-                ? productToGLAccountMappingRepository.findAllPenaltyMappings(wcLoanProductId,
-                        PortfolioProductType.WORKING_CAPITAL_LOAN.getValue())
-                : productToGLAccountMappingRepository.findAllFeeMappings(wcLoanProductId,
-                        PortfolioProductType.WORKING_CAPITAL_LOAN.getValue());
+        // The charge name lives on m_charge, not on ProductToGLAccountMapping (which only stores charge_id), and the
+        // accounting module is decoupled from the charge domain, so we read the charge -> income-account rows via
+        // JdbcTemplate rather than a JPA native query (EclipseLink does not bind Spring Data ":name" markers in native
+        // queries). Column order: [0] gl_account_id, [1] charge id, [2] charge name. The penalty flag is known from the
+        // method argument, so it is inlined as a boolean literal (never user input) rather than bound.
```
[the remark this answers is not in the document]
adamsaghy: Copilot is right, also, see above: this is not the way...

adamsaghy (changes requested):
Please review my concerns and lets find a different approach...

fineract-accounting/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/service/ProductToGLAccountMappingReadPlatformServiceImpl.java
vidakovic: First thing that jumps at you is the string concatenation (just to confirm @Aman-Mittal 's observation). Even if that particular snippet doesn't seem to contain any of the usual SQL injection issues... at the very least we should use multiline strings (still beware of any use of unchecked variables!); if it's a simple query then use Spring Data JDBC repository functions (`findByXXX`).

vidakovic:
Just a general suggestion: maybe let's go easy with the use of Copilot here... I don't think that the code base is in any shape to have predictable results here (which is anyway not really possible with these tools). And especially with this task (defining proper domain boundaries) I think Copilot might not be that useful, because we have to surgically make the needed changes.

fineract-accounting/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/service/ProductToGLAccountMappingHelper.java
```diff
@@ -613,10 +613,18 @@ private void savePaymentChannelToFundSourceMapping(final Long productId, final L
      */
     private void saveChargeToFundSourceMapping(final Long productId, final Long chargeId, final Long incomeAccountId,
             final PortfolioProductType portfolioProductType, final boolean isPenalty) {
-        final Charge charge = this.chargeRepositoryWrapper.findOneWithNotFoundDetection(chargeId);
 
+        // The accounting module no longer loads the Charge entity (it is decoupled from the charge domain), so the
```
vidakovic: Is this something inserted by Copilot? If not: the code itself is maybe not the proper place to put documentation like this. I know, it's there are not really hard criteria on how much you should or could comment, but this here looks rather large and in fact the changed code should speak for itself... if it doesn't then maybe something is not entirely clear in general.

fineract-accounting/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/service/ProductToGLAccountMappingReadPlatformServiceImpl.java
```diff
@@ -257,21 +259,23 @@ public List<ChargeToGLAccountMapper> fetchPenaltyToIncomeAccountMappingsForSavin
...
-            final GLAccountData gLAccountData = new GLAccountData().setId(mapping.getGlAccount().getId())
-                    .setName(mapping.getGlAccount().getName()).setGlCode(mapping.getGlAccount().getGlCode());
-            final ChargeData chargeData = ChargeData.builder().id(mapping.getCharge().getId()).name(mapping.getCharge().getName())
-                    .penalty(mapping.getCharge().isPenalty()).build();
-            final ChargeToGLAccountMapper chargeToGLAccountMapper = new ChargeToGLAccountMapper().setCharge(chargeData)
-                    .setIncomeAccount(gLAccountData);
-            chargeToGLAccountMappers.add(chargeToGLAccountMapper);
-        }
-        return chargeToGLAccountMappers;
+        // The accounting module is decoupled from the charge domain, so we cannot navigate a Charge association in
```
vidakovic: Same here, see comment about lengthy explanations in sources.

fineract-accounting/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/service/WorkingCapitalLoanProductAdvancedAccountingReadHelper.java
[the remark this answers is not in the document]
vidakovic: That would be a regression, agreed.

fineract-accounting/src/test/java/org/apache/fineract/accounting/AccountingCrossFeatureBoundaryTest.java
```diff
@@ -0,0 +1,148 @@
...
+            sourceTypeToTargets.computeIfAbsent(sourceType, key -> new TreeSet<>()).add(label);
+
+            if (foundation) {
+                allowedFromCore.add(featureKey + " (" + artifact + ")");
+            } else {
+                violationFeatures.add(featureKey + " (" + artifact + ")");
+            }
+        });
+
+        LOG.info("==== Accounting cross-feature dependency report ====");
```
vidakovic: Use Lombok annotation `@Slf4j`... these uppercase `LOG.xxx(...)` will become lowercase `log.xxx(...)`. Just to stay consistent with the rest of the codebase.

fineract-accounting/dependencies.gradle
```diff
@@ -16,7 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-
+ext['spring-modulith.version'] = '1.4.11'
```
vidakovic: We should probably put this (dependency management) where we manage all dependencies: see https://github.com/apache/fineract/blob/develop/buildSrc/src/main/groovy/org.apache.fineract.dependencies.gradle

fineract-accounting/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/service/ProductToGLAccountMappingHelper.java
mansi75: I'll remove it. Charge id validation belongs upstream, before the id reaches the mapping helper.

fineract-accounting/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/service/WorkingCapitalLoanProductAdvancedAccountingReadHelper.java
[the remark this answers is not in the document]
mansi75: Got it. I'll rework on it

fineract-accounting/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/service/ProductToGLAccountMappingReadPlatformServiceImpl.java
mansi75: @vidakovic Got it

adamsaghy: Lets not use native sql queries and jdbctemplate calls. Whenever a module needs information from an another module, have a service for that.

mansi75: @adamsaghy I agree. The native SQL moves the coupling to the schema rather than removing it.
'financialAccountType' already distinguishes these on our own row, and the update path in 'ProductToGLAccountMappingHelper' already uses it so the read path can too, no charge navigation. Only the charge name needs the charge module.
I'd propose charge depends on accounting, interface in accounting, impl in charge (ChargeData is already in core). Is this right to proceed with?

buildSrc/src/main/groovy/org.apache.fineract.dependencies.gradle
```diff
@@ -37,6 +37,7 @@ dependencyManagement {
         mavenBom 'io.github.resilience4j:resilience4j-bom:2.4.0'
         mavenBom 'org.testcontainers:testcontainers-bom:1.21.4'
         mavenBom 'org.glassfish.jersey:jersey-bom:3.1.11'
+        mavenBom 'org.springframework.modulith:spring-modulith-bom:1.4.11'
```
adamsaghy, line 40: Is this relevant here?

fineract-accounting/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/service/AccountingChargeReadService.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+package org.apache.fineract.accounting.producttoaccountmapping.service;
+
+import java.util.Collection;
+import java.util.List;
+import org.apache.fineract.portfolio.charge.data.ChargeData;
+
+public interface AccountingChargeReadService {
+
+    List<ChargeData> findChargesByIds(Collection<Long> chargeIds);
```
adamsaghy: Use `List` or `Set`, no Collection

fineract-accounting/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/service/ProductToGLAccountMappingReadPlatformServiceImpl.java
```diff
@@ -258,18 +262,23 @@ public List<ChargeToGLAccountMapper> fetchPenaltyToIncomeAccountMappingsForSavin
     private List<ChargeToGLAccountMapper> fetchChargeToIncomeAccountMappings(final PortfolioProductType portfolioProductType,
             final Long loanProductId, final boolean penalty) {
         final List<ProductToGLAccountMapping> mappings = penalty
-                ? productToGLAccountMappingRepository.findAllPenaltyMappings(loanProductId, portfolioProductType.getValue())
-                : productToGLAccountMappingRepository.findAllFeeMappings(loanProductId, portfolioProductType.getValue());
+                ? productToGLAccountMappingRepository.findAllPenaltyToIncomeAccountMappings(loanProductId, portfolioProductType.getValue())
+                : productToGLAccountMappingRepository.findAllFeeToIncomeAccountMappings(loanProductId, portfolioProductType.getValue());
+        if (mappings.isEmpty()) {
+            return null;
```
adamsaghy: Previously it was returning empty list, now it return `null`. Seems incorrect to me.

fineract-accounting/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/service/WorkingCapitalLoanProductAdvancedAccountingReadHelper.java
```diff
@@ -74,20 +79,26 @@ public List<AdvancedMappingToExpenseAccountData> fetchWriteOffReasonMappings(fin
...
+        final int productType = PortfolioProductType.WORKING_CAPITAL_LOAN.getValue();
         final List<ProductToGLAccountMapping> mappings = penalty
-                ? productToGLAccountMappingRepository.findAllPenaltyMappings(wcLoanProductId,
-                        PortfolioProductType.WORKING_CAPITAL_LOAN.getValue())
-                : productToGLAccountMappingRepository.findAllFeeMappings(wcLoanProductId,
-                        PortfolioProductType.WORKING_CAPITAL_LOAN.getValue());
+                ? productToGLAccountMappingRepository.findAllPenaltyToIncomeAccountMappings(wcLoanProductId, productType)
+                : productToGLAccountMappingRepository.findAllFeeToIncomeAccountMappings(wcLoanProductId, productType);
+        if (mappings.isEmpty()) {
+            return null;
```
adamsaghy: is it safe to return `null`?

fineract-accounting/src/test/java/org/apache/fineract/accounting/AccountingCrossFeatureBoundaryTest.java
```diff
@@ -0,0 +1,148 @@
...
+import java.util.regex.Pattern;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.EnabledIfSystemProperty;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.modulith.core.ApplicationModule;
+import org.springframework.modulith.core.ApplicationModuleDependency;
+import org.springframework.modulith.core.ApplicationModules;
+
+class AccountingCrossFeatureBoundaryTest {
```
adamsaghy, line 40: I am not too fond of this idea...

Any reason we want to do this?

fineract-accounting/build.gradle
```diff
@@ -48,6 +48,11 @@ configurations {
 
 apply from: 'dependencies.gradle'
 
+test {
```
adamsaghy, line 51: Any reason for this?

fineract-accounting/dependencies.gradle
```diff
@@ -25,7 +24,8 @@ dependencies {
     // implementation dependencies are directly used (compiled against) in src/main (and src/test)
     //
     implementation(project(path: ':fineract-core'))
-    implementation(project(path: ':fineract-charge'))
+    testImplementation 'org.springframework.modulith:spring-modulith-core'
+    testImplementation("org.osgi:org.osgi.core:6.0.0")
```
adamsaghy: osgi?

fineract-charge/src/main/java/org/apache/fineract/portfolio/charge/service/AccountingChargeReadServiceImpl.java
```diff
@@ -0,0 +1,47 @@
...
+ * the accounting module depends only on the interface it owns, and the charge module supplies the details at runtime.
+ */
+@Service
+@RequiredArgsConstructor
+public class AccountingChargeReadServiceImpl implements AccountingChargeReadService {
+
+    private final ChargeRepository chargeRepository;
+
+    @Override
+    public List<ChargeData> findChargesByIds(final Collection<Long> chargeIds) {
```
adamsaghy: Dont use `Collection`. Its either a `List` or `Set`

fineract-accounting/src/test/java/org/apache/fineract/accounting/AccountingCrossFeatureBoundaryTest.java
mansi75: The intent was to assert at the module level that fineract-accounting no longer reaches into other modules after this decoupling. But AllModulesCrossFeatureBoundaryTest (FINERACT-2645) already covers cross-feature boundaries globally, so we can drop this if you want. Let me know if you'd prefer that.

buildSrc/src/main/groovy/org.apache.fineract.dependencies.gradle
mansi75: It's only here to pin the spring-modulith-core version used by AccountingCrossFeatureBoundaryTest. If we drop that test (per your other comment), I'll remove this BOM entry too.

fineract-accounting/dependencies.gradle
mansi75: Test-only. Spring-modulith-core's module scanning fails at runtime without an OSGi class on the classpath, so it's needed to run the boundary test.

fineract-accounting/build.gradle
mansi75: That block only configures the run for the new boundary test. It comes out along with AccountingCrossFeatureBoundaryTest if we drop it.

buildSrc/src/main/groovy/org.apache.fineract.dependencies.gradle
vidakovic: Let's leave it, this will be part of the self-documentation based on the current layout of the source code.

fineract-accounting/dependencies.gradle
vidakovic: Hmmm... this is suspicious... can you tell with which error this fails? OSGi shouldn't be necessary... only Eclipse is using this technology (and Apache Karaf). The Spring ecosystem bets entirely on Java modules.

fineract-accounting/src/main/java/org/apache/fineract/accounting/producttoaccountmapping/service/ProductToGLAccountMappingReadPlatformServiceImpl.java
vidakovic: @adamsaghy the old code made no sense... checking for an empty collection: if it's not empty then returns null (why?!?) else if it's empty then create a new empty list (that is just waste and makes no sense whatsoever).

@mansi75 if I'm not reading this the wrong way, your code is reversing the above "contract"... which is also not correct. Your code translates to: if  the collection is empty then return null... which is the exact opposite of the original code (flawed as it might be).

Why not just return an empty list and be done with it... null-safe... ?

adamsaghy:
## Summary

The core idea is sound. `ProductToGLAccountMapping.charge` becomes a plain `chargeId` column, the JPQL is
rewritten accordingly, and charge lookups move behind an `AccountingChargeReadService` port implemented in
`fineract-charge`. The batched `findChargesByIds` also removes the old per-mapping eager `@ManyToOne` fetch,
which is a genuine improvement.

However, the write-side validation that the removed `ChargeRepositoryWrapper` was implicitly providing has not
been fully re-homed, and several read paths change their response shape. Details below.

---

## Findings

### 1. Charge-existence validation is lost for working-capital loan products

`fineract-accounting/.../service/ProductToGLAccountMappingHelper.java:611`

Removing `chargeRepositoryWrapper.findOneWithNotFoundDetection(chargeId)` from `saveChargeToFundSourceMapping`
drops validation for **every** caller that does not go through
`fineract-provider`'s `ProductToGLAccountMappingWritePlatformServiceImpl` — where the new `validateChargesExist()`
guard lives.

The working-capital loan product path does not go through that service:

```
WorkingCapitalProductAccountingMappingServiceImpl.createAccountMapping
  -> WorkingCapitalLoanProductToGLAccountMappingHelper.saveAdvancedMappings   (fineract-working-capital-loan)
    -> ProductToGLAccountMappingHelper.saveChargesToGLAccountMappings
      -> saveChargeToFundSourceMapping
```

**Scenario:** `POST /workingcapitalloanproducts` with
`accountingRule = ACCRUAL_WITH_DEFERRED_REVENUE_AMORTIZATION` and
`feeToIncomeAccountMappings: [{ "chargeId": 999999, "incomeAccountId": <valid> }]`.

- **Before:** 404 `ChargeNotFoundException`.
- **After:** `saveAndFlush` hits the DB FK `FK_acc_product_mapping_m_charge`
  (`0001_initial_schema.xml`, changeSet 693) → raw constraint violation / HTTP 500.

The same gap applies to `WorkingCapitalLoanProductToGLAccountMappingHelper.updateAdvancedMappings`.

**Suggested fix:** keep the validation inside `saveChargeToFundSourceMapping` (via the new
`AccountingChargeReadService`, so no charge-domain dependency returns), rather than duplicating it in one
caller. That preserves the guarantee for all current and future callers.

---

### 2. Nested `charge` object grows from 3 fields to the full `ChargeData`

`fineract-accounting/.../service/ProductToGLAccountMappingReadPlatformServiceImpl.java:273`
(and the equivalent code in `WorkingCapitalLoanProductAdvancedAccountingReadHelper`)

Previously each entry was built inline as a stub:

```java
ChargeData.builder().id(...).name(...).penalty(...).build()
```

Now `AccountingChargeReadServiceImpl` returns `Charge.toData()`, which populates `amount`, `currency`,
`chargeTimeType`, `chargeAppliesTo`, `chargeCalculationType`, `chargePaymentMode`, `active`, `minCap`/`maxCap`,
`incomeOrLiabilityAccount`, `taxGroup`, and more.

**Scenario:** `GET /loanproducts/{id}` on a product with fee mappings. Any client or integration test asserting
the shape of `feeToIncomeAccountMappings[].charge` breaks, and the payload grows noticeably for products with
many charge mappings.

**Suggested fix:** project only `{id, name, penalty}` in the port (e.g. a dedicated lightweight DTO or a
JPQL constructor expression)

---

### 3. Fee/penalty classification switches basis, with no data migration

`fineract-accounting/.../domain/ProductToGLAccountMappingRepository.java:53`

`findAllFeeMappings` / `findAllPenaltyMappings` (filtered on `mapping.charge.penalty`) are replaced by
`findAllFeeToIncomeAccountMappings` / `findAllPenaltyToIncomeAccountMappings` (filtered on
`financialAccountType` 4/5).

**Scenario:** a charge with `penalty = true` was configured through `feeToIncomeAccountMappings`.
`saveChargeToFundSourceMapping` never validated fee-vs-penalty — the deleted `// TODO Vishwas: Need to validate
if given charge is fee or Penalty` said exactly that — so the row is stored with `financial_account_type = 4`
while `m_charge.is_penalty = true`.

- **Before:** `GET /loanproducts/{id}` returned it under `penaltyToIncomeAccountMappings`.
- **After:** it is returned under `feeToIncomeAccountMappings`.

A client doing GET → PUT round-trips the mappings and silently moves them between buckets.

The new behaviour is arguably the *more* consistent one — it matches what `updateChargeToIncomeAccountMappings`
already used on the write side. But it is an unannounced behaviour change on existing data and should at
minimum be called out in the PR description.

---

### 4. Empty charge mappings now serialize as `[]` instead of being omitted

`fineract-accounting/.../service/WorkingCapitalLoanProductAdvancedAccountingReadHelper.java:110`
(also `ProductToGLAccountMappingReadPlatformServiceImpl.java:289`)

`return result.isEmpty() ? null : result;` became `return result;`. Gson omits `null` fields, so the key
previously did not appear in the JSON at all; it now appears as `[]`.

Worse, the sibling methods in the *same class* — `fetchPaymentTypeToFundSourceMappings` (line 63) and
`fetchReasonMappings` (line 122) — still return `null` on empty. A single response now mixes both conventions.

**Suggested fix:** revert to `null` for consistency

---

### 5. `validateChargesExist` runs even when the mappings are discarded

`fineract-provider/.../service/ProductToGLAccountMappingWritePlatformServiceImpl.java:66`

The call sits at the top of `createLoanProductToGLAccountMapping`, before `accountingRuleType` is decoded.

**Scenario:** `POST /loanproducts` with `accountingRule = 1` (NONE) and a leftover
`feeToIncomeAccountMappings` block containing a since-deleted `chargeId` — e.g. a UI that keeps the
advanced-accounting section populated when the user switches the rule to *None*. The `switch` hits
`case NONE: break;` and never persists or reads those arrays, so the request previously succeeded. It now
throws `ChargeNotFoundException` (404).

**Suggested fix:** move the call into the branches that actually persist the mappings, or guard it on
`accountingRuleType != NONE`.

---

## Design notes

**Dependency direction.** The PR removes `accounting → charge` but adds `charge → accounting` in
`fineract-charge/dependencies.gradle`, purely so `AccountingChargeReadServiceImpl` can implement an interface
owned by the accounting module. `ChargeData` already lives in `fineract-core` — putting the read contract there
too would decouple both modules without introducing a new edge in the opposite direction. As it stands, the
coupling is inverted rather than removed, and the new boundary test guards only one of the two directions.

**`AccountingCrossFeatureBoundaryTest`.** A few things worth tightening:

- It only sees `fineract-accounting`'s own test classpath, so it can only catch a re-added Gradle dependency.
  Worth stating that limitation in a class comment so the guarantee isn't over-read.
- `private static ApplicationModules modules` is lazily initialized without synchronization — not safe under
  parallel JUnit execution.
- `owningArtifact` falls back to `"(unknown-source)"` when `JavaClass.getSource()` is empty, and that value is
  not in `FOUNDATION_ARTIFACTS` — so it counts as a violation. False-failure risk.
- `printAccountingCrossFeatureDependencyReport` is permanently disabled behind an undocumented system property
  (`-Daccounting.boundary.report=true`). Either document it or drop it.
- `com.tngtech.archunit` is imported directly but only arrives transitively via `spring-modulith-core` —
  declare it explicitly as a test dependency.

**Build.** The `test { maxHeapSize = '2g'; jvmArgs += ['-XX:MaxMetaspaceSize=1g'] }` block in
`fineract-accounting/build.gradle` is a whole-module cost incurred by one test. Please add a comment explaining
it is for the Spring Modulith full-classpath scan.

**Nit.** The blank line after the license header in `fineract-accounting/dependencies.gradle` was removed for no
apparent reason — restore it to match every other module.

mansi75:
@adamsaghy I agree with all the points. Thank you for sharing the findings, I will work on it

PR #6116, opened 2026-07-10
FINERACT-2682: Fix bulk client upload failure when custom datatables …

Samer-Melhem-FOO:
## Problem

Bulk Client Upload (Excel import) fails once a custom datatable is registered
against the Client entity and an Entity Datatable Check makes it mandatory
for client creation (Entity = Client, Status = Create). The bulk import
workbook populator and import handlers have no support for datatable
columns: the downloaded template doesn't render columns for datatables
configured on the Client entity, and the import handlers have no way to
validate, parse, or submit datatable data as part of client creation — so a
mandatory datatable check blocks every row in the upload.

## Fix

- `ClientEntityWorkbookPopulator` / `ClientPersonWorkbookPopulator` now
  render a column for every datatable configured against the Client entity
  (including ones with no existing check, so they can be optionally
  populated), with support for dropdown, date, datetime, and boolean column
  types. Columns from single-row datatables are marked mandatory (`*`) to
  match Entity Datatable Check requirements; multi-row (repeatable child)
  datatable columns are always optional.
- `ImportHandlerUtils` gains helpers to parse and validate datatable columns
  from the workbook, and `validateRequiredDatatables()` now skips multi-row
  datatables, since they represent repeatable child data rather than a
  single mandatory value.
- `ClientEntityImportHandler` / `ClientPersonImportHandler` submit parsed
  datatable data as part of the client-create command, with clearer error
  reporting when validation fails.
- `DatatableData` gains an `isMultiRow()` flag, threaded through
  `DatatableReadServiceImpl` and `ReadSurveyServiceImpl` (which shares the
  same `DatatableData.create(...)` factory).

## Affected APIs

- `GET /v1/clients/downloadtemplate?legalFormType=CLIENTS_PERSON|CLIENTS_ENTITY`
- `POST /v1/clients/uploadtemplate?legalFormType=CLIENTS_PERSON|CLIENTS_ENTITY`
- `GET /v1/imports`, `GET /v1/imports/downloadOutputTemplate`

## Testing

- Added unit tests for `DatatableData` covering single-row vs. multi-row
  behavior (`DatatableDataTest`).
- Manually verified end-to-end: registered a custom datatable on the Client
  entity, added an Entity Datatable Check (Create), downloaded the template
  (columns present, single-row marked mandatory), filled it in, uploaded it,
  and confirmed the client was created successfully for both Person and
  Entity legal forms, and for both single-row and multi-row datatables.
- `./gradlew :fineract-provider:compileJava`,
  `:fineract-core:compileTestJava`, `:fineract-provider:compileTestJava` all
  pass.

JIRA: https://issues.apache.org/jira/browse/FINERACT-2682

### DIFF
--- a/fineract-core/src/main/java/org/apache/fineract/infrastructure/dataqueries/data/ResultsetColumnValueData.java
+++ b/fineract-core/src/main/java/org/apache/fineract/infrastructure/dataqueries/data/ResultsetColumnValueData.java
@@ -19,13 +19,16 @@
 package org.apache.fineract.infrastructure.dataqueries.data;
 
 import java.io.Serializable;
+import lombok.Getter;
 
 /**
  * Immutable data object representing a possible value for a given resultset column.
  */
 public class ResultsetColumnValueData implements Serializable {
 
+    @Getter
     private final int id;
+    @Getter
     private final String value;
     @SuppressWarnings("unused")
     private final Integer score;
--- a/fineract-core/src/main/java/org/apache/fineract/infrastructure/dataqueries/data/ResultsetColumnValueData.java
+++ b/fineract-core/src/main/java/org/apache/fineract/infrastructure/dataqueries/data/ResultsetColumnValueData.java
@@ -19,12 +19,14 @@
 package org.apache.fineract.infrastructure.dataqueries.data;
 
 import java.io.Serializable;
+import lombok.Getter;
 
 /**
  * Immutable data object representing a possible value for a given resultset column.
  */
 public class ResultsetColumnValueData implements Serializable {
 
+    @Getter
     private final int id;
     private final String value;
     @SuppressWarnings("unused")
@@ -48,5 +50,9 @@ public class ResultsetColumnValueData implements Serializable {
 
     public boolean codeMatches(final Integer match) {
         return match.intValue() == this.id;
+    }
+
+    public String getValue() {
+        return value;
     }
 }
--- a/fineract-core/src/main/java/org/apache/fineract/portfolio/search/service/SearchUtil.java
+++ b/fineract-core/src/main/java/org/apache/fineract/portfolio/search/service/SearchUtil.java
@@ -271,7 +271,15 @@ public class SearchUtil {
                 }
                 return columnValue;
             } else if (columnHeader.isCodeLookupDisplayType()) {
-                final Integer codeLookup = Integer.valueOf(columnValue);
+                // Extract ID from display value format: "Label (ID)" if present
+                String extractedId = extractIdFromDisplayValue(columnValue);
+                Integer codeLookup;
+                try {
+                    codeLookup = Integer.valueOf(extractedId);
+                } catch (NumberFormatException e) {
+                    // If extraction didn't change the value or parsing failed, try original value
+                    codeLookup = Integer.valueOf(columnValue);
+                }
                 if (!columnHeader.isColumnCodeAllowed(codeLookup)) {
                     ApiParameterError error = ApiParameterError.parameterError("error.msg.invalid.columnValue",
                             "Value not found in Allowed Value list", columnHeader.getColumnName(), columnValue);
@@ -293,6 +301,17 @@ public class SearchUtil {
             return JsonParserHelper.convertDateTimeFrom(columnValue, columnHeader.getColumnName(), format, locale);
         }
         if (colType.isAnyIntegerType()) {
+            // Extract ID from display value format: "Label (ID)" if present
+            // This handles FK/INTEGER columns that show human-readable values in Excel
+            String extractedId = extractIdFromDisplayValue(columnValue);
+            if (!extractedId.equals(columnValue)) {
+                // ID was extracted, try to parse it
+                try {
+                    return Integer.parseInt(extractedId);
+                } catch (NumberFormatException e) {
+                    // If parsing fails, fall back to original conversion
+                }
+            }
             return helper.convertToInteger(columnValue, columnHeader.getColumnName(), locale);
         }
         if (colType.isDecimalType()) {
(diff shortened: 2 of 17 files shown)


adamsaghy:
@Samer-Melhem-FOO Please review the failing checks.

fineract-core/src/main/java/org/apache/fineract/infrastructure/dataqueries/data/ResultsetColumnHeaderData.java
```diff
@@ -149,6 +149,10 @@ public boolean hasColumnValues() {
         return columnValues != null && !columnValues.isEmpty();
     }
 
+    public List<ResultsetColumnValueData> getColumnValues() {
```
adamsaghy: You can use Lombok

fineract-core/src/main/java/org/apache/fineract/infrastructure/dataqueries/data/ResultsetColumnValueData.java
```diff
@@ -49,4 +49,12 @@ public boolean matches(final String match) {
     public boolean codeMatches(final Integer match) {
         return match.intValue() == this.id;
     }
+
+    public int getId() {
```
adamsaghy: You can use Lombok

fineract-core/src/main/java/org/apache/fineract/portfolio/search/service/SearchUtil.java
```diff
@@ -323,4 +342,28 @@ public String camelToSnake(final String camelStr) {
...
+
+    /**
+     * Extracts numeric ID from strings ending with "(id)" format. Example: "Permanent (32)" -> "32" If the string
+     * doesn't match the pattern, returns the original value.
+     *
+     * @param value
+     *            The string value that may contain an ID in parentheses
+     * @return The extracted ID as a string, or the original value if no ID pattern is found
+     */
+    private static String extractIdFromDisplayValue(String value) {
```
adamsaghy: Do we really need this magic?

Samer-Melhem-FOO:
@adamsaghy I've pushed the changes:

- Fixed the checkstyle/spotless violations (checks are green now).
- Addressed the review feedback: replaced the manual getters with Lombok `@Getter` in `ResultsetColumnHeaderData`/`ResultsetColumnValueData`, and removed the duplicated `extractIdFromDisplayValue` logic — `ImportHandlerUtils` now calls the existing method on `SearchUtil` instead of maintaining its own copy.

Please take another look when you get a chance.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/bulkimport/importhandler/client/ClientEntityImportHandler.java
```diff
@@ -73,13 +84,32 @@ private List<ClientData> readExcelFile(final Workbook workbook, final String loc
         for (int rowIndex = 1; rowIndex <= noOfEntries; rowIndex++) {
             Row row;
             row = clientSheet.getRow(rowIndex);
+            if (row != null) {
```
adamsaghy, line 87: Please remove info logging. its just noise.. make it debug level

fineract-provider/src/main/java/org/apache/fineract/infrastructure/bulkimport/importhandler/client/ClientEntityImportHandler.java
```diff
@@ -73,13 +84,32 @@ private List<ClientData> readExcelFile(final Workbook workbook, final String loc
...
+     */
+    private Long parseIdFromDisplayValue(final String displayValue, final String fieldName) {
+        if (displayValue == null) {
+            return null;
+        }
+        String idPart = SearchUtil.extractIdFromDisplayValue(displayValue);
+        if (idPart != null && idPart.matches("\\d+")) {
+            return Long.valueOf(idPart);
+        }
+        LOG.error("Bulk upload error: Could not extract ID from {}. Row data: {}", fieldName, displayValue);
```
adamsaghy: Is this a real error? I am assuming not necessarily, but logging it as an error might misleading no?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/bulkimport/service/BulkImportWorkbookPopulatorServiceImpl.java
```diff
@@ -237,21 +251,60 @@ private WorkbookPopulator populateClientWorkbook(final String entityType, final
...
+            // Convert EntityDatatableChecks to DatatableData
+            for (EntityDatatableChecks check : entityDatatableChecks) {
+                DatatableData datatable = datatableReadService.retrieveDatatable(check.getDatatableName());
+                if (datatable != null) {
+                    requiredDatatables.add(datatable);
+                }
+            }
+        } catch (Exception e) {
+            LOG.warn("Error fetching required datatables for bulk import template", e);
+            // Continue without datatables if there's an error
```
adamsaghy: Is it safe to just ignore it?

adamsaghy (changes requested):
Please review my concerns and squash your commits.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/bulkimport/importhandler/client/ClientEntityImportHandler.java
Samer-Melhem-FOO: Done, dropped to debug level.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/bulkimport/importhandler/client/ClientEntityImportHandler.java
Samer-Melhem-FOO: Fair point, not a hard failure since it returns null and processing continues. Changed to warn.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/bulkimport/service/BulkImportWorkbookPopulatorServiceImpl.java
Samer-Melhem-FOO: No, agreed — silently continuing without the datatables would reintroduce the exact bug this PR fixes (missing mandatory datatable columns), just less visibly. Removed the try/catch so it propagates, consistent with the other populate*Workbook helpers in this class.

Samer-Melhem-FOO:
Squashed the commits into one, rebased onto develop, and force-pushed.

fineract-core/src/main/java/org/apache/fineract/infrastructure/dataqueries/data/ResultsetColumnValueData.java
```diff
@@ -49,4 +51,8 @@ public boolean matches(final String match) {
     public boolean codeMatches(final Integer match) {
         return match.intValue() == this.id;
     }
+
+    public String getValue() {
```
adamsaghy: You can also use Lombok getter here...

adamsaghy:
@Samer-Melhem-FOO Please let us know, once it is updated

fineract-core/src/main/java/org/apache/fineract/infrastructure/dataqueries/data/ResultsetColumnValueData.java
Samer-Melhem-FOO: Done, switched to Lombok `@Getter` for `value` too, consistent with `id`.

Samer-Melhem-FOO:
@adamsaghy Pushed the fix — `value` in `ResultsetColumnValueData` now also uses Lombok `@Getter` instead of the manual getter. PR is up to date, please take another look when you get a chance.

PR #6134, opened 2026-07-14
FINERACT-2593: Fix NullPointerException in DelinquencyReadPlatformServiceImpl when loan product is null

AshharAhmadKhan:
### JIRA
https://issues.apache.org/jira/browse/FINERACT-2593

### Problem
`DelinquencyReadPlatformServiceImpl.calculateLoanCollectionData` had an inverted null check (`== null ||` instead of `!= null &&`) in the pending/approved/cancelled branch. When `loan.getLoanProduct()` returned null, that inverted condition actually routed execution straight into `calculateAvailableDisbursementAmountWithOverApplied(loan)`, which then dereferenced `loanProduct` with no guard at all. That gave two separate crash surfaces coming from the same root cause, both producing an unhandled NPE (HTTP 500).

### Fix
- Corrected the operator in the pending/approved/cancelled branch, so the helper is only called when a `LoanProduct` is actually present.
- Added a null guard inside `calculateAvailableDisbursementAmountWithOverApplied` itself. This method is part of the public `DelinquencyReadPlatformService` interface and is also called directly by `LoanReadPlatformServiceImpl.retrieveApprovalTemplate` and `retrieveDisbursalTemplate`, so guarding it there protects those call sites too, not just this one.

### Why no other guards were added
Every call path we could find loads the `Loan` and reads `getLoanProduct()` inside the same `@Transactional(readOnly = true)` method, so the lazy-loaded `loanProduct` proxy should always be initializable in practice. Rather than sprinkling defensive null checks everywhere, the fix stays scoped to the two places that were actually broken.

### Tests
- Null product loan hitting the pending/approved/cancelled branch: no exception, over-applied field stays at its default.
- Null product loan calling the helper directly: no exception, returns a sensible value instead of crashing.
- A regression test with a non-null product and over-apply enabled, to make sure the fix didn't break the working case.

### DIFF
--- a/fineract-loan/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyReadPlatformServiceImpl.java
+++ b/fineract-loan/src/main/java/org/apache/fineract/portfolio/delinquency/service/DelinquencyReadPlatformServiceImpl.java
@@ -156,7 +156,7 @@ public class DelinquencyReadPlatformServiceImpl implements DelinquencyReadPlatfo
 
             // If the Loan is not Active yet or is cancelled (rejected or withdrawn), return template data
             if (loan.isSubmittedAndPendingApproval() || loan.isApproved() || loan.isCancelled()) {
-                if (loan.getLoanProduct() == null || loan.getLoanProduct().isAllowApprovedDisbursedAmountsOverApplied()) {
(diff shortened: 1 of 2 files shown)


AshharAhmadKhan:
@adamsaghy please have a look when you get the chance !

adamsaghy (approved):
LGTM

PR #6158, opened 2026-07-21
FINERACT-2649: batch migrate Tier 4 loan tests to feign

DeathGun44:
## What this PR does

fourth batch of the loan integration-test migration. It moves **54 test classes (655 test methods)** off REST-assured onto the typed Feign client.

None of those 54 classes reference REST-assured any more. With this merged, 151 integration-test classes run on the typed client.

99 files changed:

- 54 test classes (655 test methods)
- 2 shared base classes (`FeignLoanTestBase`, `BaseLoanIntegrationTest`)
- 34 shared Feign helpers and builders (16 of them new)

The tests assert the same things as before. This changes how they talk to the server, not what they check. The few assertions that did change are listed under "Bugs found" and each one is a fix.

## Commits

**26 commits**, each a group of related tests (loan charges, COB, external asset owner, delinquency, advanced payment allocation, and so on).

## What's covered

- **Client loan lifecycle** - charges of every calculation type, credit balance refund, external IDs, backdated disbursement, overpaid status, modifying approved amounts, floating-rate interest recalculation, date validation.
- **Close of Business (COB)** - partitioning, catch-up, inline COB, account locks.
- **External asset owner** - transfer, cancel, search, initiate/undisbursed.
- **Advanced Payment Allocation** - repayment schedules, interest refunds, charges, reprocessing.
- **Charges** - specific due date, after maturity, instalment fees, penalties, charge-off accounting.
- **Delinquency** - buckets, business events, classification.
- **Batch API, scheduler jobs, journal reversal, SQL-injection reporting.**

## Bugs found

The typed client is stricter than REST-assured, so a few real problems showed up:

- **Wrong charge calculation type.** `ChargeRequestBuilders` had `CHARGE_CALCULATION_TYPE_PERCENTAGE_AMOUNT_AND_INTEREST = 4`, but 4 is percent-of-interest. Percent-of-amount-and-interest is 3, which is what `ChargesHelper` uses and what the tests meant. Every test using that builder was charging on the wrong base and still passing, because the amounts were plausible. Now uses `ChargeCalculationType.PERCENT_OF_AMOUNT_AND_INTEREST`.
- **A comparison that could never be true.** In `testLoanCharges_INSTALMENT_FEE` the instalment numbers are `Long` (the request model requires it) but the schedule reports its period as an `Integer`. `Long.equals(Integer)` is always false, so the "this is the waived instalment" and "this is the paid instalment" branches never ran and the test compared against the wrong expected values. Now compared numerically.
- **A crash hidden in a helper.** `modifyLoanApplication` passed `command` through `Map.of(...)`, which rejects nulls. Three charge tests in `ClientLoanIntegrationTest` pass null here and were throwing `NullPointerException` before reaching the server. Now uses the typed query parameter, which drops a null.
- **A dropped product setting.** `LoanProductTestBuilder.withDisallowExpectedDisbursements(true)` also quietly turned on a 100% over-applied disbursement allowance. The migrated builder set it to `false`, which broke `chargeOff` because it disburses two 1000 tranches against a 1000 approval. Restored.
- **A COB step missing its prerequisite.** `CobPartitioningTest` registered the external-asset-owner COB step but had lost the `ASSET_TRANSFER` GL mapping that step needs. The step list is tenant-wide and never restored, so this broke an unrelated later test.
- **Weak assertions.** The SQL-injection tests caught bare `RuntimeException` as "the server rejected this", so a client-side connection failure counted as a pass. They also looked for "404" in the exception message, which includes the request path, so a report whose ID contained `404` hid a real failure. Both now check the HTTP status.
- **Interest calculation changed.** One charges test had moved from `SAME_AS_REPAYMENT_PERIOD` to `DAILY`. Reverted.
- **Locale changed.** Account transfers were posted with `en` instead of `en_GB`. Restored.
- **A reversal check that never fired.** `verifyTransactions` compared `manuallyReversed` with `Objects.equals(item.getManuallyReversed(), tr.reversed)`, but the API omits that field entirely for non-reversed transactions, so the comparison was `Objects.equals(null, false)` - always false, making the whole predicate unreachable. The check now treats null and false as the same state, and per review feedback it exists in **both** validators: `LoanTestValidators` (Feign) and `BaseLoanIntegrationTest` (REST-assured), where there was previously no check at all.

## Shared test infrastructure

- **`LoanChargeCommandsApi`** and **`ClientChargeCommandsApi`** (new). Two hand-written Feign nterfaces, bound to the *generated* response models, for the three request shapes the generated request models cannot express: a bodyless `?command=waive` POST (the generated client always serialises at least `{}`, which that endpoint answers with a 500), a locale-formatted decimal amount (`"50,05"`), and a client-charge amount with decimal places. No JSON is built by hand. Retyping the shared models instead was rejected: `PostLoansLoanIdChargesRequest.amount` has 118 call sites passing a number, and `PostClientsClientIdChargesRequest.amount` is an `Integer` whose correction is a breaking OpenAPI change (swagger-brake R010) that belongs in its own PR.
- **`LoanRequestBuilders`** - added a type-safe `paymentAllocationOrder(PaymentAllocationType...)` overload, replacing copies of the same private helper in individual tests.
- **`ChargeRequestBuilders`** - dropped its three local `CHARGE_CALCULATION_TYPE_*` constants. All call sites now use the `ChargeCalculationType` enum, so the name and the value can't drift apart again.

## Checklist

### DIFF
--- a/integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanAccrualTransactionOnChargeSubmittedDateTest.java
+++ b/integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanAccrualTransactionOnChargeSubmittedDateTest.java
@@ -22,894 +22,463 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
-import io.restassured.builder.RequestSpecBuilder;
-import io.restassured.builder.ResponseSpecBuilder;
-import io.restassured.http.ContentType;
-import io.restassured.specification.RequestSpecification;
-import io.restassured.specification.ResponseSpecification;
+import java.math.BigDecimal;
 import java.time.LocalDate;
-import java.time.format.DateTimeFormatter;
-import java.time.format.DateTimeFormatterBuilder;
-import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
 import java.util.UUID;
-import org.apache.fineract.client.models.GetLoanProductsProductIdResponse;
 import org.apache.fineract.client.models.GetLoansLoanIdRepaymentPeriod;
 import org.apache.fineract.client.models.GetLoansLoanIdResponse;
-import org.apache.fineract.client.models.PostCreateRescheduleLoansRequest;
-import org.apache.fineract.client.models.PostCreateRescheduleLoansResponse;
+import org.apache.fineract.client.models.GetLoansLoanIdTransactions;
+import org.apache.fineract.client.models.PostLoanProductsRequest;
+import org.apache.fineract.client.models.PostLoansLoanIdChargesRequest;
 import org.apache.fineract.client.models.PostLoansLoanIdTransactionsRequest;
 import org.apache.fineract.client.models.PostLoansLoanIdTransactionsResponse;
+import org.apache.fineract.client.models.PostLoansRequest;
 import org.apache.fineract.client.models.PutGlobalConfigurationsRequest;
-import org.apache.fineract.infrastructure.businessdate.domain.BusinessDateType;
 import org.apache.fineract.infrastructure.configuration.api.GlobalConfigurationConstants;
-import org.apache.fineract.infrastructure.core.service.DateUtils;
-import org.apache.fineract.integrationtests.common.BusinessDateHelper;
-import org.apache.fineract.integrationtests.common.ClientHelper;
-import org.apache.fineract.integrationtests.common.LoanRescheduleRequestHelper;
-import org.apache.fineract.integrationtests.common.SchedulerJobHelper;
+import org.apache.fineract.integrationtests.client.feign.FeignLoanTestBase;
+import org.apache.fineract.integrationtests.client.feign.modules.LoanRequestBuilders;
+import org.apache.fineract.integrationtests.client.feign.modules.LoanTestData;
 import org.apache.fineract.integrationtests.common.Utils;
-import org.apache.fineract.integrationtests.common.accounting.Account;
-import org.apache.fineract.integrationtests.common.accounting.AccountHelper;
-import org.apache.fineract.integrationtests.common.accounting.PeriodicAccrualAccountingHelper;
-import org.apache.fineract.integrationtests.common.charges.ChargesHelper;
-import org.apache.fineract.integrationtests.common.loans.LoanApplicationTestBuilder;
-import org.apache.fineract.integrationtests.common.loans.LoanProductTestBuilder;
-import org.apache.fineract.integrationtests.common.loans.LoanRescheduleRequestTestBuilder;
-import org.apache.fineract.integrationtests.common.loans.LoanTransactionHelper;
-import org.apache.fineract.integrationtests.inlinecob.InlineLoanCOBHelper;
-import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Order;
 import org.junit.jupiter.api.Test;
 
 @Order(1)
-public class LoanAccrualTransactionOnChargeSubmittedDateTest extends BaseLoanIntegrationTest {
+public class LoanAccrualTransactionOnChargeSubmittedDateTest extends FeignLoanTestBase {
 
-    private ResponseSpecification responseSpec;
-    private RequestSpecification requestSpec;
-    private LoanTransactionHelper loanTransactionHelper;
-    private ClientHelper clientHelper;
-    private DateTimeFormatter dateFormatter = new DateTimeFormatterBuilder().appendPattern("dd MMMM yyyy").toFormatter();
-    private AccountHelper accountHelper;
-    private InlineLoanCOBHelper inlineLoanCOBHelper;
-
-    @BeforeEach
-    public void setup() {
-        Utils.initializeRESTAssured();
-        this.requestSpec = new RequestSpecBuilder().setContentType(ContentType.JSON).build();
-        this.requestSpec.header("Authorization", "Basic " + Utils.loginIntoServerAndGetBase64EncodedAuthenticationKey());
-        this.requestSpec.header("Fineract-Platform-TenantId", "default");
-        this.responseSpec = new ResponseSpecBuilder().expectStatusCode(200).build();
-        this.loanTransactionHelper = new LoanTransactionHelper(this.requestSpec, this.responseSpec);
-        this.clientHelper = new ClientHelper(this.requestSpec, this.responseSpec);
-        this.accountHelper = new AccountHelper(this.requestSpec, this.responseSpec);
-        this.inlineLoanCOBHelper = new InlineLoanCOBHelper(this.requestSpec, this.responseSpec);
-    }
+    private static final String STRATEGY = "mifos-standard-strategy";
+    private static final String CHARGE_ACCRUAL_DATE_SUBMITTED = "submitted-date";
+    private static final String CHARGE_ACCRUAL_DATE_DUE = "due-date";
 
     @Test
     public void loanAccrualTransactionOnChargeSubmittedTest_Accrual_Accounting_Api() {
         try {
-
-            // Accounts oof periodic accrual
-            final Account assetAccount = this.accountHelper.createAssetAccount();
-            final Account incomeAccount = this.accountHelper.createIncomeAccount();
-            final Account expenseAccount = this.accountHelper.createExpenseAccount();
-            final Account overpaymentAccount = this.accountHelper.createLiabilityAccount();
-
             // Set business date
-            LocalDate currentDate = LocalDate.of(2023, 3, 3);
-            final String accrualRunTillDate = dateFormatter.format(currentDate);
-
-            globalConfigurationHelper.updateGlobalConfiguration(GlobalConfigurationConstants.ENABLE_BUSINESS_DATE,
-                    new PutGlobalConfigurationsRequest().enabled(true));
-            BusinessDateHelper.updateBusinessDate(BusinessDateType.BUSINESS_DATE, currentDate);
-            globalConfigurationHelper.updateGlobalConfiguration(GlobalConfigurationConstants.CHARGE_ACCRUAL_DATE,
-                    new PutGlobalConfigurationsRequest().stringValue("submitted-date"));
+            enableSubmittedDateChargeAccrual(LocalDate.of(2023, 3, 3));
             // Loan ExternalId
             String loanExternalIdStr = UUID.randomUUID().toString();
-
             // Client and Loan account creation
-
-            final Integer clientId = clientHelper.createClient(ClientHelper.defaultClientCreationRequest()).getClientId().intValue();
-            final GetLoanProductsProductIdResponse getLoanProductsProductResponse = createLoanProduct(loanTransactionHelper, assetAccount,
-                    incomeAccount, expenseAccount, overpaymentAccount);
-            assertNotNull(getLoanProductsProductResponse);
-
-            final Integer loanId = createLoanAccount(clientId, getLoanProductsProductResponse.getId(), loanExternalIdStr);
+            final Long clientId = createClient();
+            // Accounts oof periodic accrual
+            final Long loanProductId = createLoanProduct(singleRepaymentAccrualProduct());
+            final Long loanId = createSingleRepaymentLoan(clientId, loanProductId, loanExternalIdStr);
 
             // Add Charge Penalty
-            Integer penalty = ChargesHelper.createCharges(requestSpec, responseSpec,
-                    ChargesHelper.getLoanSpecifiedDueDateJSON(ChargesHelper.CHARGE_CALCULATION_TYPE_FLAT, "10", true));
-
-            LocalDate targetDate = LocalDate.of(2023, 3, 10);
-            final String penaltyCharge1AddedDate = dateFormatter.format(targetDate);
-
-            Integer penalty1LoanChargeId = this.loanTransactionHelper.addChargesForLoan(loanId,
-                    LoanTransactionHelper.getSpecifiedDueDateChargesForLoanAsJSON(String.valueOf(penalty), penaltyCharge1AddedDate, "10"));
-
-            assertNotNull(penalty1LoanChargeId);
-
+            addPenaltyCharge(loanId, "10 March 2023", 10.0);
             // Add Charge Fee
-            Integer feeCharge = ChargesHelper.createCharges(requestSpec, responseSpec,
-                    ChargesHelper.getLoanSpecifiedDueDateJSON(ChargesHelper.CHARGE_CALCULATION_TYPE_FLAT, "10", false));
-
-            targetDate = LocalDate.of(2023, 3, 14);
-            final String feeChargeAddedDate = dateFormatter.format(targetDate);
-            Integer feeLoanChargeId = this.loanTransactionHelper.addChargesForLoan(loanId,
-                    LoanTransactionHelper.getSpecifiedDueDateChargesForLoanAsJSON(String.valueOf(feeCharge), feeChargeAddedDate, "10"));
-
-            assertNotNull(feeLoanChargeId);
+            addFeeCharge(loanId, "14 March 2023", 10.0);
 
             // Run accrual for charge created date
-            PeriodicAccrualAccountingHelper.runPeriodicAccrualAccounting(accrualRunTillDate);
-
+            runPeriodicAccrualAccounting("03 March 2023");
             // verify accrual transaction created for charges create date
-            checkAccrualTransaction(currentDate, 0.0f, 10.0f, 10.0f, loanId);
+            checkAccrualTransaction(LocalDate.of(2023, 3, 3), 0.0, 10.0, 10.0, loanId);
 
             // Set business date
-            LocalDate futureDate = LocalDate.of(2023, 3, 4);
-            final String nextAccrualRunDate = dateFormatter.format(futureDate);
-
-            BusinessDateHelper.updateBusinessDate(BusinessDateType.BUSINESS_DATE, futureDate);
-
+            updateBusinessDate("04 March 2023");
             // make repayment
-            final PostLoansLoanIdTransactionsResponse repaymentTransaction_1 = loanTransactionHelper.makeLoanRepayment(loanExternalIdStr,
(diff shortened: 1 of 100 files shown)


integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/modules/LoanTestValidators.java
```diff
@@ -96,8 +96,9 @@ public static void verifyTransactions(GetLoansLoanIdResponse loanDetails, LoanTe
                             && Objects.equals(Utils.getDoubleValue(item.getFeeChargesPortion()), tr.feePortion)
                             && Objects.equals(Utils.getDoubleValue(item.getPenaltyChargesPortion()), tr.penaltyPortion)
                             && Objects.equals(Utils.getDoubleValue(item.getOverpaymentPortion()), tr.overpaymentPortion)
-                            && Objects.equals(Utils.getDoubleValue(item.getUnrecognizedIncomePortion()), tr.unrecognizedPortion)
-                            && (tr.reversed == null || Objects.equals(item.getManuallyReversed(), tr.reversed)));
+                            && Objects.equals(Utils.getDoubleValue(item.getUnrecognizedIncomePortion()), tr.unrecognizedPortion));
```
budaidev: I don't think your thought about this was wrong, since you followed the pattern we had, but I think it would be better to keep the possibility to be able to distinguish the reverse transactions from the others. 
In certain cases it could contain valuable information

budaidev (dismissed):
I don't see any larger issues, it looks good.

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/modules/LoanTestValidators.java
DeathGun44: Thanks! I actually had the reversed check in there originally, then removed it to match the base verifyTransactions(Long, TransactionExt...) in BaseLoanIntegrationTest, which doesn't check reversed either ,I wanted the Feign version to behave exactly like the one it's replacing so the migration stays a faithful port.

There's also a practical reason the base skips it: TransactionExt.reversed defaults to false, but the API returns manuallyReversed = null for non-reversed transactions, so a plain equals check would fail every normal transaction. Happy to add it back with null-handling if you'd like!

DeathGun44:
Hi @adamsaghy can you please take a look at this one ,when you have a moment!

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/modules/LoanTestValidators.java
adamsaghy: @DeathGun44 I think we should have the check both places.

DeathGun44: Done!

DeathGun44:
Rebased ,also removed feignhttp calls as per the reviews on https://github.com/apache/fineract/pull/6194 and updated pr description reflecting the current state

DeathGun44:
test seems flaky, should pass on rerun

DeathGun44:
@adamsaghy please take a look when you have a moment!

budaidev:
please resolve the conflict

adamsaghy:
@DeathGun44 Sorry for being slow. Let me review it now

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/ClientChargeCommandsApi.java
```diff
@@ -0,0 +1,104 @@
...
+ * what {@code ClientChargeRoundingTest} exercises - but {@code ClientChargesApiResourceSwagger} declares it as an
+ * {@code Integer}, so the generated {@code PostClientsClientIdChargesRequest} cannot carry 19.876 or 0.55.
+ *
+ * <p>
+ * The Swagger DTO is simply wrong here and correcting it is the right fix, but it is a breaking OpenAPI change
+ * (swagger-brake R010, {@code integer} to {@code number}) and belongs in its own change rather than in a test
+ * migration. Until then this keeps the call typed end to end, binding the response to the generated model.
+ */
+@Headers({ "Accept: application/json", "Content-Type: application/json" })
+public interface ClientChargeCommandsApi {
```
adamsaghy: Please rather change the incorrect swagger definition than workaround with this.

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/FeignUserHelper.java
```diff
@@ -0,0 +1,83 @@
...
+    private static FineractFeignClient simpleUserWithoutBypassPermissionClient;
+
+    private FeignUserHelper() {}
+
+    /**
+     * Lazily creates a simple user whose role lacks the loan-checker bypass permission (but can read and repay loans
+     * and manage loan reschedules) and returns a Feign client authenticated as that user. Mirrors
+     * {@code UserHelper.getSimpleUserWithoutBypassPermission}. The user and client are created once per JVM and reused.
+     */
+    public static synchronized FineractFeignClient getSimpleUserWithoutBypassPermissionClient() {
```
adamsaghy: any reason for marking it as `synchronized`?

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/LoanChargeCommandsApi.java
```diff
@@ -0,0 +1,115 @@
...
+     * <p>
+     * The generated {@code executeLoanCharge} always serialises at least <code>{}</code>, which this endpoint answers
+     * with a 500. The pre-migration test sent an empty body here, so this reproduces it exactly. A request model cannot
+     * express "no body at all", which is why this needs its own declaration rather than a Swagger DTO change.
+     *
+     * <p>
+     * Waiving a single installment of an instalment fee <em>does</em> take a body and goes through the generated
+     * client's typed {@code waiveLoanCharge} instead.
+     */
+    @RequestLine("POST /v1/loans/{loanId}/charges/{loanChargeId}?command=waive")
```
adamsaghy: I am not sure about this... not setting request should send it as `null`

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/LoanProductCommandsApi.java
```diff
@@ -0,0 +1,60 @@
...
+ * properties are {@code @JsonInclude(USE_DEFAULTS)} and {@code ObjectMapperFactory} sets {@code NON_NULL} globally, so
+ * a null field is dropped from the body and the server never takes the clearing branch.
+ *
+ * <p>
+ * Marking the field {@code nullable} in the Swagger DTO does not help - the generator emits a plain
+ * {@code @Nullable Long}, not a {@code JsonNullable}, so serialisation is unchanged. A property-level
+ * {@code @JsonInclude(ALWAYS)} is what actually overrides the mapper default, hence this request model.
+ */
+@Headers({ "Accept: application/json", "Content-Type: application/json" })
+public interface LoanProductCommandsApi {
```
adamsaghy, line 47: I wonder whether there is a way to handle such scenarios gracefully...

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanChargeTaxIntegrationTest.java
```diff
@@ -352,202 +286,137 @@ public void testTaxGroup_appearsInRetrieveAllList() {
...
             assertEquals(100.0, charge2.getAmount(), 0.01, "Loan 2 charge amount must remain 100; tax stored separately");
         });
     }
 
     // -----------------------------------------------------------------------
-    // 10. Cash-based accounting: fee with tax splits income vs. tax liability
+    // 9. Full charge list on a loan reflects tax-inflated amounts
     // -----------------------------------------------------------------------
 
-    /**
```
adamsaghy, line 398: Any reason to remove the comments?

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanChargeTaxIntegrationTest.java
```diff
@@ -212,33 +180,26 @@ public void testLoanChargeAmount_remainsUnchanged_whenTaxGroupIsConfigured() {
     // 3. Happy path – multiple tax components
     // -----------------------------------------------------------------------
 
-    /**
```
adamsaghy, line 215: any reason to remove the comments?

adamsaghy (changes requested):
@DeathGun44 Ultimately it looks good. Great job (huge PR..). 

Can you please review my concerns and advise on them?

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/ClientChargeCommandsApi.java
DeathGun44: Agreed, the swagger is wrong - amount should be number, same as the loan version of this field.
But that change fails our CI. swagger-brake flags it:
 POST /v1/clients/{clientId}/charges
       amount type was changed from integer to number
  
I think it's a false alarm - widening a request field to number can't break any client, swagger-brake just ignores direction. 
should i fix it here and accept the red check if it arrises?

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/FeignUserHelper.java
DeathGun44: no reason ,looking into it i found our tests don't run in parallel, and the original UserHelper method isn't synchronized either. Removing it!

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/ClientChargeCommandsApi.java
adamsaghy: its fine to break it if we are fixing a bug. That check is just to highlight there was such change, but we can still accept the change...it just tells, lets double check ;) 

TLDR: lets fix it.

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/LoanChargeCommandsApi.java
DeathGun44: You're right, this isn't needed. I tested against a live server - waiving a whole charge with a {} body returns 200 (flat fee, percentage fee, penalty). I'd noted a 500 earlier but can't reproduce it; the server never reads the body for a non-installment charge anyway.
Removing it and using the generated client!

Small note: passing null doesn't work - Feign throws Body parameter N was null before sending. An empty request object is how it sends {}.

I'd like to keep the second method (German "50,05") - that field is a JSON number in the spec, and the test exists to check comma parsing. Let me know if you disagree?

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/LoanProductCommandsApi.java
DeathGun44: Yes, and my comment in the file was likely wrong.
We already do this elsewhere: DelinquencyRange.id is nullable and comes out as JsonNullable<Long>.

Checked both risks: existing callers still compile (normal getter/setter stay), and swagger-brake reports no breaking change.
So I'll mark the field nullable and delete this file. Good catch.

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/ClientChargeCommandsApi.java
DeathGun44: Fixed! PostClientsClientIdChargesRequest.amount is now BigDecimal and the workaround interface is deleted — ClientChargeRoundingTest posts 19.875 and asserts 19.88 straight through the generated client.
That retype is the only non-additive change in the PR. Across the nine touched production files the diff is additive apart from three lines: this one, LoanTransactionsApiResourceSwagger going public so InteropApiResource can reference it, and a copy-paste fix to the loanrepayment operation summary (develop had it as "Disburse Loan by Account Id"). So R010 on amount should be the only thing the API check reports.

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/FeignUserHelper.java
DeathGun44: Removed!

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/LoanChargeCommandsApi.java
DeathGun44: Fixed!

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/LoanProductCommandsApi.java
DeathGun44: Correction - I checked the generated sources and my precedent only half holds. DelinquencyRange.id is JsonNullable<Long> in fineract-client-feign, but fineract-client (Retrofit) emits a plain Long with @javax.annotation.Nullable, and integration-tests declares fineract-client first (dependencies.gradle:40), so the Retrofit model wins the name clash here. Marking the field nullable is still right for SDK consumers, but it can't reach this test: .delinquencyBucketId(null) would still serialise to {} and the bucket would silently never clear. So I've kept the minimal interface with that measurement in its javadoc - happy to add the spec change too, but the workaround has to stay either way.

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanChargeTaxIntegrationTest.java
DeathGun44: Restored.

DeathGun44: The file now diffs as a pure migration - zero comment lines added or removed against develop, original section ordering included. I swept the other migrated files too; what's left are labels whose statement no longer exists, e.g. in BatchApiTest // Create a createClient Request sat over a single BatchHelper.createClientRequest(...) call that's now a request-id constant plus a BatchRequestBuilders entry.

DeathGun44:
@adamsaghy please take a look when you have a moment!

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanAccrualTransactionOnChargeSubmittedDateTest.java
```diff
@@ -22,892 +22,456 @@
...
-                    new PutGlobalConfigurationsRequest().stringValue("submitted-date"));
+            updateBusinessDate("04 March 2023");
+            makeLoanRepayment(loanExternalIdStr, repaymentRequest("4 March 2023", 100.0));
             // Loan ExternalId
-            String loanExternalIdStr = UUID.randomUUID().toString();
+            addFeeCharge(loanId, "21 March 2023", 10.0);
 
+            runPeriodicAccrualAccounting("04 March 2023");
             // Client and Loan account creation
-
```
adamsaghy, line 113: Any reason to remove half of the test steps?

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanAccrualTransactionOnChargeSubmittedDateTest.java
```diff
@@ -22,892 +22,456 @@
...
-            final Integer clientId = clientHelper.createClient(ClientHelper.defaultClientCreationRequest()).getClientId().intValue();
-            final GetLoanProductsProductIdResponse getLoanProductsProductResponse = createLoanProduct(loanTransactionHelper, assetAccount,
-                    incomeAccount, expenseAccount, overpaymentAccount);
-            assertNotNull(getLoanProductsProductResponse);
-
-            final Integer loanId = createLoanAccount(clientId, getLoanProductsProductResponse.getId(), loanExternalIdStr);
-
             // Add Charge Penalty
-            Integer penalty = ChargesHelper.createCharges(requestSpec, responseSpec,
```
adamsaghy, line 416: same

adamsaghy (changes requested):
Please review the changes in the test cases. I have noticed multiple places many of the test steps got removed....

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanAccrualTransactionOnChargeSubmittedDateTest.java
DeathGun44: Nothing was removed here - that comment is just in the wrong place, and I'm sorry it read that way.

When I restored the comments after the last review, they were re-inserted mechanically and landed
away from the code they describe. So `// Client and Loan account creation` ended up with no statement
under it, which makes it look like the step was deleted.

All the steps are still in the test, a few lines higher!

DeathGun44: I have moved every comment in this file back above the statement it describes. The file now has
exactly the same number of comment lines as before the migration (92)

integration-tests/src/test/java/org/apache/fineract/integrationtests/LoanAccrualTransactionOnChargeSubmittedDateTest.java
DeathGun44: Same cause, same fix. The charge steps are still there - they moved into two small helpers so the
same two lines are not repeated in all 7 tests!

DeathGun44:
@adamsaghy rebased again due to recent shift in develop!

adamsaghy:
@DeathGun44 Please rebase one last time and i will merge it

DeathGun44:
@adamsaghy Done!

DeathGun44:
the failed test seems flaky

adamsaghy (approved):
LGTM

PR #6165, opened 2026-07-23
FINERACT-2702: Make client search case-insensitive

Samer-Melhem-FOO:
## Summary
- `GET /v2/clients/search` matched `accountNumber`, `displayName`, `externalId`, `mobileNo`, and identifier `documentKey` using case-sensitive `LIKE` predicates, so a search term differing only in letter case from the stored value would not match.
- `SearchingClientRepositoryImpl.searchByText` now lower-cases the incoming search text and wraps each compared field with `CriteriaBuilder.lower(...)` (via `cb.function("LOWER", ...)` for the `externalId` path), making the match case-insensitive.

## Test plan
- [x] `./gradlew :fineract-core:compileJava` — builds clean
- [x] `./gradlew :fineract-core:compileTestJava :fineract-provider:compileTestJava` — builds clean
- [ ] Manual verification: create a client with a mixed-case `displayName`/`accountNumber`, call `GET /v2/clients/search?searchText=<lowercase-substring>` and confirm the client is now returned

https://issues.apache.org/jira/browse/FINERACT-2702

### DIFF
--- a/fineract-core/src/main/java/org/apache/fineract/portfolio/client/domain/search/SearchingClientRepositoryImpl.java
+++ b/fineract-core/src/main/java/org/apache/fineract/portfolio/client/domain/search/SearchingClientRepositoryImpl.java
@@ -74,10 +74,12 @@ public class SearchingClientRepositoryImpl implements SearchingClientRepository 
             List<Predicate> predicates = new ArrayList<>();
             predicates.add(cb.like(o.get("hierarchy"), hierarchyLikeValue));
 
-            String searchLikeValue = "%" + searchText + "%";
-            predicates.add(cb.or(cb.like(r.get("accountNumber"), searchLikeValue), cb.like(r.get("displayName"), searchLikeValue),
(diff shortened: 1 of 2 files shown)


adamsaghy (changes requested):
Please cover with test cases

Samer-Melhem-FOO:
@adamsaghy I've pushed a fix addressing the test coverage request:

- Added 3 new test cases to `ClientSearchTest.java` covering case-insensitive search by display name, external ID, and client identifier document key.

Also looked into the 2 failing CI checks (`run-e2e-tests` Shard 6 and `run-integrations-tests-on-mysql` Shard 4) — both are unrelated to this change (Working Capital Loan charges/delinquency tests failing on timeouts/500s, nothing to do with client search). checkstyle and spotless are green.

Please take another look when you get a chance.

IOhacker (approved):
LGTM

IOhacker (approved):

adamsaghy (approved):
LGTM

PR #6069, opened 2026-06-30
FINERACT-2621: Upgrade dependencies to Spring Boot 4.1

avivijay19:
## Description

Upgrades the platform to Spring Boot 4.1, and with it Spring Framework 7, Spring Batch 6, Jackson 3, EclipseLink 5, Jersey 4 and JUnit 6.

Most of the 363 changed files are mechanical: package relocations (`org.springframework.batch.core.*` split into `.job` / `.step` / `.infrastructure.*`, `com.fasterxml.jackson` → `tools.jackson`, `org.springframework.lang` → `org.jspecify`, Boot's autoconfigure packages) and API renames. The sections below cover everything that is **not** mechanical — please concentrate review there.

### Dependencies

| | From | To |
| --- | --- | --- |
| Spring Boot | 3.5.15 | 4.1.0 |
| Spring Framework | 6.2 | 7.0 (via BOM) |
| Spring Security | 6.5 | 7.0 (via BOM) |
| Spring Batch | 5.x | 6.0.4 |
| Jackson | 2.x | 3.x (`tools.jackson`) |
| EclipseLink | 4.0.9 | 5.0.1 |
| Jersey | 3.1.11 | 4.0.2 |
| JUnit | 5.14.4 | 6.0.3 |
| springdoc-openapi | 2.8.17 | 3.0.3 |
| REST Assured | 5.5.7 | 6.0.1 |

Notes:
- **Jackson 2 is retained** (`jackson-bom:2.22.1`) for the Jersey REST layer, which stays on Jackson 2 deliberately. Boot 4 no longer auto-configures a Jackson 2 HTTP message converter, so `JerseyJacksonObjectArgumentHandler` binds to the Jersey `ObjectMapper` directly.
- `fineract-client` is pinned back to JUnit 5, because the generated SDK targets Java 8 and JUnit 6 requires JVM 17+.
- `spring-core` is pinned to 6.2.18 **on the buildscript classpath only** — the hierynomus license plugin's transitive `license-maven-plugin` needs a `PropertyPlaceholderHelper` constructor removed in Framework 7. No runtime effect; remove once that plugin ships a compatible release.
- `resilience4j-spring-boot4` is pinned explicitly; the resilience4j BOM does not manage the Boot 4 starter variant.

---

### Database: no migration, restart in any order

**This PR contains no Liquibase changelog.** Spring Batch 6 is made to work against the existing schema instead, so a Batch 5 and a Batch 6 instance can share one database and instances may be restarted in any order.

[...]

PR #6175, opened 2026-07-25
FINERACT-2621: Add CI-only diagnostics for missing external events in e2e

avivijay19:
## Description

Extracted from #6069, where review asked for these e2e changes to be raised in a separate PR since they are not part of the Spring Boot upgrade itself.

- When an event assertion times out, the step now queries the internal external events endpoint and logs whether the event was never persisted to `m_external_event` or was persisted but never delivered to the broker. A timeout alone cannot tell these two stages apart, which makes intermittent CI failures hard to attribute. Controlled by `EVENT_FAILURE_DIAGNOSTICS_ENABLED`, off by default and set only in the e2e workflow, and it only runs after an assertion has already failed, so passing runs are unaffected.
- `loanAccountDataV1Check` consumes the latest received event of the asserted type, which can be a stale predecessor when earlier steps raised events of the same type for the same loan. The verification is now retried so each attempt consumes the next event as it arrives.

## Checklist

### DIFF
--- a/fineract-e2e-tests-core/src/test/java/org/apache/fineract/test/messaging/EventFailureDiagnostics.java
+++ b/fineract-e2e-tests-core/src/test/java/org/apache/fineract/test/messaging/EventFailureDiagnostics.java
@@ -1,0 +1,132 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.test.messaging;
+
+import static org.apache.fineract.client.feign.util.FeignCalls.ok;
+
+import com.google.gson.Gson;
+import com.google.gson.reflect.TypeToken;
+import java.lang.reflect.Type;
+import java.util.List;
+import java.util.Map;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.fineract.client.feign.FineractFeignClient;
+import org.apache.fineract.test.messaging.config.EventProperties;
+import org.apache.fineract.test.messaging.store.EventStore;
+import org.springframework.stereotype.Component;
+
+/**
+ * Explains why an event assertion timed out.
+ *
+ * <p>
+ * An event only reaches the test through two independent stages: it must be persisted into {@code m_external_event} by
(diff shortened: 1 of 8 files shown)


fineract-e2e-tests-core/src/test/java/org/apache/fineract/test/messaging/EventFailureDiagnostics.java
```diff
@@ -0,0 +1,129 @@
...
+            return;
+        }
+        try {
+            final String rawJson = ok(() -> fineractClient.internalExternalEvents().getExternalEvents(eventName, aggregateRootId));
+            final List<Map<String, Object>> persisted = GSON.fromJson(rawJson, EVENT_LIST_TYPE);
+
+            log.error("=== EVENT DIAGNOSTICS: {} (aggregateRootId={}, expected payLoad.id={}) ===", eventName, aggregateRootId,
+                    expectedPayloadId);
+
+            if (persisted == null || persisted.isEmpty()) {
```
Aman-Mittal, line 83: like this may be better 

if (persisted == null || persisted.isEmpty()) {
    log.error("""
            VERDICT: NOT PERSISTED - server has no '{}' event for aggregateRootId={}.
            The event was never written to m_external_event, so the loss is upstream of the sender job
            (event not raised, or discarded before/at transaction commit).
            """,
            eventName, aggregateRootId);
} else if (containsPayloadId(persisted, expectedPayloadId)) {
    log.error("""
            VERDICT: PERSISTED BUT NOT DELIVERED - server persisted a '{}' event with payLoad.id={},
            but it never reached the test's JMS listener within {}ms.
            The loss is in delivery (sender job selection, broker publish, or test-side consumption).
            """,
            eventName, expectedPayloadId, timeoutMs);
}

fineract-e2e-tests-core/src/test/java/org/apache/fineract/test/messaging/event/EventCheckHelper.java
```diff
@@ -184,6 +184,24 @@ private void loanAccountDataV1Check(Class<? extends AbstractLoanEvent> eventClaz
         GetLoansLoanIdResponse body = ok(() -> fineractClient.loans().retrieveOneLoan(loanId,
                 Map.of("staffInSelectedOfficeOnly", false, "associations", "all", "exclude", "", "fields", "")));
 
+        // Earlier steps may have raised events of the same type for this loan; assertEvent consumes the
+        // latest one received, which is a stale predecessor when the newest event is still in transit.
+        // Each attempt consumes one event, so retrying waits for the next one to arrive.
+        int attemptsLeft = 3;
+        while (true) {
```
Aman-Mittal: using for loop may be more cleaner compare to while loop and extract to helper function

fineract-e2e-tests-core/src/test/java/org/apache/fineract/test/messaging/EventFailureDiagnostics.java
avivijay19: Done, switched to text blocks. Kept the third `else` branch as well — that's the case where the event is persisted but under a different `payLoad.id`.

fineract-e2e-tests-core/src/test/java/org/apache/fineract/test/messaging/event/EventCheckHelper.java
avivijay19: Done, extracted to `retryOnAssertionError(maxAttempts, Runnable)` with a for loop.

Aman-Mittal (dismissed):
LGTM

adamsaghy (approved):
LGTM

PR #6176, opened 2026-07-25
FINERACT-2621: Fix SchedulerJobsTest business date job race

avivijay19:
## Description

Replaces the original approach in this PR (retrying the business date increase in a `REQUIRES_NEW` transaction), which @adamsaghy pushed back on. He was right: a new transaction decouples the date write from the job status, and a rerun after a false FAILED would increase the date twice — worse than the duplicate key it avoided.

The problem was in the test, not in production code. `SchedulerJobsTest.testTriggeringManualExecutionOfAllSchedulerJobs` enables business date and then runs every scheduled job in parallel. `INCREASE_BUSINESS_DATE_BY_1_DAY` and `INCREASE_COB_DATE_BY_1_DAY` both write `m_business_date`, so they race to insert the initial row while none exists yet, and the business date job also writes the COB date row when COB date adjustment is enabled. One of them then fails on `uq_business_date_type`, which surfaced as an intermittent failure on a MySQL CI shard of #6069.

Both date jobs now run on their own, after the parallel batch. They are excluded rather than the rows being pre-seeded, because seeding only closes the insert race — `BusinessDate` carries an `@Version` field, so two jobs writing the COB date row could still collide on optimistic locking.

## Checklist

### DIFF
--- a/integration-tests/src/test/java/org/apache/fineract/integrationtests/SchedulerJobsTest.java
+++ b/integration-tests/src/test/java/org/apache/fineract/integrationtests/SchedulerJobsTest.java
@@ -20,6 +20,7 @@ package org.apache.fineract.integrationtests;
 
 import static org.awaitility.Awaitility.await;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import io.restassured.builder.RequestSpecBuilder;
 import io.restassured.http.ContentType;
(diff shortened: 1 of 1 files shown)


AshharAhmadKhan:
@avivijay19 please check the failing log.

fineract-core/src/main/java/org/apache/fineract/infrastructure/businessdate/service/BusinessDateWritePlatformServiceImpl.java
```diff
@@ -52,6 +54,7 @@ public BusinessDateDTO updateBusinessDate(BusinessDateDTO businessDateDto) {
     }
 
     @Override
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
```
adamsaghy: I dont think this should be in new transaction. This way we detach the result from the job status and business date can be increased while the job can still fail for any reasons...

adamsaghy (changes requested):
I dont think it is a problem we need to address. I see no problem if one of them fails... they can be rerun based on the job status.

adamsaghy (dismissed):
LGTM

adamsaghy (approved):
LGTM

PR #6180, opened 2026-07-26
FINERACT-2708: Fix duplicate defined name 500 in bulk-import transaction templates

oluexpert99:
- Downloading the savings, recurring-deposit or fixed-deposit transaction template
        failed with HTTP 500 (IllegalArgumentException: The workbook already contains this
        name: Account_<client>_<id>_) on tenants where one client holds more than one
        account. The template is unusable for the whole tenant; there is no API workaround.
      - setNames builds one Account_<client>_<id>_ defined name per client, collecting
        clients by walking the account lookup table and starting a new run whenever the
        client name changes. That run detection uses a case-sensitive String.equals, but the
        table it walks was just sorted with SavingsAccountData.ClientNameComparator, which
        compares the names case-INSENSITIVELY (both upper-cased). Two clients whose names
        differ only in case therefore compare equal to the sort and may be interleaved, which
        splits one client's accounts into two runs; the second run re-adds the same client
        and the per-client loop then asks POI to create a defined name that already exists.
      - Guard the collection so a client name that has already been recorded is not added
        again, mirroring the guard LoanRepaymentWorkbookPopulator.setNames already has. Each
        client contributes exactly one defined name, whatever its account count; nothing
        changes for tenants that never hit the collision.
      - Apply the same guard to all three affected populators
        (SavingsTransactionsWorkbookPopulator, RecurringDepositTransactionWorkbookPopulator,
        FixedDepositTransactionWorkbookPopulator), which carry identical copies of the loop.
      - Add a unit test per populator: three accounts (client 18, client 19 whose name
        differs from 18's only in case, then client 18 again) reproduce the split-run
        ordering. Each test fails on the unfixed code with the exact production exception and
        asserts one defined name per client, still distinct when upper-cased since Excel
        defined names are case-insensitive.
## Description

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/bulkimport/populator/fixeddeposits/FixedDepositTransactionWorkbookPopulator.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/bulkimport/populator/fixeddeposits/FixedDepositTransactionWorkbookPopulator.java
@@ -193,8 +193,12 @@ public class FixedDepositTransactionWorkbookPopulator extends AbstractWorkbookPo
                 startIndex = i + 2;
                 clientName = savingsAccounts.get(i).getClientName();
                 clientId = savingsAccounts.get(i).getClientId();
-                clientsWithActiveSavings.add(clientName);
-                clientIdsWithActiveSavings.add(clientId);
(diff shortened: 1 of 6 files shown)


AshharAhmadKhan:
@oluexpert99 please check the failing log.

adamsaghy (approved):
LGTM

PR #6181, opened 2026-07-26
FINERACT-2709: Fix chart-of-accounts template guarding the wrong variable

oluexpert99:
- ChartOfAccountsWorkbook.setNames builds a Tags named range per account type from the
        lookup-index entry, but null-checks the map it read the entry from rather than the
        entry itself: it reads Integer[] tagValueBeginEndIndexes from
        accountTypeToBeginEndIndexesofAccountNames, then tests the map for null and
        dereferences tagValueBeginEndIndexes[0] and [1] inside the block.
      - The map is an instance field that was already dereferenced on the line above (.get(i)),
        so it is never null at that point: the guard is always true and protects nothing, while
        reading as though it protects the array access. The value that can be null is the
        per-iteration array, from a Map.get miss.
      - The sibling block immediately below, which builds the AccountName named range, shows the
        intended pattern and guards the array it dereferences. The Tags block appears to be a
        copy of it with the guarded variable left unchanged.
      - Guard tagValueBeginEndIndexes instead, so a missing lookup-index entry skips the Tags
        named range rather than dereferencing null, matching the sibling block.
      - Note on reachability and on the test: with well-formed GL-account data the null branch
        cannot be reached. accountTypesNoDuplicatesList and the index map are both derived from
        the same glAccounts list, and setLookupTable only skips a put -- leaving a gap -- for an
        account type whose account-name list is empty, which cannot happen for a type that came
        from an actual account. This is therefore a dead guard and a latent NPE rather than a
        reproducible failure, and the added ChartOfAccountsWorkbookTest is a characterisation
        test: it pins the working path (one Tags_<type> range per account type, alongside the
        AccountName_<type> range) and would catch an incorrect fix, but it passes both with and
        without this change, since no input reachable through the public API can produce the
        null entry. Reaching the null branch would require reflecting into the populator's
        internal map, which would test an implementation detail rather than behaviour.

## Description

## Checklist

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/bulkimport/populator/chartofaccounts/ChartOfAccountsWorkbook.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/bulkimport/populator/chartofaccounts/ChartOfAccountsWorkbook.java
@@ -158,7 +158,7 @@ public class ChartOfAccountsWorkbook extends AbstractWorkbookPopulator {
         for (Integer i = 0; i < accountTypesNoDuplicatesList.size(); i++) {
             Name tags = chartOfAccountsWorkbook.createName();
             Integer[] tagValueBeginEndIndexes = accountTypeToBeginEndIndexesofAccountNames.get(i);
-            if (accountTypeToBeginEndIndexesofAccountNames != null) {
+            if (tagValueBeginEndIndexes != null) {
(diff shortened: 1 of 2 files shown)


AshharAhmadKhan:
@oluexpert99 please check the failing log.

adamsaghy (approved):
LGTM

PR #6188, opened 2026-07-27
FINERACT-2714: Add ability to exclude closed clients from client search results

Samer-Melhem-FOO:
## Description

The `POST /v2/clients/search` endpoint returns matching clients regardless of status, including closed ones, with no way for the caller to filter them out.

This adds an optional `excludeClosed` boolean field to the `ClientTextSearch` search request. When `excludeClosed=true`, the search excludes clients whose status is `CLOSED` (via a `status != CLOSED` predicate in `SearchingClientRepositoryImpl`). When omitted or `false`, behavior is unchanged — fully backward compatible for existing callers.

https://issues.apache.org/jira/browse/FINERACT-2714

## Checklist

- [x] Write the commit message as per our guidelines
- [x] Acknowledge that we will not review PRs that are not passing the build ("green") - it is your responsibility to get a proposed PR to pass the build, not primarily the project's maintainers.
- [x] Follow our coding conventions.
- [ ] Add required Swagger annotation and update API documentation at fineract-provider/src/main/resources/static/legacy-docs/apiLive.htm with details of any API changes (not applicable — no new endpoint, just an optional field on an existing request DTO, already reflected via existing annotations)
- [x] This PR must not be a "code dump".

### DIFF
--- a/fineract-core/src/main/java/org/apache/fineract/portfolio/client/domain/search/SearchingClientRepository.java
+++ b/fineract-core/src/main/java/org/apache/fineract/portfolio/client/domain/search/SearchingClientRepository.java
@@ -23,5 +23,5 @@ import org.springframework.data.domain.Pageable;
 
 public interface SearchingClientRepository {
 
-    Page<SearchedClient> searchByText(String searchText, Pageable pageable, String officeHierarchy);
+    Page<SearchedClient> searchByText(String searchText, Pageable pageable, String officeHierarchy, boolean excludeClosed);
 }
--- a/fineract-core/src/main/java/org/apache/fineract/portfolio/client/domain/search/SearchingClientRepositoryImpl.java
+++ b/fineract-core/src/main/java/org/apache/fineract/portfolio/client/domain/search/SearchingClientRepositoryImpl.java
@@ -36,6 +36,7 @@ import org.apache.fineract.infrastructure.core.jpa.CriteriaQueryFactory;
 import org.apache.fineract.organisation.office.domain.Office;
 import org.apache.fineract.portfolio.client.domain.Client;
 import org.apache.fineract.portfolio.client.domain.ClientIdentifier;
+import org.apache.fineract.portfolio.client.domain.ClientStatus;
 import org.springframework.data.domain.Page;
 import org.springframework.data.domain.Pageable;
 import org.springframework.data.jpa.domain.Specification;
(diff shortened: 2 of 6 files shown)


IOhacker (changes requested):
@Samer-Melhem-FOO why no to exted to other client status?

Samer-Melhem-FOO:
Thanks for the review! The scope here is intentionally narrow: this maps to a single UI checkbox — "Exclude closed accounts" — on the client search screen, not a general-purpose client-status filter.

The intent is just to let users hide closed clients from search results by default, since closed accounts otherwise clutter results for users who are typically looking for active/actionable clients. It isn't meant to generalize into filtering by arbitrary client status (pending, rejected, withdrawn, etc.) — that would be a bigger, separate feature with its own UI/UX (e.g. a status multi-select) and API design considerations, and isn't what this fix was scoped to solve.

adamsaghy:
> Thanks for the review! The scope here is intentionally narrow: this maps to a single UI checkbox — "Exclude closed accounts" — on the client search screen, not a general-purpose client-status filter.
> 
> The intent is just to let users hide closed clients from search results by default, since closed accounts otherwise clutter results for users who are typically looking for active/actionable clients. It isn't meant to generalize into filtering by arbitrary client status (pending, rejected, withdrawn, etc.) — that would be a bigger, separate feature with its own UI/UX (e.g. a status multi-select) and API design considerations, and isn't what this fix was scoped to solve.

I agree with @IOhacker. From a backend perspective, filtering by status would make more sense. I suggest adding a “status” filter, where the caller can select the statuses they’re interested in.

Samer-Melhem-FOO:
Screenshot 2026-08-04 141147


Looking at the screenshot of the current UI (toggle off, "Show Closed Accounts"), all clients — including closed ones — are already returned by default. That doesn't quite line up with the generic status-filtering solution being proposed in this PR.

What we'd suggest instead: keep the generic status filtering, but flip the UI semantics — rename the toggle label from "Show Closed Accounts" to "Hide Closed Accounts". In its default (off) state, all clients continue to be returned as today, including closed ones. When toggled on, we hide closed accounts by adjusting the status filter param accordingly.

This keeps default behavior unchanged, gives users an explicit way to exclude closed accounts, and removes the mismatch between the switch's current label/intent and the filtering approach proposed here.

Posting this as pending your feedback — want to align on this before we land on a fully working solution.

@adamsaghy @IOhacker

adamsaghy:
> Screenshot 2026-08-04 141147
> Looking at the screenshot of the current UI (toggle off, "Show Closed Accounts"), all clients — including closed ones — are already returned by default. That doesn't quite line up with the generic status-filtering solution being proposed in this PR.
> 
> What we'd suggest instead: keep the generic status filtering, but flip the UI semantics — rename the toggle label from "Show Closed Accounts" to "Hide Closed Accounts". In its default (off) state, all clients continue to be returned as today, including closed ones. When toggled on, we hide closed accounts by adjusting the status filter param accordingly.
> 
> This keeps default behavior unchanged, gives users an explicit way to exclude closed accounts, and removes the mismatch between the switch's current label/intent and the filtering approach proposed here.
> 
> Posting this as pending your feedback — want to align on this before we land on a fully working solution.
> 
> @adamsaghy @IOhacker

To be honest, I would not change UI to fetch closed clients by default. Lets have it as an extra step.

From backend point of view, filtering by status is the better thing and no filtering means return all entries.

PR #6189, opened 2026-07-27
FINERACT-2715: Mark Activation Date as required in Client Entity import template

Samer-Melhem-FOO:
## Description

The Client Entity bulk import template did not visually indicate that "Activation Date" is a required column. `ClientEntityImportHandler` requires a valid Activation Date whenever the "Active" column is set to `true`, so imports failed on rows left blank without any indication from the template that the field was mandatory.

This changes the column header in `ClientEntityWorkbookPopulator` from `"Activation Date "` to `"Activation Date* "`, consistent with how other conditionally-required columns (e.g. `"Active*"`) are marked on the same sheet.

## JIRA

https://issues.apache.org/jira/browse/FINERACT-2715

## Test plan

- [x] `./gradlew :fineract-provider:compileJava` passes
- [x] Verified no existing test asserts the previous header string (integration test `ClientEntityWorkbookPopulatorTest` does not hardcode it)


Samer-Melhem-FOO:
The only failing check here is `run-e2e-tests / E2E Tests (Shard 4 of 20)`. Looking at the log, it fails at the `jibDockerBuild` step before any tests run, with:

```
I/O error for image [registry-1.docker.io/azul/zulu-openjdk-alpine]:
java.net.SocketException
Network is unreachable
```

This looks like a transient Docker Hub registry connectivity issue on the runner, unrelated to this PR's change (a single header-label fix in `ClientEntityWorkbookPopulator`). All other 19/20 E2E shards, all 15/15 DB-matrix shards (MariaDB/MySQL/PostgreSQL), and every other check (checkstyle, spotless, spotbugs, etc.) passed.

Could a maintainer re-run the failed job? I don't have the permissions to trigger a re-run myself. Thanks!

IOhacker (approved):

PR #6191, opened 2026-07-27
FINERACT-2717: Check all currency usages before allowing currency removal

Samer-Melhem-FOO:
## Description

`CurrencyWritePlatformServiceJpaRepositoryImpl#updateAllowedCurrencies` allows removing a currency from the organisation's allowed-currency list. The existing in-use check only looked at loan products, savings products, and charges, so a currency still referenced by actual loans, savings/share accounts, client transactions, account transfers, cashier transactions, or GL journal entries could be silently removed.

This broadens the check to a `JdbcTemplate`-based scan across all relevant tables (`m_product_loan`, `m_savings_product`, `m_share_product`, `m_charge`, `m_loan`, `m_savings_account`, `m_share_account`, `m_client_transaction`, `m_account_transfer_transaction`, `m_cashier_transactions`, `acc_gl_journal_entry`) and improves `CurrencyInUseException`'s message to be clearer to end users.

## JIRA

https://issues.apache.org/jira/browse/FINERACT-2717

## Test plan

- [x] `./gradlew :fineract-core:compileJava :fineract-provider:compileJava` passes
- [x] Added `CurrencyWritePlatformServiceJpaRepositoryImplTest` (no prior test existed for this class), covering:
  - currency removed successfully when unused
  - `CurrencyInUseException` thrown when the removed currency is still referenced
  - the usage check is skipped entirely for currencies that remain allowed
- [x] `./gradlew :fineract-provider:test --tests "...CurrencyWritePlatformServiceJpaRepositoryImplTest"` — all 3 tests pass

### DIFF
--- a/fineract-core/src/main/java/org/apache/fineract/organisation/monetary/exception/CurrencyInUseException.java
+++ b/fineract-core/src/main/java/org/apache/fineract/organisation/monetary/exception/CurrencyInUseException.java
@@ -27,7 +27,7 @@ public class CurrencyInUseException extends AbstractPlatformDomainRuleException 
 
     public CurrencyInUseException(final String currencyCode) {
         super("error.msg.currency.currencyCode.inUse",
-                "Currency cannot be removed because it is already used in existing transactions or accounts.", currencyCode);
+                "Cannot remove currency with identifier " + currencyCode + " while it is still in use", currencyCode);
     }
 
 }
--- a/fineract-core/src/main/java/org/apache/fineract/organisation/monetary/exception/CurrencyInUseException.java
+++ b/fineract-core/src/main/java/org/apache/fineract/organisation/monetary/exception/CurrencyInUseException.java
@@ -27,7 +27,7 @@ public class CurrencyInUseException extends AbstractPlatformDomainRuleException 
 
     public CurrencyInUseException(final String currencyCode) {
         super("error.msg.currency.currencyCode.inUse",
-                "Cannot remove currency with identifier " + currencyCode + " while it is still in use", currencyCode);
+                "Currency cannot be removed because it is already used in existing transactions or accounts.", currencyCode);
     }
 
 }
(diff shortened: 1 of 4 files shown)


Samer-Melhem-FOO:
Two checks are currently failing here, both unrelated to this PR's change (`CurrencyWritePlatformServiceJpaRepositoryImpl`, `CurrencyInUseException`, `OrganisationMonetaryConfiguration`, plus a new unit test — no Liquibase/schema changes):

- `run-integrations-tests-on-mariadb / MariaDB Tests (Shard 4 of 15)`: fails at `jibDockerBuild` with `java.net.SocketException: Network is unreachable` pulling `azul/zulu-openjdk-alpine` — the same transient Docker Hub registry issue seen on other PRs recently.
- `run-integrations-tests-on-mysql / MySQL Tests (Shard 5 of 15)`: the app container built and started fine (Spring context came up, DB connection pool started), but the tenant Liquibase migration stalled for ~22 minutes with no progress between two log lines, and the health-check retry loop eventually timed out. This PR doesn't touch any Liquibase changelogs, so this looks like CI-runner resource contention/slowness rather than something caused by the change.

Could a maintainer re-run the failed jobs? Thanks!

fineract-core/src/main/java/org/apache/fineract/organisation/monetary/exception/CurrencyInUseException.java
```diff
@@ -27,7 +27,7 @@ public class CurrencyInUseException extends AbstractPlatformDomainRuleException
 
     public CurrencyInUseException(final String currencyCode) {
         super("error.msg.currency.currencyCode.inUse",
-                "Cannot remove currency with identifier " + currencyCode + " while it is still in use", currencyCode);
+                "Currency cannot be removed because it is already used in existing transactions or accounts.", currencyCode);
```
adamsaghy: Any reason to change the exception message?

Samer-Melhem-FOO: No good reason — it wasn't needed for the actual fix and it drops the currency code from the human-readable message, which is a regression. Reverted to the original message.

adamsaghy (approved):
LGTM

PR #6194, opened 2026-07-28
FINERACT-2718: Migrate group center tests to feign

DeathGun44:
## What this PR does

Migrates the **group and center integration tests** from the old REST-assured HTTP client to the typed Feign SDK client

## Why

The old tests call the server with hand-written REST-assured requests and read untyped JSON. Moving them onto the generated Feign client gives typed requests and responses, so the tests are safer to change and consistent with the rest of the migrated suite.

## What changed

**Test classes migrated**
- `CenterIntegrationTest` 
- `GroupTest` 
- `GroupLoanIntegrationTest` 

**New typed helpers**:
- `FeignGroupHelper`, `FeignCenterHelper`, `FeignStaffHelper`, `FeignGlimHelper`

**Removed the old stopgap:**
- Deleted `FeignGroupCenterHelper` (a raw-HTTP helper from an earlier PR) and repointed its two users `LoanReschedulingWithinCenterTest` and `ExternalIdSupportIntegrationTest` - at the new typed helpers.

**Production change (minimal, additive):**
- Added the missing fields to `PostGroupsRequest` and `PostCentersRequest` in the Swagger DTOs so groups and centers can be created through the typed model. These are additive only - the API wire format is unchanged.

**Production change (minimal, additive):**
- Added the missing fields to the Swagger DTOs so groups, centers, clients and loans can be driven through the typed model: the creation fields, the `?command=...` request fields, the `changes` objects those commands return, the retrieval fields the tests assert on, and the group / GLIM fields on the loan application and approve bodies. Touches `GroupsApiResourceSwagger`, `CentersApiResourceSwagger`, `ClientsApiResourceSwagger` and `LoansApiResourceSwagger`.
- These are additive only - no field removed, no type or `required` changed, and the API wire format is unchanged. The backward-compatibility check against develop reports no breaking changes.

## Raw HTTP is gone

No `FeignRawHttpHelper` and no REST-assured remain in the migrated code - every request and response goes through the generated Feign models, including the `?command=...` endpoints and the GLIM application / approve / retrieval flow. Retrieval no longer needs the REST-assured-era `CenterDomain` POJO either.

One case is still bound by hand, but stays typed: `GET /centers` and `GET /groups` return a plain JSON array unless `paged=true`, and an operation can only declare one response schema (the paged one), so the generated `retrieveAllCenters` / `retrieveAllGroups` can't decode the array. Rather than document the endpoint incorrectly, those requests are bound to the generated page-item models through a small Feign interface, `NonPagedListingApi` - the same pattern as the existing `InternalExternalEventsApi`.

`paged=true` is not an equivalent substitute there: `retrievePagedAll` ignores `orphansOnly`, so the paged variant returns every group of the office. That is documented on the interface.

No assertion was weakened and no REST-assured was introduced.

## Checklist


Aman-Mittal (dismissed):
Code wise seems ok

adamsaghy (changes requested):
Why the `FeignRawHttpHelper` usages? We would like to avoid serializing back and forth json objects...

adamsaghy:
@DeathGun44 Please rebase

DeathGun44:
> Why the `FeignRawHttpHelper` usages? We would like to avoid serializing back and forth json objects...

Done!
One thing I couldn't express in the spec. GET /centers and GET /groups return a plain JSON array unless you pass paged=true, but an operation can only declare one response schema (the paged one), so the generated client can't decode the array. Rather than document the endpoint incorrectly, I bound those requests to the generated page-item models through a small Feign interface (NonPagedListingApi) - same pattern as the existing InternalExternalEventsApi. Still fully typed, no JSON parsing.

To be straight about it, only one of its three methods is really forced:
- listOrphanGroups has to stay non-paged, because retrievePagedAll ignores orphansOnly. I checked against a running server: after attaching every group of an office to a center, the non-paged call returns [] while the paged call returns both groups with centerId set. That looks like a server-side inconsistency - happy to raise a separate ticket if you agree.

- The two center listings could use paged=true. I kept them non-paged only so testListCenters still compares the two shapes rather than comparing a call to itself. Glad to switch if you'd rather.
 I also relaxed testListCenters to compare by id - neither branch applies an ORDER BY unless you ask for one, so the element-wise comparison was passing by luck.

 Rebased onto current develop

DeathGun44:
the test seems flaky, also updated the pr description

budaidev (approved):

adamsaghy (approved):
LGTM

PR #6195, opened 2026-07-28
FINERACT-2719: Add submittedOnDateTime to journal entry responses

Samer-Melhem-FOO:
## Description

The Journal Entries API only returns `submittedOnDate` (date-only), so any UI reading it always shows a time of `00:00:00`, even though the journal entry's actual creation timestamp is stored in `acc_gl_journal_entry.created_on_utc`.

This adds a new optional `submittedOnDateTime` field to `JournalEntryData` (Fineract-style `[year, month, day, hour, minute, second]` array), sourced from `created_on_utc` in `JournalEntryReadPlatformServiceImpl`. `submittedOnDate` is left unchanged for backward compatibility.

JIRA: https://issues.apache.org/jira/browse/FINERACT-2719

## Checklist

- [x] This PR must not be a "code dump". Large changes can be made in a branch, with assistance.

Note: no existing unit/integration test covers `JournalEntryReadPlatformServiceImpl`'s row-mapping logic (it's a plain JDBC `RowMapper`, not easily unit-testable in isolation), so none were added — happy to add integration-test coverage if a reviewer points to the right harness for it.

### DIFF
--- a/fineract-accounting/src/main/java/org/apache/fineract/accounting/journalentry/JournalEntryMapper.java
+++ b/fineract-accounting/src/main/java/org/apache/fineract/accounting/journalentry/JournalEntryMapper.java
@@ -46,6 +46,7 @@ public interface JournalEntryMapper {
     @Mapping(target = "entityType", source = "entityType", qualifiedByName = "entityType")
     @Mapping(target = "entityId", source = "entityId")
     @Mapping(target = "submittedOnDate", source = "submittedOnDate")
+    @Mapping(target = "submittedOnDateTime", ignore = true)
     @Mapping(target = "transactionId", source = "transactionId")
     @Mapping(target = "currency", source = "currencyCode")
     @Mapping(target = "manualEntry", source = "manualEntry")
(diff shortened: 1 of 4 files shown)


adamsaghy (changes requested):
I see no benefit of adding this field. If you want to have a sorted list, feel free to sort by created date time / and return that...

Samer-Melhem-FOO:
> I see no benefit of adding this field. If you want to have a sorted list, feel free to sort by created date time / and return that...

The benefit isn't sorting — it's the displayed value itself. `submittedOnDate` is date-only, so any UI that reads it always renders a time of `00:00:00`, even though the actual creation timestamp exists in `acc_gl_journal_entry.created_on_utc`. Without `submittedOnDateTime`, there's no way for a client to show *when* a journal entry was actually created, only *what day*.

`submittedOnDate` is left untouched for backward compatibility, so existing consumers (including anything sorting by it) are unaffected — this is purely additive for clients that want the real timestamp.

adamsaghy:
> > I see no benefit of adding this field. If you want to have a sorted list, feel free to sort by created date time / and return that...
> 
> The benefit isn't sorting — it's the displayed value itself. `submittedOnDate` is date-only, so any UI that reads it always renders a time of `00:00:00`, even though the actual creation timestamp exists in `acc_gl_journal_entry.created_on_utc`. Without `submittedOnDateTime`, there's no way for a client to show _when_ a journal entry was actually created, only _what day_.
> 
> `submittedOnDate` is left untouched for backward compatibility, so existing consumers (including anything sorting by it) are unaffected — this is purely additive for clients that want the real timestamp.

Disclose the created date and time, but there’s no need to introduce a new field when one already exists with the necessary data.

PR #6197, opened 2026-07-28
FINERACT-2720: Scope entity-to-entity mapping results to user office hierarchy

aya-abdallah-FOO:
## Description                                                                                                                                                                    
                                                                                                                                                                                    
  When the **All** option is selected in the Office dropdown on the Entity-to-Entity Mapping screen (System → Entity to Entity Mapping), the API returns mappings across all offices
  in the system, including branches the authenticated user has no access to. This violates office-level access restrictions.                                                        
                                                                                                                                                                                    
  **Root cause:** `retrieveEntityToEntityMappings` treats `fromId = 0` (the "All" sentinel) as a pass-through, so the SQL returns every row regardless of the caller's office       
  hierarchy.                                                                                                                                                                        
                                                                                                                                                                                    
  ## Fix                                                                                                                                                                            
                                                                                                                                                                                    
  When `fromId == 0` and the relation's `from_entity_type` is `1` (office), a recursive CTE (`WITH RECURSIVE office_descendants`) scopes results to only the authenticated user's   
  office and its descendants. Non-office relation types fall through to the existing query unchanged.                                                                               
                                                                                                                                                                                    
  ## Testing                                                                                                                                                                        
                                                                                                                                                                                    
  - A user in a non-root office sees only their office (and child offices) in the Entity-to-Entity Mapping dropdown — not "All" or unrelated branches.                              
  - `:fineract-provider:compileJava` and `:fineract-provider:compileTestJava` pass clean
  
                                                                                                                                                                                   
                                                                                                                                        
                                                                                                                                                                                    
  FINERACT-2720

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/entityaccess/service/FineractEntityAccessReadServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/entityaccess/service/FineractEntityAccessReadServiceImpl.java
@@ -204,6 +204,19 @@ public class FineractEntityAccessReadServiceImpl implements FineractEntityAccess
     @Override
     public Collection<FineractEntityToEntityMappingData> retrieveEntityToEntityMappings(Long mapId, Long fromId, Long toId) {
 
+        if (fromId == 0) {
+            final String fromEntityTypeSql = "SELECT er.from_entity_type FROM m_entity_relation er WHERE er.id = ?";
+            Integer fromEntityType = jdbcTemplate.queryForObject(fromEntityTypeSql, Integer.class, mapId);
+            if (fromEntityType != null && fromEntityType == 1) {
+                final AppUser currentUser = this.context.authenticatedUser();
+                final Long userOfficeId = currentUser.getOffice().getId();
+
+                EntityToEntityMapper entityToEntityMapper = new EntityToEntityMapper();
+                String officeScopedSql = entityToEntityMapper.schemaWithOfficeScopedFromId();
+                return this.jdbcTemplate.query(officeScopedSql, entityToEntityMapper, new Object[] { userOfficeId, mapId, toId, toId });
+            }
+        }
+
         EntityToEntityMapper entityToEntityMapper = new EntityToEntityMapper();
         String sql = entityToEntityMapper.schema();
         final Collection<FineractEntityToEntityMappingData> mapTypes = this.jdbcTemplate.query(sql, entityToEntityMapper,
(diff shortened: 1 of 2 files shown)


adamsaghy:
```
Execution failed for task ':fineract-provider:spotlessJavaCheck'.
      src/main/java/org/apache/fineract/infrastructure/entityaccess/service/FineractEntityAccessReadServiceImpl.java
          @@ -213,8 +213,7 @@
           
           ················EntityToEntityMapper·entityToEntityMapper·=·new·EntityToEntityMapper();
           ················String·officeScopedSql·=·entityToEntityMapper.schemaWithOfficeScopedFromId();
          -················return·this.jdbcTemplate.query(officeScopedSql,·entityToEntityMapper,
          -························new·Object[]·{·userOfficeId,·mapId,·toId,·toId·});
          +················return·this.jdbcTemplate.query(officeScopedSql,·entityToEntityMapper,·new·Object[]·{·userOfficeId,·mapId,·toId,·toId·});
           ············}
           ········}
           
  Run './gradlew :fineract-provider:spotlessApply' to fix these violations.

**Review by adamsaghy — CHANGES_REQUESTED** [review]
Can you add an integration / E2E test case to cover this?

**Comment by aya-abdallah-FOO** [comment]
> Can you add an integration / E2E test case to cover this?

@adamsaghy  Done! Added EntityToEntityMappingIntegrationTest with two integration tests covering this fix:                                                                                    
                                                                                                                                                                                    
  1. testOfficeUserSeesOnlyOwnOfficeMappingsWhenFromIdIsZero — Creates two sibling child offices (A and B), maps both to a loan product, then queries GET                           
  /entitytoentitymapping/{relId}/0/0 as a user belonging to Office A. Asserts that only Office A's mapping is returned and Office B's is not visible.                               
  2. testHeadOfficeUserSeesAllDescendantOfficeMappingsWhenFromIdIsZero — Same setup, queried as the superadmin (Head Office). Asserts both mappings are visible since Head Office is
  the root of the hierarchy.

**Comment by rymghosn** [comment]
@adamsaghy the requested changes (integration test coverage and the spotless formatting fix) have been addressed and the branch is rebased on latest `develop`. Could you take another look?

File: fineract-provider/src/main/java/org/apache/fineract/infrastructure/entityaccess/service/FineractEntityAccessReadServiceImpl.java [hunk 1]
```diff
@@ -204,6 +204,19 @@ public FineractEntityRelationData mapRow(final ResultSet rs, @SuppressWarnings("
     @Override
     public Collection<FineractEntityToEntityMappingData> retrieveEntityToEntityMappings(Long mapId, Long fromId, Long toId) {
 
+        if (fromId == 0) {
+            final String fromEntityTypeSql = "SELECT er.from_entity_type FROM m_entity_relation er WHERE er.id = ?";
```
adamsaghy, line 208: Can we use JQPL and repository here instead of native query?

adamsaghy:
@aya-abdallah-FOO Can you please rebase this PR? It seems failing for irrelevant reasons

PR #6199, opened 2026-07-29
FINERACT-2723: Rename ambiguous domain-enum getters in StandingInstructionData

Samer-Melhem-FOO:
## Description

`StandingInstructionData` declares several fields of type `EnumOptionData` (`fromAccountType`, `toAccountType`, `transferType`, `instructionType`, `recurrenceType`, `recurrenceFrequency`). The class also declares same-named getter methods that return a different, internal domain-enum type instead, e.g.:

```java
public PortfolioAccountType getFromAccountType() {
    return Optional.ofNullable(this.fromAccountType).map(e -> PortfolioAccountType.fromInt(e.getId().intValue())).orElse(null);
}
```

This is confusing to read and maintain: a field and its same-named "getter" disagree on type. **No behavior or API change**: `StandingInstructionApiResource` serializes this class via `DefaultToApiJsonSerializer`, which is Gson-based and reads the raw field directly regardless of these methods, so today's JSON output is unaffected either way — but the naming collision is still a latent correctness hazard (e.g. for anyone who later touches serialization, or adds a getter-based tool/library) worth removing.

**Change:** rename the six domain-enum helper methods to a distinct `*Enum` suffix (`getFromAccountTypeEnum()`, `getToAccountTypeEnum()`, `getTransferTypeEnum()`, `getInstructionTypeEnum()`, `getRecurrenceTypeEnum()`, `getRecurrenceFrequencyEnum()`) so they no longer shadow the field name, and add an explicit getter on each underlying field for consistency with the rest of the class. Internal callers (`StandingInstructionApiResource`, `ExecuteStandingInstructionsTasklet`) are updated to use the renamed methods, with added null-safety in `toTransferType()` since the domain-enum helper can return `null` when the underlying `EnumOptionData` field is `null`.

JIRA: https://issues.apache.org/jira/browse/FINERACT-2723

## Checklist

- [x] This PR must not be a "code dump". Large changes can be made in a branch, with assistance.

Note: pure internal rename with no behavior change, so no new test coverage was added; existing Standing Instruction tests continue to pass unmodified. No API documentation update needed since there is no API response change.

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/account/api/StandingInstructionApiResource.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/account/api/StandingInstructionApiResource.java
@@ -212,10 +212,10 @@ public class StandingInstructionApiResource {
             if (associationParameters.contains("template")) {
                 final StandingInstructionData templateData = standingInstructionReadPlatformService.retrieveTemplate(
                         standingInstructionData.getFromClient().getOfficeId(), standingInstructionData.getFromClient().getId(),
-                        standingInstructionData.getFromAccount().getId(), standingInstructionData.getFromAccountType().getValue(),
+                        standingInstructionData.getFromAccount().getId(), standingInstructionData.getFromAccountTypeEnum().getValue(),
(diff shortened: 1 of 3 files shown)


adamsaghy (approved):
LGTM

PR #6205, opened 2026-07-31
FINERACT-2455: WC breach/delinquency reschedule, re-date current period on frequency change + endDate validation

oleksii-novikov-onix:
## Description

## Checklist

### DIFF
--- a/fineract-e2e-tests-runner/src/test/resources/features/WorkingCapitalDelinquencyReschedule.feature
+++ b/fineract-e2e-tests-runner/src/test/resources/features/WorkingCapitalDelinquencyReschedule.feature
@@ -1875,6 +1875,7 @@ Feature: Working Capital Delinquency Reschedule Action
       | 4            | 01 April 2026 | 30 April 2026 | 150            | 0          | 150               |                       |
     Then Admin closes the Working Capital loan with a full repayment on "17 April 2026"
 
+  @TestRailId:C93975
   Scenario: Verify delinquency reschedule frequency change is rejected when resulting period end date is in the past
     When Admin sets the business date to "01 January 2026"
     When Admin creates a client with random data
@@ -1897,6 +1898,7 @@ Feature: Working Capital Delinquency Reschedule Action
     Then Admin fails to create WC delinquency reschedule action with error containing "Frequency change results a delinquency period endDate before current businessDate is not allowed" and the following parameters:
       | frequency | frequencyType |
       | 19        | DAYS          |
+    # --- Delinquency reschedule ---
     When Admin creates WC delinquency reschedule action with the following parameters:
       | frequency | frequencyType |
       | 20        | DAYS          |
@@ -1904,8 +1906,10 @@ Feature: Working Capital Delinquency Reschedule Action
       | periodNumber | fromDate        | toDate           | expectedAmount | paidAmount | outstandingAmount | minPaymentCriteriaMet |
       | 1            | 01 January 2026 | 20 January 2026  | 300            | 0          | 300               | false                 |
       | 2            | 21 January 2026 | 09 February 2026 | 300            | 0          | 300               |                       |
+    # --- Close loan ---
     Then Admin closes the Working Capital loan with a full repayment on "20 January 2026"
 
+  @TestRailId:C93975
   Scenario: Verify delinquency reschedule validates the recalculated period end date against the resumed pause length
     When Admin sets the business date to "01 January 2026"
     When Admin creates a client with random data
@@ -1919,12 +1923,14 @@ Feature: Working Capital Delinquency Reschedule Action
     When Admin runs inline COB job for Working Capital Loan
     When Admin sets the business date to "10 January 2026"
     When Admin runs inline COB job for Working Capital Loan
+    # --- Delinquency pause ---
     And Admin initiate a Working Capital loan delinquency pause with startDate "10 January 2026" and endDate "19 January 2026"
     Then WC loan delinquency range schedule has the following periods:
       | periodNumber | fromDate        | toDate           | expectedAmount | paidAmount | outstandingAmount | minPaymentCriteriaMet |
       | 1            | 01 January 2026 | 09 February 2026 | 300            | 0          | 300               |                       |
     When Admin sets the business date to "15 January 2026"
     When Admin runs inline COB job for Working Capital Loan
+    # --- Delinquency resume ---
     And Admin initiate a Working Capital loan delinquency resume with startDate "15 January 2026"
     Then WC loan delinquency range schedule has the following periods:
       | periodNumber | fromDate        | toDate           | expectedAmount | paidAmount | outstandingAmount | minPaymentCriteriaMet |
@@ -1933,6 +1939,7 @@ Feature: Working Capital Delinquency Reschedule Action
     Then Admin fails to create WC delinquency reschedule action with error containing "Frequency change results a delinquency period endDate before current businessDate is not allowed" and the following parameters:
       | frequency | frequencyType |
       | 25        | DAYS          |
+    # --- Delinquency reschedule ---
     When Admin creates WC delinquency reschedule action with the following parameters:
       | frequency | frequencyType |
       | 26        | DAYS          |
@@ -1940,4 +1947,5 @@ Feature: Working Capital Delinquency Reschedule Action
       | periodNumber | fromDate         | toDate           | expectedAmount | paidAmount | outstandingAmount | minPaymentCriteriaMet |
       | 1            | 01 January 2026  | 01 February 2026 | 300            | 0          | 300               | false                 |
       | 2            | 02 February 2026 | 27 February 2026 | 300            | 0          | 300               |                       |
+    # --- Close loan ---
     Then Admin closes the Working Capital loan with a full repayment on "01 February 2026"
--- a/fineract-e2e-tests-runner/src/test/resources/features/WorkingCapitalDelinquencyReschedule.feature
+++ b/fineract-e2e-tests-runner/src/test/resources/features/WorkingCapitalDelinquencyReschedule.feature
@@ -1875,7 +1875,7 @@ Feature: Working Capital Delinquency Reschedule Action
       | 4            | 01 April 2026 | 30 April 2026 | 150            | 0          | 150               |                       |
     Then Admin closes the Working Capital loan with a full repayment on "17 April 2026"
 
-  @TestRailId:C93975
+  @TestRailId:C93974
   Scenario: Verify delinquency reschedule frequency change is rejected when resulting period end date is in the past
     When Admin sets the business date to "01 January 2026"
     When Admin creates a client with random data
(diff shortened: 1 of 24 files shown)


adamsaghy:
AC3 is not met — 8 existing TestRail scenarios had their inputs changed to dodge the new validation, no?

Breach C85273, C85274, C85275, C85280, C85588: business date 01 June 2026 → 15 May 2026
Breach C85282, C85283: business date 10 March 2019 → 15 January 2019
Delinquency C85583: reschedule frequency 15 DAYS → 30 DAYS, with a new comment stating outright it's so the recalculated end date doesn't fall before the business date

Each of those originally exercised "reschedule deep inside a long period", which is now precisely the AC2 error case. Rather than moving the date, keep the original inputs and flip the assertion to the new validation error, then add a separate happy-path scenario at the earlier date. That both satisfies AC3 honestly and gives AC2 coverage across several product configs instead of one. Scenarios where only the expected tables changed (e.g. C74496, C74497, C85488, C85501) are legitimate consequences of AC1.

The pause-drop isn't stable. It's action.getFrequency() != null  "a frequency was supplied", not "the frequency changed". Re-sending the same frequency still re-dates the period and discards any pause extension, which the new "ignoring an overlapping pause" scenario locks in. It isn't durable: a later PAUSE/RESUME calls recalculatePeriodsForPauses, which rebuilds every period from period 1 using the effective frequency and re-applies all recorded pauses: resurrecting the extension that was just dropped. Same loan, different boundaries depending on action order.

The validator and the service compute the same date two different ways. The validator uses action.getFrequency()/getFrequencyType(); the service uses params.frequency() from resolveEffectiveRescheduleParams. They agree only because the action is saved before the service runs and is the latest with a frequency group. Both also do their own findCurrentOpenPeriod lookup. Extract one helper both call, so the check can't drift from the mutation it's guarding.

findCurrentOpenPeriod returning Optional is stricter than the loop it replaces. The old code tolerated multiple matches (last one won); the derived query throws IncorrectResultSizeDataAccessException if two open periods ever cover the business date. findTop…OrderByPeriodNumberDesc would preserve the old tolerance.

The delinquency queries add AND s.reset = false, which the replaced loops did not have. A behaviour change bundled into a refactor. Unreachable today (resetPeriods only touches periods with toDate < resetDate), but if a reset period ever landed in the future window, updateFuturePeriods renumbers sequentially and would skip it → duplicate period numbers. Also note the breach queries have no equivalent filter, so the two schedules now diverge.

Three different "period expired" predicates coexist in WorkingCapitalLoanBreachScheduleServiceImpl: evaluateBreachOnDate uses businessDate >= toDate, evaluateExpiredPeriods uses toDate < businessDate, recomputeBreach uses businessDate > toDate. Pre-existing, but this PR makes the boundary reachable by design, since the validation deliberately permits a re-dated period ending exactly on the business date. The "equals the current business date" scenario closes the loan immediately and never runs another COB, so what that period does at the next evaluation is unasserted. Add a COB step and pin it.

Minor

The breach validator now mixes two error mechanisms (failWithCodeNoParameterAddedToErrorCode and the new failGeneralValidation) and hardcodes "workingCapitalLoanBreachAction" in two places. The delinquency validator has a VALIDATION_RESOURCE constant — mirror it.

No unit tests. calculateToDate and both validateFrequencyDoesNotEndBeforeBusinessDate methods are cheap to cover directly, especially the == businessDate / -1 day boundary that currently only has slow e2e coverage.

galovics (dismissed):
LGTM

oleksii-novikov-onix:
> AC3 is not met — 8 existing TestRail scenarios had their inputs changed to dodge the new validation, no?
> 
> Breach C85273, C85274, C85275, C85280, C85588: business date 01 June 2026 → 15 May 2026 Breach C85282, C85283: business date 10 March 2019 → 15 January 2019 Delinquency C85583: reschedule frequency 15 DAYS → 30 DAYS, with a new comment stating outright it's so the recalculated end date doesn't fall before the business date
> 
> Each of those originally exercised "reschedule deep inside a long period", which is now precisely the AC2 error case. Rather than moving the date, keep the original inputs and flip the assertion to the new validation error, then add a separate happy-path scenario at the earlier date. That both satisfies AC3 honestly and gives AC2 coverage across several product configs instead of one. Scenarios where only the expected tables changed (e.g. C74496, C74497, C85488, C85501) are legitimate consequences of AC1.
> 
> The pause-drop isn't stable. It's action.getFrequency() != null "a frequency was supplied", not "the frequency changed". Re-sending the same frequency still re-dates the period and discards any pause extension, which the new "ignoring an overlapping pause" scenario locks in. It isn't durable: a later PAUSE/RESUME calls recalculatePeriodsForPauses, which rebuilds every period from period 1 using the effective frequency and re-applies all recorded pauses: resurrecting the extension that was just dropped. Same loan, different boundaries depending on action order.
> 
> The validator and the service compute the same date two different ways. The validator uses action.getFrequency()/getFrequencyType(); the service uses params.frequency() from resolveEffectiveRescheduleParams. They agree only because the action is saved before the service runs and is the latest with a frequency group. Both also do their own findCurrentOpenPeriod lookup. Extract one helper both call, so the check can't drift from the mutation it's guarding.
> 
> findCurrentOpenPeriod returning Optional is stricter than the loop it replaces. The old code tolerated multiple matches (last one won); the derived query throws IncorrectResultSizeDataAccessException if two open periods ever cover the business date. findTop…OrderByPeriodNumberDesc would preserve the old tolerance.
> 
> The delinquency queries add AND s.reset = false, which the replaced loops did not have. A behaviour change bundled into a refactor. Unreachable today (resetPeriods only touches periods with toDate < resetDate), but if a reset period ever landed in the future window, updateFuturePeriods renumbers sequentially and would skip it → duplicate period numbers. Also note the breach queries have no equivalent filter, so the two schedules now diverge.
> 
> Three different "period expired" predicates coexist in WorkingCapitalLoanBreachScheduleServiceImpl: evaluateBreachOnDate uses businessDate >= toDate, evaluateExpiredPeriods uses toDate < businessDate, recomputeBreach uses businessDate > toDate. Pre-existing, but this PR makes the boundary reachable by design, since the validation deliberately permits a re-dated period ending exactly on the business date. The "equals the current business date" scenario closes the loan immediately and never runs another COB, so what that period does at the next evaluation is unasserted. Add a COB step and pin it.
> 
> Minor
> 
> The breach validator now mixes two error mechanisms (failWithCodeNoParameterAddedToErrorCode and the new failGeneralValidation) and hardcodes "workingCapitalLoanBreachAction" in two places. The delinquency validator has a VALIDATION_RESOURCE constant — mirror it.
> 
> No unit tests. calculateToDate and both validateFrequencyDoesNotEndBeforeBusinessDate methods are cheap to cover directly, especially the == businessDate / -1 day boundary that currently only has slow e2e coverage.

1. AC3 - eight scenarios had their inputs adjusted to stay valid under the new rule.

Not changed. Those scenarios were written to cover reschedule inheritance and repeated reschedules, not the end date rule. Moving the business date keeps each of them testing what it was written for, and the new rule has its own dedicated scenarios on both the breach and the delinquency side.

2. Re-dating drops a pause extension, and a later pause resurrects it.

Fixed.

3. The end date is computed in two places.

Fixed. 

4. findCurrentOpenPeriod returning Optional is stricter than the loop it replaced.

Fixed.

5. AND s.reset = false in the delinquency query.

Fixed.

6. Three predicates for "period expired", and the boundary is not asserted.

The boundary part is fixed: the scenario now runs COB on the day after the recalculated end date and pins the resulting schedule, so the behaviour at that boundary is locked in.

On the predicates, there are two rules, not three. evaluateExpiredPeriods and recomputeBreach are the same exclusive condition written two ways. The repository query and evaluateBreachOnDate are the same inclusive condition, and the query is just a prefilter for that check. The exclusive rule belongs to the reprocess path and the inclusive one to the COB evaluation path, which runs as of the previous business date. Unifying them would change breach timing logic.

7. Minor - mixed error mechanisms in the validator.

Fixed.

8. Minor - no unit tests for the new date arithmetic.

Not added as unit tests, extended current e2e test scenarios.

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/domain/WorkingCapitalLoanDelinquencyPauseUtils.java
```diff
@@ -58,6 +58,27 @@ public static LocalDate resolveEffectivePauseEnd(final WorkingCapitalLoanDelinqu
                 .map(WorkingCapitalLoanDelinquencyAction::getStartDate).min(Comparator.naturalOrder()).orElse(pauseEnd);
     }
 
+    /**
+     * Extends an inclusive period end date by every recorded pause overlapping the period, mirroring the pause replay
+     * performed by the schedule rebuild.
+     */
+    public static LocalDate extendToDateByRecordedPauses(final LocalDate fromDate, final LocalDate baseToDate,
```
adamsaghy: The two extendToDateByRecordedPauses implementations have different semantics under identical javadoc (medium, maintainability)
Breach version mutates toDate inside the loop, so the overlap check (pauseStart.isAfter(toDate)) sees the growing end date — a pause starting past the base end can still be absorbed.
Delinquency version sums against the fixed baseToDate, so it cannot.
Each mirrors its own applyRecordedPauses (breach's uses live period.getToDate(), delinquency's captures periodToDate up front), so both are probably correct today. But the doc comment is copy-pasted verbatim, which invites someone to treat them as interchangeable or "fix" one into the other. Neither replays the fromDate shift that applyRecordedPauses performs; 

@oleksii-novikov-onix Is this correct to have different behaviour?

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/validator/WorkingCapitalLoanDelinquencyActionParseAndValidator.java
```diff
@@ -250,6 +252,24 @@ private void validateReschedule(final WorkingCapitalLoanDelinquencyAction action
...
+                validateFrequencyDoesNotEndBeforeBusinessDate(action, workingCapitalLoan, dataValidator);
+            }
+        }
+    }
+
+    /**
+     * Rejects a frequency change whose resulting period end date falls before the current business date. The candidate
+     * end date comes from the range schedule service, so it is the very date the re-date would persist.
+     */
+    private void validateFrequencyDoesNotEndBeforeBusinessDate(final WorkingCapitalLoanDelinquencyAction action,
```
adamsaghy, line 265: Both validators now inject the write-side schedule service purely to call calculateRescheduledCurrentPeriodToDate, widening a mutation-oriented interface with a query method and creating a validator→service edge (no cycle today — both impls depend only on repositories — but a fragile direction in a module where the write service already depends on the validator).

validateAndParse(command, loan, existing) already receives the full action list, and the validators already hold both repositories. So the same "one implementation" guarantee is available without the new interface method or the new dependency:

breachScheduleRepository.findCurrentOpenPeriodsOrderByPeriodNumberDesc(loanId, businessDate).stream().findFirst()
    .map(p -> WorkingCapitalLoanBreachScheduleEvaluationUtils.calculateRescheduledToDate(
            p.getFromDate(), action.getFrequency(), action.getFrequencyType(), existing))
That also removes the redundant second findByWorkingCapitalLoanIdOrderById fetch — currently every reschedule request loads all breach actions twice (validator, then service).

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/service/WorkingCapitalLoanBreachScheduleServiceImpl.java
```diff
@@ -260,6 +255,23 @@ public void rescheduleMinimumPayment(final WorkingCapitalLoan loan) {
...
 
+    @Override
+    public Optional<LocalDate> calculateRescheduledCurrentPeriodToDate(final Long loanId, final Integer frequency,
+            final WorkingCapitalLoanPeriodFrequencyType frequencyType) {
+        return findCurrentOpenPeriod(loanId, DateUtils.getBusinessLocalDate())
+                .map(currentPeriod -> resolveRescheduledToDate(loanId, currentPeriod, frequency, frequencyType));
+    }
+
+    private Optional<WorkingCapitalLoanBreachSchedule> findCurrentOpenPeriod(final Long loanId, final LocalDate businessDate) {
+        return repository.findCurrentOpenPeriodsOrderByPeriodNumberDesc(loanId, businessDate).stream().findFirst();
```
adamsaghy: pass Limit.of(1) so the DB truncates.

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/repository/WorkingCapitalLoanBreachScheduleRepository.java
```diff
@@ -41,4 +43,22 @@ Optional<WorkingCapitalLoanBreachSchedule> findByLoanIdAndFromDateLessThanEqualA
...
     Optional<WorkingCapitalLoanBreachSchedule> findTopByLoanIdAndBreachTrueOrderByFromDateAsc(Long loanId);
+
+    @Query("""
+            SELECT s FROM WorkingCapitalLoanBreachSchedule s
+            WHERE s.loan.id = :loanId
+              AND s.breach IS NULL
+              AND s.fromDate <= :businessDate
+              AND s.toDate >= :businessDate
+            ORDER BY s.periodNumber DESC""")
+    List<WorkingCapitalLoanBreachSchedule> findCurrentOpenPeriodsOrderByPeriodNumberDesc(@Param("loanId") Long loanId,
```
adamsaghy: Its never more than 1 item, no?

fineract-e2e-tests-runner/src/test/resources/features/WorkingCapitalDelinquencyReschedule.feature
```diff
@@ -1861,6 +1871,81 @@ Feature: Working Capital Delinquency Reschedule Action
...
+      | frequency | frequencyType |
+      | 20        | DAYS          |
+    Then WC loan delinquency range schedule has the following periods:
+      | periodNumber | fromDate        | toDate           | expectedAmount | paidAmount | outstandingAmount | minPaymentCriteriaMet |
+      | 1            | 01 January 2026 | 20 January 2026  | 300            | 0          | 300               | false                 |
+      | 2            | 21 January 2026 | 09 February 2026 | 300            | 0          | 300               |                       |
+    # --- Close loan ---
+    Then Admin closes the Working Capital loan with a full repayment on "20 January 2026"
+
+  @TestRailId:C93975
```
adamsaghy, line 1912: @peter-kovacs-dpc duplicate `C93975`

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/repository/WorkingCapitalLoanBreachScheduleRepository.java
oleksii-novikov-onix: I have rechecked, it can never be more than one, so breach could return `Optional`.

The delinquency schedule is built the same way, so the same change would make sense there to keep the two in sync. What do you think?

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/domain/WorkingCapitalLoanDelinquencyPauseUtils.java
oleksii-novikov-onix: I think this is a degradation, not a deliberate difference. I will fix it.

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/service/WorkingCapitalLoanBreachScheduleServiceImpl.java
oleksii-novikov-onix: Probably we can ignore this, as we have another [comment](https://github.com/apache/fineract/pull/6205#discussion_r3749096549) that is more valid.

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/repository/WorkingCapitalLoanBreachScheduleRepository.java
adamsaghy: yeah. lets keep it in sync

fineract-e2e-tests-runner/src/test/resources/features/WorkingCapitalDelinquencyReschedule.feature
oleksii-novikov-onix: Fixed

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/repository/WorkingCapitalLoanBreachScheduleRepository.java
oleksii-novikov-onix: Fixed

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/validator/WorkingCapitalLoanDelinquencyActionParseAndValidator.java
oleksii-novikov-onix: Fixed

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/domain/WorkingCapitalLoanDelinquencyPauseUtils.java
oleksii-novikov-onix: Fixed

adamsaghy (approved):
LGTM

PR #6206, opened 2026-07-31
FINERACT-2726: Saving accounts - Unable to create, edit a saving product

rymghosn:
**Issue: FineractEntityAccessReadServiceImpl – Invalid SQL Query When User Has No Entity Access Mappings**

**Root Cause:**
In `FineractEntityAccessReadServiceImpl#getSQLQueryInClause_WithListOfIDsForEntityAccess`, when a user has no entity-access mappings for a specific entity type (for example, office-restricted savings products), the method generates a comma-separated ID list to be used inside a SQL `IN (...)` clause.

When the access list is empty, the method appends the literal value `"false"` instead of a numeric ID. Since the SQL query expects numeric IDs, this results in an invalid query and database failure.

**Example Error:**

```
PostgreSQL: invalid input syntax for type bigint: "false"
```

**Impact:**
With entity/office-specific access restrictions enabled, users without mapped offices for a given entity type cannot create or edit entities filtered through this helper (such as savings products). The operation fails instead of returning an empty/restricted result set.

**Fix:**
Replace `"false"` with `"-1"` when the access list is empty.

`-1` is a valid numeric value that does not match any existing entity ID, allowing the SQL query to execute successfully while returning zero rows.
This PR  :(https://issues.apache.org/jira/browse/FINERACT-2726)

### DIFF
--- a/fineract-e2e-tests-core/src/test/java/org/apache/fineract/test/stepdef/saving/SavingsAccountStepDef.java
+++ b/fineract-e2e-tests-core/src/test/java/org/apache/fineract/test/stepdef/saving/SavingsAccountStepDef.java
@@ -23,8 +23,10 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.fail;
 
 import io.cucumber.datatable.DataTable;
+import io.cucumber.java.After;
 import io.cucumber.java.en.And;
 import io.cucumber.java.en.Then;
+import io.cucumber.java.en.When;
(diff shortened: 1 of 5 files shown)


adamsaghy (changes requested):
Can you please add an integration / E2E test case to cover the fix?

rymghosn:
@adamsaghy  Added an integration/E2E test case to cover the fix.

adamsaghy (approved):
LGTM

PR #6207, opened 2026-07-31
FINERACT-2725: Fix case-insensitive LIMIT/OFFSET removal in countQueryResult

aya-abdallah-FOO:
Fixes FINERACT-2725.                                                                                                                                                              
                                                                                                                                                                                    
  `PaginationParameters.limitSql()` generates lowercase SQL clauses (`limit`/`offset`).                                                                                             
  `DatabaseSpecificSQLGenerator.countQueryResult` was stripping these using                                                                                                         
  case-sensitive regex, so on PostgreSQL the LIMIT clause survived into the                                                                                                         
  `SELECT COUNT(*) FROM (...) AS temp` wrapper — causing paginated endpoints                                                                                                        
  to return a `totalFilteredRecords` value capped at the page size instead of                                                                                                       
  the true total.                                                                                                                                                                   
                                                                                                                                                                                    
  Made both patterns case-insensitive (`(?i)`) and use `\s+` for whitespace.

### DIFF
--- a/fineract-core/src/test/java/org/apache/fineract/infrastructure/core/service/database/DatabaseSpecificSQLGeneratorTest.java
+++ b/fineract-core/src/test/java/org/apache/fineract/infrastructure/core/service/database/DatabaseSpecificSQLGeneratorTest.java
@@ -82,4 +82,18 @@ public class DatabaseSpecificSQLGeneratorTest {
         Mockito.verify(connection).createArrayOf("bigint", new Long[] { 1L, 2L, 3L });
         Mockito.verify(connection).close();
     }
+
+    @Test
+    public void testCountQueryResultOnSqlWithLowercaseLimit() {
+        String sql = "SELECT 1 FROM test_table WHERE asd=2 limit 25";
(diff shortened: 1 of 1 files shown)


adamsaghy:
@aya-abdallah-FOO Please rebase

aya-abdallah-FOO:
> @aya-abdallah-FOO Please rebase

done

adamsaghy:
@aya-abdallah-FOO Only test cases?

aya-abdallah-FOO:
> @aya-abdallah-FOO Only test cases?

@adamsaghy I originally included both the implementation fix (making the LIMIT/OFFSET regex case-insensitive with (?i)) and the test cases in this PR. However, after rebasing on the latest develop, I noticed the implementation change disappeared from the diff ; it turns out a community contributor independently submitted the same fix which was already merged into develop. As a result, only the test cases are now visible in this PR, since the implementation is already covered by the base branch.

adamsaghy (approved):
LGTM

PR #6210, opened 2026-07-31
FINERACT-2729: Accept .xlsx files in bulk import, not just legacy .xls

rymghosn:
Bulk import validated the uploaded file's MIME type but only accepted the legacy Excel (`.xls`) format, rejecting valid OOXML (`.xlsx`) files such as those downloaded from Google Sheets. Additionally, the implementation always parsed uploaded files using `HSSFWorkbook`, which only supports the `.xls` format.

This change adds support for the OOXML spreadsheet MIME type and replaces the hardcoded `HSSFWorkbook` with `WorkbookFactory.create()`, allowing Apache POI to automatically detect and parse both `.xls` and `.xlsx` workbooks.

This PR : https://issues.apache.org/jira/browse/FINERACT-2729


adamsaghy (approved):
LGTM

PR #6211, opened 2026-08-01
FINERACT-2730: add loan and savings transaction search

AnvayKharb:
## Description

This PR extends the global search functionality to support searching loan repayment and savings transactions.

Users can now search transactions using existing transaction identifiers without changing the current search endpoint or affecting existing search behavior.

### Supported searches

**Loan repayments**
- Transaction ID
- External ID

**Savings deposits and withdrawals**
- Transaction ID
- External ID
- Reference Number (refNo)

### Changes

- Added `loanTransactions` and `savingsTransactions` search resources.
- Extended the global search SQL to return matching loan repayment and savings transaction results.
- Added transaction metadata to the search response, including transaction ID, transaction type, account ID, account number, account type, external ID, and reference number where applicable.
- Preserved backward compatibility by adding nullable response fields without changing the existing API contract.
- Added integration tests covering transaction searches by transaction ID, external ID, and reference number.

## Related Issue

FINERACT-2730

## Screenshots

N/A

## Checklist

- [ ] If you have multiple commits, they have been squashed into a single commit.
- [ ] Read and understood the contribution guidelines.
- [ ] Added or updated tests for the implemented functionality.
- [ ] Verified that existing search functionality continues to work.

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/search/service/SearchReadServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/search/service/SearchReadServiceImpl.java
@@ -35,9 +35,11 @@ import org.apache.fineract.organisation.office.service.OfficeReadPlatformService
 import org.apache.fineract.portfolio.client.domain.ClientEnumerations;
 import org.apache.fineract.portfolio.group.domain.GroupingTypeEnumerations;
 import org.apache.fineract.portfolio.loanaccount.data.LoanStatusEnumData;
+import org.apache.fineract.portfolio.loanaccount.domain.LoanTransactionType;
 import org.apache.fineract.portfolio.loanproduct.data.LoanProductData;
 import org.apache.fineract.portfolio.loanproduct.service.LoanEnumerations;
 import org.apache.fineract.portfolio.loanproduct.service.LoanProductReadPlatformService;
+import org.apache.fineract.portfolio.savings.SavingsAccountTransactionType;
 import org.apache.fineract.portfolio.savings.data.SavingsAccountStatusEnumData;
 import org.apache.fineract.portfolio.savings.service.SavingsEnumerations;
 import org.apache.fineract.portfolio.search.SearchConstants;
@@ -79,6 +81,10 @@ public class SearchReadServiceImpl implements SearchReadService {
         } else {
             params.addValue("search", "%" + searchConditions.getSearchQuery() + "%");
         }
+        params.addValue("searchTransactionId", parseSearchTransactionId(searchConditions.getSearchQuery()));
+        params.addValue("loanRepaymentTransactionType", LoanTransactionType.REPAYMENT.getValue());
+        params.addValue("savingsDepositTransactionType", SavingsAccountTransactionType.DEPOSIT.getValue());
+        params.addValue("savingsWithdrawalTransactionType", SavingsAccountTransactionType.WITHDRAWAL.getValue());
         return namedParameterJdbcTemplate.query(searchSchema(searchConditions), params, rm);
     }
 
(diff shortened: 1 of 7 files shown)


fineract-provider/src/main/java/org/apache/fineract/portfolio/search/service/SearchReadServiceImpl.java
```diff
@@ -152,13 +163,54 @@ public String searchSchema(final SearchConditions searchConditions) {
...
+                CASE WHEN g.id is null THEN 'client' ELSE 'group' END as parentType, \
+                lt.id as transactionId, 'repayment' as transactionType, lt.external_id as transactionExternalId, \
+                null as transactionRefNo, l.id as accountId, l.account_no as accountNo, 'loan' as accountType \
+                from m_loan_transaction lt join m_loan l on l.id = lt.loan_id \
+                left join m_client c on l.client_id = c.id \
+                left join m_group g ON l.group_id = g.id \
+                left join m_office o on o.id = coalesce(c.office_id, g.office_id) \
+                left join m_product_loan pl on pl.id=l.product_id \
+                where o.hierarchy like :hierarchy \
+                and lt.transaction_type_enum = 2 \
```
adamsaghy: Why is there a hardcoded filter for the **repayment** type?

fineract-provider/src/main/java/org/apache/fineract/portfolio/search/service/SearchReadServiceImpl.java
```diff
@@ -152,13 +163,54 @@ public String searchSchema(final SearchConditions searchConditions) {
...
+                st.id as transactionId, CASE WHEN st.transaction_type_enum = 1 THEN 'deposit' ELSE 'withdrawal' END as transactionType, \
+                st.external_id as transactionExternalId, st.ref_no as transactionRefNo, \
+                s.id as accountId, s.account_no as accountNo, 'savings' as accountType \
+                from m_savings_account_transaction st join m_savings_account s on s.id = st.savings_account_id \
+                left join m_client c on s.client_id = c.id \
+                left join m_group g ON s.group_id = g.id \
+                left join m_office o on o.id = coalesce(c.office_id, g.office_id) \
+                left join m_savings_product sp on sp.id=s.product_id \
+                where o.hierarchy like :hierarchy \
+                and st.transaction_type_enum in (1, 2) \
```
adamsaghy: Why is there a hardcoded filter for the **Deposit and Withdrawal** type?

fineract-provider/src/main/java/org/apache/fineract/portfolio/search/service/SearchReadServiceImpl.java
AnvayKharb: Hi @adamsaghy Thanks for pointing this out. I’ve updated the search SQL to avoid hardcoded transaction type enum values.

The loan transaction search now uses `LoanTransactionType.REPAYMENT`, and savings transaction search uses `SavingsAccountTransactionType.DEPOSIT` and `SavingsAccountTransactionType.WITHDRAWAL` as named SQL parameters.

The search scope is unchanged: loan repayments, savings deposits, and savings withdrawals only.

adamsaghy: Why to filter on these types?

AnvayKharb: @adamsaghy I kept this filter because the current scope is limited to loan repayments, savings deposits, and savings withdrawals. If you'd prefer it to cover all transaction types, I can update it.

IOhacker (approved):
LGTM

PR #6226, opened 2026-08-04
FINERACT-2737: Return collateral name, unit price and percentage on l…

rymghosn:
Loan collateral responses previously exposed only the collateral quantity and calculated values (`quantity`, `total`, and `totalCollateral`), while omitting collateral metadata that is already available in the client collateral API.

This change adds the following fields to the loan collateral response:

* `name` – the collateral type name
* `unitPrice` – the base price per unit
* `pctToBase` – the percentage applied to the base price when calculating the collateral value

This aligns the loan collateral API with the client collateral API, provides a consistent representation of collateral data across endpoints, and removes the need for clients to make additional API calls to retrieve collateral type and pricing information.
PR:(https://issues.apache.org/jira/browse/FINERACT-2737)

### DIFF
--- a/fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/data/LoanCollateralManagementData.java
+++ b/fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/data/LoanCollateralManagementData.java
@@ -19,11 +19,9 @@
 package org.apache.fineract.portfolio.loanaccount.data;
 
 import java.math.BigDecimal;
-import lombok.AllArgsConstructor;
 import lombok.Data;
 
 @Data
-@AllArgsConstructor
 public class LoanCollateralManagementData {
 
     private Long clientCollateralId;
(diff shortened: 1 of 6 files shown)


rymghosn:
@adamsaghy Could you please review the code?

fineract-loan/src/main/java/org/apache/fineract/portfolio/loanaccount/data/LoanCollateralManagementData.java
```diff
@@ -35,4 +33,22 @@ public class LoanCollateralManagementData {
...
 
     private Long id;
+
+    private String name;
+
+    private BigDecimal unitPrice;
+
+    private BigDecimal pctToBase;
+
+    public LoanCollateralManagementData(Long clientCollateralId, BigDecimal quantity, BigDecimal total, BigDecimal totalCollateral, Long id,
```
adamsaghy, line 43: We prefer @AllArgsConstructor over manual constructor. Please undo this change

adamsaghy (changes requested):
Please add proper integration or E2E test to cover the correct behaviour.

Please review my request to avoid using manual constructors

PR #6235, opened 2026-08-06
FINERACT-2743: Fix standing instruction not created at loan disbursem…

rymghosn:
Fixes two issues in the Standing Instruction (SI) feature:

1. **Standing Instruction not created during loan disbursement**

   * Fixed the lookup logic in `StandingInstructionReadPlatformServiceImpl` to correctly identify the loan account based on the transfer type.
   * For `LOAN_REPAYMENT` transfers, the loan account is stored in `to_loan_account_id` rather than `from_loan_account_id`.
   * Added a defensive fallback to check both columns when the transfer type is unknown, ensuring the standing instruction is found and persisted correctly.

2. **Standing Instruction batch job failure on PostgreSQL**

   * Fixed PostgreSQL-incompatible boolean comparisons by replacing integer comparisons (`<> 1`) with proper boolean predicates (`IS FALSE` / `= false`).
   * This resolves the batch job failure while maintaining compatibility with PostgreSQL.

Additionally, added a Liquibase changeset to remove duplicate `m_permission` entries for `CREATE_STANDINGINSTRUCTION`, `UPDATE_STANDINGINSTRUCTION`, and `DELETE_STANDINGINSTRUCTION` that differ only by a trailing space in their permission code. The changeset is PostgreSQL-only, idempotent, and safely executes only when duplicates exist.

PR:(https://issues.apache.org/jira/browse/FINERACT-2743)

### Testing

* `./gradlew :fineract-provider:compileJava` — Passed
* `./gradlew :fineract-core:compileTestJava :fineract-provider:compileTestJava` — Passed
* Verified existing tests (`StandingInstructionDataSerializationTest` and `ExecuteOverdueAndCurrentStandingInstructionsTest`) all pass.
* Manually verified that enabling **Create Standing Instructions at Disbursement**, activating, and disbursing a loan correctly creates and persists the standing instruction, and that the SI batch job executes successfully on PostgreSQL.

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/account/service/StandingInstructionReadPlatformServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/account/service/StandingInstructionReadPlatformServiceImpl.java
@@ -299,8 +299,21 @@ public class StandingInstructionReadPlatformServiceImpl implements StandingInstr
                 sqlBuilder.append(" fromsavacc.id=? ");
                 paramObj.add(standingInstructionDTO.fromAccount());
             } else if (PortfolioAccountType.LOAN.equals(accountType)) {
-                sqlBuilder.append(" fromloanacc.id=? ");
-                paramObj.add(standingInstructionDTO.fromAccount());
+                // For LOAN_REPAYMENT transfers, loan is stored in to_loan_account_id (FROM=SAVINGS, TO=LOAN)
+                // For other transfer types, loan is stored in from_loan_account_id
+                // Defensive fallback: if transferType is null, check both columns to handle UI inconsistencies
+                Integer transferTypeValue = standingInstructionDTO.transferType();
+                if (transferTypeValue != null && transferTypeValue.equals(AccountTransferType.LOAN_REPAYMENT.getValue())) {
+                    sqlBuilder.append(" toloanacc.id=? ");
+                    paramObj.add(standingInstructionDTO.fromAccount());
+                } else if (transferTypeValue == null) {
+                    sqlBuilder.append(" (toloanacc.id=? OR fromloanacc.id=?) ");
+                    paramObj.add(standingInstructionDTO.fromAccount());
+                    paramObj.add(standingInstructionDTO.fromAccount());
+                } else {
+                    sqlBuilder.append(" fromloanacc.id=? ");
+                    paramObj.add(standingInstructionDTO.fromAccount());
+                }
             }
             addAndCaluse = true;
         }
@@ -358,7 +371,7 @@ public class StandingInstructionReadPlatformServiceImpl implements StandingInstr
     public StandingInstructionDuesData retriveLoanDuesData(final Long loanId) {
         final StandingInstructionLoanDuesMapper rm = new StandingInstructionLoanDuesMapper();
         final String sql = "select " + rm.schema() + " where ml.id= ? and ls.duedate <= " + sqlGenerator.currentBusinessDate()
-                + " and ls.completed_derived <> 1";
+                + " and ls.completed_derived IS FALSE";
         return this.jdbcTemplate.queryForObject(sql, rm, new Object[] { loanId }); // NOSONAR
     }
 
--- a/fineract-provider/src/main/resources/db/changelog/tenant/parts/0245_delete_trailing_space_standinginstruction_permissions.xml
+++ b/fineract-provider/src/main/resources/db/changelog/tenant/parts/0245_delete_trailing_space_standinginstruction_permissions.xml
@@ -1,0 +1,75 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements. See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership. The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License. You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied. See the License for the
+    specific language governing permissions and limitations
+    under the License.
+
+-->
+<!--
+    Liquibase changeSet to safely delete duplicate m_permission rows
+    for Standing Instruction permissions WITHOUT trailing spaces,
+    keeping the legacy trailing-space versions.
+
+    Deletes ONLY:
+      - 'CREATE_STANDINGINSTRUCTION'
+      - 'UPDATE_STANDINGINSTRUCTION'
+      - 'DELETE_STANDINGINSTRUCTION'
+
+    Safety rules:
+    - Delete ONLY if a trailing-space version already exists.
+    - Do NOT modify or insert any permissions.
+    - Do NOT touch m_role_permission.
+    - Idempotent and safe to re-run.
+    - PostgreSQL only.
+-->
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.1.xsd">
+
+    <changeSet id="0245_delete_non_spaced_standinginstruction_permissions"
+               author="Samer Melhem"
+               failOnError="true">
+
+        <!-- Run only on PostgreSQL -->
+        <preConditions onFail="MARK_RAN" onError="HALT">
+            <dbms type="postgresql"/>
+        </preConditions>
+
+        <sql>
+            DELETE FROM m_permission p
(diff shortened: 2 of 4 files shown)


IOhacker (approved):
LGTM

rymghosn:
@adamsaghy Could you please review the code? If everything looks good, kindly approve and merge it. Thanks!

fineract-provider/src/main/resources/db/changelog/tenant/parts/0245_delete_trailing_space_standinginstruction_permissions.xml
```diff
@@ -0,0 +1,75 @@
...
+    - Do NOT modify or insert any permissions.
+    - Do NOT touch m_role_permission.
+    - Idempotent and safe to re-run.
+    - PostgreSQL only.
+-->
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.1.xsd">
+
+    <changeSet id="0245_delete_non_spaced_standinginstruction_permissions"
```
adamsaghy, line 43: Why to delete the non-spaced over the trailing space one? 

Having trailing space in the permissions are clearly incorrect.

Should we consider removing the one with extra space instead?

fineract-provider/src/main/java/org/apache/fineract/portfolio/account/service/StandingInstructionReadPlatformServiceImpl.java
```diff
@@ -299,8 +299,21 @@ public Page<StandingInstructionData> retrieveAll(final StandingInstructionDTO st
                 sqlBuilder.append(" fromsavacc.id=? ");
                 paramObj.add(standingInstructionDTO.fromAccount());
             } else if (PortfolioAccountType.LOAN.equals(accountType)) {
-                sqlBuilder.append(" fromloanacc.id=? ");
-                paramObj.add(standingInstructionDTO.fromAccount());
+                // For LOAN_REPAYMENT transfers, loan is stored in to_loan_account_id (FROM=SAVINGS, TO=LOAN)
```
adamsaghy, line 302: This look incorrect. In my understanding you are completely mixing the from and to accounts here...

fineract-provider/src/main/java/org/apache/fineract/portfolio/account/service/StandingInstructionReadPlatformServiceImpl.java
```diff
@@ -299,8 +299,21 @@ public Page<StandingInstructionData> retrieveAll(final StandingInstructionDTO st
...
+                // For other transfer types, loan is stored in from_loan_account_id
+                // Defensive fallback: if transferType is null, check both columns to handle UI inconsistencies
+                Integer transferTypeValue = standingInstructionDTO.transferType();
+                if (transferTypeValue != null && transferTypeValue.equals(AccountTransferType.LOAN_REPAYMENT.getValue())) {
+                    sqlBuilder.append(" toloanacc.id=? ");
```
adamsaghy, line 307: why to use the `standingInstructionDTO.fromAccount()` for `toloanacc.id`?

fineract-provider/src/main/java/org/apache/fineract/portfolio/account/service/StandingInstructionReadPlatformServiceImpl.java
```diff
@@ -299,8 +299,21 @@ public Page<StandingInstructionData> retrieveAll(final StandingInstructionDTO st
...
+                    paramObj.add(standingInstructionDTO.fromAccount());
+                } else if (transferTypeValue == null) {
+                    sqlBuilder.append(" (toloanacc.id=? OR fromloanacc.id=?) ");
```
adamsaghy, line 310: why to check for `to` and `from` account as well the `standingInstructionDTO.fromAccount()`?

adamsaghy (changes requested):
Please see my concerns

fineract-provider/src/main/resources/db/changelog/tenant/parts/0245_delete_trailing_space_standinginstruction_permissions.xml
rymghosn: Great question — the trailing space is definitely wrong, but it's actually the *original* value, not the new one.

`0002_initial_data.xml` (the base seed data every tenant has run since Fineract's very first release) inserts `'CREATE_STANDINGINSTRUCTION '`, `'UPDATE_STANDINGINSTRUCTION '` and `'DELETE_STANDINGINSTRUCTION '` **with** the trailing space (ids 450-452). The non-spaced duplicate was introduced much later, by accident, in `0194_fix_missing_permission.xml` (changeSets 20-22): that changeset's precondition does an exact-match check (`code = 'CREATE_STANDINGINSTRUCTION'`), which — because the existing row already has the trailing space — never matches, so it always reports "0 rows found" and inserts a second, non-spaced row on every tenant that has run through the changelog.

Since `m_role_permission` references a permission by its numeric id, not by its code, any tenant that's been running since before changeset 0194 could already have granted Create/Update/Delete Standing Instruction to a custom role against the *original* (spaced) id. This changeset explicitly promises not to touch `m_role_permission`, so the only row we can safely delete without needing a role-permission migration is the newer accidental duplicate — which is what it does.

It's also not just cosmetic: `PermissionRepository.findOneByCode()` already TRIM/LOWER-cases both sides of the comparison specifically because of this bug (see the comment on that method) — so with both rows present, any code path resolving this permission "by code" today throws a `NonUniqueResultException`, not just an ugly space.

Properly fixing the underlying ~decade-old trailing space (e.g. `UPDATE m_permission SET code = TRIM(code)` on the surviving row) is a good idea, but it's a separate, larger change — it'd need a sweep to confirm nothing else compares against the exact `'... '` string, and probably deserves its own ticket rather than riding along with this SI disbursement fix. Happy to file that as a follow-up if useful.

fineract-provider/src/main/java/org/apache/fineract/portfolio/account/service/StandingInstructionReadPlatformServiceImpl.java
rymghosn: Totally fair to double check — the naming here is genuinely confusing, but it's not a mix-up.

`standingInstructionDTO.fromAccount()` / `fromAccountType()` at this point in the method are the API's generic search filter ("give me standing instructions involving this account"), not literally "the source side of the money movement." Which physical column a loan account actually lives in depends on the transfer type, not on how the caller happened to name the query parameter.

For the standing instructions this PR is about — the ones auto-created at loan disbursement — `LoanWritePlatformServiceJpaRepositoryImpl.createStandingInstruction()` builds the transfer via:

```java
AccountTransferDetails.savingsToLoanTransfer(fromOffice, fromClient, linkedSavingsAccount, toOffice, toClient, loan, transferType);
```

That factory always stores the savings account as `fromSavingsAccount` and the loan as `toLoanAccount` (`from_loan_account_id` is left `null`). So for `LOAN_REPAYMENT` standing instructions, the loan is *always* on the `to_loan_account_id` side in the database — never `from_loan_account_id`.

The old code compared the caller's loan-id filter against `fromloanacc.id` (i.e. `from_loan_account_id`), which is always null for these rows, so it could never find the auto-created standing instruction for a given loan. That's the actual "not found" symptom from the bug report. The fix just routes the comparison to whichever column the loan is genuinely stored in, based on the same `transferType` the write path used to create it.

fineract-provider/src/main/java/org/apache/fineract/portfolio/account/service/StandingInstructionReadPlatformServiceImpl.java
rymghosn: Because `fromAccount()` here is the loan id the *caller* is searching for (e.g. "show me the standing instruction for loan #42"), not literally the from-side of the money movement.

As shown in `AccountTransferDetails.savingsToLoanTransfer(...)` (used by `LoanWritePlatformServiceJpaRepositoryImpl.createStandingInstruction()`), for `LOAN_REPAYMENT` transfers the loan is persisted as `toLoanAccount` / `to_loan_account_id` — `from_loan_account_id` is left null for these rows. So to find the standing instruction for loan #42 we have to match `toloanacc.id = 42`, even though the caller supplies that same loan id via the `fromAccount` parameter.

I agree the parameter name is misleading — it really means "the account of interest" rather than "the from side" — but renaming it would touch the public search API's query params, which is a bigger, more invasive change than this bugfix needs.

fineract-provider/src/main/java/org/apache/fineract/portfolio/account/service/StandingInstructionReadPlatformServiceImpl.java
rymghosn: That branch only runs when `transferType` is `null`, i.e. the caller didn't filter by transfer type at all — they just asked for standing instructions touching a given loan account, regardless of type.

Since which column (`to_loan_account_id` vs `from_loan_account_id`) holds the loan id depends on the transfer type (see the LOAN_REPAYMENT case above, where it's always `to_loan_account_id`), and we don't know the transfer type in this branch, we can't know in advance which single column to check — so we check both with `OR`, using the same caller-supplied loan id on each side. If a transfer type *is* given, we're back in one of the other two branches where we already know which single column applies.

adamsaghy:
@rymghosn Are you planning to work on this?

PR #6241, opened 2026-08-07
FINERACT-2732: migrate the savings and deposit integration tests to the Feign client (Tier 1)

DeathGun44:
## Description

Tier 1 of moving the savings and deposit integration tests onto the generated Feign client, so they call the API through typed models instead of raw JSON. It covers the 14 smaller savings classes.

## Production changes

Limited to four Swagger files.

**Most are additive** - optional fields that typed clients need but the schema does not
describe:

- savings product accounting mappings and overdraft fields
- a transaction `note`
- `transactionId` on two datatable responses
- `withdrawalFeeForTransfers`

The endpoints already accept and return these; only the documented shape changes.

**One is not additive.** `SavingsAccountTransactionsSearchResponse.content` is a `List` rather than a `Set`. The endpoint returns a page ordered by the request's `orderBy`/`sortOrder`, and a set throws that order away in generated clients with no way to get it back. The response is a JSON array either way, so the wire format is unchanged.

## Backward compatibility

The CI gate was reproduced locally against a spec built from the merge base. No breaking changes. The two specs differ in 52 places - 51 additions and one removal, the `uniqueItems` flag a list does not carry.

## Notes on a few choices

- Each test class resets every global configuration to its default after each test and verifies it, so no test leaks configuration into the next.
- The scheduler helper asserts that the run it awaited was triggered by the test itself and finished with status `success`, so a failed job cannot look like a passing one. The helper is shared with `FeignLoanTestBase`, so the loan suites get the same check.
- The accrual accounting test gives every GL mapping its own account, so an assertion about one mapping cannot be satisfied by another.
- Boolean checks on generated getters are null-safe, since those getters are nullable.
- The batch endpoint answers 200 even when a sub-request fails, so the tests assert on each sub-request's own status code.
## Checklist

### DIFF
--- a/integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/FeignSavingsTransactionHelper.java
+++ b/integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/FeignSavingsTransactionHelper.java
@@ -21,14 +21,13 @@ package org.apache.fineract.integrationtests.client.feign.helpers;
 import static org.apache.fineract.client.feign.util.FeignCalls.fail;
 import static org.apache.fineract.client.feign.util.FeignCalls.ok;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
 import feign.FeignException;
 import java.math.BigDecimal;
 import java.math.RoundingMode;
 import java.util.List;
 import java.util.Map;
 import org.apache.fineract.client.feign.FineractFeignClient;
-import org.apache.fineract.client.feign.ObjectMapperFactory;
+import org.apache.fineract.client.feign.services.SavingsAccountTransactionsApi.SearchSavingsAccountTransactionsQueryParams;
 import org.apache.fineract.client.feign.util.CallFailedRuntimeException;
 import org.apache.fineract.client.models.CommandProcessingResult;
 import org.apache.fineract.client.models.PostSavingsAccountBulkReversalTransactionsRequest;
@@ -108,14 +107,8 @@ public class FeignSavingsTransactionHelper {
                 command));
     }
 
-    /** The generated client types this response as a bare {@code String}, so the JSON is mapped here. */
     public SavingsAccountTransactionData getTransaction(Long savingsId, Long transactionId) {
-        String json = ok(() -> fineractClient.savingsAccountTransactions().retrieveOneSavingsAccountTransaction(savingsId, transactionId));
-        try {
-            return ObjectMapperFactory.getShared().readValue(json, SavingsAccountTransactionData.class);
-        } catch (JsonProcessingException e) {
-            throw new IllegalStateException("Failed to parse savings transaction " + transactionId, e);
-        }
+        return ok(() -> fineractClient.savingsAccountTransactions().retrieveOneSavingsAccountTransaction(savingsId, transactionId));
     }
 
     public List<SavingsAccountTransactionData> getTransactions(Long savingsId) {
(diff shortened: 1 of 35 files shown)


integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/SavingsTransactionCommandsApi.java
```diff
@@ -0,0 +1,39 @@
...
+import org.apache.fineract.client.models.CommandProcessingResult;
+import org.apache.fineract.client.models.PostSavingsAccountBulkReversalTransactionsRequest;
+
+/**
+ * Adjusts a savings transaction - {@code reverse}, {@code undo} or {@code releaseAmount}. Hand-written because
+ * {@code SavingsAccountTransactionsApiResource} documents the 200 as an array of request models, so the generated
+ * client cannot read the {@code resourceId} these tests assert on. Correcting that {@code @ApiResponse} is a breaking
+ * OpenAPI change, so it is prepared separately rather than carried by a test migration.
+ */
+@Headers({ "Accept: application/json", "Content-Type: application/json" })
```
adamsaghy, line 30: Lets correct the swagger if it is incorrect.

adamsaghy (changes requested):
@DeathGun44 Ultimately it looks okay. Please review my only concern.

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/SavingsTransactionCommandsApi.java
DeathGun44: Done!

DeathGun44:
@adamsaghy rebased and ready for review!

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/FeignSavingsTransactionHelper.java
```diff
@@ -48,4 +62,93 @@ public PostSavingsAccountTransactionsResponse withdraw(Long savingsId, PostSavin
...
+    /** All three commands are driven by the query parameter alone and read no payload, hence the empty body. */
+    private CommandProcessingResult adjustTransaction(Long savingsId, Long transactionId, String command) {
+        PostSavingsAccountBulkReversalTransactionsRequest request = new PostSavingsAccountBulkReversalTransactionsRequest();
+        return ok(() -> fineractClient.savingsAccountTransactions().adjustSavingsAccountTransaction(savingsId, transactionId, request,
+                command));
+    }
+
+    /** The generated client types this response as a bare {@code String}, so the JSON is mapped here. */
+    public SavingsAccountTransactionData getTransaction(Long savingsId, Long transactionId) {
+        String json = ok(() -> fineractClient.savingsAccountTransactions().retrieveOneSavingsAccountTransaction(savingsId, transactionId));
```
adamsaghy: Instead of keeping String, it would be better to update the swagger to return the valid data type

adamsaghy (changes requested):
Please ensure that wherever a concrete data type can be set as a request or response object (except for datatables), do so. For instance, returning JSON as a String should be changed.

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/helpers/FeignSavingsTransactionHelper.java
DeathGun44: The endpoint had no @ApiResponse at all, so the spec fell back to the method's Java return type and documented the response as a string. Added the real type to all four retrieveOne variants. SavingsAccountTransactionData was already in the spec, so no new model. The helper is now a plain typed call. The API compatibility check goes red with R015 on those four paths - response type string → object. The wire format is unchanged.

DeathGun44: only two others return a String, both on datatables. Those stay, because a datatable's columns are defined at runtime and that endpoint returns two different shapes - I added a javadoc saying so. I also replaced a hand-built map of 15 query-string keys in the search test with the generated query-params builder.

adamsaghy (approved):
LGTM

PR #6253, opened 2026-08-10
FINERACT-2752: Charge - Income from charge list is not the same on add and edit mode

rymghosn:
When creating a charge with chargeAppliesTo = Client and selecting an item from the "Income from Charge" (GL income account) list, editing that  
  charge afterwards showed the previously selected income account as missing, and reopening the dropdown showed fewer available options than       
  before.                                                                                                                                          
                                                                                                                                                   
  Root cause: ChargeWritePlatformServiceJpaRepositoryImpl#updateCharge persisted the updated charge with a plain chargeRepository.save(). This     
  defers the actual flush to Hibernate's own timing, so when the same request cycle immediately re-read the charge to build the response (or to    
  populate the edit-mode dropdown), it could still see the pre-update state — making the income account selection look reverted or absent.         
                                                                                                                                                   
  Fix:                                                                                                                                             
  - updateCharge now calls chargeRepository.saveAndFlush() instead of save(), so the change is flushed before the charge is re-read.               
  - Added the missing incomeAccountId field to ChargeRequest so the API request DTO accurately reflects the incomeAccountId parameter the service  
  already accepts (ChargesApiConstants.glAccountIdParamName).
  PR:(https://issues.apache.org/jira/browse/FINERACT-2752)

### DIFF
--- a/fineract-charge/src/main/java/org/apache/fineract/portfolio/charge/request/ChargeRequest.java
+++ b/fineract-charge/src/main/java/org/apache/fineract/portfolio/charge/request/ChargeRequest.java
@@ -52,5 +52,6 @@ public class ChargeRequest implements Serializable {
     private BigDecimal minCap;
     private BigDecimal maxCap;
     private Long taxGroupId;
+    private Long incomeAccountId;
 
 }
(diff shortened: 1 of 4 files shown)


adamsaghy (approved):
LGTM

adamsaghy (changes requested):
There is no read after the `save` so once transaction finished, it will be committed. I see absolute no reason what `saveAndFlush` would change at this point... 

Dont misunderstand me, it is not wrong, just unnecessary...

Your PR description contradicts with what happens in Fineract... the update and the read are two separate transactions. Flush does not play any part here...

PR #6255, opened 2026-08-10
FINERACT-2753: Improve Maker-Checker Action Endpoint Handling and Permission Responses

rymghosn:
Today, maker-checker enabled actions have two rough edges:                                                                                       
                                                                                                                                                   
  1. A maker can resubmit the same action on the same resource any number of times while an earlier submission is still AWAITING_APPROVAL, silently
  creating multiple redundant pending entries for the same underlying change.                                                                      
  2. A user who only holds the checker permission for a task (e.g. DISBURSE_LOAN_CHECKER, but not DISBURSE_LOAN) and who tries to submit that      
  action directly currently falls through to the generic NoAuthorizationException ("User has no authority to: ...") — which reads like a           
  permissions-configuration bug rather than telling the user they should be approving a pending entry instead of initiating one.                   
                                                                                                                                                   
  This PR adds two targeted, backwards-compatible checks in PortfolioCommandSourceWritePlatformServiceImpl#logCommandSource, only for tasks that   
  have maker-checker enabled:                                                                                                                      
                                                                                                                                                   
  - Duplicate pending submission: if the current user has the base permission (a "maker") but not the task's _CHECKER permission, and a command    
  already exists with status AWAITING_APPROVAL for the same action name, entity name and resource id, the new submission is rejected with a new    
  MakerCheckerDuplicatePendingSubmissionException (HTTP 409) instead of being queued as another pending entry.                                     
  - Checker-only initiation: if the current user lacks the base permission but holds the task's _CHECKER permission (or is a CHECKER_SUPER_USER),  
  the request is rejected with a new MakerCheckerCheckerOnlyInitiationException (HTTP 403) telling them to use the approval flow instead.
  Users who hold both the base and _CHECKER permission for a task, and any non maker-checker-enabled action, are unaffected. This PR intentionally 
  does not change approveEntry or introduce any new self-approval semantics — it only tightens the initiation path.  
  PR:(https://issues.apache.org/jira/browse/FINERACT-2753)

### DIFF
--- a/fineract-core/src/main/java/org/apache/fineract/commands/domain/CommandSourceRepository.java
+++ b/fineract-core/src/main/java/org/apache/fineract/commands/domain/CommandSourceRepository.java
@@ -19,6 +19,7 @@
 package org.apache.fineract.commands.domain;
 
 import java.time.OffsetDateTime;
+import java.util.List;
 import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.data.jpa.repository.JpaSpecificationExecutor;
 import org.springframework.data.jpa.repository.Modifying;
@@ -28,6 +29,9 @@ import org.springframework.data.repository.query.Param;
 public interface CommandSourceRepository extends JpaRepository<CommandSource, Long>, JpaSpecificationExecutor<CommandSource> {
 
     CommandSource findByActionNameAndEntityNameAndIdempotencyKey(String actionName, String entityName, String idempotencyKey);
+
+    List<CommandSource> findByActionNameAndEntityNameAndResourceIdAndStatus(String actionName, String entityName, Long resourceId,
+            Integer status);
 
     @Modifying(flushAutomatically = true)
     @Query("delete from CommandSource c where c.status = :status and c.madeOnDate is not null and c.madeOnDate <= :dateForPurgeCriteria")
--- a/fineract-core/src/main/java/org/apache/fineract/commands/exception/MakerCheckerCheckerOnlyInitiationException.java
+++ b/fineract-core/src/main/java/org/apache/fineract/commands/exception/MakerCheckerCheckerOnlyInitiationException.java
@@ -1,0 +1,34 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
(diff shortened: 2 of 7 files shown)


adamsaghy (changes requested):
Can you review the below concerns where the code does not deliver what's described in the PR description?

**1. The 409 is only half a 409. The description says the duplicate is "rejected with … (HTTP 409)".** MakerCheckerDuplicatePendingSubmissionExceptionMapper sets Response.status(CONFLICT), but builds the body with ApiGlobalErrorResponse.domainRuleViolation(...), which hardcodes SC_FORBIDDEN into the serialized httpStatusCode field (ApiGlobalErrorResponse.java:139) and stamps userMessageGlobalisationCode = "validation.msg.domain.rule.violation". So the response is 409 on the wire with "httpStatusCode": "403" in the payload. The batch API propagates both. -> **I think this should be 403, not 409**

**2. Stated problem #1 is not fixed for the configuration it describes.**
 The description opens with "A maker can resubmit the same action on the same resource any number of times while an earlier submission is still AWAITING_APPROVAL". The implementation exempts anyone holding _CHECKER — and holding _CHECKER does not stop a submission from becoming a pending entry. Only isCheckerSuperUser() does that (CommandSourceService.java:131). So a user with DISBURSE_LOAN + DISBURSE_LOAN_CHECKER is still a maker, still queues pending entries, and can still stack duplicates. That's the standard role shape — the repo's own MakercheckerTest.java:103 grants base + _CHECKER on one role. The exemption is stated in the description, so the code matches the description; the description's exemption just doesn't leave much of problem #1 fixed.

**3. "the same underlying change" is not what the key identifies.**
The description justifies the check as preventing "multiple redundant pending entries for the same underlying change", matched on "action name, entity name and resource id". The code does exactly that — but for commands where entityId is the parent and subentityId identifies the target, that key collapses distinct changes into one. UNDOTRANSACTION / REVERSETRANSACTION / ADJUSTTRANSACTION on SAVINGSACCOUNT set entityId = accountId, subentityId = transactionId (CommandWrapperBuilder.java:2072-2105); same shape for the loan-transaction variants (2726, 2845), UPDATE_CODEVALUE/DELETE_CODEVALUE (1729-1757), guarantor funding (580), cashier (3124). Two pending undos of different transactions on one savings account are different underlying changes, and the second gets a 409.

**4. "A maker can resubmit" — the code doesn't check who the maker is.** 

findByActionNameAndEntityNameAndResourceIdAndStatus has no maker filter, so maker B is blocked by maker A's pending entry. That may be the behaviour you want, but it isn't the behaviour described, and it isn't covered by a test.

**5. Undescribed carve-out.** 
wrapper.getEntityId() != null silently exempts every CREATE action from the duplicate check. Defensible (there's no resource id yet), but the description doesn't mention it and nothing in the code says why.

PR #6258, opened 2026-08-11
FINERACT-2755: Remove unimplemented Teller/Cashier read endpoints

AshharAhmadKhan:
JIRA
https://issues.apache.org/jira/browse/FINERACT-2755

Problem
Five methods on TellerManagementReadPlatformServiceImpl have been unimplemented stubs returning null since the original Cash Management commit in 2013: getCashierData, findTellerTransaction, fetchTellerTransactionsByTellerId, getJournals, fetchTellerJournals.

None of the three backing resource classes (CashierApiResource, TellerApiResource, TellerJournalApiResource) null-check the return value, so all five endpoints silently return HTTP 204 regardless of what is requested, including for resource IDs that do not exist.

Fix
CashierApiResource and TellerJournalApiResource have no other endpoints and are deleted entirely. TellerApiResource keeps its other working endpoints; only the three broken methods and their now-unused imports are removed from it.

Raised on the dev list before removal, no objections received:
https://lists.apache.org/thread/m972c6dgr46lob01smkryqwq6dxn7lms

Verification
All 15 existing TellerWritePlatformServiceJpaImplTest tests pass unchanged. compileJava and spotlessCheck both pass clean.


IOhacker (changes requested):
@AshharAhmadKhan why remove and not to complete?

AshharAhmadKhan:
> @AshharAhmadKhan why remove and not to complete?

No replies on dev mailing list and also james give his opinion on removal on matrix.

@IOhacker do you know anything about these endpoints? Your help is appreciated.

AshharAhmadKhan:
Investigated all three failures:

1. `api-compatibility-check` - expected. This is a breaking API change (endpoint removal), so this check is correctly flagging it.

2. `generated-client-api-method-compatibility-check` - also expected. The script has no allowlist for intentional removals, it hard-fails on any generated method name that disappears, which will always happen when endpoints are deleted instead of kept.

3. `run-integrations-tests-on-mysql / MySQL Tests (Shard 1 of 15)` - unrelated to this PR. Looks flaky, could someone re-trigger this check?

IOhacker (approved):

AshharAhmadKhan:
hey @adamsaghy, the remaining compatibility checks are expected due to the intentional endpoint removals. Could you please take a look when you get a chance? thanks a lot!

adamsaghy (approved):
LGTM

PR #6264, opened 2026-08-12
FINERACT-2455: payment rate change EIR history

budaidev:
## Description

## Checklist


fineract-working-capital-loan/src/main/resources/db/changelog/tenant/module/workingcapitalloan/parts/0075_wc_period_payment_rate_change_eir_history.xml
```diff
@@ -0,0 +1,46 @@
...
+         from a schedule model that has since been rewritten. The EIR keeps 12 decimals because two rate changes only
+         differ from the ninth decimal onwards. -->
+    <changeSet author="fineract" id="wcl-0070-1">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <columnExists tableName="m_wc_loan_period_payment_rate_change" columnName="eir"/>
+            </not>
+        </preConditions>
+        <addColumn tableName="m_wc_loan_period_payment_rate_change">
+            <column name="eir" type="DECIMAL(19,12)"/>
```
adamsaghy: Let's store annual effective interest rate, and use `DECIMAL(19,6)`.

galovics (changes requested):
The design is well thought through overall - the as-booked snapshot semantics, the additive/nullable migration, and the test coverage (7 integration tests plus solid e2e assertions) are all good. Two things I'd like fixed before this merges, plus a couple of questions.

**1. `rateSegmentAt` can silently resolve to the wrong rate change's segment.**

```java
public RateSegment rateSegmentAt(final LocalDate date) {
    return segmentForDay(splitDayIndexFor(date));
}
```

`splitDayIndexFor` clamps to `scheduleTerm()`, and `segmentForDay` returns the *last* segment whose `startDayIndex() <= dayIndex`. A rate increase shortens the schedule term, so a later-effective change that's already been clamped can resolve to the same split index as the change being booked right now. `applyRateChange` then removes segments at-or-after that index and drops the wrong one, and `recordCalculatedValues` persists another change's EIR/balance/term into this row - with no error, no log, nothing. Since `applyRateChange` always adds its own segment at exactly `splitDayIndex`, an exact-match guard closes this off cheaply:

```java
final int split = splitDayIndexFor(date);
final RateSegment seg = segmentForDay(split);
return seg != null && seg.startDayIndex() == split ? seg : null;
```

The existing `segment == null` fallback in `recordCalculatedValues` already handles the null case gracefully (leaves the snapshot unset rather than wrong).

**2. EIR is rounded with the tenant's money rounding mode, but a rate isn't money.** `MoneyHelper.getRoundingMode()` is tenant-configurable (UP/DOWN/HALF_*), so two tenants with identical inputs will store different EIRs for the same rate change. The javadoc right above this code says scales are fixed specifically so "API responses and event payloads carry the same value whichever database the tenant runs on" - the rounding mode undermines that same stated goal. A fixed `RoundingMode.HALF_UP` (which the test helper itself already uses) would match the comment's intent.

Smaller things: the snapshot in `recordCalculatedValues` relies on dirty checking rather than an explicit save, unlike every other write in that method - would be good for consistency and so a future refactor (e.g. splitting the regenerate call into its own transaction) can't silently drop it. Also, the row mixes a *restated* `previousRate` (rewritten by `restatePreviousRates` when a backdated change slots in) with a *never-restated* `eir` snapshot computed against whatever the predecessor was at booking time - your own feature file shows a row where `previousRate: 19.0` but the EIR is the one computed against a `previousRate` of `11.0`. Either both should be as-booked or both restated, otherwise I don't think a reader of the history can trust the row.

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/calc/ProjectedAmortizationScheduleModel.java
```diff
@@ -78,6 +78,9 @@ public final class ProjectedAmortizationScheduleModel {
      */
     public static final int MAX_CALCULABLE_TOTAL_DAYS = 100_000;
 
+    /** Compounding periods a daily EIR is raised over to express it annually. */
+    private static final int ANNUALIZATION_DAYS = 365;
```
adamsaghy: This hardcoded value seems incorrect:
npvDayCount from Loan product should be used.

NPV day count and number of days to calculate annual eir must be the same

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/domain/WorkingCapitalLoanPeriodPaymentRateChange.java
```diff
@@ -58,6 +64,26 @@ public class WorkingCapitalLoanPeriodPaymentRateChange extends AbstractAuditable
     @Column(name = "reversed_on_date")
     private LocalDate reversedOnDate;
 
+    /**
+     * Snapshot of what the amortization schedule computed for the segment this change opened, taken once when the
+     * change was booked. Nullable: changes booked before the snapshot existed never had one, and it cannot be
+     * reconstructed from a schedule model that has since been rewritten.
+     */
+    @Column(name = "eir", scale = EIR_SCALE, precision = 19)
+    private BigDecimal eir;
```
adamsaghy: Lets store the annual eir and use the same scaling and precision as all the other BigDecimal fields.

adamsaghy (changes requested):
Kindly review my concerns

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/calc/ProjectedAmortizationScheduleModel.java
budaidev: fixed

adamsaghy:
@budaidev Please rebase

adamsaghy:
> The design is well thought through overall - the as-booked snapshot semantics, the additive/nullable migration, and the test coverage (7 integration tests plus solid e2e assertions) are all good. Two things I'd like fixed before this merges, plus a couple of questions.
> 
> **1. `rateSegmentAt` can silently resolve to the wrong rate change's segment.**
> 
> ```java
> public RateSegment rateSegmentAt(final LocalDate date) {
>     return segmentForDay(splitDayIndexFor(date));
> }
> ```
> 
> `splitDayIndexFor` clamps to `scheduleTerm()`, and `segmentForDay` returns the _last_ segment whose `startDayIndex() <= dayIndex`. A rate increase shortens the schedule term, so a later-effective change that's already been clamped can resolve to the same split index as the change being booked right now. `applyRateChange` then removes segments at-or-after that index and drops the wrong one, and `recordCalculatedValues` persists another change's EIR/balance/term into this row - with no error, no log, nothing. Since `applyRateChange` always adds its own segment at exactly `splitDayIndex`, an exact-match guard closes this off cheaply:
> 
> ```java
> final int split = splitDayIndexFor(date);
> final RateSegment seg = segmentForDay(split);
> return seg != null && seg.startDayIndex() == split ? seg : null;
> ```
> 
> The existing `segment == null` fallback in `recordCalculatedValues` already handles the null case gracefully (leaves the snapshot unset rather than wrong).
> 
> **2. EIR is rounded with the tenant's money rounding mode, but a rate isn't money.** `MoneyHelper.getRoundingMode()` is tenant-configurable (UP/DOWN/HALF_*), so two tenants with identical inputs will store different EIRs for the same rate change. The javadoc right above this code says scales are fixed specifically so "API responses and event payloads carry the same value whichever database the tenant runs on" - the rounding mode undermines that same stated goal. A fixed `RoundingMode.HALF_UP` (which the test helper itself already uses) would match the comment's intent.
> 
> Smaller things: the snapshot in `recordCalculatedValues` relies on dirty checking rather than an explicit save, unlike every other write in that method - would be good for consistency and so a future refactor (e.g. splitting the regenerate call into its own transaction) can't silently drop it. Also, the row mixes a _restated_ `previousRate` (rewritten by `restatePreviousRates` when a backdated change slots in) with a _never-restated_ `eir` snapshot computed against whatever the predecessor was at booking time - your own feature file shows a row where `previousRate: 19.0` but the EIR is the one computed against a `previousRate` of `11.0`. Either both should be as-booked or both restated, otherwise I don't think a reader of the history can trust the row.
> 

@budaidev Have you had the chance to review these concerns?

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/domain/WorkingCapitalLoanPeriodPaymentRateChange.java
budaidev: fixed

budaidev:
> > The design is well thought through overall - the as-booked snapshot semantics, the additive/nullable migration, and the test coverage (7 integration tests plus solid e2e assertions) are all good. Two things I'd like fixed before this merges, plus a couple of questions.
> > **1. `rateSegmentAt` can silently resolve to the wrong rate change's segment.**
> > ```java
> > public RateSegment rateSegmentAt(final LocalDate date) {
> >     return segmentForDay(splitDayIndexFor(date));
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `splitDayIndexFor` clamps to `scheduleTerm()`, and `segmentForDay` returns the _last_ segment whose `startDayIndex() <= dayIndex`. A rate increase shortens the schedule term, so a later-effective change that's already been clamped can resolve to the same split index as the change being booked right now. `applyRateChange` then removes segments at-or-after that index and drops the wrong one, and `recordCalculatedValues` persists another change's EIR/balance/term into this row - with no error, no log, nothing. Since `applyRateChange` always adds its own segment at exactly `splitDayIndex`, an exact-match guard closes this off cheaply:
> > ```java
> > final int split = splitDayIndexFor(date);
> > final RateSegment seg = segmentForDay(split);
> > return seg != null && seg.startDayIndex() == split ? seg : null;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The existing `segment == null` fallback in `recordCalculatedValues` already handles the null case gracefully (leaves the snapshot unset rather than wrong).
> > **2. EIR is rounded with the tenant's money rounding mode, but a rate isn't money.** `MoneyHelper.getRoundingMode()` is tenant-configurable (UP/DOWN/HALF_*), so two tenants with identical inputs will store different EIRs for the same rate change. The javadoc right above this code says scales are fixed specifically so "API responses and event payloads carry the same value whichever database the tenant runs on" - the rounding mode undermines that same stated goal. A fixed `RoundingMode.HALF_UP` (which the test helper itself already uses) would match the comment's intent.
> > Smaller things: the snapshot in `recordCalculatedValues` relies on dirty checking rather than an explicit save, unlike every other write in that method - would be good for consistency and so a future refactor (e.g. splitting the regenerate call into its own transaction) can't silently drop it. Also, the row mixes a _restated_ `previousRate` (rewritten by `restatePreviousRates` when a backdated change slots in) with a _never-restated_ `eir` snapshot computed against whatever the predecessor was at booking time - your own feature file shows a row where `previousRate: 19.0` but the EIR is the one computed against a `previousRate` of `11.0`. Either both should be as-booked or both restated, otherwise I don't think a reader of the history can trust the row.
> > Recommendation: CHANGES_REQUESTED
> 
> @budaidev Have you had the chance to review these concerns?

1. `rateSegmentAt` — implemented as you suggested
2. Fixed.
+1 . Question — restated `previousRate` vs as-booked `eir`
  Of the two options you offered, I don't think either is quite right on its own:
  
  - *Both as-booked* (stop restating `previousRate`) is the cleanest audit model, but restatement exists
    deliberately so the history reads as one chain in effective-date order, and dropping it changes the meaning
    of an existing REST/Avro field.
  - *Both restated* isn't reliably possible — the EIR snapshot can't be recomputed once the schedule model has
    been rewritten, which is the whole reason it's stored rather than derived.
    
    What I'd propose instead is to keep both facts and stop them being confusable: add an immutable
  `asBookedPreviousRate` next to the restated `previousRate`, giving two explicit tuples: 
  - as-booked audit: `asBookedPreviousRate` + `newRate` + EIR and the derived snapshot
  - current effective chain: `previousRate` + `newRate` + `effectiveDate`

adamsaghy:
> > > The design is well thought through overall - the as-booked snapshot semantics, the additive/nullable migration, and the test coverage (7 integration tests plus solid e2e assertions) are all good. Two things I'd like fixed before this merges, plus a couple of questions.
> > > **1. `rateSegmentAt` can silently resolve to the wrong rate change's segment.**
> > > ```java
> > > public RateSegment rateSegmentAt(final LocalDate date) {
> > >     return segmentForDay(splitDayIndexFor(date));
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > `splitDayIndexFor` clamps to `scheduleTerm()`, and `segmentForDay` returns the _last_ segment whose `startDayIndex() <= dayIndex`. A rate increase shortens the schedule term, so a later-effective change that's already been clamped can resolve to the same split index as the change being booked right now. `applyRateChange` then removes segments at-or-after that index and drops the wrong one, and `recordCalculatedValues` persists another change's EIR/balance/term into this row - with no error, no log, nothing. Since `applyRateChange` always adds its own segment at exactly `splitDayIndex`, an exact-match guard closes this off cheaply:
> > > ```java
> > > final int split = splitDayIndexFor(date);
> > > final RateSegment seg = segmentForDay(split);
> > > return seg != null && seg.startDayIndex() == split ? seg : null;
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > The existing `segment == null` fallback in `recordCalculatedValues` already handles the null case gracefully (leaves the snapshot unset rather than wrong).
> > > **2. EIR is rounded with the tenant's money rounding mode, but a rate isn't money.** `MoneyHelper.getRoundingMode()` is tenant-configurable (UP/DOWN/HALF_*), so two tenants with identical inputs will store different EIRs for the same rate change. The javadoc right above this code says scales are fixed specifically so "API responses and event payloads carry the same value whichever database the tenant runs on" - the rounding mode undermines that same stated goal. A fixed `RoundingMode.HALF_UP` (which the test helper itself already uses) would match the comment's intent.
> > > Smaller things: the snapshot in `recordCalculatedValues` relies on dirty checking rather than an explicit save, unlike every other write in that method - would be good for consistency and so a future refactor (e.g. splitting the regenerate call into its own transaction) can't silently drop it. Also, the row mixes a _restated_ `previousRate` (rewritten by `restatePreviousRates` when a backdated change slots in) with a _never-restated_ `eir` snapshot computed against whatever the predecessor was at booking time - your own feature file shows a row where `previousRate: 19.0` but the EIR is the one computed against a `previousRate` of `11.0`. Either both should be as-booked or both restated, otherwise I don't think a reader of the history can trust the row.
> > > Recommendation: CHANGES_REQUESTED
> > 
> > 
> > @budaidev Have you had the chance to review these concerns?
> 
> 1. `rateSegmentAt` — implemented as you suggested
> 2. Fixed.
>    +1 . Question — restated `previousRate` vs as-booked `eir`
>    Of the two options you offered, I don't think either is quite right on its own:
> 
> * _Both as-booked_ (stop restating `previousRate`) is the cleanest audit model, but restatement exists
>   deliberately so the history reads as one chain in effective-date order, and dropping it changes the meaning
>   of an existing REST/Avro field.
> * _Both restated_ isn't reliably possible — the EIR snapshot can't be recomputed once the schedule model has
>   been rewritten, which is the whole reason it's stored rather than derived.
>   What I'd propose instead is to keep both facts and stop them being confusable: add an immutable
>   `asBookedPreviousRate` next to the restated `previousRate`, giving two explicit tuples:
> * as-booked audit: `asBookedPreviousRate` + `newRate` + EIR and the derived snapshot
> * current effective chain: `previousRate` + `newRate` + `effectiveDate`

Sounds good to me to store these values as part of the previous rate to avoid confusion and unnecessary recalculation.

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/domain/WorkingCapitalLoanPeriodPaymentRateChange.java
```diff
@@ -40,6 +41,20 @@
...
+    private static final int AMOUNT_SCALE = 6;
+
+    /** The annualised rate is a headline figure; six decimals match the rate columns either side of it. */
+    private static final int ANNUAL_EIR_SCALE = 6;
+
+    /**
+     * Fixed rather than the tenant's money rounding mode: these are rates and derived figures, not money, and the whole
+     * point of pinning the scales here is that the same inputs yield the same stored value for every tenant.
+     */
+    private static final RoundingMode SNAPSHOT_ROUNDING = RoundingMode.HALF_UP;
```
adamsaghy: HALF_EVEN is better for monetary operations:
_"It is widely known as Banker's Rounding or convergent rounding, and it represents the default rounding logic specified by the IEEE 754 standard for floating-point math"_

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/modules/WorkingCapitalLoanRateHistoryValidators.java
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Asserts a numeric field equals the expected decimal, comparing at the expected value's scale (HALF_UP) so an
+     * implementation carrying more precision than asserted (the 6-dp columns returning {@code 8060.040000}) still
+     * matches.
+     */
+    public static void validateDecimal(final String label, final String field, final BigDecimal actual, final String expected) {
+        assertNotNull(actual, label + ": expected '" + field + "' = " + expected + " but it was null");
+        final BigDecimal expectedValue = new BigDecimal(expected);
+        final BigDecimal rounded = actual.setScale(expectedValue.scale(), RoundingMode.HALF_UP);
```
adamsaghy: HALF_EVEN is better for monetary operations:
_"It is widely known as Banker's Rounding or convergent rounding, and it represents the default rounding logic specified by the IEEE 754 standard for floating-point math"_

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/tests/FeignWorkingCapitalLoanRateChangeEirHistoryTest.java
```diff
@@ -0,0 +1,387 @@
...
+        final String uniqueShortName = Utils.uniqueRandomStringGenerator("", 4);
+        // discountDefault override lets the application state the discount of 1000 explicitly
+        return productHelper
+                .createWorkingCapitalLoanProduct(new WorkingCapitalLoanProductTestBuilder().withName(uniqueName)
+                        .withShortName(uniqueShortName).withAllowAttributeOverrides(Map.of("discountDefault", Boolean.TRUE)).build())
+                .getResourceId();
+    }
+
+    private static BigDecimal round(final BigDecimal value, final int scale) {
+        return value == null ? null : value.setScale(scale, RoundingMode.HALF_UP);
```
adamsaghy: HALF_EVEN is better for monetary operations:
_"It is widely known as Banker's Rounding or convergent rounding, and it represents the default rounding logic specified by the IEEE 754 standard for floating-point math"_

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/domain/WorkingCapitalLoanPeriodPaymentRateChange.java
budaidev: Agreed — changed to `HALF_EVEN`.

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/modules/WorkingCapitalLoanRateHistoryValidators.java
budaidev: Done — `HALF_EVEN`.

The javadoc now says the comparison uses the same rounding the snapshot is stored with, which is the actual reason the test should follow production here rather than picking a mode of its own.

integration-tests/src/test/java/org/apache/fineract/integrationtests/client/feign/tests/FeignWorkingCapitalLoanRateChangeEirHistoryTest.java
budaidev: Done — same change to the local `round(...)` helper, for the same reason as the validator above.

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/domain/WorkingCapitalLoanPeriodPaymentRateChange.java
adamsaghy: yeah.. those are incorrect. let me fix them in separate PR :(

adamsaghy:
@budaidev Please pull latest changes for your branch then rebase with `develop` branch.

adamsaghy:
@budaidev Please review the failing checks.

adamsaghy:
@budaidev Please review the failing test:

```
org.apache.fineract.integrationtests.client.feign.tests.FeignWorkingCapitalLoanRateChangeEirHistoryTest
  
    Test A future-dated rate change snapshots its calculated values already at booking time PASSED
    Test A backdated rate change leaves the earlier-booked change's snapshot as-booked (audit trail, no restatement) FAILED
  
    org.opentest4j.AssertionFailedError: 20% change (2019-02-01) booked after the 17% one: 'eir' expected 0.001169946129 but was 0.001185871519 ==> expected: <0> but was: <-1>
        at app//org.apache.fineract.integrationtests.client.feign.tests.FeignWorkingCapitalLoanRateChangeEirHistoryTest.lambda$backdatedRateChange_keepsEarlierBookedSnapshotAsBooked$0(FeignWorkingCapitalLoanRateChangeEirHistoryTest.java:329)

File: fineract-e2e-tests-runner/src/test/resources/features/WorkingCapitalPeriodPaymentRate.feature [hunk 7]
```diff
@@ -371,6 +371,10 @@ Feature: Working Capital Period Payment Rate
       | Effective Date  | Previous Rate | New Rate | Reversed |
       | 10 January 2026 | 18.0          | 11.0     | false    |
       | 20 January 2026 | 11.0          | 20.0     | false    |
+    And Working capital loan account has the correct period payment rate history data:
+      | effectiveDate   | previousRate | newRate | eir            | calculatedAnnualEIR | dailyPaymentAmount | discountAmount | outstandingBalanceAtSplit | segmentTerm | reversed |
+      | 10 January 2026 | 18.0         | 11.0    | 0.000653823564 | 0.265288            | 30.56              | 1000.0         | 9000.0                    | 328         | false    |
+      | 20 January 2026 | 11.0         | 20.0    | 0.001185944716 | 0.532173            | 55.56              | 1000.0         | 9000.0                    | 180         | false    |
```
adamsaghy: `calculatedAnnualEIR` looks way too low.

fineract-e2e-tests-runner/src/test/resources/features/WorkingCapitalPeriodPaymentRate.feature
```diff
@@ -371,6 +371,10 @@ Feature: Working Capital Period Payment Rate
       | Effective Date  | Previous Rate | New Rate | Reversed |
       | 10 January 2026 | 18.0          | 11.0     | false    |
       | 20 January 2026 | 11.0          | 20.0     | false    |
+    And Working capital loan account has the correct period payment rate history data:
+      | effectiveDate   | previousRate | newRate | calculatedAnnualEIR | dailyPaymentAmount | segmentTerm | reversed |
+      | 10 January 2026 | 18.0         | 11.0    | 0.265288            | 30.56              | 328         | false    |
```
adamsaghy: 0.265288 looks way too low... it should percentage, like 26.5288... that way we truly can have the required 6 precision after the decimal point

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/domain/WorkingCapitalLoanPeriodPaymentRateChange.java
```diff
@@ -62,6 +62,16 @@ public class WorkingCapitalLoanPeriodPaymentRateChange extends AbstractAuditable
     @Column(name = "submitted_on_date", nullable = false)
     private LocalDate submittedOnDate;
 
+    /** Snapshot taken when the change was booked, never restated; null for changes booked before it existed. */
+    @Column(name = "calculated_annual_eir", scale = 6, precision = 19)
+    private BigDecimal calculatedAnnualEir;
```
adamsaghy, line 70: This should store the percentage amount like "26.123456"

adamsaghy (changes requested):
Kindly review my concerns

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/domain/WorkingCapitalLoanPeriodPaymentRateChange.java
budaidev: fixed

fineract-e2e-tests-runner/src/test/resources/features/WorkingCapitalPeriodPaymentRate.feature
budaidev: fixed to percentage

adamsaghy:
@budaidev Please rebase

galovics (dismissed):
Both blockers from last round are properly fixed, and the `rateSegmentAt` fix is better than what I originally asked for - it now records exactly what each rate change solved to, keyed by its own effective date, and returns null on a miss rather than a neighbor's numbers, with a test pinning that exact case. The tenant-rounding concern is resolved by removing the contradictory rescale entirely.

But fixing it surfaced a new problem: nothing explicitly rounds the EIR before persisting anymore, so the value depends on whatever the DB does with a 19-significant-digit `BigDecimal` against a `DECIMAL(19,6)` column. Concretely, the business event is raised in the same transaction as the calculation and reads the still-managed, unrounded entity (then widens to 8dp for Avro), while a later GET reads back the DB-rounded 6dp value - so the event and the API can report different numbers for the same rate change. A single explicit `setScale(6, RoundingMode.HALF_EVEN)` at the point of calculation would close this.

The `previousRate`/EIR inconsistency I raised originally (a history row can carry a restated `previousRate` next to a never-restated EIR snapshot computed against a different predecessor) is still present - the Swagger/Avro docs now explain the as-booked semantics, which helps a reader, but the underlying row still carries two different definitions of "before." I'd want an explicit decision here (restate both or neither) rather than leaving it documented-but-unresolved.

And the collision with #6343 is still open, and it's compounded - both PRs are re-deriving the same formula in different classes, and now #6343 is also making `calculatedAnnualEir` a percentage on this same endpoint while this PR keeps it a fraction. Whoever merges second is going to have a bad time, and one of the two fixes to the underlying `pow(365)` bug is going to get lost.

Two smaller things: there's an unrelated CI workflow change (`SKIP_SDK_GEN`) bundled in that skips building the avro-schemas SDK, which seems risky specifically in a PR that adds a new avro schema - and a test using `ReflectionTestUtils.invokeMethod` on a private method by string name, which will silently stop testing anything on the next rename.

Recommendation: CHANGES_REQUESTED (down from before - the original two issues are resolved, but the #6343 collision plus the new rounding-divergence bug keep this from being clean)

budaidev:
> Both blockers from last round are properly fixed, and the `rateSegmentAt` fix is better than what I originally asked for - it now records exactly what each rate change solved to, keyed by its own effective date, and returns null on a miss rather than a neighbor's numbers, with a test pinning that exact case. The tenant-rounding concern is resolved by removing the contradictory rescale entirely.
> 
> But fixing it surfaced a new problem: nothing explicitly rounds the EIR before persisting anymore, so the value depends on whatever the DB does with a 19-significant-digit `BigDecimal` against a `DECIMAL(19,6)` column. Concretely, the business event is raised in the same transaction as the calculation and reads the still-managed, unrounded entity (then widens to 8dp for Avro), while a later GET reads back the DB-rounded 6dp value - so the event and the API can report different numbers for the same rate change. A single explicit `setScale(6, RoundingMode.HALF_EVEN)` at the point of calculation would close this.
> 
> The `previousRate`/EIR inconsistency I raised originally (a history row can carry a restated `previousRate` next to a never-restated EIR snapshot computed against a different predecessor) is still present - the Swagger/Avro docs now explain the as-booked semantics, which helps a reader, but the underlying row still carries two different definitions of "before." I'd want an explicit decision here (restate both or neither) rather than leaving it documented-but-unresolved.
> 
> And the collision with #6343 is still open, and it's compounded - both PRs are re-deriving the same formula in different classes, and now #6343 is also making `calculatedAnnualEir` a percentage on this same endpoint while this PR keeps it a fraction. Whoever merges second is going to have a bad time, and one of the two fixes to the underlying `pow(365)` bug is going to get lost.
> 
> Two smaller things: there's an unrelated CI workflow change (`SKIP_SDK_GEN`) bundled in that skips building the avro-schemas SDK, which seems risky specifically in a PR that adds a new avro schema - and a test using `ReflectionTestUtils.invokeMethod` on a private method by string name, which will silently stop testing anything on the next rename.
> 
> Recommendation: CHANGES_REQUESTED (down from before - the original two issues are resolved, but the #6343 collision plus the new rounding-divergence bug keep this from being clean)

@galovics thanks. The rounding divergence is fixed: annualEirPercentage is now rounded at the point of calculation (setScale(6, HALF_EVEN), constants on ProjectedAmortizationScheduleModel), so the entity the business event serialises already carries the value the column stores, and it is the same convention #6343 uses. On previousRate vs the EIR snapshot, the decision was made but never reached the PR: previousRate stays the restated chain link in effective-date order, while the three snapshot fields (calculatedAnnualEir, dailyPaymentAmount, segmentTerm) are as-booked and never restated, because the schedule model is rewritten by the change and cannot be recomputed later; the Swagger/Avro docs say so.  The SKIP_SDK_GEN workflow change is removed from this PR and recordCalculatedValues is package-private now with the test calling it directly, so a rename breaks the compile instead of the test.

fineract-e2e-tests-runner/src/test/resources/features/WorkingCapitalPeriodPaymentRate.feature
```diff
@@ -631,6 +653,10 @@ Feature: Working Capital Period Payment Rate
       | 1         | 2026-01-02 | 47.22                 | 8952.91         | 0.13                       | 11.87                      |
       | 221       | 2026-08-10 | 52.78                 | 8781.12         | 0.14                       | 11.42                      |
       | 388       | 2027-01-24 | 31.06                 | 0.00            | 0.00                       | 0.00                       |
+    And Working capital loan account has the correct period payment rate history data:
+      | effectiveDate  | previousRate | newRate | calculatedAnnualEIR | dailyPaymentAmount | segmentTerm | reversed |
```
adamsaghy: `0.560191` This seems way too low.

fineract-e2e-tests-runner/src/test/resources/features/WorkingCapitalPeriodPaymentRate.feature
```diff
@@ -1840,6 +1873,9 @@ Feature: Working Capital Period Payment Rate
       | 04 January 2026 | Repayment                 | 50.0              | 50.0             | 0.0               | 0.0                   | false    |
       | 04 January 2026 | Discount Fee Amortization | 8.45              |                  |                   |                       | false    |
       | 21 January 2026 | Repayment                 | 55.56             | 55.56            | 0.0               | 0.0                   | false    |
+    And Working capital loan account has the correct period payment rate history data:
+      | effectiveDate   | previousRate | newRate | calculatedAnnualEIR | dailyPaymentAmount | segmentTerm | reversed |
+      | 20 January 2026 | 18.0         | 20.0    | 2740.680185         | 55.56              | 19          | false    |
```
adamsaghy: This seems way too big

fineract-e2e-tests-runner/src/test/resources/features/WorkingCapitalPeriodPaymentRate.feature
```diff
@@ -1944,6 +1980,9 @@ Feature: Working Capital Period Payment Rate
       | 01 January 2026 | Disbursement              | 1000.0            | 1000.0           | 0.0               | 0.0                   | false    |
       | 01 January 2026 | Discount Fee              | 100.0             | 100.0            | 0.0               | 0.0                   | false    |
       | 21 January 2026 | Repayment                 | 55.56             | 55.56            | 0.0               | 0.0                   | false    |
+    And Working capital loan account has the correct period payment rate history data:
+      | effectiveDate   | previousRate | newRate | calculatedAnnualEIR | dailyPaymentAmount | segmentTerm | reversed |
+      | 20 January 2026 | 18.0         | 20.0    | 2741.704245         | 55.56              | 20          | false    |
```
adamsaghy: Same here

adamsaghy (approved):
LGTM

PR #6282, opened 2026-08-18
FINERACT-2762: Client Approval/Activation Fails for Branch Manager Despite Having Required Permissions

rymghosn:
For a maker-checker-enabled task, a user holding only the <TASK>_CHECKER permission (no base permission) was refused with a generic "not         
  authorized" response when calling the entity's normal action endpoint — even when a matching maker submission was already awaiting exactly this  
  user's approval. The only way such a user could act was through the separate checker-inbox approval endpoint (POST                               
  .../commands/{id}?command=approve); there was no fallback when a checker-only user instead called the same action endpoint a maker would use     
  (e.g. POST /clients/{clientId}?command=activate), which is the natural path from the entity's own screen.                                        
                                                                                                                                                   
  PortfolioCommandSourceWritePlatformServiceImpl#logCommandSource now checks, for a checker-only caller, whether a pending CommandSource exists for
  the same action/entity/resource, and approves it directly if found. If none exists, the call is refused with the existing permission-denied      
  response, unchanged. Maker flows and users holding both the base and checker permission are unaffected.      
  PR:(https://issues.apache.org/jira/browse/FINERACT-2762)

### DIFF
--- a/fineract-core/src/main/java/org/apache/fineract/commands/domain/CommandSourceRepository.java
+++ b/fineract-core/src/main/java/org/apache/fineract/commands/domain/CommandSourceRepository.java
@@ -19,6 +19,7 @@
 package org.apache.fineract.commands.domain;
 
 import java.time.OffsetDateTime;
+import java.util.List;
 import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.data.jpa.repository.JpaSpecificationExecutor;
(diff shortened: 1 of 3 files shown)


adamsaghy:
@rymghosn Is this duplicate with https://github.com/apache/fineract/pull/6255?

PR #6283, opened 2026-08-18
FINERACT-2455: working capital breach reset split period undo

budaidev:
## Description

## Checklist

### DIFF
--- a/fineract-e2e-tests-runner/src/test/resources/features/WorkingCapitalBreachReset.feature
+++ b/fineract-e2e-tests-runner/src/test/resources/features/WorkingCapitalBreachReset.feature
@@ -818,3 +818,259 @@ Feature: Working Capital Breach Reset and Undo Reset
       | action | startDate       |
       | RESET  | 07 January 2026 |
     Then Admin closes the Working Capital loan with a full repayment on "07 January 2026"
+
+  @TestRailId:C102384
+  Scenario: Verify WC breach reset split current period: undo reset with backdated repayment reverts the period split created by the restart period option - UC6
+    When Admin sets the business date to "01 January 2026"
+    And Admin creates a client with random data
+    And Admin creates a Working Capital Loan Product with custom breach config and overrides enabled:
+      | breachFrequency | breachFrequencyType | breachAmountCalculationType | breachAmount | delinquencyGraceDays |
+      | 6               | DAYS                | PERCENTAGE                  | 50           |                      |
+    And Admin creates a working capital loan using created product with the following data:
+      | submittedOnDate | expectedDisbursementDate | principalAmount | totalPaymentVolume | periodPaymentRate | discount |
+      | 01 January 2026 | 01 January 2026          | 800             | 10000              | 18                | 0        |
+    And Admin successfully approves the working capital loan on "01 January 2026" with "800" amount and expected disbursement date on "01 January 2026"
+    When Admin successfully disburse the Working Capital loan on "01 January 2026" with "800" EUR transaction amount
+    And Admin runs inline COB job for Working Capital Loan by loanId
+    When Admin sets the business date to "03 January 2026"
+    And Customer makes repayment on "03 January 2026" with 200.0 transaction amount on Working Capital loan
+    When Admin sets the business date to "05 January 2026"
+    And Admin runs inline COB job for Working Capital Loan by loanId
+    And Customer makes repayment on "05 January 2026" with 100.0 transaction amount on Working Capital loan
+    When Admin sets the business date to "08 January 2026"
+    And Admin runs inline COB job for Working Capital Loan by loanId
+    When Admin sets the business date to "10 January 2026"
+    And Admin runs inline COB job for Working Capital Loan by loanId
+    # --- Breach reset with restart period from reset date splits period 2 ---
+    And Admin creates WC breach reset action with restart period from reset date
+    Then Working Capital loan breach schedule has the following data:
+      | periodNumber | fromDate   | toDate     | numberOfDays | minPaymentAmount | outstandingAmount | breach | reset |
+      | 1            | 2026-01-01 | 2026-01-06 | 6            | 400.00           | 100.00            | true   | false |
+      | 2            | 2026-01-07 | 2026-01-09 | 3            | 400.00           | 400.00            | true   | false |
+      | 3            | 2026-01-10 | 2026-01-15 | 6            | 400.00           | 400.00            | null   | true  |
+    And Working Capital loan balance has breach past due amount "0"
+    When Admin sets the business date to "16 January 2026"
+    And Admin runs inline COB job for Working Capital Loan by loanId
+    And Customer makes repayment on "14 January 2026" with 150.0 transaction amount on Working Capital loan
+    Then Working Capital loan breach schedule has the following data:
+      | periodNumber | fromDate   | toDate     | numberOfDays | minPaymentAmount | outstandingAmount | breach | reset |
+      | 1            | 2026-01-01 | 2026-01-06 | 6            | 400.00           | 100.00            | true   | false |
+      | 2            | 2026-01-07 | 2026-01-09 | 3            | 400.00           | 400.00            | true   | false |
+      | 3            | 2026-01-10 | 2026-01-15 | 6            | 400.00           | 250.00            | true   | true  |
+      | 4            | 2026-01-16 | 2026-01-21 | 6            | 350.00           | 350.00            | null   | false |
+    And Working Capital loan balance has breach past due amount "250"
+    When Admin sets the business date to "17 January 2026"
+    And Admin runs inline COB job for Working Capital Loan by loanId
+    # --- Undo the reset: the period split created by the restart option is reverted, period 3 is restored ---
+    When Admin creates WC breach undo reset action
+    Then Working Capital loan breach schedule has the following data:
+      | periodNumber | fromDate   | toDate     | numberOfDays | minPaymentAmount | outstandingAmount | breach | reset |
+      | 1            | 2026-01-01 | 2026-01-06 | 6            | 400.00           | 100.00            | true   | false |
+      | 2            | 2026-01-07 | 2026-01-12 | 6            | 400.00           | 400.00            | true   | false |
+      | 3            | 2026-01-13 | 2026-01-18 | 6            | 400.00           | 250.00            | null   | false |
+    And Working Capital loan balance has breach past due amount "500"
+    And WC loan breach actions have the following data:
+      | action     | startDate       |
+      | RESET      | 10 January 2026 |
+      | UNDO_RESET | 17 January 2026 |
+    Then Admin closes the Working Capital loan with a full repayment on "17 January 2026"
+
+  @TestRailId:C102385
+  Scenario: Verify WC breach reset split current period: undo reset with repayment reverts the period split created by the restart period option - UC7
+    When Admin sets the business date to "01 January 2026"
+    And Admin creates a client with random data
+    And Admin creates a Working Capital Loan Product with custom breach config and overrides enabled:
+      | breachFrequency | breachFrequencyType | breachAmountCalculationType | breachAmount | delinquencyGraceDays |
+      | 6               | DAYS                | PERCENTAGE                  | 50           |                      |
+    And Admin creates a working capital loan using created product with the following data:
+      | submittedOnDate | expectedDisbursementDate | principalAmount | totalPaymentVolume | periodPaymentRate | discount |
+      | 01 January 2026 | 01 January 2026          | 800             | 10000              | 18                | 0        |
+    And Admin successfully approves the working capital loan on "01 January 2026" with "800" amount and expected disbursement date on "01 January 2026"
+    When Admin successfully disburse the Working Capital loan on "01 January 2026" with "800" EUR transaction amount
+    And Admin runs inline COB job for Working Capital Loan by loanId
+    When Admin sets the business date to "03 January 2026"
+    And Customer makes repayment on "03 January 2026" with 200.0 transaction amount on Working Capital loan
+    When Admin sets the business date to "05 January 2026"
+    And Admin runs inline COB job for Working Capital Loan by loanId
+    And Customer makes repayment on "05 January 2026" with 100.0 transaction amount on Working Capital loan
+    When Admin sets the business date to "08 January 2026"
+    And Admin runs inline COB job for Working Capital Loan by loanId
+    When Admin sets the business date to "10 January 2026"
+    And Admin runs inline COB job for Working Capital Loan by loanId
+    # --- Breach reset with restart period from reset date splits period 2 ---
+    And Admin creates WC breach reset action with restart period from reset date
+    Then Working Capital loan breach schedule has the following data:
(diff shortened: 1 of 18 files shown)


galovics (commented):
The undo path is well covered - unit tests on both the resolver and the schedule methods, plus six integration tests including the pause-extended and payment-reassignment cases. That's usually the part I have to ask for, so nice work there.

My concern is with how the undo figures out what to revert. `restoreSplitPeriod` reverse-engineers the pre-split state from the schedule as it stands now plus the current breach config, using three guard clauses to decide whether the split it's looking at is actually the one this undo owns. Two of those guards fall through to a `log.warn` and a `false`, and the caller then just recalculates past due and returns 200 - so the user gets a success response for an undo that only lifted a flag and left the split in place.

Concretely: a RESCHEDULE recorded between the reset and the undo changes `resolveEffectiveRescheduleParams`, so the restore recomputes a boundary that never existed. And any PAUSE/RESUME between reset and undo rewrites every period contiguously from the schedule's first `fromDate` without knowing about the split, so the split disappears from the schedule entirely while the RESET row still says `restartPeriodFromResetDate = true` - the undo then silently falls into the flag-only branch.

Wouldn't it be simpler and deterministic to record on the RESET action row what the reset actually changed (the pre-split `toDate`, maybe the period id), and have the undo just write that back? Then there's nothing to guess and nothing that can silently degrade. The same reconstruction problem shows up in `findLatestFlaggedPeriod`'s "last flagged period by period number" fallback in the reset service - it assumes period-number order tracks reset-date order, which is exactly the assumption the boundary rewrites can break.

Two smaller things while I was in there:
- `splitPeriodAtReset` only guards `resetDate == fromDate`; `resetDate < fromDate` falls through and can persist a period with `toDate < fromDate` / negative `numberOfDays`.
- The RESET-undo replay in `WorkingCapitalLoanActiveBreachResetResolver` is ordered by `startDate, id` - since reset/undo dates are always "the current business date," if the business date is ever moved backwards between two actions the replay order diverges from creation order, and the `break`-on-id can stop before the reset it's meant to cancel.

This is a draft so nothing's blocking yet, but I'd like to settle the design question (store vs. reconstruct) before it comes out of draft - it affects several of the methods above at once.

Recommendation: COMMENT

adamsaghy:
**Can you please verify the below use cases?**

WorkingCapitalLoanBreachScheduleServiceImpl.java:277 : a PAUSE recorded after a restart reset makes recalculatePeriodsForPauses re-date every period from the schedule start, wiping the split while leaving reset=true on a period that no longer contains the reset date. Probe: P2 goes back to 70 days and P3 shifts to [2026-05-11..2026-07-09] still flagged, so past due is computed from the wrong row. Pre-existing code, but it makes the split non-durable and this PR is what gives the split meaning: the pause/resume handler should re-derive flags via the new applyResetFlags.
WorkingCapitalLoanBreachResetServiceImpl.java:53; PLAUSIBLE (also line 76): applyResetFlags runs before reprocessBreachSchedule regenerates periods, so an active reset whose date falls outside the mid-operation schedule gets pinned to the wrong row by resolveEvaluationPeriod's fallback. Needs a backwards business-date move to trigger, which the activeResets javadoc explicitly says can happen. Apply the flags after the schedule is final.

Some minor:

restoreSplitPeriod filters reschedules to those recorded before the reset (beforeActionId) but passes the unfiltered pause list to applyPauses. That happens to be consistent today because a pause already re-dates all periods, but the asymmetry is worth a comment.
The WorkingCapitalLoanAccount.feature scenario C102391 (modify with invalid product id) and its step def are unrelated to breach reset undo. I traced the path and the expectation holds: validateForUpdate collects the longGreaterThanZero() error but productRepository.findById(...).orElseThrow(WorkingCapitalLoanProductNotFoundException) fires first, so 0 and -1 do return 404, but it's scope creep in a PR the checklist asks not to be a code dump. WorkingCapitalBreachReset.feature also lost its trailing newline (not spotless-enforced for .feature, so cosmetic only).

adamsaghy:
@budaidev Please rebase

budaidev:
> **Can you please verify the below use cases?**
> 
> WorkingCapitalLoanBreachScheduleServiceImpl.java:277 : a PAUSE recorded after a restart reset makes recalculatePeriodsForPauses re-date every period from the schedule start, wiping the split while leaving reset=true on a period that no longer contains the reset date. Probe: P2 goes back to 70 days and P3 shifts to [2026-05-11..2026-07-09] still flagged, so past due is computed from the wrong row. Pre-existing code, but it makes the split non-durable and this PR is what gives the split meaning: the pause/resume handler should re-derive flags via the new applyResetFlags. WorkingCapitalLoanBreachResetServiceImpl.java:53; PLAUSIBLE (also line 76): applyResetFlags runs before reprocessBreachSchedule regenerates periods, so an active reset whose date falls outside the mid-operation schedule gets pinned to the wrong row by resolveEvaluationPeriod's fallback. Needs a backwards business-date move to trigger, which the activeResets javadoc explicitly says can happen. Apply the flags after the schedule is final.

1. PAUSE after a restart reset (`recalculatePeriodsForPauses`)** - confirmed, exactly your numbers.
`recalculatePeriodsForPauses` is unchanged on this branch (identical to develop), so the erasure is pre-existing, but agreed that this PR is what gives the split meaning. Fixed: the pause recalculation now re-derives the flags (via `applyResetFlags` over the persisted actions) before it recalculates the past due, so the flag and the past-due anchor follow the reset date to its re-dated row.

2. applyResetFlags fixed by moving the derivation into the schedule service, which is the one reshaping the schedule: `reprocessBreachSchedule` derives the flags right after it generated the periods up to the business date and before it replays the payments and recalculates the past due. The reset service no longer derives flags ahead of the split/restore; on the flag-only paths (plain reset, its undo) it asks the schedule service to derive them and recalculates the past due. One derivation, always over the persisted actions, always on the final schedule. The same re-derivation now also runs after a reschedule re-dates the periods and after `generateNextPeriodIfNeeded` created periods (COB, charge extension), so a reset placed on the last period by the fallback moves to its containing period once that period exists.

> 
> Some minor:
> 
> restoreSplitPeriod filters reschedules to those recorded before the reset (beforeActionId) but passes the unfiltered pause list to applyPauses. That happens to be consistent today because a pause already re-dates all periods, but the asymmetry is worth a comment. The WorkingCapitalLoanAccount.feature scenario C102391 (modify with invalid product id) and its step def are unrelated to breach reset undo. I traced the path and the expectation holds: validateForUpdate collects the longGreaterThanZero() error but productRepository.findById(...).orElseThrow(WorkingCapitalLoanProductNotFoundException) fires first, so 0 and -1 do return 404, but it's scope creep in a PR the checklist asks not to be a code dump. WorkingCapitalBreachReset.feature also lost its trailing newline (not spotless-enforced for .feature, so cosmetic only).

- C102391 (invalid product id) and its step definition: dropped from this PR. QA can re-add it on its own ticket.
- Trailing newline in `WorkingCapitalBreachReset.feature`: restored.

budaidev:
> The undo path is well covered - unit tests on both the resolver and the schedule methods, plus six integration tests including the pause-extended and payment-reassignment cases. That's usually the part I have to ask for, so nice work there.
> 
> My concern is with how the undo figures out what to revert. `restoreSplitPeriod` reverse-engineers the pre-split state from the schedule as it stands now plus the current breach config, using three guard clauses to decide whether the split it's looking at is actually the one this undo owns. Two of those guards fall through to a `log.warn` and a `false`, and the caller then just recalculates past due and returns 200 - so the user gets a success response for an undo that only lifted a flag and left the split in place.
> 
> Concretely: a RESCHEDULE recorded between the reset and the undo changes `resolveEffectiveRescheduleParams`, so the restore recomputes a boundary that never existed. And any PAUSE/RESUME between reset and undo rewrites every period contiguously from the schedule's first `fromDate` without knowing about the split, so the split disappears from the schedule entirely while the RESET row still says `restartPeriodFromResetDate = true` - the undo then silently falls into the flag-only branch.
> 
> Wouldn't it be simpler and deterministic to record on the RESET action row what the reset actually changed (the pre-split `toDate`, maybe the period id), and have the undo just write that back? Then there's nothing to guess and nothing that can silently degrade. The same reconstruction problem shows up in `findLatestFlaggedPeriod`'s "last flagged period by period number" fallback in the reset service - it assumes period-number order tracks reset-date order, which is exactly the assumption the boundary rewrites can break.
> 
> Two smaller things while I was in there:
> 
> * `splitPeriodAtReset` only guards `resetDate == fromDate`; `resetDate < fromDate` falls through and can persist a period with `toDate < fromDate` / negative `numberOfDays`.
> * The RESET-undo replay in `WorkingCapitalLoanActiveBreachResetResolver` is ordered by `startDate, id` - since reset/undo dates are always "the current business date," if the business date is ever moved backwards between two actions the replay order diverges from creation order, and the `break`-on-id can stop before the reset it's meant to cancel.
> 
> This is a draft so nothing's blocking yet, but I'd like to settle the design question (store vs. reconstruct) before it comes out of draft - it affects several of the methods above at once.
> 
> Recommendation: COMMENT

Thanks for the review, and sorry for the late answer here: the design question was settled with Adam and the implementation you reviewed has been replaced, so the concerns below no longer apply to the current head.

adamsaghy:
I think only one gap remained:

WorkingCapitalLoanBreachScheduleServiceImpl.java:346: restoreSplitPeriod derives the pre-split end as calculateToDate(...) + graceDays for period 1, but fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/service/WorkingCapitalLoanBreachScheduleServiceImpl.java:297 — which runs on every PAUSE/RESUME — re-dates period 1 without grace days. On a loan with breachGraceDays > 0 that has ever been paused, the undo restores the wrong boundary. Probe over the PR's own harness (grace 3, freq 60, pause 02-10..02-14): P1 stands at [01-01..03-06] (65 days), a restart reset on 02-20 splits it correctly, and the undo restores [01-01..03-09] (68 days) instead of 65 — every later boundary shifts by the grace days. Same mismatch defeats the new early-out at line 352: after that pause, a restart reset on P2's fromDate splits nothing (correctly), but the undo re-dates P1 and deletes P2 without regenerating it, leaving no period covering the business date — which contradicts the PR's own e2e scenario "Undo of a no-split restart reset is flag-only: schedule structure unchanged". The grace omission in recalculatePeriodsForPauses is pre-existing, but restoreSplitPeriod is new and depends on agreeing with it; both should compute the period end through one shared helper.

galovics (commented):
Thanks @budaidev, this addresses everything I raised in the draft review. The LIFO replay now goes by recording order (not business-date order) with the reasoning documented directly on the method, the negative-period guard is in, and the reschedule-after-reset case is handled by filtering to reschedules in force at the time of the reset. Test coverage went from thin to substantial - unit tests on the resolver, the flag derivation, and the split/restore methods, plus a large Feign integration suite and new e2e scenarios.

One thing I want called out explicitly rather than left implicit: two of your own new integration tests demonstrate that a split can still be silently lost. `pauseAfterRestartReset_keepsTheFlagOnThePeriodHoldingTheResetDate` shows a pause recorded after a restart-reset re-expanding the truncated boundary back to its natural length - the reset flag survives but the split itself is gone, so the later undo has nothing to restore. And `undoAfterBackwardsBusinessDateMove_flagsTheRegeneratedPeriodOfTheStillActiveReset` shows a second restart-reset deleting an earlier, still-active reset's boundary permanently, even though that earlier reset stays on the active stack.

Both of these read like accepted/intentional behavior in the tests as written, and I don't think either needs to block this PR - the acute failure modes (heuristic drift, stale reschedule params, "success" on a no-op undo) are genuinely closed. But if the split is meant to be durable across a pause or a stacked reset, these two cases say otherwise, and I'd rather that be a documented decision (or a follow-up ticket to persist the pre-split boundary on the reset action itself, which would close both at once) than something a future bug report rediscovers.

Recommendation: COMMENT (would be APPROVE if the two cases above get an explicit answer either in this PR's description or as a tracked follow-up)

budaidev:
> I think only one gap remained:
> 
> WorkingCapitalLoanBreachScheduleServiceImpl.java:346: restoreSplitPeriod derives the pre-split end as calculateToDate(...) + graceDays for period 1, but fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/service/WorkingCapitalLoanBreachScheduleServiceImpl.java:297 — which runs on every PAUSE/RESUME — re-dates period 1 without grace days. On a loan with breachGraceDays > 0 that has ever been paused, the undo restores the wrong boundary. Probe over the PR's own harness (grace 3, freq 60, pause 02-10..02-14): P1 stands at [01-01..03-06] (65 days), a restart reset on 02-20 splits it correctly, and the undo restores [01-01..03-09] (68 days) instead of 65 — every later boundary shifts by the grace days. Same mismatch defeats the new early-out at line 352: after that pause, a restart reset on P2's fromDate splits nothing (correctly), but the undo re-dates P1 and deletes P2 without regenerating it, leaving no period covering the business date — which contradicts the PR's own e2e scenario "Undo of a no-split restart reset is flag-only: schedule structure unchanged". The grace omission in recalculatePeriodsForPauses is pre-existing, but restoreSplitPeriod is new and depends on agreeing with it; both should compute the period end through one shared helper.

Confirmed, reproduced with the numbers. Fixed the way you suggested: one private helper computes the natural period end. Test coverage has been added

budaidev:
> Thanks @budaidev, this addresses everything I raised in the draft review. The LIFO replay now goes by recording order (not business-date order) with the reasoning documented directly on the method, the negative-period guard is in, and the reschedule-after-reset case is handled by filtering to reschedules in force at the time of the reset. Test coverage went from thin to substantial - unit tests on the resolver, the flag derivation, and the split/restore methods, plus a large Feign integration suite and new e2e scenarios.
> 
> One thing I want called out explicitly rather than left implicit: two of your own new integration tests demonstrate that a split can still be silently lost. `pauseAfterRestartReset_keepsTheFlagOnThePeriodHoldingTheResetDate` shows a pause recorded after a restart-reset re-expanding the truncated boundary back to its natural length - the reset flag survives but the split itself is gone, so the later undo has nothing to restore. And `undoAfterBackwardsBusinessDateMove_flagsTheRegeneratedPeriodOfTheStillActiveReset` shows a second restart-reset deleting an earlier, still-active reset's boundary permanently, even though that earlier reset stays on the active stack.
> 
> Both of these read like accepted/intentional behavior in the tests as written, and I don't think either needs to block this PR - the acute failure modes (heuristic drift, stale reschedule params, "success" on a no-op undo) are genuinely closed. But if the split is meant to be durable across a pause or a stacked reset, these two cases say otherwise, and I'd rather that be a documented decision (or a follow-up ticket to persist the pre-split boundary on the reset action itself, which would close both at once) than something a future bug report rediscovers.
> 
> Recommendation: COMMENT (would be APPROVE if the two cases above get an explicit answer either in this PR's description or as a tracked follow-up)

Since your review, one more fix from Adam's last comment went in: the pause re-dating, the initial period and the undo restore now compute the period end through one shared helper, so period 1's grace days can no longer drift between them; unit and Feign scenarios cover it.

adamsaghy (approved):
LGTM

PR #6284, opened 2026-08-18
FINERACT-2763: Checker Users Unable to Approve Pending Maker-Checker Tasks Without Direct Operational Permissions and Missing Tasks in Checker Inbox

rymghosn:
Currently, when a client, loan, or savings account has one or more maker-checker actions awaiting checker approval, there is no way to discover this from the corresponding GET API response. A caller has to separately query the checker inbox/pending-audit endpoints and correlate entries by entity, which is easy to miss and leaves API consumers unaware that the record they just fetched has a pending change awaiting approval.

This PR adds a pendingMakerCheckerApprovals field to the ClientData, LoanAccountData, SavingsAccountData, and DepositAccountData (fixed/recurring deposit) response objects. When a GET request for one of these entities is served, the field is populated with any CommandSource entries still AWAITING_APPROVAL for that entity (id, action name, entity name, permission code, maker username, and submission date), or omitted (null) when there is nothing pending.
Changes:                                                                                                                                         
  - New PendingMakerCheckerData DTO and MakerCheckerReadService/MakerCheckerReadServiceImpl to look up pending CommandSource rows by               
  loan/client/savings id.                                                                                                                          
  - New CommandSourceRepository.findPendingBy{Loan,Client,Savings}Id queries (join fetch c.maker to avoid N+1 lookups for the maker username).     
  - Wired into ClientReadPlatformServiceImpl#retrieveOne, LoanReadPlatformServiceImpl#retrieveOne,                                                 
  SavingsAccountReadPlatformServiceImpl#retrieveOne, and DepositAccountReadPlatformServiceImpl#retrieveOne.                                        
                                                                                                                                                   
  This is a read-only, additive change — it does not alter maker-checker approval/permission logic itself, only exposes existing pending-approval  
  state that was previously only visible via the checker inbox. 
  PR:(https://issues.apache.org/jira/browse/FINERACT-2763)

### DIFF
--- a/fineract-core/src/main/java/org/apache/fineract/commands/data/PendingMakerCheckerData.java
+++ b/fineract-core/src/main/java/org/apache/fineract/commands/data/PendingMakerCheckerData.java
@@ -1,0 +1,49 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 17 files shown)


adamsaghy:
@rymghosn I dont think i agree with this approach, neither attach "irrelevant" details into the target APIs.

Would you mind to raise this question to the Fineract DEV mail list to hear other's opinions as well?

My 2 cents:
- Checker API is there for this purpose, i dont see why we should not use that...

If you want something global, i dont mind to introduce endpoint for that purpose.

PR #6286, opened 2026-08-18
FINERACT-2765: Tax Component and Tax Group names should be unique

rymghosn:
Currently nothing prevents creating a `TaxComponent` or `TaxGroup` with a                                                                        
  name that's already in use, or renaming one to collide with another. This                                                                        
  also allowed submitting a `TaxGroup` with the same tax component listed                                                                          
  more than once in its `taxComponents` array.
  This PR adds:                                                                                                                                    
                                                                                                                                                   
  - On tax component create/update: reject the request if another tax                                                                              
    component already has the given `name` (case-sensitive match against                                                                           
    `TaxComponentRepository.findByName`, relying on DB collation for exact                                                                         
    matching), returning `tax.component.name.must.be.unique`.                                                                                      
  - On tax group create/update: reject the request if another tax group                                                                            
    already has the given `name` (case-insensitive, via                                                                                            
    `TaxGroupRepository.existsByNameIgnoreCase` /                                                                                                  
    `existsByNameIgnoreCaseAndIdNot`), returning                                                                                                   
    `tax.group.name.must.be.unique`.                                                                                                               
  - On tax group create/update: reject a `taxComponents` payload that                                                                              
    references the same `taxComponentId` more than once, returning                                                                                 
    `validation.msg.tax.group.duplicate.component`. 
                                                                                                                                                       
  Update checks only trigger when the `name` (or, for groups, a component                                                                          
  list) is actually being changed, so updates that leave the name untouched                                                                        
  are unaffected.
 PR (https://issues.apache.org/jira/browse/FINERACT-2765).

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/tax/service/TaxWritePlatformServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/tax/service/TaxWritePlatformServiceImpl.java
@@ -18,12 +18,18 @@
  */
 package org.apache.fineract.portfolio.tax.service;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import lombok.RequiredArgsConstructor;
 import org.apache.fineract.infrastructure.core.api.JsonCommand;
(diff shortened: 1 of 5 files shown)


adamsaghy (changes requested):
Adding a unique constraint on the name fields would be better and cleaner solution.

PR #6290, opened 2026-08-19
FINERACT-2769: Tax group – Start Date field validation/behavior issues on edit mode if the group is unlinked

rymghosn:
Tax group updates could never actually carry a start date change through to                                                                      
  an existing component mapping, and a JSON-key bug made updates fail                                                                              
  unnecessarily:                                                                                                                                   
                                                                                                                                                   
  - validateForTaxGroupUpdate extracted a submitted mapping's id using the                                                                         
    taxComponentId JSON key instead of id. Since taxComponentId is                                                                                 
    immutable and normally omitted on update, this misclassified a                                                                                 
    legitimate "update existing mapping" request as "add a new component"                                                                          
    and wrongly demanded a taxComponentId.                                                                                                         
  - TaxAssembler/TaxGroupMappings/TaxGroup always discarded any                                                                                    
    submitted startDate on update — only endDate was ever applied, even                                                                            
    though startDate was already an accepted request parameter.                                                                                    
  - Adding a brand-new tax component to an existing group via update never                                                                         
    set the new mapping's taxGroup relationship, which would violate the                                                                           
    tax_group_id NOT NULL constraint on insert.
      This PR:                                                                                                                                         
                                                                                                                                                   
  - Fixes the JSON-key bug.                                                                                                                        
  - Lets an existing component's start date be edited, but only while it                                                                           
    hasn't taken effect yet (i.e. its current start date is still in the                                                                           
    future) — once a start date is on or before today it's locked, and only                                                                        
    the end date remains editable.                                                                                                                 
  - Fixes the missing taxGroup FK assignment for newly-added components.                                                                           
  - Exposes groupEditable (on GET /taxes/group/{id}?template=true) and                                                                             
    per-component componentEditable so clients can tell which fields are                                                                           
    still open for editing: a group linked to charge products is only                                                                              
    editable if at least one of its components hasn't taken effect yet.
    PR:(https://issues.apache.org/jira/browse/FINERACT-2769)

### DIFF
--- a/fineract-charge/src/main/java/org/apache/fineract/portfolio/charge/domain/ChargeRepository.java
+++ b/fineract-charge/src/main/java/org/apache/fineract/portfolio/charge/domain/ChargeRepository.java
@@ -28,4 +28,7 @@ public interface ChargeRepository extends JpaRepository<Charge, Long>, JpaSpecif
 
     @Query("select lc.id from WorkingCapitalLoanCharge lc where lc.charge.id = :chargeId and lc.active = true")
     Optional<Long> isAnyWorkingCapitalLoansAssociateWithThisCharge(@Param("chargeId") Long chargeId);
+
+    @Query("select case when count(c) > 0 then true else false end from Charge c where c.taxGroup.id = :taxGroupId")
+    boolean existsByTaxGroupId(@Param("taxGroupId") Long taxGroupId);
 }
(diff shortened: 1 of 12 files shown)


adamsaghy (changes requested):
Can you please cover with integration or E2E test cases this behaviour?

PR #6292, opened 2026-08-19
FINERACT-2771: Tax - User should be able to set end date for a tax group

rymghosn:
A tax group component's end date could only ever be set once: `TaxValidator.validateTaxGroupEndDateAndTaxComponent()`                            
  rejected any attempt to change an end date to a different value if it had                                                                        
  already been set before, even when the new value was a perfectly valid                                                                           
  future date after the component's start date. This made it impossible to                                                                         
  correct or extend an end date once entered.                                                                                                      
                                                                                                                                                   
  This PR removes that one-time restriction. An existing tax group                                                                                 
  component's end date can now be updated repeatedly, with the only remaining                                                                      
  constraint being that each new end date must fall after the component's                                                                          
  start date.
  PR:(https://issues.apache.org/jira/browse/FINERACT-2771)


IOhacker (approved):
LGTM

adamsaghy (changes requested):

PR #6293, opened 2026-08-19
FINERACT-2772: Tax group - User should be able to set end date if start date was in the past

rymghosn:
`TaxGroup.findOneBy(TaxGroupMappings)` threw `TaxMappingNotFoundException` as soon as the **first** existing mapping in the set didn't match the one being looked up, instead of checking the rest of the set. With a tax group that has more than one component mapping, this meant editing/adding a component could randomly fail with an Internal Server Error depending on iteration order, even though a matching mapping existed further down the set. The same method could also throw an NPE via `groupMappings.getId().equals(...)` when an existing mapping's id was itself `null`.

This PR:
- Fixes `findOneBy` to scan the full set of existing mappings and only throw `TaxMappingNotFoundException` if none of them match, using `Objects.equals` to avoid the NPE on a null id.
- Exposes a `businessDate` field on `TaxGroupData` (populated from `DateUtils.getBusinessLocalDate()` when fetching a single tax group), so a client can validate a new tax component's start date against the system business date before submitting, instead of relying on the server round-trip to reject it.
 ## Changes                                                                                                                                       
                                                                                                                                                   
  - `TaxGroup#findOneBy`: iterate over all mappings before giving up, `Objects.equals` instead of `.getId().equals(...)`                           
  - `TaxGroupData`: new `businessDate` field + `withBusinessDate(TaxGroupData, LocalDate)` factory (mirrors the existing `template(...)`           
  copy-factory pattern); existing `lookup()`/`template()` factories updated to carry the field through                                             
  - `TaxGroupMapper`: `businessDate` marked `ignore` in the MapStruct mapping (it's computed, not entity data)                                     
  - `TaxReadPlatformServiceImpl#retrieveTaxGroupData`: populates `businessDate` on the returned `TaxGroupData`   
  PR:(https://issues.apache.org/jira/browse/FINERACT-2772)

### DIFF
--- a/fineract-core/src/main/java/org/apache/fineract/portfolio/tax/data/TaxGroupData.java
+++ b/fineract-core/src/main/java/org/apache/fineract/portfolio/tax/data/TaxGroupData.java
@@ -19,6 +19,7 @@
 package org.apache.fineract.portfolio.tax.data;
 
 import java.io.Serializable;
+import java.time.LocalDate;
 import java.util.Collection;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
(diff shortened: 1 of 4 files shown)


adamsaghy (changes requested):

PR #6331, opened 2026-08-25
FINERACT-2786: Fix NullPointerException on /v1/email CREATE

AshharAhmadKhan:
https://issues.apache.org/jira/browse/FINERACT-2786

EmailMessageAssembler.assembleFromJson() called EmailMessage.pendingEmail() with a hardcoded null emailCampaign. pendingEmail() unconditionally calls emailCampaign.getStatus() with no null check, throwing NullPointerException on every /v1/email CREATE request regardless of which valid fields are supplied.

The three other callers of pendingEmail() (in EmailCampaignWritePlatformCommandHandlerImpl and UpdateEmailOutboundWithCampaignMessageTasklet) always pass a real EmailCampaign, and email_campaign_id is nullable at the JPA/DB level, confirming campaign-less emails were always intended to be valid. EmailMessage already has a sibling factory, instance(), which takes an explicit EmailMessageStatusType instead of deriving it from emailCampaign.getStatus(), so it is null safe by construction and was apparently built for exactly this case, just never wired up. Routing the /v1/email assembler to instance() fixes the NPE with zero changes to EmailMessage.java itself, so it carries zero risk to the three working campaign driven callers of pendingEmail().

Adds a regression test confirming assembleFromJson() no longer throws and produces a PENDING EmailMessage when only clientId is supplied.

Related: FINERACT-2785 (validator fix, separate bug, same resource, found during the same investigation)

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/campaigns/email/domain/EmailMessageAssembler.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/campaigns/email/domain/EmailMessageAssembler.java
@@ -81,7 +81,8 @@ public class EmailMessageAssembler {
         final String message = this.fromApiJsonHelper.extractStringNamed(EmailApiConstants.messageParamName, element);
         final String emailSubject = this.fromApiJsonHelper.extractStringNamed(EmailApiConstants.subjectParamName, element);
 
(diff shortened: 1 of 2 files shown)


AshharAhmadKhan:
Please anyone trigger checks whenever possible, thanks!

AshharAhmadKhan:
hey @adamsaghy, this is ready for review. Please check it out when possible, thanks!

adamsaghy (approved):
LGTM

PR #6332, opened 2026-08-25
FINERACT-2455: WC - Real-time accrual posting for charges

somasorosdpc:
## Description

## Checklist

### DIFF
--- a/fineract-core/src/main/java/org/apache/fineract/infrastructure/configuration/api/GlobalConfigurationConstants.java
+++ b/fineract-core/src/main/java/org/apache/fineract/infrastructure/configuration/api/GlobalConfigurationConstants.java
@@ -95,6 +95,7 @@ public final class GlobalConfigurationConstants {
     public static final String RETAINED_EARNING_USED_BY_REPORT_NAME = "retained-earning-used-by-report-name";
     public static final String OFFICE_ID = "office-id";
     public static final String DISALLOW_BACKDATED_TRANSACTIONS = "disallow-backdated-transactions";
+    public static final String WCL_CHARGE_ACCRUAL_TIME = "wcl-charge-accrual-time";
 
     private GlobalConfigurationConstants() {}
 }
--- a/fineract-e2e-tests-core/src/test/java/org/apache/fineract/test/stepdef/common/GlobalConfigurationStepDef.java
+++ b/fineract-e2e-tests-core/src/test/java/org/apache/fineract/test/stepdef/common/GlobalConfigurationStepDef.java
@@ -47,6 +47,7 @@ public class GlobalConfigurationStepDef {
     @After("@WorkingCapitalLoanChargeAccrualFeature")
     public void restoreChargeAccrualDateConfig() {
         globalConfigurationHelper.setGlobalConfigValueString("charge-accrual-date", "due-date");
+        globalConfigurationHelper.setGlobalConfigValueString("wcl-charge-accrual-time", "eod");
     }
 
     @After("@BusinessDateDisabledCheck")
--- a/fineract-e2e-tests-runner/src/test/resources/features/LoanAccrualTransaction.feature
+++ b/fineract-e2e-tests-runner/src/test/resources/features/LoanAccrualTransaction.feature
@@ -1995,7 +1995,7 @@ Feature: LoanAccrualTransaction
     Then LoanAccrualTransactionCreatedBusinessEvent is raised on "17 November 2025"
     Then LoanTransactionAccrualActivityPostBusinessEvent is raised on "18 November 2025"
 
-  @TestRailId:C85643
+  @TestRailId:C85645
   Scenario: Core parity - accrual is posted on early loan closure even when charge-accrual-date is submitted-date
     When Global config "charge-accrual-date" value set to "due-date"
     When Admin sets the business date to "01 June 2023"
--- a/fineract-e2e-tests-runner/src/test/resources/features/WorkingCapitalLoanChargeAccrual.feature
+++ b/fineract-e2e-tests-runner/src/test/resources/features/WorkingCapitalLoanChargeAccrual.feature
@@ -55,6 +55,7 @@ Feature: Working Capital Loan Charge Accrual
       | LoanProduct         | submittedOnDate | expectedDisbursementDate | principalAmount | totalPayment | periodPaymentRate | discount |
       | WCLP_ACC_DEF_REV_AM | 01 February 2026 | 01 February 2026         | 9000            | 100000       | 18                | 0        |
     When Global config "charge-accrual-date" value set to "submitted-date"
+    When Global config "wcl-charge-accrual-time" value set to "real-time"
     And Admin sets the business date to "10 February 2026"
     And Admin runs inline COB job for Working Capital Loan by loanId
     And Admin adds "WORKING_CAPITAL_SPECIFIED_DUE_DATE_PENALTY" specified due date charge to working capital loan with "20 February 2026" due date and 30.0 transaction amount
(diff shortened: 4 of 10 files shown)


galovics (changes requested):
The idempotency guard (`isAlreadyAccrued`) does hold across all three posting paths, so I'm not worried about double-posting or duplicate journal entries. Two other things bother me though, both real regressions rather than draft incompleteness.

**1. Accrual date silently changed from `charge.getDueDate()`/`charge.getSubmittedOnDate()` to the COB `businessDate`, in both branches.** For any charge whose due/submitted date is in the past relative to when COB actually runs on it (backdated charge, a skipped COB day, a catch-up run), the accrual now lands on the wrong date - and this has nothing to do with real-time posting, it's a change to the existing EOD behavior. The `charge-accrual-date` config name literally promises "accrue on the due date" - if this is a deliberate policy change to avoid backdated GL entries, it needs to be called out explicitly and tested, because right now it silently changes existing due-date-mode behavior for every WC tenant.

**2. An unrecognized config value silently disables all charge accrual.** Both `if (X.equalsIgnoreCase(...))` checks have no `else` - a typo or an unexpected string value entered via the global-configuration API (e.g. `"EOD "`, `"realtime"`) makes both the real-time and the EOD path no-ops. Charges then never accrue until loan closure, with no error and no log. Worth making this default-safe (e.g. `isRealTimeAccrual()` boolean with the EOD path as the fallback) rather than two independent string comparisons that can both come back false.

Also, the new e2e scenarios set `wcl-charge-accrual-time` to `eod` via a trailing step in the scenario body rather than through an `@After` restore hook - that step won't run if the scenario fails partway through, so the config leaks into every subsequent scenario in the serially-tagged feature file and can cascade failures across the whole shard. The existing `charge-accrual-date` config already has exactly this pattern solved with an `@After` hook and an explicit comment about why - worth extending that hook to cover the new config too, rather than the trailing-step approach.

Smaller: the accrual date config lives in a WC-specific accessor that's now been added to the platform-wide `ConfigurationDomainService` interface, while the sibling `charge-accrual-date` config in the same class reads through the generic repository wrapper directly - two mechanisms for the same kind of thing in one file. And it's still a draft, so I know tests/docs are pending, but the config matrix (real-time/EOD × due-date/submitted-date) has zero unit test coverage right now, only e2e.

adamsaghy:
@somasorosdpc Can you please let us know whether you had the chance to review Arnold's concerns and whether you have addressed them?

somasorosdpc:
@galovics 
* if eod is configured COB job is updated to make the accruals on submitted on date / due date depending on configs.
* if customer incorrectly configures the global value, we have to think that they want to miss configure on purpose.
* I moved the trailing back configuration to @After to reset it to the default.
* I fixed the access to the global configuration, I removed it from the  ConfigurationDomainService since it is a wc specific. Also I follow the already available access mechanic to read the value.

adamsaghy:
**Resolved in code**

- Accrual date no longer the COB businessDate — :82-89 uses submittedOnDate/dueDate per charge-accrual-date.
- e2e config leak — restored via the @After hook, GlobalConfigurationStepDef.java:49; trailing in-scenario resets removed.
- The ConfigurationDomainService interface change is gone; the new config reads through GlobalConfigurationRepositoryWrapper like its sibling.

**Open**

1. Dead field. :25 + :66 still inject ConfigurationDomainService, never read. Soma believes this was removed; only the interface method was. One-line delete, checkstyle won't flag it.
2. eod → real-time switch gap (substantive, and not addressed by any of the four replies). A charge added under eod with the config flipped to real-time before the next COB is skipped by the sweep permanently (:81) and only accrues at closure, dated the closing date. This is a valid configuration, not a typo, so "misconfigured on purpose" doesn't cover it. Fix: let the COB sweep always run on the charge-accrual-date rule and have real-time only additionally post at add time — isAlreadyAccrued (:154) already makes that idempotent.
3. Unknown config value — Soma's call to leave it, which I'd accept, but a .trim() and a log.warn are worth it: the failure mode is fee/penalty income silently unrecognized tenant-wide until closure, and nothing validates the value on write.
4. Undocumented behavior changes — that real-time ignores charge-accrual-date, and that existing submitted-date tenants now get the accrual at the next COB (or, if the loan closes first, dated the closing date rather than submittedOnDate). Both belong in the PR description / release note.

galovics (dismissed):
Two of the three things from last round are properly fixed - accruals are dated off `charge.getDueDate()`/`getSubmittedOnDate()` again rather than the COB business date, and the e2e config now gets restored via the `@After` hook instead of a trailing scenario step that wouldn't run on a mid-scenario failure.

The unrecognized-config-value issue is still open though: `retrieveWCChargeAccrualDateConfig()` only falls back to the default on null/blank, so any other value (a typo, `"realtime"`, `"EOD "`) makes both the real-time and the EOD branches false, and charge income silently stops being recognized until the loan closes - no error, no log. I'd really like this to be an enum with an explicit fallback (or at least an `else` that logs and defaults), same ask as before.

Two new things from this round: there's now an unused `ConfigurationDomainService` field injected into the class (looks like a leftover from an earlier version of the fix), and the submitted-date branch of the EOD path is missing the "not after business date" guard that the due-date branch has - so during a COB catch-up run, a charge submitted on a later business date than the one currently being processed could get accrued with a date after the COB processing date. Worth adding the same filter there.

mariiaKraievska:
@galovics I addressed the open review points in the latest push:

Switch eod → real-time gap
COB now always sweeps by charge-accrual-date, regardless of wcl-charge-accrual-time. Real-time only adds an immediate post-on-add path; isAlreadyAccrued keeps both idempotent. Added e2e UC9 (due-date pending after flip) and UC10 (submitted-date pending before next COB).

Unrecognized config value
Unknown / mistyped values now fall back to EOD (after trim), instead of making both posting paths no-ops.

Dead ConfigurationDomainService field
Removed.

Missing submitted-date COB guard
EOD, submitted-date path now filters with !submittedOnDate.isAfter(businessDate), same idea as the due-date branch, so catch-up COB cannot post an accrual dated after the day being processed.

@galovics PTAL

adamsaghy:
@mariiaKraievska Can you please check the below two things?

WorkingCapitalLoanChargeAccrualService.java:75 — with the new default (EOD) plus charge-accrual-date=submitted-date, a charge added to an overpaid loan is never accrued. processOnChargeAdded is now a no-op, applyChargeDrivenLifecycle leaves the loan OVERPAID or closes it (:293-311), COB excludes closed/overpaid loans (WorkingCapitalLoanRetrieveIdServiceImpl.java:43), and createLoanCharge never calls accrueOnClosure. The fee is settled from the overpayment but no accrual/JE is ever posted — a regression, since the pre-PR on-add path covered it. An accrueOnClosure(loan, businessDate) after the lifecycle block would close the gap.

WorkingCapitalLoanChargeAccrualService.java:90 — retrieveChargeAccrualDateConfig() is called twice (both branches), and both branches repeat the same charge query. Two config lookups per loan per COB run; hoist the mode and the charge list, and vary only the date extractor.

mariiaKraievska:
> Can you please check the below two things?
> 
> WorkingCapitalLoanChargeAccrualService.java:75 — with the new default (EOD) plus charge-accrual-date=submitted-date, a charge added to an overpaid loan is never accrued. processOnChargeAdded is now a no-op, applyChargeDrivenLifecycle leaves the loan OVERPAID or closes it (:293-311), COB excludes closed/overpaid loans (WorkingCapitalLoanRetrieveIdServiceImpl.java:43), and createLoanCharge never calls accrueOnClosure. The fee is settled from the overpayment but no accrual/JE is ever posted — a regression, since the pre-PR on-add path covered it. An accrueOnClosure(loan, businessDate) after the lifecycle block would close the gap.
> 
> WorkingCapitalLoanChargeAccrualService.java:90 — retrieveChargeAccrualDateConfig() is called twice (both branches), and both branches repeat the same charge query. Two config lookups per loan per COB run; hoist the mode and the charge list, and vary only the date extractor.

Thanks for catching these. Both fixed in the latest push.

Overpaid + EOD + submitted-date — added accrueOnClosure in createLoanCharge after processOnChargeAdded, plus e2e UC11 covering the gap.
COB duplicate lookup — moved the accrual-date mode and charge list into local variables so they are read once

adamsaghy (approved):
LGTM

PR #6368, opened 2026-09-01
FINERACT-789: Fix fee accounting for disbursements to savings

udayagiri-sivasai-dev:
## Summary

- **Bug:** Repayment-at-disbursement fees for `disburseToSavings` used `FUND_SOURCE` instead of `LIABILITY_TRANSFER`.
- **Root cause:** The `isAccountTransfer` context was lost, and `false` was hardcoded when posting the fee transaction.
- **Fix:** Propagate the existing `isAccountTransfer` flag to the fee journal posting.

## Tests

- Focused FINERACT-789 integration test passed.
- Related disbursement-to-savings and ordinary fee-accounting regression tests passed.

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanDisbursementService.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/loanaccount/service/LoanDisbursementService.java
@@ -215,7 +215,8 @@ public class LoanDisbursementService {
         return disburseAmount;
     }
 
-    public void handleDisbursementTransaction(final Loan loan, final LocalDate disbursedOn, final PaymentDetail paymentDetail) {
+    public void handleDisbursementTransaction(final Loan loan, final LocalDate disbursedOn, final PaymentDetail paymentDetail,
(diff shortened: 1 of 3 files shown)


adamsaghy (approved):
LGTM

@bharathgowda Please review

bharathcgowda:
Looks Good

TianHengZhuang (commented):
Good bug fix with a regression test! A few observations:

1. **Root cause**: Hardcoding `false` for `isAccountTransfer` instead of propagating it would cause incorrect accounting entries for disbursements to savings — a subtle but real issue.
2. **Method signature change**: Correct — new parameter added at the end, only the internal call site is updated.
3. **Integration test**: The new `LoanAccountDisbursementToSavingsAccountingTest` is the right level of test. Consider adding an explicit assertion for `LIABILITY_TRANSFER` account type in the journal entries to make the test self-documenting.

Thanks for the thorough test coverage. ✅

TianHengZhuang (approved):
LGTM — correct propagation of `isAccountTransfer` flag, clean signature update, and integration test makes this a solid PR. ✅

PR #6385, opened 2026-09-03
FINERACT-2455: WC - Transaction - Prepay Loan

somasorosdpc:
## Description

## Checklist


galovics (dismissed):
The `transactionDate` parameter is accepted, validated, parsed, and echoed back in the response - but never actually used to compute anything. The template amount is just `balance.getTotalOutstanding()` as it stands right now, so for a WC loan where penalties and discount fee accrue daily and backdated transactions are allowed, a payoff quote for any date other than today is going to be wrong with no way for the caller to tell. Either the payoff needs to be computed as of `transactionDate` (which is what I'd expect from something called "prepay" - including whatever the intended treatment of unearned discount fee is on an early settlement), or the parameter should be dropped until it does something.

Two more concrete issues:
- `repository.getReferenceById(loanId)` returns a lazy proxy - an unknown `loanId` won't fail here, it'll throw `EntityNotFoundException` on first property access (`getCurrency()`), which has no exception mapper and surfaces as a 500 instead of a 404. `retrieveOne` in the same class does this correctly via `findByIdWithFullDetails(...).orElseThrow(...)` - worth reusing that pattern here, especially since `getCurrency()` and the balance can both be null in general.
- The new Swagger response DTO (`WorkingCapitalLoanTransactionTemplateResponse`) doesn't match what the endpoint actually returns - it declares `id` (which the data class doesn't have) and omits `currency` (which it does). The generated client will always get `null` for a field that never exists and won't expose the one that does.

Smaller things: the permission check happens after loan resolution rather than before (every other endpoint in this resource does it the other way), and this also introduces a second, parallel "template" mechanism on the same resource (`{loanId}/transactions/template?command=X` vs. the existing `{loanId}/template?templateType=X`) with different query param names and different backing services for the same concept - worth confirming that's intentional rather than folding `prepayLoan` into the existing one.

Cocoa-Puffs:
@galovics I have addressed the concerns!

> The template amount is just balance.getTotalOutstanding() as it stands right now, so for a WC loan where penalties and discount fee accrue daily and backdated transactions are allowed, a payoff quote for any date other than today is going to be wrong with no way for the caller to tell.

Working Capital loans employ what is essentially a cash-like accounting, meaning that fee accrued is only dependent on how much money was repaid, not the time elapsed. Every outstanding bucket is `due − paid − writtenOff`, all plain columns; there is no time term anywhere. `applyDisbursement` folds the whole discount into principal at disbursement, and `realizedIncomeFromDiscountFee` only moves income recognition, it never touches `getTotalOutstanding()`. Only a payment or a charge moves the number, and both are explicit commands. A date-parameterised recomputation would return the same value for every date. I have added multiple scenarios into `WorkingCapitalLoanPrepayment.feature`

For the same reason we only echo transactionDate. I have not added a validation for being future-dated, because the repayment validator rejects future dated transactions and as far as I can see validations are unprecedented when it comes to transaction templates, no other template does similar validations, leaving it to the various transaction processing validators to reject the transactions if they are sent with the wrong value.

> repository.getReferenceById(loanId) returns a lazy proxy - an unknown loanId won't fail here, it'll throw EntityNotFoundException on first property access (getCurrency()), which has no exception mapper and surfaces as a 500 instead of a 404. retrieveOne in the same class does this correctly via findByIdWithFullDetails(...).orElseThrow(...) - worth reusing that pattern here, especially since getCurrency() and the balance can both be null in general.

The method now throws `WorkingCapitalLoanNotFoundException` if a loan is not found with the given id, mirroring other methods.

> The new Swagger response DTO (WorkingCapitalLoanTransactionTemplateResponse) doesn't match what the endpoint actually returns - it declares id (which the data class doesn't have) and omits currency (which it does). The generated client will always get null for a field that never exists and won't expose the one that does.

I have modified the swagger response DTO to match what the endpoint actually returns.

> Smaller things: the permission check happens after loan resolution rather than before (every other endpoint in this resource does it the other way), and this also introduces a second, parallel "template" mechanism on the same resource ({loanId}/transactions/template?command=X vs. the existing {loanId}/template?templateType=X) with different query param names and different backing services for the same concept - worth confirming that's intentional rather than folding prepayLoan into the existing one.

Moved the permission check to before the loan resolution. As for the template mechanism, I believe it mirrors core behaviour.

Term loans already have two template types split between lifecycle vs transaction operations.
|endpoint | core supports | WC supports|
-- | -- | --
{loanId}/template?templateType= | approval — one value | approve, disburse, repayment, goodwillCredit, creditBalanceRefund, recoveryPayment, discountFee, discountFeeAdjustment, chargeOff — nine
{loanId}/transactions/template?command= | repayment, prepayLoan, chargeOff, writeOff, foreclosure, payoutRefund, goodwillCredit, … | prepayLoan

The problem is more the opposite. WC currently crams in 8 transaction templates into the lifecycle endpoint. That is why the split looks arbitrary today. I have migrated the 8 transaction templates to the proper transaction template endpoint to mirror core behaviour.

|endpoint | commands|
-- | --
{loanId}/template?templateType= | approve
{loanId}/transactions/template?command= | disburse, repayment, goodwillCredit, creditBalanceRefund, recoveryPayment, discountFee, discountFeeAdjustment, chargeOff, prepayLoan

I can revert this and include `prePayment` in the already existing endpoint if that is more desirable.

Cocoa-Puffs:
api backwards compatibility is failing due to the migration of the transactions from the existing template endpoint

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/api/WorkingCapitalLoanTransactionsApiResourceSwagger.java
```diff
@@ -264,4 +265,48 @@ private ExecuteWorkingCapitalLoanTransactionCommandResponse() {}
...
+        @Schema(example = "1")
+        public Long wcLoanId;
+        @Schema(description = "Loan currency")
+        public CurrencyData currency;
+        @Schema(description = "Transaction type")
+        public LoanTransactionEnumData type;
+        @Schema(example = "[2024, 2, 1]")
+        public LocalDate transactionDate;
+        @Schema(example = "10000.00", description = "prepayLoan only: total payoff amount, principal + fee + penalty outstanding")
+        public BigDecimal transactionAmount;
```
adamsaghy, line 283: Lets remove this field and use `expectedAmount` instead. We dont need to introduce different kind of amount fields...

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/api/WorkingCapitalLoanTransactionsApiResourceSwagger.java
```diff
@@ -264,4 +265,48 @@ private ExecuteWorkingCapitalLoanTransactionCommandResponse() {}
...
+
+        @Schema(example = "10000.00", description = "prepayLoan only: outstanding principal portion of the payoff amount")
+        public BigDecimal principalPortion;
+        @Schema(example = "0.00", description = "prepayLoan only: outstanding fee portion of the payoff amount")
+        public BigDecimal feeChargesPortion;
+        @Schema(example = "0.00", description = "prepayLoan only: outstanding penalty portion of the payoff amount")
+        public BigDecimal penaltyChargesPortion;
+
+        @Schema(example = "10000.00", description = "chargeOff only: auto-calculated outstanding balance, read-only")
+        public BigDecimal chargeOffAmount;
```
adamsaghy, line 301: Same here.

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/service/WorkingCapitalLoanTransactionReadPlatformServiceImpl.java
```diff
@@ -78,52 +95,68 @@ public WorkingCapitalLoanCommandTemplateData retrieveLoanTransactionTemplate(fin
...
+     * command, with the repayment allocation and accounting treatment that comes with it.
+     */
+    private WorkingCapitalLoanTransactionTemplateData prePaymentTemplate(
+            final WorkingCapitalLoanTransactionTemplateData.WorkingCapitalLoanTransactionTemplateDataBuilder template,
+            final WorkingCapitalLoanBalance balance, final LocalDate transactionDate) {
+        // The balance row is created when the application is submitted, so every loan reached through the normal flow
+        // has one and the amounts below are real. Reading it defensively anyway reports a missing balance as absent
+        // amounts rather than as a payoff of zero.
+        final Optional<WorkingCapitalLoanBalance> present = Optional.ofNullable(balance);
+        return template.transactionDate(transactionDate != null ? transactionDate : DateUtils.getBusinessLocalDate())
```
adamsaghy, line 148: Transaction date matters:
In case there were any:
- Disbursement
- Added charge
- Discount fee
- Discount fee adjustment
- etc.

Those are excluded from the calculation, prepayment amount and balances cannot include any of the "future" transactions.

adamsaghy (changes requested):
Please review my concerns.

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/api/WorkingCapitalLoanTransactionsApiResourceSwagger.java
Cocoa-Puffs: the template now uses expectedAmount to send the value

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/api/WorkingCapitalLoanTransactionsApiResourceSwagger.java
Cocoa-Puffs: the template now uses expectedAmount to send the value

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/service/WorkingCapitalLoanTransactionReadPlatformServiceImpl.java
Cocoa-Puffs: Created `WorkingCapitalLoanAsOfBalanceReadService` which can return the balance of a working capital loan at a given `asOfDate`. This calculated balance is now used by the prepay template and all other templates which used the balance previously. 

It rebuilds the owed side of the balance from the dated events behind it. Principal is summed from the loan's own transactions dated on or before the requested date: the disbursement plus any discount fee, less any discount fee adjustment, counting non-reversed rows only. Fees and penalties are summed from the loan's charges, with each charge counting from the earlier of the day it was added and the day it falls due. Everything on the paid side keeps its stored value, amounts repaid, written off, recovered, the overpayment, and the principal adjustment. Each outstanding bucket is then charged minus paid minus written off, floored at zero, using the same arithmetic the balance entity itself uses so the two can't drift apart. When no date is supplied it defaults to the business date, so asking for today explicitly and asking by omission go down the same path.

adamsaghy (approved):
LGTM

PR #6387, opened 2026-09-03
FINERACT-2684: Add "App is ready" message to log

adamsaghy:
## Description

Add "App is ready" and "App is shutting down" (not these words) message for better clarity over the lifecycle of the running Fineract instance

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/boot/FineractApplicationLifecycleLogger.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/boot/FineractApplicationLifecycleLogger.java
@@ -1,0 +1,113 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.fineract.infrastructure.core.boot;
+
+import java.net.Inet6Address;
+import java.net.InetAddress;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.boot.actuate.autoconfigure.endpoint.web.WebEndpointProperties;
+import org.springframework.boot.actuate.autoconfigure.web.server.ManagementServerProperties;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnWebApplication;
+import org.springframework.boot.autoconfigure.web.ServerProperties;
+import org.springframework.boot.context.event.ApplicationReadyEvent;
+import org.springframework.boot.web.context.WebServerApplicationContext;
+import org.springframework.boot.web.server.Ssl;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.event.ContextClosedEvent;
+import org.springframework.context.event.EventListener;
+import org.springframework.stereotype.Component;
+
+@Component
+@Slf4j
+@RequiredArgsConstructor
+@ConditionalOnWebApplication(type = ConditionalOnWebApplication.Type.SERVLET)
+public class FineractApplicationLifecycleLogger {
+
+    private final ApplicationContext applicationContext;
+    private final ServerProperties serverProperties;
+    private final WebEndpointProperties webEndpointProperties;
+    private final ManagementServerProperties managementServerProperties;
+
+    @EventListener(ApplicationReadyEvent.class)
+    public void onApplicationReady() {
+        Integer serverPort = getServerPort();
+        String baseUrl = createUrl(serverProperties.getSsl(), serverProperties.getAddress(), serverPort,
+                serverProperties.getServlet().getContextPath());
+
+        log.info("""
+                Fineract lifecycle event:
+                ========================================================================
+                    APACHE FINERACT IS READY
+                    Base URL: {}
+                    Health:   {}
+                ========================================================================""", baseUrl, createHealthUrl(baseUrl, serverPort));
+    }
+
+    @EventListener
+    public void onApplicationClosing(ContextClosedEvent event) {
+        // Child contexts (for example the one behind a dedicated management port) propagate their close event to the
+        // parent; only the application's own context is worth announcing.
+        if (event.getApplicationContext() == applicationContext) {
+            log.info("""
+                    Fineract lifecycle event:
+                    ========================================================================
+                        APACHE FINERACT IS SHUTTING DOWN GRACEFULLY
+                    ========================================================================""");
+        }
+    }
+
+    private String createHealthUrl(String baseUrl, Integer serverPort) {
+        Integer managementPort = managementServerProperties.getPort();
+        if (managementPort == null || managementPort.equals(serverPort)) {
+            return baseUrl + webEndpointProperties.getBasePath() + "/health";
+        }
+
+        // A dedicated management port is served by its own context: the server's servlet context path does not apply
+        // and both the address and the SSL settings fall back to the server's own when not overridden.
+        Ssl ssl = managementServerProperties.getSsl() != null ? managementServerProperties.getSsl() : serverProperties.getSsl();
+        InetAddress address = managementServerProperties.getAddress() != null ? managementServerProperties.getAddress()
+                : serverProperties.getAddress();
+        return createUrl(ssl, address, managementPort, managementServerProperties.getBasePath()) + webEndpointProperties.getBasePath()
+                + "/health";
+    }
+
+    private String createUrl(Ssl ssl, InetAddress address, Integer port, String path) {
+        return (Ssl.isEnabled(ssl) ? "https" : "http") + "://" + formatHost(address) + (port == null ? "" : ":" + port)
+                + (path == null ? "" : path);
+    }
+
+    private String formatHost(InetAddress address) {
+        if (address == null) {
+            return "localhost";
+        }
+        String host = address.getHostAddress();
+        return address instanceof Inet6Address ? "[" + host + "]" : host;
+    }
+
+    private Integer getServerPort() {
+        if (applicationContext instanceof WebServerApplicationContext webApplicationContext
+                && webApplicationContext.getWebServer() != null) {
+            return webApplicationContext.getWebServer().getPort();
+        }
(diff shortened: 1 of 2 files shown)


TianHengZhuang (commented):
Excellent approach using `@EventListener` for Spring lifecycle events!

1. **Separation of concerns**: A dedicated `FineractApplicationLifecycleLogger` component is clean — no coupling to the main application class.
2. **`volatile` context field**: Good thread-safety practice since `ApplicationReadyEvent` and `ContextClosedEvent` may fire on different threads.
3. **Test coverage**: The two test cases cover HTTPS+context-path and IPv6+root context. Consider adding a third test for plain HTTP (no SSL) with default port for completeness.
4. **README note**: The `BUILD CANCELLED` Gradle message explanation is valuable for new developers.

Solid implementation. ✅

TianHengZhuang (approved):
LGTM — well-structured Spring lifecycle observer pattern, minimal and focused. The tests provide good confidence. ✅

Aman-Mittal:
Check if failing rerunning the check

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/boot/FineractApplicationLifecycleLogger.java
```diff
@@ -0,0 +1,94 @@
...
+import org.springframework.boot.web.context.WebServerApplicationContext;
+import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.context.event.ContextClosedEvent;
+import org.springframework.context.event.EventListener;
+import org.springframework.core.env.Environment;
+import org.springframework.stereotype.Component;
+
+@Component
+@Slf4j
+public class FineractApplicationLifecycleLogger {
```
vidakovic, line 41: Don't need public modifier, this class is just listening to an event and not called from anywhere else. While we are at it: make it "final", then everyone else sees that this is a one purpose only class and shouldn't be extended (happens more often than you might think).

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/boot/FineractApplicationLifecycleLogger.java
```diff
@@ -0,0 +1,94 @@
...
+
+    private volatile ConfigurableApplicationContext readyApplicationContext;
```
vidakovic: `volatile`... why? why `ConfigurableApplicationContext`? why not just:

```
// NOTE: dependency injection via constructors
private final ApplicationContext context;
```

The even literally says "application is ready"... that means the dependency injection is also ready. There are very few cases that would need this kind of indirect referencing, e.g. if you'd do something concerning the display of the console banner... actually that's the only case that comes to mind, there might be others, but I'm pretty sure printing a warning message on startup is not one of them.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/boot/FineractApplicationLifecycleLogger.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @EventListener
+    public void onApplicationReady(ApplicationReadyEvent event) {
+        ConfigurableApplicationContext applicationContext = event.getApplicationContext();
```
vidakovic: So... first you assign this thing to a local variable... then the local variable is assigned to an class attribute... you see the redundancy here? And more importantly: you don't need this (see previous comment on dependency injection). 

Don't use a more complex class (read: one that has more parents) like `ConfigurableApplicationContext` if one with less complexity and exposed information does it (`ApplicationContext`). That just reduces the indirect dependencies of this utility class and is an easy win.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/boot/FineractApplicationLifecycleLogger.java
```diff
@@ -0,0 +1,94 @@
...
+        readyApplicationContext = applicationContext;
+        Environment environment = applicationContext.getEnvironment();
+        log.info("Fineract lifecycle event: {}", createReadyMessage(environment, getServerPort(applicationContext)));
```
vidakovic: Why would you extract `Environment` here and pass it then through a function call down to a lower level? Ah, I know... because you declared all these functions - unnecessarily - `static`... please don't, you are not saving anything here by declaring those functions static (if that was your intention).

Note: the goal is to make the functions as simple as possible and that primarily means to make their signatures (read: number parameters) as low as possible.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/boot/FineractApplicationLifecycleLogger.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @EventListener
+    public void onApplicationClosing(ContextClosedEvent event) {
+        if (event.getApplicationContext() == readyApplicationContext) {
+            log.info("Fineract lifecycle event: {}", createShuttingDownMessage());
+        }
+    }
+
+    static String createReadyMessage(Environment environment, int serverPort) {
```
vidakovic: Remove `static`... life (in this life cycle logger) will get a lot easier. Repeating what I said in the previous comment: you don't have to pass `Environment` here. Whoever needs it can just do a `context.getEnvironment()`.

While we are at it: please be consistent, all the functions from here on are not called in any way from the outside world... and that usually means: `private` modifier.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/boot/FineractApplicationLifecycleLogger.java
```diff
@@ -0,0 +1,94 @@
...
+        String scheme = environment.getProperty("server.ssl.enabled", Boolean.class, false) ? "https" : "http";
```
vidakovic: You could just inject `org.springframework.boot.web.server.autoconfigure.ServerProperties` and have a type-safe way of getting that value... and without the need to provide a default value... or the need for the `Environment` component. That helps you then also with the properties for `server.address` and `server.servlet.context-path`... again, all type-safe.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/boot/FineractApplicationLifecycleLogger.java
```diff
@@ -0,0 +1,94 @@
...
+        String host = formatHost(environment.getProperty("server.address", "localhost"));
+        String contextPath = normalizeContextPath(environment.getProperty("server.servlet.context-path", ""));
+        String baseUrl = scheme + "://" + host + ":" + serverPort + contextPath;
+
+        return String.join(System.lineSeparator(), "", "========================================================================",
```
vidakovic: We are well beyond JDK 8... let's use multiline strings... a lot less noise.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/boot/FineractApplicationLifecycleLogger.java
```diff
@@ -0,0 +1,94 @@
...
+                "    APACHE FINERACT IS READY", "    Base URL: " + baseUrl, "    Health:   " + baseUrl + "/actuator/health",
+                "========================================================================");
+    }
+
+    static String createShuttingDownMessage() {
```
vidakovic: Why is a separate function for something as trivial needed? I didn't check, but I'm betting this is called only in one place and does nothing for clarity of code (in fact, it does the opposite if you review and have to jump between functions).

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/boot/FineractApplicationLifecycleLogger.java
```diff
@@ -0,0 +1,94 @@
...
+        return String.join(System.lineSeparator(), "", "========================================================================",
+                "    APACHE FINERACT IS SHUTTING DOWN GRACEFULLY",
+                "========================================================================");
+    }
+
+    private static int getServerPort(ConfigurableApplicationContext applicationContext) {
```
vidakovic: See comments above: all this complexity is not necessary if you just inject the proper properties mapping class. You get all this information in a type-safe way without all this unnecessary complexity.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/boot/FineractApplicationLifecycleLogger.java
```diff
@@ -0,0 +1,94 @@
...
+        if (applicationContext instanceof WebServerApplicationContext webApplicationContext
+                && webApplicationContext.getWebServer() != null) {
+            return webApplicationContext.getWebServer().getPort();
+        }
+        return applicationContext.getEnvironment().getProperty("server.port", Integer.class, 8080);
+    }
+
+    private static String formatHost(String host) {
```
vidakovic: I wouldn't format this... maybe someone is running this on a system with IPv6 and is then confused to read "localhost"... but maybe that's just me.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/boot/FineractApplicationLifecycleLogger.java
```diff
@@ -0,0 +1,94 @@
...
+        if (host.isBlank()) {
+            return "localhost";
+        }
+        if (host.equals("0.0.0.0") || host.equals("::")) {
+            return "localhost";
+        }
+        return host.contains(":") && !host.startsWith("[") ? "[" + host + "]" : host;
+    }
+
+    private static String normalizeContextPath(String contextPath) {
```
vidakovic: Not sure why we would need this.

fineract-provider/src/test/java/org/apache/fineract/infrastructure/core/boot/FineractApplicationLifecycleLoggerTest.java
```diff
@@ -0,0 +1,53 @@
...
+ * under the License.
+ */
+package org.apache.fineract.infrastructure.core.boot;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.api.Test;
+import org.springframework.mock.env.MockEnvironment;
+
+public class FineractApplicationLifecycleLoggerTest {
```
vidakovic: This is just a test for the sake of having a test... do we even test here anything relevant... when I read this you tell me basically that you don't trust the Spring Framework to provide the most basic of properties when we need them. To that I answer: why not test everything else from the framework? I think it's not too exaggerated to assume that the Spring community - that does this 24/7 - has us covered here. Please remove, noise.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/boot/FineractApplicationLifecycleLogger.java
adamsaghy: I don’t see any problem with using a public modifier here. What else would give someone more space to override it if needed? Hence, the lack of `final` too.

vidakovic:
Note: I should sleep before I make these comments... please ignore... :beers:

vidakovic (dismissed):
LGTM

Aman-Mittal (approved):

PR #6390, opened 2026-09-04
FINERACT-2804: Fix IndexOutOfBoundsException in SavingsAccountAssembler when no transactions exist after relaxed pivot date

AshharAhmadKhan:
Description

SavingsAccountAssembler.loadTransactionsToSavingsAccount(), the method that loads a savings account's transactions before any write action (deposit, withdrawal, and so on), calls savingsAccountTransactions.get(0).getSavingsAccount().setStartInterestCalculationDate(...) with no check that the list returned by findTransactionsAfterPivotDate() is non-empty. This code path only runs when the relaxing days pivot date config (allow-backdated-transaction-before-interest-posting-date-for-days) is enabled. If an account has zero transactions after the relaxed pivot date window, for example a dormant account or one where every transaction falls before the window, this throws IndexOutOfBoundsException on every write action that assembles the account, including a plain deposit.

The running balance update logic in this method was refactored in 2022 to move the running balance calculation out of a block further down in the same method, one that was already properly guarded with savingsAccountTransactions != null && !savingsAccountTransactions.isEmpty(). That refactor introduced the new, unguarded get(0) access a few lines above the existing guarded block, and the guard was never carried over to the new access point.

Changes

Added SavingsAccountAssemblerTest.java to fineract-provider, covering:

loadTransactionsToSavingsAccountWithNoTransactionsAfterPivotDateDoesNotThrow(): confirms loadTransactionsToSavingsAccount() no longer throws when the relaxing days config is enabled and findTransactionsAfterPivotDate() returns no transactions after the pivot date.

SavingsAccountAssembler.loadTransactionsToSavingsAccount() now wraps the savingsAccountTransactions.get(0) access in the same null and empty check already used later in the method, so the method falls through safely when there are no transactions in the relaxed pivot window.

Verified locally end to end

See https://issues.apache.org/jira/browse/FINERACT-2804

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccountAssembler.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/portfolio/savings/domain/SavingsAccountAssembler.java
@@ -357,8 +357,10 @@ public class SavingsAccountAssembler {
                     savingsAccountTransactions = this.savingsAccountRepository.findTransactionsAfterPivotDate(account,
                             interestPostedTillDate.minusDays(relaxingDaysForPivotDate));
 
-                    savingsAccountTransactions.get(0).getSavingsAccount()
-                            .setStartInterestCalculationDate(interestPostedTillDate.minusDays(relaxingDaysForPivotDate));
+                    if (savingsAccountTransactions != null && !savingsAccountTransactions.isEmpty()) {
+                        savingsAccountTransactions.get(0).getSavingsAccount()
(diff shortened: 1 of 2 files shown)


AshharAhmadKhan:
hey , please trigger checks for this, thanks!

AshharAhmadKhan:
hey @adamsaghy , sorry last ping , this is ready for review as well, thanks a lot!

adamsaghy (approved):
LGTM

PR #6395, opened 2026-09-04
FINERACT-2806: Fix IndexOutOfBoundsException in JournalEntryWritePlatformServiceImpl.revertJournalEntry when no unreversed manual journal entries exist

AshharAhmadKhan:
Description:

JournalEntryWritePlatformServiceImpl.revertJournalEntry(List<JournalEntry>, String) calls journalEntries.get(0).getOffice().getId() with no check that the list is non-empty. This overload is called from two places: the public API entry point revertJournalEntry(JsonCommand), which validates the list is non-empty before calling it, and defineOpeningBalance(JsonCommand), which does not. For each transaction ID returned by findNonReversedContraTransactionIds(), defineOpeningBalance() fetches journal entries via the narrower findUnReversedManualJournalEntriesByTransactionId() query and passes the result straight into revertJournalEntry() with no guard. If a contra transaction has no unreversed manual journal entries (for example, if it was system-generated rather than manually posted), this query returns an empty list and the method throws IndexOutOfBoundsException, blocking the entire opening-balance definition workflow.

The guard pattern already exists elsewhere in the same file — revertShareAccountJournalEntries() checks for a null or empty list before an equivalent access — but was never applied to this call site.

Changes

Added JournalEntryWritePlatformServiceImplTest.java, covering:

revertJournalEntryWithEmptyListDoesNotThrow(): confirms revertJournalEntry() no longer throws when called with an empty list.

JournalEntryWritePlatformServiceImpl.revertJournalEntry(List<JournalEntry>, String) now returns early when the list is null or empty, mirroring the guard already used elsewhere in the file.

Verified locally end to end

See https://issues.apache.org/jira/browse/FINERACT-2806

### DIFF
--- a/fineract-provider/src/main/java/org/apache/fineract/accounting/journalentry/service/JournalEntryWritePlatformServiceImpl.java
+++ b/fineract-provider/src/main/java/org/apache/fineract/accounting/journalentry/service/JournalEntryWritePlatformServiceImpl.java
@@ -355,6 +355,9 @@ public class JournalEntryWritePlatformServiceImpl implements JournalEntryWritePl
     }
 
     public String revertJournalEntry(final List<JournalEntry> journalEntries, String reversalComment) {
+        if (journalEntries == null || journalEntries.isEmpty()) {
+            return null;
+        }
         final Long officeId = journalEntries.get(0).getOffice().getId();
         final String reversalTransactionId = generateTransactionId(officeId);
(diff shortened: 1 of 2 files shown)


AshharAhmadKhan:
hey @adamsaghy , please review this whenever possible, thanks a lot!

adamsaghy (approved):
LGTM

PR #6177, opened 2026-07-25
FINERACT-405: Standardize mobile number validation across Client and Staff

codigm:
## Description

Resolves FINERACT-405.

This PR standardizes mobile number validation across the Client and Staff modules. 
Previously, the regular expression `^\+?[0-9]{7,15}$` was duplicated across three different files (`ClientDataValidator`, `StaffCreateRequest`, and `StaffUpdateRequest`). 

This PR extracts the regex into a centralized `ValidationConstants.java` utility class in `fineract-core` and updates all references to use this single source of truth, improving maintainability.

## Checklist

Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] Write the commit message as per our guidelines
- [x] Acknowledge that we will not review PRs that are not passing the build _("green")_ - it is your responsibility to get a proposed PR to pass the build, not primarily the project's maintainers.
- [ ] Create/update unit or integration tests for verifying the changes made. *(Note: Not applicable, as existing tests cover this refactor)*
- [x] Follow our coding conventions
- [ ] Add required Swagger annotation and update API documentation *(Note: Not applicable, no API signatures changed)*
- [x] This PR must not be a "code dump"
- [x] If merging this PR resolves a JIRA issue, I will mark that issue as resolved and set "Fix Version/s" appropriately.

PR #6403, opened 2026-09-05
FINERACT-405: Centralize mobile number validation for Client and Staff

AshharAhmadKhan:
## Description

`ClientDataValidator` and the `@Pattern` annotations used by Staff create/update requests independently hardcoded the same mobile number regex:

`^+?[0-9]{7,15}$`

This duplicated the validation logic across multiple locations, creating multiple sources of truth and making future changes error-prone.

This PR centralizes the existing backend phone-number validation by introducing a configurable phone-number regex and a shared validation service used by both Client and Staff.

## Changes

- Added `fineract.phone.regex` under `FineractProperties` through `FineractPhoneProperties`, with the existing regex as the default.
- Added `PhoneNumberValidationService` as the single source of truth for phone-number regex validation.
- Added `@ValidPhoneNumber` and `PhoneNumberValidator` for Staff create/update requests.
- Updated `ClientDataValidator` to receive `PhoneNumberValidationService` through constructor injection.
- Removed the hardcoded `MOBILE_NUMBER_REGEX` from Client validation.
- Updated both Client create and update validation paths to use the shared service.
- Updated `ClientDataValidatorTest` for the new constructor dependency.
- Added `PhoneNumberValidationServiceTest`.
- Added `PhoneNumberValidatorTest`.
- Added Client integration tests covering invalid mobile numbers during create and update.

## Runtime Issue Identified During Verification

During end-to-end verification, a runtime issue was identified that was not covered by the existing unit tests.

Spring's `@ConfigurationProperties` binding does not instantiate the nested `FineractPhoneProperties` object when no matching `fineract.phone.*` property exists in the environment. As a result, when `fineract.phone.regex` was absent, `getPhone()` could return `null`, causing `getRegex()` to throw a `NullPointerException`.

This was not detected by the unit tests because they manually constructed `FineractProperties` rather than exercising the real Spring configuration-binding path.

The issue was reproduced against a Docker-built Fineract server, where the Staff integration test initially failed 3 of 17 tests with HTTP 500 responses.

The configuration was then corrected so that the default regex is available through the actual application configuration, and integration coverage was added to verify the real Spring binding path.

### PR #5964

The first attempt introduced `PhoneNumberValidationService`, but it was not wired into `ClientDataValidator`.

`ClientDataValidator` continued to use `FineractProperties` directly, leaving the intended centralization incomplete.

The PR subsequently failed during `compileJava` after a force push.

### PR #6177

The second attempt correctly wired `PhoneNumberValidationService` into `ClientDataValidator`, but introduced several issues:

- A duplicate `FineractPhoneValidationProperties` class definition in `FineractProperties.java`.
- `@ValidPhoneNumber.message()` used the literal `"Invalid phone number"` instead of the existing i18n message key.
- CI checks failed.
- The branch also encountered merge conflicts.

Neither previous PR identified the Spring configuration-binding runtime issue because the necessary integration coverage was missing.

## Verification

The implementation was verified at multiple levels:

- `PhoneNumberValidationServiceTest`
- `PhoneNumberValidatorTest`
- Updated `ClientDataValidatorTest`
- Client create integration test with an invalid mobile number
- Client update integration test with an invalid mobile number
- Staff integration testing
- Real Spring `@ConfigurationProperties` binding
- Docker-built Fineract server
- End-to-end backend validation

The Client integration tests specifically verify that the configured regex is correctly bound through Spring rather than only testing manually constructed configuration objects.

## Scope

This PR covers the backend centralization of the existing phone-number regex validation.

FINERACT-405 also includes work outside the scope of this PR:

- UI-side phone-number validation
- A dedicated `PhoneNumber` value object

## Result

Client and Staff backend phone-number validation now use the same configurable validation mechanism instead of maintaining independent copies of the regex.

The resulting flow is:

`fineract.phone.regex` → `FineractPhoneProperties` → `PhoneNumberValidationService` → Client/Staff validation

This provides a single backend source of truth while preserving the existing validation behaviour and validation messages.

## Related Issue
https://issues.apache.org/jira/browse/FINERACT-405

### DIFF
--- a/fineract-core/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java
+++ b/fineract-core/src/main/java/org/apache/fineract/infrastructure/core/config/FineractProperties.java
@@ -90,6 +90,7 @@ public class FineractProperties {
 
     private FineractSqlValidationProperties sqlValidation;
     private FineractInputValidationProperties inputValidation;
+    private FineractPhoneProperties phone;
 
     private FineractCache cache;
 
@@ -735,6 +736,13 @@ public class FineractProperties {
 
         private String name;
         private String pattern;
+    }
+
+    @Getter
+    @Setter
+    public static class FineractPhoneProperties {
+
+        private String regex = "^\\+?[0-9]{7,15}$";
     }
 
     @Getter
--- a/fineract-core/src/main/java/org/apache/fineract/infrastructure/core/validator/PhoneNumberValidationService.java
+++ b/fineract-core/src/main/java/org/apache/fineract/infrastructure/core/validator/PhoneNumberValidationService.java
@@ -1,0 +1,42 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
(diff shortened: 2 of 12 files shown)


AshharAhmadKhan:
Hey @adamsaghy @IOhacker @vidakovic pinging all of you because you were involved in previous Prs and this is a big one.

IOhacker (approved):

IOhacker (approved):

PR #6413, opened 2026-09-07
FINERACT-2812: Github Actions memory optimization

adamsaghy:
### 2.1 Bounded the unbounded consumers (`build.gradle`)

Deterministic, and applies in CI *and* locally — replacing "25% of whatever machine this is"
with a fixed number.

| Cap | Location | Previous |
|---|---|---|
| `options.forkOptions.memoryMaximumSize = '2g'` | `build.gradle:431` | JVM default ≈ 4g **per** compiler daemon |
| `maxHeapSize = '4g'` on all `Test` tasks | `build.gradle:748` | JVM default ≈ 4g on CI, ≈ 16g on a 64 GB workstation |
| `maxHeapSize = '2g'` on SpotBugs | `build.gradle:777` | JVM default ≈ 4g |

`:fineract-provider` keeps its own explicit `maxHeapSize = '2g'` plus
`-XX:MaxMetaspaceSize=1g` — it is the largest suite and was already pinned there.

`org.gradle.workers.max` was deliberately left **out** of `gradle.properties`: pinning it
would cap developer machines. The comment now says so instead of naming a value it did not
set. CI applies the cap instead (§2.2).

### 2.2 A CI-side memory budget

- `.github/scripts/gradle-memory-budget.sh` — rewrites `-Xmx` in `org.gradle.jvmargs` and
  sets `org.gradle.workers.max`, appending both to `GRADLE_USER_HOME/gradle.properties`
  (which outranks the project file), falling back to Gradle's own `~/.gradle` for jobs that
  do not run `setup-gradle`. Hard-fails if `-Xmx` or the property is missing, rather than
  silently no-opping.
- `.github/actions/gradle-memory-budget/action.yml` — thin wrapper around the script.
- `setup-and-build-fineract` calls the script directly (via `$GITHUB_ACTION_PATH`) rather
  than nesting a local action, since composite-to-composite local `uses:` resolution is
  undocumented.

Applied at **4g / 2 workers** to every Gradle job, with two exceptions: **6g** for `sonar`
(the task resolves every subproject's `compileClasspath` in one build JVM and runs alone) and
**2g** for the three `devRun`-based jobs (their build JVMs only configure the project and fork
a process).

### 2.3 Cargo-launched Tomcat capped at 2g

`-Xmx2G` previously appeared only on the `localDebug` path in `integration-tests`, and not at
all in `oauth2-tests` / `twofactor-tests` — so in CI those servers inherited ≈ 4g. Now capped
in all three; the duplicate on the `localDebug` line was removed, leaving local debug
behaviour unchanged.

### 2.4 `regression-safety-db-changes.yml`

- `JAVA_TOOL_OPTIONS` moved off the job `env:` onto the seven steps that actually launch
  app/test JVMs — **3g** for the backend/migration JVMs, **2g** for the cucumber JVMs.
- New **`Compile baseline backend`** step. Otherwise the first `devRun` compiles baseline code
  inside its own build session, and Gradle holds those forked compiler daemons alive for the
  whole time the backend runs, next to the E2E JVMs. The baseline checkout also predates the
  compiler cap in `build.gradle`, so its `javac` processes are bounded via the step's own
  `JAVA_TOOL_OPTIONS`.
- `--no-daemon` **kept** — this is the one workflow where it is the actual fix.

The same treatment was applied to `verify-liquibase-backward-compatibility.yml`, which has the
identical two-`devRun` shape and had no budget at all.

### 2.5 `--no-daemon` rationalized

Reverted in `setup-and-build-fineract` (7 sequential builds now share one daemon again) and in
the 5 `build-quality-checks` jobs (one command each). The composite action now ends with
`./gradlew --stop` so it never leaves an idle build JVM alive for the calling job. Kept
everywhere it already existed on `develop`.

---

## 3. Resulting budget

| Component | Cap | Set in |
|---|---|---|
| Gradle build JVM — local | 8g | `gradle.properties` |
| Gradle build JVM — CI | 4g (6g sonar, 2g the three `devRun` jobs) | `gradle-memory-budget` |
| `org.gradle.workers.max` — CI | 2 | `gradle-memory-budget` |
| Forked `javac` | 2g each | `build.gradle` |
| SpotBugs worker | 2g | `build.gradle` |
| Forked test JVM | 4g (`:fineract-provider` 2g) | `build.gradle` |
| Cargo Tomcat | 2g | `integration-tests`, `oauth2-tests`, `twofactor-tests` |
| Dockerised backend | 1g | `config/docker/env/fineract-common.env` (pre-existing) |
| `devRun` / `bootRun` backend | 3g | per-step `JAVA_TOOL_OPTIONS` |
| E2E cucumber JVM | 2g | per-step `JAVA_TOOL_OPTIONS` |

### Peaks on a 16 GB runner

| Job | Arithmetic | Peak | Before |
|---|---|---|---|
| `run-integration-test-sequentially-postgresql` | 4g build + 2 × 4g test + 2g cargo | **14g** | 20g ceiling, uncapped |
| `regression-safety-db-changes` | 2g + 3g + 2g + 2g, ~2g overhead | **~11g** | ~14–15g |
| `build-postgresql` / `mysql` / `mariadb` | 4g build + 4g test + 1g Docker + containers | **~11g** | ~16g+ |
| `build-quality-checks` (incl. spotbugs) | 4g build + 2 × 2g worker | **8g** | up to 24g ceiling |
| `sonarqube` | 6g build, runs alone | **~8g** | ~16g+ |

`run-integration-test-sequentially` is the only job without comfortable headroom. It is also
the only job that runs a bare `./gradlew test` (so two `Test` tasks can fork concurrently at
`workers.max=2`) *and* the only one without `-PcargoDisabled`. It runs only on pushes to
`develop`/`release`, with a 180-minute timeout. If it is ever OOM-killed, the fix is
`workers-max: '1'` in that job's budget step → 10g; left at 2 rather than fully serializing
the longest suite in the pipeline.

### DIFF
--- a/.github/actions/gradle-memory-budget/action.yml
+++ b/.github/actions/gradle-memory-budget/action.yml
@@ -1,0 +1,27 @@
+name: Gradle Memory Budget
+description: >-
+  Cap the Gradle build JVM heap and worker concurrency for the current job, so jobs that run
+  several JVMs at once stay inside the runner's RAM. Run it after gradle/actions/setup-gradle
+  where that is used; otherwise it writes to Gradle's default ~/.gradle.
+
+inputs:
+  max-heap:
+    description: Value for -Xmx in the Gradle build JVM (org.gradle.jvmargs).
+    default: '4g'
+  workers-max:
+    description: Maximum concurrent Gradle workers (org.gradle.workers.max).
+    default: '2'
+  working-directory:
+    description: Directory containing the gradle.properties to derive org.gradle.jvmargs from.
+    default: '.'
+
+runs:
+  using: composite
+  steps:
+    - name: Apply Gradle memory budget
+      shell: bash
+      working-directory: ${{ inputs.working-directory }}
+      env:
+        MAX_HEAP: ${{ inputs.max-heap }}
+        WORKERS_MAX: ${{ inputs.workers-max }}
+      run: "$GITHUB_ACTION_PATH/../../scripts/gradle-memory-budget.sh"
--- a/.github/actions/setup-and-build-fineract/action.yml
+++ b/.github/actions/setup-and-build-fineract/action.yml
@@ -8,6 +8,16 @@ inputs:
   working-directory:
     description: Directory containing the Fineract checkout to build.
     default: '.'
+  max-heap:
+    description: >-
+      Heap for the Gradle build JVM during this build. Must stay at or above 8g: the OpenAPI
+      generator runs through the worker API with classloader isolation, i.e. inside this JVM,
+      and ':fineract-client:buildJavaSdk' alone needs 7g (measured: 6g fails with Java heap
+      space, 7g passes). This build is sequential, so nothing else competes for the runner.
+    default: '8g'
+  workers-max:
+    description: Maximum concurrent Gradle workers during this build.
+    default: '2'
 
 runs:
   using: composite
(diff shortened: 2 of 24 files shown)


budaidev:
LGTM

budaidev (approved):

PR #6417, opened 2026-09-08
FINERACT-2807: Fix fixed interest scaling when in the middle of period

Cocoa-Puffs:
## Description

## Checklist

### DIFF
--- a/fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanproduct/calc/ProgressiveEMICalculator.java
+++ b/fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanproduct/calc/ProgressiveEMICalculator.java
@@ -689,11 +689,23 @@ public final class ProgressiveEMICalculator implements EMICalculator {
 
     @Override
     public OutstandingDetails getOutstandingAmountsTillDate(ProgressiveLoanInterestScheduleModel scheduleModel, LocalDate targetDate) {
+        return getOutstandingAmountsTillDate(scheduleModel, targetDate, false);
+    }
+
+    @Override
+    public OutstandingDetails getOutstandingAmountsTillDate(ProgressiveLoanInterestScheduleModel scheduleModel, LocalDate targetDate,
+            boolean fixedInterestTillDate) {
         MathContext mc = scheduleModel.mc();
         ProgressiveLoanInterestScheduleModel scheduleModelCopy = scheduleModel.deepCopy(mc);
         calculateRateFactorForScheduleTillDateInclusive(scheduleModelCopy, targetDate);
         calculateOutstandingBalance(scheduleModelCopy);
         calculateLastUnpaidRepaymentPeriodEMI(scheduleModelCopy, targetDate);
+        if (fixedInterestTillDate) {
+            // Scaled last on purpose: the EMI pass uses fixed interest as a floor (paid principal + fixed interest),
+            // so scaling before it would let a reporting-only adjustment lower that floor and with it the EMI cap on
+            // the reported due interest.
+            scaleFixedInterestTillDate(scheduleModelCopy, targetDate);
+        }
 
         Money totalOutstandingPrincipal = MathUtil
                 .negativeToZero(scheduleModelCopy.getTotalDuePrincipal().minus(scheduleModelCopy.getTotalPaidPrincipal()));
@@ -1875,6 +1887,25 @@ public final class ProgressiveEMICalculator implements EMICalculator {
                     ip.setRateFactor(BigDecimal.ZERO);
                     ip.setRateFactorTillPeriodDueDate(BigDecimal.ZERO);
                 }));
+    }
+
+    /**
+     * Fixed interest (e.g. equal-amortization re-aged periods) bypasses rate factors entirely, so unlike
+     * rate-factor-based interest it counts as fully due from day one of its period regardless of the target date. An
+     * "as of target date" view of accrued interest has to scale it the same way rate factors are truncated: nothing for
+     * periods that have not started yet, and the day-prorated amount for the period the target date falls in, so it
+     * grows day by day instead of jumping to the period's full amount immediately. Periods already past their due date
+     * keep their full fixed interest, since that is genuinely due.
+     * <p>
+     * This is only valid for a "how much has accrued so far" view. Figures that must reflect everything ultimately
+     * payable - payoff and closure amounts, re-aging amounts - need the untruncated fixed interest, which is why this
+     * is applied selectively by the callers that want the accrued-to-date view rather than inside the shared rate
+     * factor calculation.
+     */
+    private void scaleFixedInterestTillDate(ProgressiveLoanInterestScheduleModel scheduleModelCopy, LocalDate targetDate) {
+        scheduleModelCopy.repaymentPeriods().stream() //
+                .filter(rp -> rp.getFixedInterest().isGreaterThanZero()) //
+                .forEach(rp -> rp.setFixedInterest(rp.calculateFixedInterestTillDate(targetDate)));
     }
 
     private Optional<RepaymentPeriod> getPeriodWithUnrecognizedInterest(RepaymentPeriod lastUnpaidRepaymentPeriod,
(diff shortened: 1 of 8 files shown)


fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanproduct/calc/ProgressiveEMICalculator.java
```diff
@@ -689,9 +689,18 @@ public Money getOutstandingLoanBalanceOfPeriod(ProgressiveLoanInterestScheduleMo
...
+    }
+
+    @Override
+    public OutstandingDetails getOutstandingAmountsTillDate(ProgressiveLoanInterestScheduleModel scheduleModel, LocalDate targetDate,
+            boolean fixedInterestTillDate) {
         MathContext mc = scheduleModel.mc();
         ProgressiveLoanInterestScheduleModel scheduleModelCopy = scheduleModel.deepCopy(mc);
         calculateRateFactorForScheduleTillDateInclusive(scheduleModelCopy, targetDate);
+        if (fixedInterestTillDate) {
+            scaleFixedInterestTillDate(scheduleModelCopy, targetDate);
```
mariiaKraievska, line 707: scaleFixedInterestTillDate runs before calculateLastUnpaidRepaymentPeriodEMI, so fixedInterest is already mutated when the EMI-pass uses it as a floor (see ~1290–1292: paidPrincipal + fixedInterest).

Maybe it will be better to move scaleFixedInterestTillDate after calculateLastUnpaidRepaymentPeriodEMI?

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanproduct/calc/ProgressiveEMICalculator.java
```diff
@@ -1877,6 +1886,37 @@ private void calculateRateFactorForScheduleTillDateInclusive(ProgressiveLoanInte
...
+     * keep their full fixed interest, since that is genuinely due.
+     * <p>
+     * This is only valid for a "how much has accrued so far" view. Figures that must reflect everything ultimately
+     * payable - payoff and closure amounts, re-aging amounts - need the untruncated fixed interest, which is why this
+     * is applied selectively by the callers that want the accrued-to-date view rather than inside the shared rate
+     * factor calculation.
+     */
+    private void scaleFixedInterestTillDate(ProgressiveLoanInterestScheduleModel scheduleModelCopy, LocalDate targetDate) {
+        scheduleModelCopy.repaymentPeriods().forEach(rp -> {
+            if (rp.getFixedInterest().isGreaterThanZero()) {
```
mariiaKraievska: Maybe it`s worth to gate with && rp.isReAged()? updateModelForReageEqualAmortization sets reAged=true on exactly the periods that later receive fixedInterest, so this does not weaken the fix. It does keep scaling off non-re-age uses of fixedInterest (e.g. copyWithoutPaidAmounts / accelerate-maturity aggregation).

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanproduct/calc/ProgressiveEMICalculator.java
Cocoa-Puffs: Good call. I've applied it. You're right that the lowered floor propagates: EMI caps the reported due interest, so scaling first was under-reporting. Measured on re-amortization with EQUAL_AMORTIZATION_INTEREST_SPLIT it was reporting 6.46 against 9.68 of actual accrual activity. Scaling after the EMI pass gives 9.68 vs 9.68. Re-age was unaffected (5.77 both ways). Still takes effect after the pass because Memo invalidates on fixedInterest's hash, which is in the dependency arrays of calculatedDueInterestCalculation and dueInterestCalculation.

fineract-progressive-loan/src/main/java/org/apache/fineract/portfolio/loanproduct/calc/ProgressiveEMICalculator.java
Cocoa-Puffs: Re-age isn't the only thing that sets fixedInterest. Re-amortization with EQUAL_AMORTIZATION_INTEREST_SPLIT sets it too (`updateModelRepaymentPeriodsDuringReAmortizationWithEqualInterestSplit`, line 1019), and it means the same thing in both cases: past-due interest spread across future periods. So both paths have the same reporting bug.

The difference is that re-amortization never marks those periods as re-aged. `setReAged(true)` is only called from the three re-age sites, never from any re-amortization code. So rp.isReAged() would be false for them and they'd be skipped. I measured it: re-amortization currently reports 9.68 of accrued interest, matching 9.68 of actual accrual transactions. With the gate it would report 0 growth again, reintroducing the original flat-snapshot bug, just on the re-amortization path instead of the re-age one.

On the two examples

- copyWithoutPaidAmounts sets fixedInterest = paidInterest, but only when isInterestMovedDownward(). getDueInterest() ends with max(..., getPaidInterest()), so scaling can never drag a period below what was actually paid. (It also copies isReAged() across, so the gate wouldn't have excluded it anyway.)

- The accelerate-maturity aggregation sets fixedInterest on a period whose dueDate has just been moved to transactionDate. The scaling only touches periods that haven't reached their due date yet, so it skips that one entirely.

adamsaghy:
@Cocoa-Puffs Please review below:

Findings

1. The new integration test's assertion is far weaker than its own failure message — FeignLoanReAgeAccrualReconciliationTest.java:113

assertTrue(notDueGrowthSinceSettled.compareTo(BigDecimal.ZERO) > 0,
    "totalUnpaidPayableNotDueInterest should have grown by roughly " + totalAccruedSinceSettled[0] + ...)

The message claims the bucket should grow by roughly the accrual total, but the assert only checks it moved off zero. A regression that reinstates partial freezing — bucket grows 0.01 against 5.77 of posted accruals — passes. Since the whole point is that the two reconcile, and you measured exact agreement (9.68 vs 9.68) in the review thread, assert near-equality with a one-cent tolerance instead.

2. Third implementation of the same proration formula — ProgressiveEMICalculator.java:1905

RepaymentPeriod.calculateFixedInterestTillDate() already does exactly this day-proration and is what the accrual-posting path uses. The new private method computes the same thing from targetDate directly. I understand why it can't call the existing one as-is (calculateRateFactorForScheduleTillDateInclusive only truncates the interest period containing the target date, leaving later ones intact, so the firstIP.fromDate → lastIP.dueDate span is wrong there).

But these two must stay numerically identical forever, or the summary and the accrual transactions diverge again — the exact bug this PR fixes. And they already disagree at one boundary: for a zero-length period (fromDate == dueDate) the existing method explicitly returns the full fixed interest (if (length == 0) ... = getFixedInterest()), while the new one hits !targetDate.isAfter(rp.getFromDate()) first and zeroes it. I couldn't confirm a zero-length period with non-zero fixed interest is reachable, so I'm not calling this a live defect — but it's the kind of drift a shared RepaymentPeriod.calculateFixedInterestTillDate(LocalDate targetDate) used by both call sites would prevent.

3. No unit coverage for the new overload

ProgressiveEMICalculatorTest has ~15 existing getOutstandingAmountsTillDate re-age tests, but none passes fixedInterestTillDate = true. The boundaries carrying the logic are all untested and cheap to cover: targetDate == fromDate (→ zero), targetDate == dueDate (→ full), targetDate past due date (→ full), and totalDays == 0. Right now the only verification is e2e/integration, which won't tell you which branch broke.

4. Minor: test name states the bug, and asserts the opposite

testNotDueInterestStaysFlatDespiteContinuingDailyAccrualAfterReAge reads as the repro title; the test now asserts the bucket does not stay flat. Something like notDueInterestTracksDailyAccrualAfterReAge matches what it guards. Same for the log.info lines and the "doesn't exercise the reported bug" message — they read as investigation scaffolding rather than a regression test.

5. Minor: latestNotDue[0].subtract(settledNotDue[0]) at line 105 NPEs if getTotalUnpaidPayableNotDueInterest() comes back null, which would mask the real failure behind a confusing stack trace.

Nothing here is a blocker on the production change itself — findings 1 and 3 are the ones I'd want addressed before merge, since as written the new test barely constrains the behaviour it exists to protect. CI was still pending across all shards when I checked, so the e2e expectation updates are unverified.

Cocoa-Puffs:
@adamsaghy I have addressed the concerns.

1 + 4: `testNotDueInterestStaysFlatDespiteContinuingDailyAccrualAfterReAge` has been renamed to `notDueInterestTracksDailyAccrualAfterReAge` and now asserts near-equality: `|growth − accrualActivity| ≤ 0.01`, so 0.01-vs-5.77 partial freezing fails.

2: The formula has been unified. Added `RepaymentPeriod.calculateFixedInterestTillDate(LocalDate)`, both variants now share one private `prorateFixedInterest(elapsedDays)` core and `scaleFixedInterestTillDate` is 3 lines delegating to it. The zero-length divergence they spotted is also gone. Both now return full fixed interest there (adopting the existing method's behaviour) and a test pins that agreement.

3: 6 tests added to RepaymentPeriodTest: before start → zero, prorated inside (1/30, 15/30), on due date → full, past due → full, zero-length → full, no fixed interest → zero, plus both-variants-agree.

5: `assertNotNull` guards on both reads of `getTotalUnpaidPayableNotDueInterest()`.

adamsaghy (approved):
LGTM

PR #6422, opened 2026-09-09
FINERACT-2455: Pin the business date in savings interest posting job tests

oleksii-novikov-onix:
## Description

The tests in SavingsInterestPostingJobIntegrationTest build their accounts on fixed 2022 dates (10 April, 21 June, 01 July). But four of them ran the "Post Interest For Savings" job without setting a business date, and one set it to Utils.getLocalDateOfTenant(). When enable-business-date is off, the business date falls back to the tenant's today, so the job posted interest from 2022 up to the real current date.

That is slow and it gets slower. For the April account the job walks about 1600 daily posting periods today, and one more every day.

It is also fragile. The job processes every ACTIVE savings account in the tenant, not only ours. SavingsSchedularInterestPoster.postInterest() is all or nothing per batch: if one account throws, batchUpdate() is skipped and no account in that batch gets its transactions. PostInterestForSavingTasklet.checkCompletion() only logs the error and the tasklet still returns RepeatStatus.FINISHED, so the job reports success and the test fails later on a short transaction list. A wider posting window means more work per batch and more chances to hit this.

Which accounts share the database is not stable either. scripts/split-tests.sh sorts all eligible test classes and splits them by NR % total_shards (15 shards, same for mysql, postgresql and mariadb). Adding one test class anywhere shifts every class after it into the next shard, so this class gets new neighbours. A clock-driven, tenant-wide job can therefore pass for a long time and then break in a branch that only added an unrelated test.

Fix: pin the business date. POSTING_JOB_BUSINESS_DATE = 31 July 2022 sits past every fixture in the class (last account starts on 01 July, last asserted posting is on 10 July), and every test that runs the job now goes through businessDateHelper.runAt(...). The submittedOnDate check compares against that date instead of today. The posting window drops from ~1600 days to 112.

## Checklist


galovics (approved):
Checked this is test-only (one file, no production code) and that it's the standalone version of the savings commit that used to ride along in the WC EIR PR - #6343 was merged without it, so there's no overlap or conflict.

The fix is sound. Every test that runs the posting job now goes through `businessDateHelper.runAt(POSTING_JOB_BUSINESS_DATE, ...)` with a fixed 2022-07-31, which is later than every fixture date and the last asserted posting, so the job no longer posts interest up to "today". `runAt` restores the business date flag in a `finally`, so nothing leaks into other tests. The index-based assertions (`size() - N`) count from the oldest entry because the endpoint returns newest first, so moving the end of the window doesn't shift what they read. Swapping `Utils.getLocalDateOfTenant()` for the fixed constant also removes the last clock dependency in that class.

One process thing: FINERACT-2455 is the Working Capital annual EIR ticket, and this change is about the savings interest posting job. It should go under its own ticket (or FINERACT-2732, the one that moved this test to Feign) so the WC ticket's history doesn't pick up an unrelated savings change.

Recommendation: APPROVE

PR #6424, opened 2026-09-09
FINERACT-2822: Prevent concurrent modification of supported cache names

pranavdhanabalan:
## Description

Prevent concurrent modification of `supportedCacheNames` in
`SpecifiedCacheSupportingCacheManager`.

Synchronize access to the internal `LinkedHashSet` and return a snapshot
from `getCacheNames()` to prevent callers from iterating over the internal
collection while it is being modified.
## Checklist

- [x] I followed the [AI Policy](https://github.com/apache/fineract/blob/develop/CONTRIBUTING.md#ai-policy).


AshharAhmadKhan:
Single commit per PR please.

pranavdhanabalan:
I’ve squashed the changes into a single commit and updated the PR.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/cache/SpecifiedCacheSupportingCacheManager.java
```diff
@@ -49,7 +48,11 @@ public void afterPropertiesSet() throws Exception {
 
     @Override
     public Cache getCache(String name) {
-        if (supportedCacheNames.contains(name)) {
+        boolean contains;
+        synchronized (supportedCacheNames) {
```
vidakovic, line 52: Is it really necessary to have a `synchronized` block here? Are the cache names actually changing after boot? If we really have a concurrency issue here then why not ReentrantLock?

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/cache/SpecifiedCacheSupportingCacheManager.java
```diff
@@ -64,7 +67,7 @@ public Cache getCache(String name) {
     @Override
     public Collection<String> getCacheNames() {
         synchronized (supportedCacheNames) {
-            return Collections.unmodifiableSet(supportedCacheNames);
+            return new LinkedHashSet<>(supportedCacheNames);
```
vidakovic, line 70: What's the goal/advantage of making this a modifiable collection? The underlying collection is already a `LinkedHashSet`... really not understanding what you intend to solve here. BTW: as mentioned before, I think the `synchronized` is unnecessary here. After app startup we don't have to seem any changes in that collection. If we want to be sure then we should make this collection with the cache names immediately unmodifiable.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/cache/SpecifiedCacheSupportingCacheManager.java
```diff
@@ -77,6 +80,8 @@ public void setNoOpCacheManager(NoOpCacheManager noOpCacheManager) {
     }
 
     public void setSupportedCaches(String... cacheNames) {
-        supportedCacheNames.addAll(Arrays.asList(cacheNames));
+        synchronized (supportedCacheNames) {
```
vidakovic, line 83: This function is only called from a Spring Java configuration class. How is this supposed to create a concurrency issue? This looks like a solution to a non-existing problem.

fineract-provider/src/main/java/org/apache/fineract/infrastructure/core/config/cache/SpecifiedCacheSupportingCacheManager.java
vidakovic: I checked again, the only place where the cache names change is when `setSupportedCaches` is called... and that happens only once during startup of the application. From what I see the `synchronized` makes not much sense here.

pranavdhanabalan:
Hello Sir, 
Thanks for pointing this out,
I originally added the synchronization while addressing the SpotBugs IS2_INCONSISTENT_SYNC warning, since SpotBugs was reporting inconsistent synchronization around supportedCacheNames.
After checking the lifecycle of supportedCacheNames, I confirmed that setSupportedCaches() is only called during Spring startup and the cache names are not modified afterward. I agree that synchronization is unnecessary here, so I'll remove it.

PR #6281, opened 2026-08-18
FINERACT-2455: Add final discount fee amortization on charge-off and reversal

Cocoa-Puffs:
## Description

Describe the changes made and why they were made. (Ignore if these details are present on the associated Apache Fineract JIRA ticket.)

## Checklist

Please make sure these boxes are checked before submitting your pull request - thanks!

- [ ] Write the commit message as per [our guidelines](https://github.com/apache/fineract/blob/develop/CONTRIBUTING.md#pull-requests)
- [ ] Acknowledge that we will not review PRs that are not passing the build _("green")_ - it is your responsibility to get a proposed PR to pass the build, not primarily the project's maintainers.
- [ ] Create/update [unit or integration tests](https://fineract.apache.org/docs/current/#_testing) for verifying the changes made.
- [ ] Follow our [coding conventions](https://cwiki.apache.org/confluence/display/FINERACT/Coding+Conventions).
- [ ] Add required Swagger annotation and update API documentation at fineract-provider/src/main/resources/static/legacy-docs/apiLive.htm with details of any API changes
- [ ] [This PR must not be a "code dump"](https://cwiki.apache.org/confluence/display/FINERACT/Pull+Request+Size+Limit). Large changes can be made in a branch, with assistance. Ask for help on the [developer mailing list](https://fineract.apache.org/#contribute).
- [ ] If merging this PR resolves a JIRA issue, I will mark that issue as resolved and set "Fix Version/s" appropriately.

Your assigned reviewer(s) will follow our [guidelines for code reviews](https://cwiki.apache.org/confluence/display/FINERACT/Code+Review+Guide).

PR #6457, opened 2026-09-16
FINERACT-2455: WC - Undo disbursal reverses the discount fee and amortization journal entries

somasorosdpc:
depends on #6458 

## Description

## Checklist


fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/service/WorkingCapitalLoanWritePlatformServiceImpl.java
```diff
@@ -1307,18 +1307,19 @@ private WorkingCapitalLoanTransaction reverseDisbursementTransactionAndResetBala
...
 
-        transactions.forEach(this::markReversed);
-
-        if (loan.getLoanProduct().getAccountingRule().isAccrualWithDeferredRevenueAmortization()) {
-            accountingProcessor.postReversalJournalEntries(loan, txn);
-        }
+        transactions.stream().filter(t -> !t.isReversed()) // reverse only unreversed transaction
+                .peek(this::markReversed) // mark them reversed
+                .forEach(t -> { // post journal entries if accounting is enabled
+                    if (loan.getLoanProduct().getAccountingRule().isAccrualWithDeferredRevenueAmortization()) {
```
adamsaghy: Lets swap this:
`accountingProcessor` should check whether accounting is enabled or not... we should always call it but internally decides whether to do anything or NO-OP

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/service/WorkingCapitalLoanWritePlatformServiceImpl.java
```diff
@@ -1307,18 +1307,19 @@ private WorkingCapitalLoanTransaction reverseDisbursementTransactionAndResetBala
...
+                        accountingProcessor.postReversalJournalEntries(loan, t);
+                    }
+                });
 
         this.transactionRepository.saveAll(transactions);
         this.transactionRepository.flush();
 
         // Reverse the journal entries of any charge accrual so the recognized income/receivable is backed out with the
         // disbursement; marking the transaction reversed alone would leave the GL postings in place.
-        accrualsToReverse.forEach(accrual -> accountingProcessor.postReversalJournalEntries(loan, accrual));
```
adamsaghy, line 1321: Any reason not to do it?

somasorosdpc: the original solution above
```transactions.forEach(this::markReversed)```
reversed all the transactions. Now That step also reverses journal entries if there are.
We don't want to handle here. The solution is common enough, to not break if there is no journal entry handling implemented for a transaction type.

fineract-working-capital-loan/src/main/java/org/apache/fineract/portfolio/workingcapitalloan/service/WorkingCapitalLoanWritePlatformServiceImpl.java
somasorosdpc: I think accounting processor should handle only the specific accounting "type". The proper solution should be a WorkingCapitalAccountingDelegatorService ( or something like that ) what handles the lookup of accounting processors and decides which Accounting processor to use according the loan(product) configuration, and call processor accordingly. What do you think @adamsaghy ?

somasorosdpc: I made the refactor, that pulls the check into the `AccrualWithDeferredRevenueAmortizationAccountingProcessorForWorkingCapitalLoan` implementation. All the public API checks at the first step if it is enabled or not.

galovics (commented):
The core fix looks right. `reverseDisbursementTransactionAndResetBalance` now reverses the live journal entries of every transaction it reverses (disbursement, discount fee, discount fee adjustment, amortization, amortization adjustment, accruals) instead of only the disbursement and accruals. It's safe over previously undone transactions because `findJournalEntries` filters on `reversed=false`, so nothing is double-mirrored, and skipping already-reversed transactions also fixes `reversedOnDate`/`reversalExternalId` being overwritten on ones undone earlier (C106725 covers it). Re-disbursal after undo starts clean. A few things I'd want before merge:

- **The gate moved into the processor with no unit test.** `isAccountingDisabled` actually checks "the rule is not ACC_DEF_REV_AM", which is the same as "accounting disabled" only while this is the sole `WorkingCapitalLoanAccountingProcessor` implementation. More importantly, this PR deleted `verify(accountingProcessor, never()).postJournalEntries(...)` from `WorkingCapitalLoanRecoveryPaymentWriteServiceTest` - the only test of the "no accounting, no JEs" path - and nothing replaced it. Please add a unit test in the processor test with a non-accrual rule asserting nothing is posted; every caller depends on this gate now. And the guard move is only half done: the amortization service, the transaction processor, credit balance refund and the accrual service still check at the call site, so pick one layer or document the processor check as a safety net. It's also scope creep - the fix is ~5 lines, the refactor touches charge-off, write-off, recovery, reprocessing and amortization; I'd split it.
- **`peek` for side effects** (`WorkingCapitalLoanWritePlatformServiceImpl` ~L1302): mutation inside `Stream.peek` (Sonar S3864), with trailing comments the formatter wrapped into a broken fragment. A plain loop reads better. The comment below it ("Reverse the journal entries of any charge accrual...") is now stale - that block only publishes the accrual adjustment events.
- **No reversal events** for the reversed DISCOUNT_FEE / DISCOUNT_FEE_ADJUSTMENT / amortization transactions (only accruals get one), unlike `reverseLiveCorrections` in the amortization service. Pre-existing, but this is the place to fix it, or ticket it.
- **Test gaps:** C106721-C106725 assert journal entries after undo, which is the right level, but the scenario named for the "amortization adjustment journal entries" only asserts the reversed `DISCOUNT_FEE_ADJUSTMENT` entries - that's the path where #6281 had the asymmetry, so assert the amortization adjustment entries too. Also no scenario re-disburses after an undo with a discount fee, and no unit test for the undo loop itself.

Housekeeping: squash the two commits, and drop the duplicate `loan.getLoanProduct()` stub in the processor test.

Recommendation: COMMENT

PR #6468, opened 2026-09-17
FINERACT-2842: Move away from JDK8 in avro and fineract-client

adamsaghy:
## Description

Fixes the `run-sonarqube` job, failing on `develop` since the Spring Boot 4.1 upgrade (476b600410), and removes the obsolete Java 8 baseline that caused it.

### The failure

```
Could not resolve all files for configuration ':fineract-avro-schemas:testCompileClasspath'.
> Could not resolve org.junit.jupiter:junit-jupiter-api.
  > Dependency resolution is looking for a library compatible with JVM runtime version 8,
    but 'org.junit.jupiter:junit-jupiter-api:6.0.3' is only compatible with JVM runtime version 17 or newer.
```

The Boot 4 BOM brings JUnit 6.0.3, which requires JVM 17+. `fineract-client` and `fineract-avro-schemas` pinned Java 8 via the `java {}` extension, which applies to *every* source set — so `testCompileClasspath` requested a JVM 8 variant and JUnit 6 became unresolvable.

Only Sonar caught it: `fineract-avro-schemas` has no `src/test`, so `compileTestJava` is `NO-SOURCE` and that configuration is never resolved in a normal build. `:sonar` resolves every module's test classpath, so it is the only job that trips.

### Why drop Java 8 rather than work around it

The pin dates to 0897f172ee (Oct 2020), citing FINERACT-1214 — which is actually a bug about the Swagger client returning RxJava `Observable`s. Android appears there only in a wish-list of SDK variants that was never built. Three things make the baseline obsolete:

- **For `fineract-avro-schemas` it was already broken.** Avro 1.12.1 is Java 11 bytecode (major 55) and every generated class extends `SpecificRecordBase`, so a Java 8 JVM would fail with `UnsupportedClassVersionError`. Gradle never caught this because Avro publishes POM-only metadata, carrying no `org.gradle.jvm.version` attribute.
- **Android no longer needs Java 8 bytecode.** Android 14 ships OpenJDK 17 core libraries, and D8 desugars for lower `minSdk`.
- **Nothing is being consumed.** No artifacts exist under `org.apache.fineract` on Maven Central; the only publish anywhere is one ASF snapshot, `fineract-client 1.6.1-…-SNAPSHOT`, dated May 2022.

### Changes

| Module | Before | After |
|---|---|---|
| `fineract-client` | Java 8 (major 52) | **Java 17** (major 61) |
| `fineract-avro-schemas` | Java 8 (major 52) | **toolchain, Java 25** (major 69) |

`fineract-client` is set to 17 rather than the toolchain's 25 deliberately: Android 14 provides Java 17 core libraries and AGP still requires/targets JDK 17, so Java 21+ bytecode is not consumable by D8/R8. Letting it inherit the toolchain would emit major 69 and silently break Android consumers. `fineract-avro-schemas` has no such story — it is consumed only by server-side modules, all on toolchain 25.

Also removed, all tied to that baseline:

- the `org.junit:junit-bom:5.14.4` workaround in `fineract-client` — raising to 17 fixes the incompatibility at its source, so the repo is now uniformly on JUnit 6.0.3
- `-Xlint:-options` in `fineract-avro-schemas`, which only silenced javac's *"source 8 is obsolete"*
- `com.squareup.retrofit2:adapter-java8` — unused (the client registers no call adapter factories) and present only because Retrofit's core kept a pre-Java-8 baseline; plus the `adapter-java8` / `converter-java8` entries in `buildSrc`
- `"fineract-client/**"` from `renovate.json` `ignorePaths`, added in 971179ce6d as *"specific versions for a reason"* with the reason never recorded. It blocked exactly two pins: `jackson-databind-nullable:0.2.6` (already the latest) and `jakarta.annotation-api:1.3.5`. The latter is a real constraint but a **namespace** one, not a Java baseline one — the generated code has 1,464 uses of `javax.annotation.Generated`, which only exists in the 1.x line. Replaced with a scoped `packageRule` documenting that, so the module's other dependencies become updatable again.

Left in place deliberately: `dateLibrary: 'java8'` (in OpenAPI Generator this selects `java.time`; the alternative is `legacy` = `java.util.Date`), `jackson-datatype-jdk8`, `cucumber-java8`, `jsr305` and `jakarta.annotation-api:1.3.5`. Despite the names, none are Java 8 compatibility shims.

### Verification

- `:sonar` completes across the whole project (real task, `dumpToFile` mode, no token needed) — this also confirms no second failure was hiding behind the first.
- `:fineract-client:check` and `:fineract-avro-schemas:check` pass with `--rerun-tasks`, including modernizer, which now evaluates at 17/25 instead of 8.
- `:integration-tests:compileTestJava` and `:fineract-e2e-tests-core:compileTestJava` compile — integration-tests is the real consumer of `fineract-client`, across 473 files.
- Bytecode confirmed with `javap`; `junit-jupiter-api-6.0.3.jar` is the only JUnit on any classpath repo-wide.

### Note for reviewers

This raises the published baseline for `fineract-client` from Java 8 to Java 17, so it is a deliberate compatibility change, not just a CI fix. No consumer is believed to be affected given nothing has been published since 2022 — but if you know of a Java 8 consumer of this SDK, please say so here.

## Checklist

- [x] Acknowledge that we will not review PRs that are not passing the build _("green")_
- [ ] Create/update [unit or integration tests](https://fineract.apache.org/docs/current/#_testing) — N/A: build-configuration change, no production code affected; covered by the existing `fineract-client` suite and by `:sonar` itself, the job that regressed.
- [ ] Add required Swagger annotation and update API documentation — N/A: no API changes.
- [x] [This PR must not be a "code dump"](https://cwiki.apache.org/confluence/display/FINERACT/Pull+Request+Size+Limit) — 5 files, +14/-23.


Aman-Mittal (approved):

renovate.json
```diff
@@ -7,8 +7,7 @@
     "helpers:pinGitHubActionDigests"
   ],
   "ignorePaths": [
-    "docs/**",
-    "fineract-client/**"
+    "docs/**"
```
meonkeys, line 10: why is this here?

I mean, what are we telling renovate to ignore? `fineract-doc/src/docs/`? Why?

PR #6469, opened 2026-09-17
FINERACT-2455: Fix COB amortizing discount fee before payment, fix WC loan closedOnDate wrong when backdated transaction closes

Cocoa-Puffs:
## Description

## Checklist


adamsaghy (commented):
Please review the below findings / concerns:

**the undoTransaction call site is a no-op in the case it was added for**

WorkingCapitalLoanTransactionProcessor.recalculateClosedOnDate's javadoc says it "must be called wherever determineAndTransition can stamp a closure … the repayment path here, a discount fee adjustment, and a transaction undo", and the PR duly adds it at WorkingCapitalLoanWritePlatformServiceImpl.java line 1145. That call can never fire there:

- undoTransaction transitions with DateUtils.getBusinessLocalDate(), so any closure it stamps carries today's date.

- Future-dated transactions are rejected (WorkingCapitalLoanDataValidator), so MAX(repayment date) <= today, and the isBefore guard is always false.

The reachable case it would need to fix is the mirror image: validateUndoTransaction explicitly allows OVERPAID, so undoing an overpaying repayment that leaves the loan exactly settled fires LOAN_REPAID_IN_FULL and stamps closedOnDate = the undo day, while maturedOnDate keeps the earlier date it got when the loan first went overpaid. You end up with timeline.closedOnDate after timeline.actualMaturityDate — the drift the settlementDate javadoc says these two should not have. Pre-existing, not a regression, but the javadoc currently overstates what the call does.

Smaller, same shape: recalculateClosedOnDate returns early unless isClosedObligationsMet(), so on the overpaid path maturedOnDate is never corrected. A backdated repayment tipping a part-paid loan into OVERPAID leaves actualMaturityDate at the backdated date even though recalculateOverpaidOnDate fixes overpaidOnDate and settlementDate uses the corrected value for the amortization.

Cocoa-Puffs:
Both findings turned out to be correct. `recalculateClosedOnDate(loan)` became `recalculateSettlementDates(loan, statusBeforeEvent)`, covering the OVERPAID branch as well as the closed one. I've extended the repository query to return the date of the newest live repayment-like transaction whose allocation is not entirely overpayment. 

The overpaid branch can now set this date directly. 

The closed branch now considers whether the loan was overpaid going into the current transaction processing, if prior the loan was overpaid and now it is simply closed it means that the current transaction is takins some sort of excess away. In this case the obligationMetOnDate is assigned outright, meaning they can move "earlier" as the loan is falling back onto a settlement that already happened.

Otherwise the event is what settled the loan, and it may not have been a repayment. Each date only moves later, never earlier, so a settlement completed by a non-repayment keeps the transition's own date.

I also added 2 new integration tests, that test for the mirror scenario you described.

adamsaghy:
@Cocoa-Puffs Can you please review the below concerns?

**Regression?**
-  recalculateMaturityOnOverpaid moves the maturity earlier when a non-repayment settles the loan

obligationsMetOn answers "the last repayment the loan needed", which equals "the day obligations were met" only when a repayment is what met them. recalculateClosure guards against that — it keeps the one-directional rule precisely because settlement can be completed by "a discount fee adjustment that reduces what is owed to what has already been paid". recalculateMaturityOnOverpaid has no such guard and assigns "in either direction".

Existing scenario WorkingCapitalDiscountAmortizationAdjustment.feature UC9 hits it exactly:

- 02 Jan: repayment 9900 against 10000 owed → ACTIVE, still owes 100.
- 03 Jan: discount fee adjustment of 500 drops what's owed to 9500 → OVERPAID by 400. Reprocessing reallocates the 02 Jan repayment to 9500 principal + 400 overpayment, so overpaymentPortion (400) < transactionAmount (9900) and it qualifies as "had a due portion".
- determineAndTransition correctly stamps maturedOnDate = 03 Jan (it was null). recalculateSettlementDates(loan, ACTIVE) then takes the overpaid branch and overwrites it with 02 Jan — a day the loan was ACTIVE and owed 100.
- 04 Jan: the CBR closes the loan. statusBeforeEvent is OVERPAID, so the bidirectional branch in recalculateClosure sets both closedOnDate and maturedOnDate to obligationsMetOn = 02 Jan.

Published timeline for UC9 goes from closedOnDate 2026-01-04 / actualMaturityDate 2026-01-03 to 2026-01-02 / 2026-01-02. UC9 asserts status, balances and transactions but no timeline fields, which is why CI stays green. DISCOUNT_FEE_ADJUSTMENT (46) is not in getRepaymentLikeTransactionTypes() — I checked it isn't aliased to CHARGE_ADJUSTMENT (26) — so the query can never name the day that actually settled the loan.

Working through the reachable transitions, the bidirectional assignment in recalculateMaturityOnOverpaid never helps: the case its javadoc cites ("an extra payment onto an already closed loan") is already covered upstream by determineAndTransition's if (getMaturedOnDate() == null) guard, and the backdated-overpayment case (UC18) only needs the date moved later. The same one-directional rule recalculateClosure uses would be correct in every one of them. The already-settled branch of recalculateClosure needs the same treatment — it can't fall back below a maturedOnDate that already records when obligations were met.

**Additional scope introduced?**

the credit-balance-refund behaviour change

The CBR path is new in this revision and isn't covered by either title claim. It changes a published API field: WorkingCapitalLoanRepaymentOverpayment.feature UC6 had its timeline.closedOnDate expectation edited from 2026-01-06 (the refund day) to 2026-01-04 (the day obligations were met), and loan.setMaturedOnDate(transactionDate) became conditional. This also diverges from core, where DefaultLoanLifecycleStateMachine stamps closedOnDate from the CBR's own date. The reasoning in the javadoc is coherent, but closedOnDate now names a day on which the loan's status was OVERPAID rather than closed — worth explicit product sign-off rather than riding along on a COB-dating fix.

Cocoa-Puffs:
Both findings are correct. During the retest after the previous round of findings a new scenario was found in which case the closing date turned out to be wrong. This was added as a deliberate test with testrail id: @TestRailId:C106757.

The issue was that an overpaid loan with credit balance refund set the maturity and closed on dates as the date of the CBR, which was considered wrong, as a CBR hands back money that the loan never needed, so it is not what settled it. The solution was to take the latest transactions date that paid into any outstanding amount, instead of 100% overpayment. That's what made the closed on date move in @TestRailId:C98268 as well.

Unfortunately this introduced a regression. A Discount Fee Adjustment, while not a repayment removes part of the obligation, therefore can be the transaction that closes or settles the loan.

The fix: Two changes in `WorkingCapitalLoanTransactionProcessor.java`.

`recalculateMaturityOnOverpaid` now only moves the date later, and the already-settled branch of `recalculateClosure` is floored at the maturity the loan already carries.

A floor rather than a direction ban, because the undo case still needs closedOnDate to move earlier — from the undo day back to the settlement it falls onto. The javadoc on both now says why: obligationsMetOn names the last payment the loan needed, which is the day obligations were met only when a payment is what met them, and the stamp is the only record of a settlement completed by something else.

I have also added extra validations to @TestRailId:C83084 to pin down this behaviour.

galovics (commented):
The COB fix looks right, and I checked it against the model rather than just the tests. Computing the target as of the COB date fixes both the "COB date trails the business date" case and the catch-up replay - a payment dated D+1 no longer gets amortized on COB day D. The fast path and the rebuild path agree on the same as-of date (including with a rate change), and the old `withDiscount` silently dropping rate changes is fixed as a side effect - that deserves its own unit test. UC13/UC14 and the Feign Case A/B tests reproduce the original bug.

**Things I'd like settled:**

1. **A backdated settlement that moves earlier leaves stale dates.** `recalculateMaturityOnOverpaid` only ever moves maturity later, and the refund/undo branch uses the stored maturity as a floor, so neither can go back. Example: 9000 paid on 02 Jan and 1500 on 10 Jan (overpaid, maturity 10 Jan), then a 1000 repayment dated 05 Jan settles it exactly on 05 Jan and the 10 Jan payment becomes pure surplus - maturity stays 10 Jan and a later full CBR sets `closedOnDate` to 10 Jan instead of 05 Jan. `DISCOUNT_FEE_ADJUSTMENT` isn't in `findLatestActiveTransactionDateWithDuePortion`, so a backdated adjustment on a settled loan hits the same gap. Including it and assigning the date outright would make it a pure function of history and let the one-way rule and the floor go.

2. **The as-of target isn't matched by an as-of "already posted" total** (`WorkingCapitalLoanDiscountFeeAmortizationServiceImpl` ~L71-90). `alreadyPosted` sums every amortization transaction regardless of date. If a closing payment dated D5 posted the full 1000 and is then undone before catch-up runs, catch-up on D1 computes 0 - 1000 and posts a 1000 reversal dated D1, then D5 re-books 739.90 - negative realized income for D1-D4 and a reversal dated before the income it reverses (the old code produced a smaller -260.10). Making the posted total as-of too (non-reversed, dated on or before the COB date) gives a clean D1=0, D5=-260.10.

3. **Closure income is dated inconsistently.** Amortization and `closedOnDate` now use the settlement date but `accrueOnClosure` still uses the triggering transaction's date, so on a backdated closure charge accruals can land before the settlement (or before the charge's due date), and on the CBR path after the now-earlier `closedOnDate`. Pass `settlementDate(...)` there too. Relatedly `settlementDate()` reads `overpaidOnDate` while `recalculateSettlementDates` maintains `maturedOnDate` - they can differ.

4. **Scope: the credit balance refund `closedOnDate` change** (UC6 in the overpayment feature flipped from 2026-01-06 to 2026-01-04). Undo falling back to an earlier settlement is fine, but a CBR is a real money movement - until it happened the account wasn't really closed, and core's `DefaultLoanLifecycleStateMachine` stamps the CBR date. WC now differs from core on a published API field, and an existing assertion was flipped to match. Needs explicit sign-off or its own ticket.

**Tests to add:** model unit tests for `totalActualAmortizationAsOf` (fast vs rebuild path, rate change before/after the as-of date); the backdated-repayment-into-overpaid-loan scenario with a later refund asserting `actualMaturityDate`/`closedOnDate`; a COB after the undo in UC15 asserting the amortization adjustment (UC15 stops right after the undo, which hides the closing amortization being left in place).

Minor: `asOfModel` calls `applyPayment` in a loop, rebuilding the schedule each time - add the payments first and rebuild once; the comments are far longer than the usual style; the PR description is the untouched template.

Recommendation: COMMENT

Cocoa-Puffs:
@galovics I've addressed the concerns.

## 1. A backdated settlement that moves earlier leaves stale dates — confirmed, fixed

Reproduced exactly as described. Negative control on the pre-fix build (9000 on 02 June, 1500 on 10 June, then 1000 backdated to 05 June):

```
timeline.actualMaturityDate expected: <2026-06-05> but was: <2026-06-10>
```

`recalculateSettlementDates(loan, statusBeforeEvent)` is now `recalculateSettlementDates(loan)`. The one-way rule, the floor and the `statusBeforeEvent` argument are gone; both dates are assigned outright from a new `settlementDateFromHistory(loan)`.

I didn't take the literal suggestion of adding `DISCOUNT_FEE_ADJUSTMENT` to the query, because `WorkingCapitalLoanTransactionAllocation.forDiscountFeeAdjustment` hardcodes `overpaymentPortion = 0`. The `COALESCE(a.overpaymentPortion, 0) < t.transactionAmount` filter would therefore pass for *every* adjustment, including one on an already-overpaid loan, which would then claim a settlement it didn't make — loan overpaid since 02 Jan, adjustment on 05 Jan, maturity wrongly moves to 05 Jan. Instead the state machine's stamp survives only while a non-reversed discount fee adjustment exists on that exact date. Same pure-function-of-history property you were after, without that regression.

## 2. The as-of target isn't matched by an as-of "already posted" total — confirmed, fixed

New `sumNetAmortizationAsOf`; `processDiscountFeeAmortization` now bounds the posted total by the same date as the target. Negative control on the pre-fix build:

```
transactionDate expected: <05 January 2026> but was: <02 January 2026>
```

That's the 1000 reversal landing on the first replayed day, before the income it reverses.

One correction to your figures: in the scenario we built, the closing amortization is dated on the *settlement* day rather than the payment day, so the corrected shape is +739.90 on 05 January and −1000 on 07 January, not 0 / −260.10. Net realized income is 739.90 either way; the defect and the fix are exactly as you described.

## 3. Closure income dated inconsistently — fixed

All three `accrueOnClosure` sites now date from the settlement, and `settlementDate()` reads `maturedOnDate` in both branches, so it can no longer disagree with `recalculateSettlementDates`.

The accruals use a separate `closureIncomeDate()` rather than `settlementDate()`, because the two answer different questions. The discount is measured against a schedule, and a schedule changes on the day it changes — so a discount fee adjustment landing after the settlement earns its correction on its own later day, and `settlementDate()` keeps its floor for that. A charge accrual instead recognizes income the *closure* leaves unrecognized. A write-off still falls back to the transaction date, where the closing transaction is the settlement.

In fairness this one is a consistency fix rather than a defect repair: `accrueOnClosure` already fires on whichever event settles the loan, so by the time a later non-settling event reaches it every active charge is accrued and the call is idempotent. The "accruals land before the settlement" case is the opposite direction, and the floor already handled it.

## 4. CBR scope — you're right, and we've reversed it

Agreed: a CBR is a real money movement, the account isn't closed until it happens, and it can absolutely be what closes the loan. `DefaultLoanLifecycleStateMachine.determineTransitionFromOverpaid` stamps both `closedOnDate` and `actualMaturityDate` from the refund's own date, and WC now does the same. The CBR path does no correction at all — `recalculateSettlementDates` is not called there, and the javadoc says why. Three assertions moved:

| Test | Field | Was | Now |
| --- | --- | --- | --- |
| `C106757` (UC17) | `closedOnDate`, `actualMaturityDate` | `2026-01-05` | `2026-01-09` |
| `C98268` (UC6) | `closedOnDate` | `2026-01-04` | `2026-01-06` |
| `C83084` (UC9) | `closedOnDate`, `actualMaturityDate` | `2026-01-03` | `2026-01-04` |

`C106757`'s scenario title was reworded too, since it asserted the old rule in prose. `C98268` goes back to the tester's original value.

## Tests added

All three you asked for:

- `ProjectedAmortizationScheduleAsOfTest` — 5 tests covering `totalActualAmortizationAsOf`: fast vs rebuild path, rate change before and after the as-of date, with `assertNotEquals` guards so a dropped rate change fails rather than passing silently.
- A Feign test for the backdated repayment into an overpaid loan with a later refund, asserting both `actualMaturityDate` and `closedOnDate`.
- UC19 — a COB after the undo, asserting the amortization adjustment. It carries a `# TODO: assign a TestRail ID` rather than an invented one.

## Minor

`asOfModel` now adds payments straight to the list and rebuilds once, the way `regenerate()` already does. Comments trimmed throughout.

PR #6477, opened 2026-09-18
FINERACT-2455: WC - delinquency bucket for regular loan shouldn't be available on WC loan product creation

mariiaKraievska:
## Description

## Checklist


adamsaghy:
@mariiaKraievska Please review my concerns / findings:


1. updateFrom resolves the bucket before the change check

WorkingCapitalLoanAssemblerImpl.java:420-429

if (fromApiJsonHelper.parameterExists(delinquencyBucketIdParamName, element)) {
    final Long bucketId = ...;
    final DelinquencyBucket bucket = resolver.findWorkingCapitalBucketByIdIfProvided(bucketId); // always runs
    final Long existingBucketId = ...;
    if (!Objects.equals(bucketId, existingBucketId)) { ... }
}

The lookup + type validation runs unconditionally, outside the !Objects.equals(...) guard. The three sibling blocks around it (breachId, nearBreachId, delinquencyGraceDays) all gate the lookup on the change check. Consequence: a WC loan created before this PR with a REGULAR bucket (the old code was findById(bucketId).orElse(null) with no type check, so this was reachable) can no longer be modified at all if the client resends the unchanged delinquencyBucketId — it 400s on a no-op. Same shape for a since-deleted id, which used to be silently ignored and now throws DelinquencyBucketNotFoundException.

Moving the resolve inside the if fixes both and matches the surrounding style. The WC product update path already does this correctly via isChangeInLongParameterNamed (WorkingCapitalLoanProductWritePlatformServiceImpl.java:276), as does the LP update path.

2. Error codes and unused message args

Both new errors pass the bucket id as a defaultUserMessageArgs value, but neither message has a {0} placeholder, so it's never rendered. The codes also drop the entity segment that the rest of the codebase uses (validation.msg.delinquencyBucketId.must.be.regular.type vs. the conventional validation.msg.loanproduct.delinquencyBucketId.…). The two messages are otherwise near-identical — one code parameterized by expected type would do.

3. retrieveWorkingCapitalDelinquencyBucketOptions() returns null when empty

WorkingCapitalDelinquencyBucketResolver.java:46-50. This preserves the old serialization behaviour, but it pushes null-returning into a shared service method that now has two callers (WC product template and, transitively, the WC loan template). Returning List.of() and keeping the null-ing at the response-building boundary would be safer.

4. Duplication

retrieveAllDelinquencyBuckets() and retrieveDelinquencyBucketsByType() are identical apart from the finder (DelinquencyReadPlatformServiceImpl.java:105-119). Also validateIsWorkingCapitalType is public but only called from within its own class.

5. Test-side nits

DelinquencyBucketResolver.resolveBucketId calls resolve(...) on this, so Spring's @Cacheable proxy is bypassed — every resolution re-fetches all buckets. Not a correctness issue.
WorkingCapitalLoanAccount.feature still hardcodes delinquencyBucketId | 1 in three existing scenarios. Harmless today (buildCreateLoanRequest ignores column 7, and the negative scenario fails in the validator before the assembler), but the seeded bucket ids are genuinely nondeterministic — DelinquencyGlobalInitializerStep.initialize() creates BASIC_DELINQUENCY_BUCKET and WC_DELINQUENCY_BUCKET via ParallelExecutionHelper.runInParallel. Now that resolveBucketId accepts names, those literals are worth converting before someone wires column 7 up.
LoanProduct.feature C106741 creates a loan product and leaves it behind; the WC counterpart C106744 deletes its product.

mariiaKraievska:
> @mariiaKraievska Please review my concerns / findings:
> 
> 1. updateFrom resolves the bucket before the change check
> 
> WorkingCapitalLoanAssemblerImpl.java:420-429
> 
> if (fromApiJsonHelper.parameterExists(delinquencyBucketIdParamName, element)) { final Long bucketId = ...; final DelinquencyBucket bucket = resolver.findWorkingCapitalBucketByIdIfProvided(bucketId); // always runs final Long existingBucketId = ...; if (!Objects.equals(bucketId, existingBucketId)) { ... } }
> 
> The lookup + type validation runs unconditionally, outside the !Objects.equals(...) guard. The three sibling blocks around it (breachId, nearBreachId, delinquencyGraceDays) all gate the lookup on the change check. Consequence: a WC loan created before this PR with a REGULAR bucket (the old code was findById(bucketId).orElse(null) with no type check, so this was reachable) can no longer be modified at all if the client resends the unchanged delinquencyBucketId — it 400s on a no-op. Same shape for a since-deleted id, which used to be silently ignored and now throws DelinquencyBucketNotFoundException.
> 
> Moving the resolve inside the if fixes both and matches the surrounding style. The WC product update path already does this correctly via isChangeInLongParameterNamed (WorkingCapitalLoanProductWritePlatformServiceImpl.java:276), as does the LP update path.
> 
> 2. Error codes and unused message args
> 
> Both new errors pass the bucket id as a defaultUserMessageArgs value, but neither message has a {0} placeholder, so it's never rendered. The codes also drop the entity segment that the rest of the codebase uses (validation.msg.delinquencyBucketId.must.be.regular.type vs. the conventional validation.msg.loanproduct.delinquencyBucketId.…). The two messages are otherwise near-identical — one code parameterized by expected type would do.
> 
> 3. retrieveWorkingCapitalDelinquencyBucketOptions() returns null when empty
> 
> WorkingCapitalDelinquencyBucketResolver.java:46-50. This preserves the old serialization behaviour, but it pushes null-returning into a shared service method that now has two callers (WC product template and, transitively, the WC loan template). Returning List.of() and keeping the null-ing at the response-building boundary would be safer.
> 
> 4. Duplication
> 
> retrieveAllDelinquencyBuckets() and retrieveDelinquencyBucketsByType() are identical apart from the finder (DelinquencyReadPlatformServiceImpl.java:105-119). Also validateIsWorkingCapitalType is public but only called from within its own class.
> 
> 5. Test-side nits
> 
> DelinquencyBucketResolver.resolveBucketId calls resolve(...) on this, so Spring's @Cacheable proxy is bypassed — every resolution re-fetches all buckets. Not a correctness issue. WorkingCapitalLoanAccount.feature still hardcodes delinquencyBucketId | 1 in three existing scenarios. Harmless today (buildCreateLoanRequest ignores column 7, and the negative scenario fails in the validator before the assembler), but the seeded bucket ids are genuinely nondeterministic — DelinquencyGlobalInitializerStep.initialize() creates BASIC_DELINQUENCY_BUCKET and WC_DELINQUENCY_BUCKET via ParallelExecutionHelper.runInParallel. Now that resolveBucketId accepts names, those literals are worth converting before someone wires column 7 up. LoanProduct.feature C106741 creates a loan product and leaves it behind; the WC counterpart C106744 deletes its product.

From 1-4 addressed, thanks:

1. moved bucket validation inside `isChangeInLongParameterNamed`, same pattern as breach/nearBreach and WC product update
2. Error codes — shared message template parameterized by expected type
3. null from resolver— `retrieveWorkingCapitalDelinquencyBucketOptions()` returns the list as-is. empty → `null` only at the product template response.
4. Duplication — `retrieveAll` / `byType` share `mapAndEnrich`. `validateIsWorkingCapitalType` is private.

@MarianaDmytrivBinariks told me she is working on 5th. Please correct me if I`m wrong

MarianaDmytrivBinariks:
> > @mariiaKraievska Please review my concerns / findings:
> > 
> > 1. updateFrom resolves the bucket before the change check
> > 
> > WorkingCapitalLoanAssemblerImpl.java:420-429
> > if (fromApiJsonHelper.parameterExists(delinquencyBucketIdParamName, element)) { final Long bucketId = ...; final DelinquencyBucket bucket = resolver.findWorkingCapitalBucketByIdIfProvided(bucketId); // always runs final Long existingBucketId = ...; if (!Objects.equals(bucketId, existingBucketId)) { ... } }
> > The lookup + type validation runs unconditionally, outside the !Objects.equals(...) guard. The three sibling blocks around it (breachId, nearBreachId, delinquencyGraceDays) all gate the lookup on the change check. Consequence: a WC loan created before this PR with a REGULAR bucket (the old code was findById(bucketId).orElse(null) with no type check, so this was reachable) can no longer be modified at all if the client resends the unchanged delinquencyBucketId — it 400s on a no-op. Same shape for a since-deleted id, which used to be silently ignored and now throws DelinquencyBucketNotFoundException.
> > Moving the resolve inside the if fixes both and matches the surrounding style. The WC product update path already does this correctly via isChangeInLongParameterNamed (WorkingCapitalLoanProductWritePlatformServiceImpl.java:276), as does the LP update path.
> > 
> > 2. Error codes and unused message args
> > 
> > Both new errors pass the bucket id as a defaultUserMessageArgs value, but neither message has a {0} placeholder, so it's never rendered. The codes also drop the entity segment that the rest of the codebase uses (validation.msg.delinquencyBucketId.must.be.regular.type vs. the conventional validation.msg.loanproduct.delinquencyBucketId.…). The two messages are otherwise near-identical — one code parameterized by expected type would do.
> > 
> > 3. retrieveWorkingCapitalDelinquencyBucketOptions() returns null when empty
> > 
> > WorkingCapitalDelinquencyBucketResolver.java:46-50. This preserves the old serialization behaviour, but it pushes null-returning into a shared service method that now has two callers (WC product template and, transitively, the WC loan template). Returning List.of() and keeping the null-ing at the response-building boundary would be safer.
> > 
> > 4. Duplication
> > 
> > retrieveAllDelinquencyBuckets() and retrieveDelinquencyBucketsByType() are identical apart from the finder (DelinquencyReadPlatformServiceImpl.java:105-119). Also validateIsWorkingCapitalType is public but only called from within its own class.
> > 
> > 5. Test-side nits
> > 
> > DelinquencyBucketResolver.resolveBucketId calls resolve(...) on this, so Spring's @Cacheable proxy is bypassed — every resolution re-fetches all buckets. Not a correctness issue. WorkingCapitalLoanAccount.feature still hardcodes delinquencyBucketId | 1 in three existing scenarios. Harmless today (buildCreateLoanRequest ignores column 7, and the negative scenario fails in the validator before the assembler), but the seeded bucket ids are genuinely nondeterministic — DelinquencyGlobalInitializerStep.initialize() creates BASIC_DELINQUENCY_BUCKET and WC_DELINQUENCY_BUCKET via ParallelExecutionHelper.runInParallel. Now that resolveBucketId accepts names, those literals are worth converting before someone wires column 7 up. LoanProduct.feature C106741 creates a loan product and leaves it behind; the WC counterpart C106744 deletes its product.
> 
> From 1-4 addressed, thanks:
> 
> 1. moved bucket validation inside `isChangeInLongParameterNamed`, same pattern as breach/nearBreach and WC product update
> 2. Error codes — shared message template parameterized by expected type
> 3. null from resolver— `retrieveWorkingCapitalDelinquencyBucketOptions()` returns the list as-is. empty → `null` only at the product template response.
> 4. Duplication — `retrieveAll` / `byType` share `mapAndEnrich`. `validateIsWorkingCapitalType` is private.
> 
> @MarianaDmytrivBinariks told me she is working on 5th. Please correct me if I`m wrong

@mariiaKraievska yes, you're right
5. 
 - DelinquencyBucketResolver.resolveBucketId calls resolve(...) on this, so Spring's @Cacheable proxy is bypassed — every resolution re-fetches all buckets. Not a correctness issue.  - fixed
 - WorkingCapitalLoanAccount.feature still hardcodes delinquencyBucketId | 1 in three existing scenarios. - fixed
 - LoanProduct.feature C106741 creates a loan product and leaves it behind; the WC counterpart C106744 deletes its product. - regular loan product have no delete endpoint at all, so it can't be done

galovics (dismissed):
The enforcement is solid. The bucket type is validated on every write path, not just in the dropdowns: loan product create/update (`findDelinquencyBucketIdIfProvided`), and WC product create/update plus WC loan create/modify through the new `WorkingCapitalDelinquencyBucketResolver`, in both directions (REGULAR-only for loan products, WORKING_CAPITAL-only for WC). `bucket_type` is `NOT NULL DEFAULT 'REGULAR'`, so existing rows are reliably discriminated, and every update path only re-validates when the bucket id actually changes, so existing products/loans linked to a mismatched bucket keep working as long as the client resends the same id. Core delinquency behaviour isn't regressed (`retrieveAllDelinquencyBuckets()` still returns everything for `GET /delinquency/buckets`), and creating a WC loan with an unknown bucket id now fails with a 404 instead of silently storing null - a good silent fix.

Two follow-ups, neither blocking:
- **The invariant can still be broken from the bucket side.** `updateDelinquencyBucket` lets `bucketType` change while the bucket is in use, and `DelinquencyBucketParseAndValidator.validateAndParseUpdate` defaults a missing `bucketType` to REGULAR - so a PUT on a WC bucket that just omits the field silently turns it into a REGULAR one while WC products still point at it. I'd reject a type change while the bucket is in use (the delete path already has usage checkers) or keep the current type when omitted. Worth a follow-up ticket.
- A legacy WC product created before this with a REGULAR bucket still hands it to new loans that don't send their own `delinquencyBucketId` (the assembler copies `product.getDelinquencyBucket()` without a type check). Worth a line in the JIRA so operators check existing products.

Nits: the wrong-type error code for the WC loan uses `validation.msg.WORKINGCAPITALLOANPRODUCT...` hardcoded instead of the constant, the message text is duplicated between the two modules, and the WC loan modify path has no test (nor the "legacy product still updatable when the same id is resent" case). The PR description is still the empty template and should list the API behaviour changes (filtered template options, `delinquencyBucketOptions` null when there are no WC buckets).

Recommendation: APPROVE

mariiaKraievska:
> The enforcement is solid. The bucket type is validated on every write path, not just in the dropdowns: loan product create/update (`findDelinquencyBucketIdIfProvided`), and WC product create/update plus WC loan create/modify through the new `WorkingCapitalDelinquencyBucketResolver`, in both directions (REGULAR-only for loan products, WORKING_CAPITAL-only for WC). `bucket_type` is `NOT NULL DEFAULT 'REGULAR'`, so existing rows are reliably discriminated, and every update path only re-validates when the bucket id actually changes, so existing products/loans linked to a mismatched bucket keep working as long as the client resends the same id. Core delinquency behaviour isn't regressed (`retrieveAllDelinquencyBuckets()` still returns everything for `GET /delinquency/buckets`), and creating a WC loan with an unknown bucket id now fails with a 404 instead of silently storing null - a good silent fix.
> 
> Two follow-ups, neither blocking:
> 
> * **The invariant can still be broken from the bucket side.** `updateDelinquencyBucket` lets `bucketType` change while the bucket is in use, and `DelinquencyBucketParseAndValidator.validateAndParseUpdate` defaults a missing `bucketType` to REGULAR - so a PUT on a WC bucket that just omits the field silently turns it into a REGULAR one while WC products still point at it. I'd reject a type change while the bucket is in use (the delete path already has usage checkers) or keep the current type when omitted. Worth a follow-up ticket.
> * A legacy WC product created before this with a REGULAR bucket still hands it to new loans that don't send their own `delinquencyBucketId` (the assembler copies `product.getDelinquencyBucket()` without a type check). Worth a line in the JIRA so operators check existing products.
> 
> Nits: the wrong-type error code for the WC loan uses `validation.msg.WORKINGCAPITALLOANPRODUCT...` hardcoded instead of the constant, the message text is duplicated between the two modules, and the WC loan modify path has no test (nor the "legacy product still updatable when the same id is resent" case). The PR description is still the empty template and should list the API behaviour changes (filtered template options, `delinquencyBucketOptions` null when there are no WC buckets).
> 
> Recommendation: APPROVE

Addressed the follow-ups in this PR:

1. updateDelinquencyBucket now rejects bucketType changes while the bucket is in use (same usage checks as delete), and omitted bucketType keeps the existing type instead of defaulting to REGULAR
2. WC wrong-type validation code goes through a constant. The message text is shared via DelinquencyApiConstants
3. Added e2e coverage for WC loan modify with a non-WC bucket (rejected) and resenting the same bucket id (accepted)

PR #6479, opened 2026-09-18
FINERACT-2455: Fix WorkingCapitalPeriodPaymentRate.feature long cob runs, attempt to fix flaky CobPartitioningTest

Cocoa-Puffs:
## Description

CobPartitioningTest.testLoanCOBPartitioningQuery() frequently fails in Github Actions runs with:

Exception Description: Syntax error parsing [SELECT lt.typeOf AS transactionType, ... ]. The select item is missing from the result variable declaration. SELECT lt.typeOf  [ AS transactionType ] ... The SELECT clause has 'lt.typeOf' and 'AS transactionType' that are not separated by a comma. 

The culprit is the query: findTransactionDataForForeclosureIncome() in LoanTransactionRepository. It uses "AS <alias>" in the select clause which are in practice JPQL result-variables. The parse error is EclipseLink's Hermes parser reporting that the grammar singleton it consulted has no result-variable rule registered, likely because another thread is still initializing it. My proposed fix is to bypass result-variables by using a constructor expression, that every other query uses too.

## Checklist


ruzeynalov (approved):

galovics (commented):
The e2e side is fine: I checked that the feature-file changes don't weaken the scenarios. No `Then` steps were removed, only dates and the derived numbers moved, future-dated rate changes are still exercised (UC12: business date 10 Jan, change effective 15 Jan), and UC12 is actually stronger now - it asserts the closing row (balance 0.00) plus an exact row count of 334, and the numbers are consistent (change at row 14, `segmentTerm` 320). UC21 is the same schedule shifted 10 days earlier with the same tail values and EIR. All dates are business dates, so there's no wall-clock or timezone dependence.

Two things:

**Scope.** The title says e2e speed-up, but this isn't test-only: it also changes `LoanTransactionRepository.findTransactionDataForForeclosureIncome` from `AS alias` result variables to a `SELECT new ...TransactionPortionsForForeclosure(...)` constructor expression and turns the public `TransactionPortionsForForeclosure` interface into a Lombok class (anyone implementing the interface outside the repo breaks). The behaviour is unchanged (same getters, the only caller in `LoanAccrualsProcessingServiceImpl` is fine, and it's the only `AS alias` query left in the repo), and the intent - working around a flaky `CobPartitioningTest` - is reasonable. But it's a term-loan production change for an unrelated flaky test under the WC ticket; it deserves its own ticket and PR, or at least a description covering both. The root cause ("grammar singleton not initialized by another thread") is a guess with no reproduction, so present it as a workaround and keep watching the test.

**Coverage lost in two scenarios.**
- C89816 (backdated repayment): with the rate change now on 03 Jan, `remembers ... payments before "2026-01-03"` captures only row 0 and row 1, where before it compared about 9 rows - so "reprocessing must not move the segment before the change" is checked on one meaningful row. Moving the change to ~06 Jan with repayments on 08 and 07 Jan (the second backdated) keeps several rows before it while still cutting most of the COB days.
- UC23 (change after maturity): at rate 18 on 1000 + 100 the loan matures ~23 Jan, so the new date (24 Jan) is only about one day past maturity where before it had ~9 days of margin; it now depends on how the maturity comparison treats the boundary. This scenario has no inline COB, so the date change saves little - leave some margin.

Recommendation: COMMENT

Cocoa-Puffs:
Restored coverage to the two mentioned scenarios.

## UC23, restored to original date

I put the date back to 01 February 2026 entirely, because the change was buying nothing: the scenario runs no COB. No reason to trade any margin.

## C89816, took the suggestion

Rate change → 06 Jan, repayments on 08 Jan and a backdated 07 Jan. Five settled rows now sit ahead of the change instead of one.

I believe the two changes are related, the failing integration test and the feature-file work are the same effort to get this suite running reliably and quickly.

## What was failing
CobPartitioningTest.testLoanCOBPartitioningQuery() failed intermittently on CI with a 403 from POST /loans/{id}/transactions?command=foreclosure, carrying this validation error:

Exception Description: Syntax error parsing [SELECT lt.typeOf AS transactionType, lt.interestPortion AS interestPortion,
lt.feeChargesPortion AS feeChargesPortion, lt.penaltyChargesPortion AS penaltyChargesPortion FROM LoanTransaction lt WHERE lt.loan = :loan AND lt.reversed = false AND lt.dateOf <= :tillDate AND lt.typeOf NOT IN (...) ORDER BY lt.dateOf]. [17, 17] The select item is missing from the result variable declaration. (SELECT lt.typeOf  [ AS transactionType ] ...

The error names the query verbatim. It is a JPQL parse failure inside EclipseLink, raised before any SQL is generated — which is why it is database-independent, and why it surfaces as a grammar error rather than a timeout.

## Why it is intermittent
The test drives five concurrent foreclosures. The failure appears when several threads reach this query's first parse at once. CobPartitioningTest already carries two hand-rolled warm-up calls for exactly this, added previously by someone who hit the same thing:

```
// Warm up (EclipseLink sometimes fails if JPQL cache is not warm up but concurrent queries are executed)
```
It reduces failures, but doesn't eliminate them.

The concurrency explanation for the intermittency remains a hypothesis; we have no standalone reproduction of the race itself. What is not a hypothesis is which query fails and why it is rejected, the stack trace states both.

## Compatibility concerns

TransactionPortionsForForeclosure is public, so this is a source-incompatible change for anyone who implemented the interface themselves. In practice a Spring Data projection interface is implemented by the framework, never by hand, and the type is not part of the REST API.

PR #6485, opened 2026-09-20
FINERACT-2845: Register the JPA auditing handler as a singleton

alberto-art3ch:
## Description

`jpaAuditingHandler` was registered as a prototype, and since Spring Data Commons 4.x the `AuditingHandler` constructor walks every persistent entity to build its auditing metadata — so every command triggered a full mapping-context scan (85% of JFR execution samples under a write-heavy workload; a mixed write run went from 18m 33s to 12m 58s after the fix).

The prototype scope existed only because `markCreated`/`markModified` mutated the handler's `DateTimeProvider` to pick local vs UTC timestamps per entity; that choice now travels in a thread local, so a shared instance is safe and the bean goes back to singleton.

Audit timestamp semantics are unchanged — the four pre-existing `CustomAuditingHandlerTest` cases pass unmodified — plus new tests for the thread-local isolation and for the bean scope.

[FINERACT-2845](https://issues.apache.org/jira/browse/FINERACT-2845)

## Checklist


adamsaghy:
@alberto-art3ch It looks good, but would you review the below concerns? 

- ThreadLocalDateTimeProvider.java:47: clear() removes rather than restoring a previous value. Correct today (re-entrancy is impossible, as above), but if set returned the prior provider and the finally restored it, the class would stay correct without depending on that reasoning.

- The holder is static on a singleton, while the choice conceptually belongs to CustomAuditingHandler. A non-static ThreadLocal in a provider instance owned by the handler would keep two handler instances independent — only matters for tests, so take it or leave it.

- JpaAuditingHandlerRegistrarTest.java:38 passes null for AnnotationMetadata. It works because the parameter is ignored, but the test silently depends on that; mock(AnnotationMetadata.class) would be more robust.

adamsaghy (approved):
LGTM

PR #6488, opened 2026-09-21
FINERACT-2455: Fix WorkingCapitalBreach feature files long cob runs

Cocoa-Puffs:
## Description

## Checklist


ruzeynalov (approved):

galovics (changes requested):
Test-only and I walked the rescaling scenario by scenario: the breach/near-breach windows, the disable/enable boundary pairs (including the same-day case with `endDate = startDate - 1`), and the period-edge checks (Jan 5/6/7 around a Jan 1-6 period) all still line up, no `Then` step was removed, and the month-arithmetic coverage (monthly periods across month ends, disburse on the 31st, leap year, grace days) survives in `WorkingCapitalBreachSchedule.feature`. The new step is backward compatible. One scenario no longer tests what it says though, plus two smaller things.

**UC4 (C85466) is no longer backdated.** The title is "Breach disable and enable are rejected when backdated...", but the PR moves the business date from 15 Jan to 03 Jan and leaves `startDate "10 January 2026"` on both actions. Before, 10 Jan was 5 days in the past; now it's 7 days in the future. It still passes because the "must be the current business date" message covers both directions, so CI can't catch it - which means past-date rejection for disable/enable is now untested. Change both dates to something before 03 Jan (e.g. `02 January 2026`).

**UC5 (C85276) lost its MONTHS reschedule coverage.** It was the only reschedule scenario using `frequencyType MONTHS`, and the PR changes all three actions and the expected history to `6 DAYS`. It only checks the stored action history and runs one COB, so it wasn't slow because of the frequency - moving the date from 01 June to 03 January already gives the speedup. Restore `2 | MONTHS` in the three tables and the expected actions.

**`delinquencyGraceDays 3` is added everywhere without explanation.** Every replaced product step now uses the custom-config step with grace 3, including two scenarios (C74554, C74560) that previously had it empty, so each scenario changes two things at once and alters the delinquency path during COB. Leave it empty to preserve behaviour, or say why it's needed.

Nits: the new step duplicates `createProductWithGraceDaysForLoanTest` apart from `.delinquencyBucketId(bucketId)` (a small helper would do), it reads the bucket id from `TestContext.GLOBAL` which the per-scenario hook never resets (so a missing "create bucket" step would silently reuse an old bucket and the `isNotNull()` check still passes - same pattern as the existing delinquency steps), a few tables lost their alignment, and the PR description is empty.

Cocoa-Puffs:
## UC4 (C85466) fixed

Both rejected dates moved to 02 January 2026, one day before the 03 January business date, so they are backdated again.

## UC5 (C85276) fixed, and the product step reverted too
Correct: this scenario only asserts stored action history plus one COB, so the frequency was never what made it slow, the 01 June → 03 January move is the entire saving.

I went slightly further than suggested and reverted the product step as well, back to Admin creates a new Working Capital Loan Product with breachId and overrides enabled. The custom 6-day breach product was no more load-bearing here than the 6 DAYS reschedule frequency was. So UC5's diff is now purely the date change, with 2 | MONTHS restored in all three action tables and in the expected history.

## delinquencyGraceDays 3 fixed
It is now empty in all 18 rows this PR introduced. What the replacement now changes is only breachFrequency / breachFrequencyType, 2 MONTHS → 6 DAYS, which is the actual lever.

## Nits:
- Duplicated step definition fixed. Both step methods now delegate to a private `createGraceDaysProductForLoanTest(graceDays, startType, bucketId);` a null bucket id leaves the factory default in place.

- TestContext.GLOBAL staleness — fair, and fixed. The bucket id is now also published scenario-scoped where the bucket is created, and the new step reads that, so a scenario missing its create-bucket step fails instead of silently reusing a previous scenario's bucket. Existing readers keep using the global entry, so nothing else changes behaviour.

- Table alignment: realigned in the tables touched here.

